PR #5, opened 2021-09-24
Import the WHO GHO dataset into the catalog

larsyencken:
Import an unharmonized version of WHO's Global Health Observatory dataset into our data catalog.


larsyencken:
@danyx23 This import is completely working, but it doesn't use `_dimensions.csv` or `_dim_values.csv`, meaning that the dimension data has no supporting metadata. This dataset looks like a rich goldmine for considering dimensions, entities and normalisation.

larsyencken:
Gonna merge it as is, then we can decide together where entity metadata should live in our new format.

PR #21, opened 2021-10-05
WIP: first sketch of what the grapher import could look like

larsyencken:


dag.yml
```diff
@@ -17,4 +17,6 @@ steps:
     - data://meadow/gapminder/2019-12-10/population
     - data://reference
   data://meadow/wpp/2019/standard_projections:
-    - walden://wpp/2019/standard_projections
\ No newline at end of file
+    - walden://wpp/2019/standard_projections
+  grapher://grapher/who/2021-07-01/ghe:
```
larsyencken, line 21:
```suggestion
  grapher://who/2021-07-01/ghe:
```

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+
+    path: str
+    dependencies: List[Step]
+
+    def __init__(self, path: str, dependencies: List[Step]) -> None:
+        self.path = path
+        self.dependencies = dependencies
+
+    def __str__(self) -> str:
+        return f"data://{self.path}"
```
larsyencken, line 346:
```suggestion
        return f"grapher://{self.path}"
```

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+    @property
+    def _dest_dir(self) -> Path:
+        return DATA_DIR / self.path.lstrip("/")
+
+    def _run_py(self) -> None:
+        """
+        Import the Python module for this step and call get_grapher_tables() on it.
+        """
+        module_path = self.path.lstrip("/").replace("/", ".")
+        step_module = import_module(f"etl.steps.data.{module_path}")
```
larsyencken, line 434: Suggest to move this from `etl.steps.data.XYZ` to `etl.steps.grapher.XYZ`

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+        in_order = [v for _, v in sorted(checksums.items())]
+        return hashlib.md5(",".join(in_order).encode("utf8")).hexdigest()
+
+    @property
+    def _output_dataset(self) -> catalog.Dataset:
+        "If this step is completed, return the MD5 of the output."
+        if not self._dest_dir.is_dir():
+            raise Exception("dataset has not been created yet")
+
+        return catalog.Dataset(self._dest_dir.as_posix())
```
larsyencken, line 408: There is not dataset output, so trim this method.

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+
+        found_source_checksum = catalog.Dataset(
+            self._dest_dir.as_posix()
+        ).metadata.source_checksum
+        exp_source_checksum = self.checksum_input()
+
+        if found_source_checksum != exp_source_checksum:
+            return True
+
+        return False
```
larsyencken, line 379:
```suggestion
    def is_dirty(self) -> bool:
        return True
```

We would need to store checksums in MySQL to know if we need to re-run this step. That will be important to do, but we can defer it until we're happy with the shape of things. Until then, just always say the step needs running.

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+    def _output_dataset(self) -> catalog.Dataset:
+        "If this step is completed, return the MD5 of the output."
+        if not self._dest_dir.is_dir():
+            raise Exception("dataset has not been created yet")
+
+        return catalog.Dataset(self._dest_dir.as_posix())
+
+    def checksum_output(self) -> str:
+        # This cast from str to str is IMHO unnecessary but MyPy complains about this without it...
+        return cast(str, self._output_dataset.checksum())
```
larsyencken, line 412:
```suggestion
        raise Exception('grapher steps do not output a dataset')
```

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+    def _step_files(self) -> List[str]:
+        "Return a list of code files defining this step."
+        if self._search_path.is_dir():
+            return [p.as_posix() for p in files.walk(self._search_path)]
+
+        return glob(self._search_path.as_posix() + ".*")
+
+    @property
+    def _search_path(self) -> Path:
+        return Path(STEP_DIR) / "data" / self.path
```
larsyencken, line 423:
```suggestion
        return Path(STEP_DIR) / "grapher" / self.path
```

See earlier suggestion to move the grapher steps into their own folder namespace.

etl/command.py
```diff
@@ -322,6 +327,130 @@ def _run_notebook(self) -> None:
...
+            )
+
+        # data steps
+        dataset = step_module.get_grapher_dataset()  # type: ignore
+        dataset_upsert_results = upsert_dataset(
+            dataset, dataset.metadata.namespace, dataset.metadata.sources
+        )
+        for table in step_module.get_grapher_tables(dataset):  # type: ignore
+            upsert_table(table, dataset_upsert_results)
+            # TODO: call grapher_import.upsert_table here
```
larsyencken, line 451:
```suggestion
```

etl/config.py
```diff
@@ -3,10 +3,18 @@
...
 
 S3_BUCKET = "owid-catalog"
 S3_REGION_NAME = "nyc3"
 S3_ENDPOINT_URL = "https://nyc3.digitaloceanspaces.com"
 S3_HOST = "nyc3.digitaloceanspaces.com"
 S3_ACCESS_KEY = environ.get("OWID_ACCESS_KEY")
 S3_SECRET_KEY = environ.get("OWID_SECRET_KEY")
+
+DEBUG = environ.get("DEBUG") == "True"
+GRAPHER_USER_ID = cast(int, environ.get("GRAPHER_USER_ID", -1))
```
larsyencken, line 20: Note that I changed this from `USER_ID` to `GRAPHER_USER_ID` when cleaning up. You might update your `.env` file.

PR #27, opened 2021-10-15
data: Import all of open numbers into meadow

larsyencken:
Gapminder publishes its full data catalog in Frictionless format under the Open Numbers banner.

This PR imports their full catalog into meadow, to give us quick coverage and see how the ETL system manages.

The approach taken:

1. Support shallow Github clones as dependencies via `github://`
    - The repos are cloned to `~/.owid/git/` and/or updated when the step is run
2. Find every DDF repo in the `open-numbers` org on Github
3. Add a meadow step for every repo, with a matching github dependency
4. Symlink all meadow Python steps to the `etl/open_numbers.py` module; this means that any time that is edited, all of them are triggered to rebuild
5. In the `etl/open_numbers.py` step, ensure we repack the data frames into more compact versions


larsyencken:
If you want to try running it, run `make etl` and it will start single-threaded. But to run it in parallel, which you absolutely should do if you want it to finish, you can use `xargs`:

```
# do all github clones in 8 parallel processes
grep github dag.yml | grep open-numbers | sed 's/^  - //g' | xargs -n 1 -P 8 .venv/bin/etl

# do all import steps in 8 parallel processes
grep open_numbers dag.yml | grep -v -- ' - ' | sed 's/:$//g' | sed 's/^  //g' | xargs -n 1 -P 8 nice .venv/bin/etl
```

PR #37, opened 2021-10-29
[do-not-merge] Miniconda based environment for M1 macs

larsyencken:
This PR is not meant to be merged, but serves as a temporary branch from which you can do development from on M1 Macs.

Any work that's branched off of this should be rebased onto master and tested there on another computer before being merged.


larsyencken:
It turns out you can get things working without miniconda, burying this branch.

PR #48, opened 2021-11-10
enhance: Speed up github checks for open numbers

larsyencken:
Currently when we need to check if a repo is current, we shell out twice using the `sh` module, once to check the current ref, once to do a dry-run of `git fetch`. Both of these are slow and mean it takes a long time to see when the repo is out of date, once it's built.

The net effect of this is that, once you've built everything, your next `.venv/bin/etl --dry-run` spends several minutes just trying to work out what's out of date.

This change switches to getting the local sha by just reading from disk, and doing a single HTTPS request using Git's smart HTTPS protocol to Github to get the latest remote sha.

## `repo.is_up_to_date()` / 1.9x speedup

Before:

```
In [3]: %timeit git.GithubRepo('open-numbers', 'ddf--ihme--global_burden_disease_etiology').is_up_to_date()
428 ms ± 8.22 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

After:

```
In [6]: %timeit git.GithubRepo('open-numbers', 'ddf--ihme--global_burden_disease_etiology').is_up_to_date()
223 ms ± 15.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## `.venv/bin/etl --dry-run` / 3.5x speedup

On a fully-built ETL, meaning we have c. 100 repositories to check for updates.

```
➜  etl git:(master) time .venv/bin/etl --dry-run
Running 2 steps:
1. github://open-numbers/ddf--undp--hdi...
2. data://meadow/open_numbers/latest/undp__hdi...
.venv/bin/etl --dry-run  7.97s user 7.24s system 9% cpu 2:39.30 total
```

After:

```
➜  etl git:(master) time .venv/bin/etl --dry-run
Running 2 steps:
1. github://open-numbers/ddf--undp--hdi...
2. data://meadow/open_numbers/latest/undp__hdi...
.venv/bin/etl --dry-run  4.53s user 2.35s system 15% cpu 45.378 total
```

### DIFF
--- a/etl/git.py
+++ b/etl/git.py
@@ -9,9 +9,9 @@ Helpers for working with Git in an ETL flow.
 
 from dataclasses import dataclass
 from pathlib import Path
-import re
 from typing import Any, cast
 
+import requests
 import sh
 
 
@@ -72,13 +72,11 @@ class GithubRepo:
 
     @property
     def latest_sha(self) -> str:
-        output = self._git("log", "-n", "1", '--format="%H"')
-        try:
-            (sha,) = re.findall("[0-9a-f]{40}", output)
-        except ValueError:
-            raise GitError(f"could not get latest sha from repo: {self.cache_dir}")
+        master_file = self.cache_dir / ".git/refs/heads/master"
+        with open(master_file, "r") as f:
+            sha = f.read().strip()
 
-        return cast(str, sha)
+        return sha
 
     def _git(self, *args: str, **kwargs: Any) -> str:
         "Execute a git command in the context of this repo."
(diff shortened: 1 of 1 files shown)


lucasrodes (approved):
LGTM

danyx23 (approved):
Looks good. If we want to optimize this further (e.g. because the number of git repos grows) then I think we could switch to using the github graphql api to do not one query per repo but one query per organisation. This would require some changes as the git repos would have to be collected in one go and then queried like this via graphql but it would read to ATM only 2 http requests which would probably be quite a bit faster still - I think :). But nice work to make it faster already!

larsyencken:
Thank you both!

@danyx23 So far I'm trying to avoid the number of credentials you need to work with it, which is why I've avoided using the API directly (that was an earlier implementation). We could totally go back there if it gets problematic. Likewise we could find a way to do these checks async instead of sequentially. But... that's another kettle of fish 😅

PR #49, opened 2021-11-10
feature: Add an interactive command-line country harmonizer

larsyencken:
This PR adds a new command `.venv/bin/harmonize` which can be used to generate a reference country mapping in JSON for a table.

The command takes an existing table (feather or csv) and the column which is meant to represent the country/region field. Then you interactively generate a mapping file from the existing names to OWID canonical names.

It uses the built-in `cmd` module and a command loop when it reaches an ambiguous match. The user can then do one of:

- `n`: to ignore and skip to the next
- `s`: to get suggestions for the match; the user can either select one, enter one manually, or press enter to choose none
- `q`: to quit and save progress so far

A partial mapping can be resumed on a later run. Each time a new alias is discovered, the user has the option to save it back to the reference dataset, making each dataset easier to map than the last.


etl/harmonize.py
```diff
@@ -0,0 +1,260 @@
...
+    with open(output_file, "w") as ostream:
+        json.dump(mapping, ostream, indent=2)
+
+
+def read_table(input_file: str) -> Table:
+    if input_file.endswith(".feather"):
+        return Table.read_feather(input_file)
+
+    if input_file.endswith(".csv"):
+        return Table.read_csv(input_file)
```
danyx23, line 61:
```suggestion
    metadata_filename = Path(input_file).with_suffix(".meta.json")
    if input_file.endswith(".feather"):
        if metadata_filename.exists():
            return Table.read_feather(input_file)
        else:
            return pd.read_feather(input_file)

    if input_file.endswith(".csv"):
        if metadata_filename.exists():
            return Table.read_csv(input_file)
        else:
            return pd.read_csv(input_file)
```

etl/harmonize.py
```diff
@@ -0,0 +1,260 @@
...
+    n: to ignore and skip to the next one
+
+    s: to suggest matches; you can also enter a manual match here
+
+    q: to quit and save the partially complete mapping
+
+    If a mapping file already exists, it will resume where the mapping file left off.
+    """
+    t = read_table(data_file)
+    geo_column = t.reset_index()[column]
```
danyx23, line 42:
```suggestion
    geo_column = t.reset_index()[column].astype("str")
```

danyx23 (commented):
Very nice that you are adding this! A couple of suggestions:
* I think it would be nice if this could be run on a vanilla csv file, before it has been brought even into walden. I think that sometimes you might download a csv file and want to get an idea of what the harmonization process will look like and then this would be neat (see suggestion)
* It'd be nice if it could be a more explicit 2 pass system where first it tries to do all the automatic matches, tell you how many it accomplished and then asks for interactive mappings if there are unmatched entries
* I think the "No match found for ..." message should include a brief "(n) skip, (s) suggest, q (quit)" instruction so you are reminded which keys you can press
* I tried this with a [funky dataset](https://www.accessnow.org/keepiton/) that has lot's of regions we don't have (subregions of India e.g.). This might be out of scope for now but it would be cool if there was a way to collect new geo entities that will be added as new rows to the references dataset. We mainly would need to ask for a value for the code column that is unique, the rest of the columns is more optional - but of course we might want to take more care with extending this than on the fly with this tool. But maybe a list of candidates could be drawn up here?
* The country name should be trimmed of whitespace
* We should make sure that the column is of dtype str - I had a number in it and then trying to sort throws an exception (see suggestion)

I mainly looked at the functionality now, would look more closely if I notice anything about the code in another round if that works for you?

etl/harmonize.py
larsyencken: Ah, good call. Actually, let's just use pandas directly, no need to go via `Table`. It makes it more flexible.

etl/harmonize.py
larsyencken: Did this plus a `dropna()` to avoid stringifying NaN and the like

larsyencken:
@danyx23 Changes since your review:

- Fix for numeric columns
- Fix to allow plain csv and feather files
- Summarised the number of ambiguous cases and focused the text UI only on these
- Helper cue at each step
- More spacing on the text UI, and numbered the cases to give a sense of progress
- Summary report at the end
- Whitespace trimming of manually given names

We should absolutely allow adding new regions or subregions in the future, the text UI is ripe for being extended this way.

danyx23 (approved):
Nice, I like it!

PR #51, opened 2021-11-12
data: Add first draft of food_explorer dataset

lucasrodes:
This PR adds/upgrades several data pipelines:

**Meadow**
- FAOSTAT_QCL (upgrade): `data://meadow/faostat/2021-03-18/faostat_qcl`
- FAOSTAT_FBS (upgrade): `data://meadow/faostat/2021-04-09/faostat_fbs`
- FAOSTAT_FBSH (upgrade): `data://meadow/faostat/2017-12-11/faostat_fbsh`

**Garden**
- FAOSTAT_QCL _new_: `data://garden/faostat/2021-03-18/faostat_qcl`
- FOOD_EXPLORER _new_: `data://garden/explorers/2021/food_explorer`

The current version of the _Food Explorer_ dataset only contains data from _QCL_ dataset. The next iteration should add data from other sources (FBS/FBSH).

I am saving the Food Explorer dataset under namespace _explorers_, happy to adjust.

cc. @MarcelGerber

PR #53, opened 2021-11-17
data: Upgrade food_explorer dataset and add FBSC garden dataset

lucasrodes:
This Pull request builds on the previous PR #51. It adds or upgrades the following pipelines:

**Garden**

- **FAOSTAT_FBSC** _new_: `data://garden/faostat/2021-04-09/faostat_fbsc`. This pipeline combines _FBS_ and _FBSH_ datasets (`data://meadow/faostat/2021-04-09/faostat_fbs` and `data://meadow/faostat/2017-12-11/faostat_fbsh`, respectively) into a single combined dataset.
- **FOOD_EXPLORER**: `data://garden/explorers/2021/food_explorer`. Now this pipeline integrates the data coming from the _FBSC_ garden dataset (previously it only had data from the _QCL_ garden dataset).

Therefore, this version of the _Food Explorer_ dataset contains data on both _QCL_ and _FBSC_ datasets. In the next iteration we should:

- Add population data: This should enable showing _per capita_ versions of current metrics.
- Work on region data (continents, income groups, etc.)

cc. @MarcelGerber


MarcelGerber:
Found a comment that is no longer accurate (in both Notebooks):
```At the moment, only QCL dataset is processed. In next iterations FBS+FBSH dataset will be also imported.```

MarcelGerber:
I didn't look too much at the code itself, but the structure of it is certainly great!
I love how you're telling a story of what you do & what you want to achieve throughout the two notebooks, and doing many sanity checks along the way. Absolutely amazing work in that regard 👏

larsyencken:
@lucasrodes Thanks for your work here! I saw the note about the checksum field in the dataset metadata. Actually, it will be automatically calculated and added after this notebook gets run, so you never need to set it yourself.

See the end of the `run()` method for `DataStep`, where it gets added:

https://github.com/owid/etl/blob/master/etl/steps/__init__.py#L253

PR #54, opened 2021-11-22
Adding the Living Planet Database to meadow

spoonerf:
Have done some basic cleaning and transformed the data from wide to long


larsyencken:
Hey @spoonerf, I rebased this to update it, and fixed a missing parameter. One thing that's confusing is that there's two notebooks in garden, one with 2020 date and one with 2021 date. I think the 2021 might be an accidental duplicate that you could delete.

If you fix that, I think it'll be good to merge!

spoonerf:
Hey @larsyencken I think this is good to go now. I committed `vendor/walden` too but thought I should check with you if this was correct before merging.

PR #57, opened 2021-12-01
data: add World Bank income dataset

lucasrodes:
This PR adds meadow and garden tables with income groups as defined by the World Bank.


larsyencken:
@lucasrodes Hey, I saw just now that this one slipped the net. Is it still something we want to merge?

PR #74, opened 2022-01-31
style(black): format code as pre-commit hook

Marigold:
Fix CI and make sure `make format` runs as a pre-commit hook to avoid these styling bugs in the future. (linting and isort could be added as a pre-commit hook too)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+- repo: local
+  hooks:
+    - id: format
+      name: format
+      entry: make format
+      language: system


larsyencken (approved):
Looks good 👍 

I wonder if there's a bot that can reformat and push a commit to PRs, since that can still help if people haven't set up their pre-commit hook.

Marigold:
> I wonder if there's a bot that can reformat and push a commit to PRs, since that can still help if people haven't set up their pre-commit hook.

For sure there is, but would it be maybe easier to install `pre-commit` as part of `make`? (bot commits could get messy)

Issue #71, opened 2022-01-24
Project: Easy export to grapher from data science catalog


larsyencken:
## Scope

- [x] Chat to data team and work out barriers to adoption, confirm export-to-MySQL is a big one
- [ ] ~~Make an API/process/convention for export to MySQL that makes it simple and predictable~~
- [x] #140
- [x] #141

## Why / why now?

We are one foot into the new data catalog, but it needs this change, and support for embargoed data, to be able to fully replace our old system.

## Who

Mojmir w. @bnjmacdonald 

## Required for future work

API for data, lazy explorers, browseable data catalog, everything explorer, Global Food Explorer, data download options for grapher charts, …more

## Success means

- The data team uses the new data catalog exclusively for all non-embargoed data updates

## Risks

- There may be other, as yet unknown blockers for the data team in using the new catalog; another round of feedback is needed there
- We do not yet know how metadata curation and technical text fit into our model of the data science catalog, they are big upcoming design challenges

PR #79, opened 2022-02-02
Easy export to grapher

Marigold:
First stab at more friendly grapher export. I tried to simplify it by using YAML annotation file with meta description of all variables. It's only a POC, I haven't cleaned it up yet, but it has been tested on two examples - `ghe` and `un_sdg` (which I made up)

Possible issues:
1. variable metadata is already present in Dataset loaded from catalog, so perhaps enumerating them in annotation file is not needed
2. annotation file is too limiting (could be loosened up)

Possible improvements:
1. dataframe validation (e.g. pandera), but maybe we should assume that data loaded from catalog is already clean?
2. [structlog](https://www.structlog.org/en/stable/) logging


larsyencken:
Tagging project: https://github.com/owid/etl/issues/71

Marigold:
I made helper functions more intuitive and created two examples on real datasets from garden. It supports both wide and long format & optional annotations file. I don't think I can make it any simpler for those examples, but they'll certainly evolve with more use cases. Rather than waiting for more examples, I'd try to merge what we have.

larsyencken:
Sounds good, let me know when it's ready for review

Marigold:
Rather than stacking branches on top of this one, it might be easier to just paste the code for `population_density` export to grapher here as a proof of concept that these helper functions are useful and could be merged (after cleaning up those examples).

```
from owid import catalog
from collections.abc import Iterable

from etl import grapher_helpers as gh


def get_grapher_dataset() -> catalog.Dataset:
    return catalog.Dataset(DATA_DIR / "garden/owid/latest/population_density")


def get_grapher_tables(dataset: catalog.Dataset) -> Iterable[catalog.Table]:
    table = dataset["population-density"].reset_index()

    table = (
        table.assign(entity_id=gh.country_to_entity_id(table["country"])).set_index(
            ["entity_id", "year"],
        )
    )[["population_density"]]

    yield from gh.yield_wide_table(table)
```

larsyencken (commented):
This generally looks good. My one comment is that in the example, the `annotations.yml` file is implicitly scoped to the whole dataset (the folder) rather the specific file.

Probably, in our model, a grapher step `my_step.py` might need `my_step.annotations.yml`.

Something out of scope for this PR to think about is a mechanism of grapher export that can tell us which data is out of date in grapher. Ideally we'd propagate the checksum mechanism all the way into MySQL, and use that to detect what parts of grapher are out of date and should be updated. I'll make a new issue for that.

Feel free to merge this change!

Marigold:
> Probably, in our model, a grapher step my_step.py might need my_step.annotations.yml.

Yeah, that's a good idea. I've moved all examples to a different branch anyway, but I'll make sure I get the naming right in the first use case (probably population density).

PR #85, opened 2022-02-14
feat(private_data): enable private datasets

Marigold:
Couple of things still missing:
* [x] unit tests
* [x] accompanying PRs in submodules
* [x] (raise an error if someone uses private step as dependency to a public step)
* [ ] (raise an error if we use `--private` but don't have valid credentials)

Note that I've left there parametrization of `bucket` in publish step. I found it useful for testing.

### Open questions

1. Each repo (`walden`, `etl`, `owid-catalog-py`) has its own credentials and helper functions for S3. What's the best way to DRY them? (e.g. a new package or adding to one of the libraries like `walden`)


etl/command.py
```diff
@@ -89,6 +94,13 @@ def run_dag(
     if not include_grapher:
         excludes.append("grapher://")
 
+    _validate_private_steps(dag)
+
+    if not private:
+        # NOTE: could this possibly go to `excludes`?
+        dag = _filter_private_steps(dag)
+        print("Excluding private steps...")
```
larsyencken, line 102: I think your intuition is right, `private://` could be appended to the excludes list here to get the same behaviour.

etl/publish.py
```diff
@@ -25,53 +25,60 @@ class CannotPublish(Exception):
...
     """
     Publish the generated data catalog to S3.
     """
-    sanity_checks()
-    sync_catalog_to_s3(dry_run=dry_run)
+    catalog = Path(DATA_DIR)
+    if not dry_run:
+        assert (
+            private
+        ), "You cannot publish public catalog yet, only private catalogs with flag --private are supported"
```
larsyencken, line 39: Nit: actually throw an exception here, and put it in an if statement

etl/publish.py
```diff
@@ -25,53 +25,60 @@ class CannotPublish(Exception):
...
+    sync_datasets(s3, bucket, catalog, dry_run=dry_run)
     if not dry_run:
-        update_catalog(s3)
+        update_catalog(s3, bucket, catalog)
 
 
-def is_catalog_up_to_date(s3: Any) -> bool:
-    remote = get_remote_checksum(s3, "catalog.feather")
-    local = files.checksum_file(DATA_DIR / "catalog.feather")
+def is_catalog_up_to_date(s3: Any, bucket: str, catalog: Path) -> bool:
```
larsyencken, line 63: Suggesting a comment that would help future readers

```suggestion
def is_catalog_up_to_date(s3: Any, bucket: str, catalog: Path) -> bool:
    # The catalog file is synced last -- if it is the same as our local one, then all the remote
    # files will be the same as our local ones too.
```

etl/steps/__init__.py
```diff
@@ -574,3 +584,29 @@ def run(self) -> None:
...
+        ds.metadata.is_public = False
+        ds.save()
+
+    def _run_py(self) -> None:
+        super()._run_py()
+        self._make_dataset_private(self._dest_dir.as_posix())
+
+    def _run_notebook(self) -> None:
+        super()._run_notebook()
+        self._make_dataset_private(self._dest_dir.as_posix())
```
larsyencken, line 604: Not important now, but if we added more `_run` methods, we'd probably just want an `after_run()` hook you could define in the subclass.

larsyencken (approved):
Overall it's good, only minor feedback. :shipit: 

On the PR itself, a little more description up front could make it much faster to review :) In this case, a description of the approach was missing from the PR. Something like:

> Introduces private versions of all existing steps (e.g. `data-private://...`, `walden-private://...`) that remember and enforce a dataset level `is_public` flag.
> 
> When publishing, the index is public, but tables in the index that are private are only available over s3 with appropriate credentials. `owid-catalog-py` is also updated to support these private datasets and supports fetching them over s3 transparently.

I'm gonna keep nudging you on it 😄

etl/steps/__init__.py
Marigold: That's much better, fixed it

Marigold:
> I'm gonna keep nudging you on it 😄

Please do, my writing skills have deteriorated since I dropped PhD... I've at least updated README.

Addressed all comments and also fixed `examples/`. Honestly, these are a bit weird right now with all the envs, but I didn't want to include them in the main `dag.yml` file. They work though and we can fix them later.

PR #88, opened 2022-02-16
style(mypy): disable redundant casts

Marigold:
Redundant casts have been an issue for a few people already, it's unclear what's causing it but you either get error about `redundant cast` or returning `Any`. Might be easier to be less strict about it and temporarily disable it.


Makefile
```diff
@@ -45,7 +45,9 @@ watch: .venv
 
 check-typing: .venv
 	@echo '==> Checking types'
-	.venv/bin/mypy --strict $(SRC)
+	# redundant cast is temporarily disabled because it's not consistent
+	# feel free to enable it with future versions of mypy
+	.venv/bin/mypy --strict --no-warn-redundant-casts $(SRC)
```
larsyencken, line 50: I think this is better in the `.mypy.ini` file, both the `--strict` and the `--no-warn-redundant-casts` flag.

Marigold: Eh, obviously. Fixed, thanks.

larsyencken:
Perfect! :shipit:

PR #90, opened 2022-02-17
Population data is out of sync with importers

Marigold:


dag.yml
```diff
@@ -34,10 +42,13 @@ steps:
     - data://meadow/faostat/2022-02-10/faostat_metadata
   data://garden/faostat/2021-06-17/faostat_rl:
     - data://meadow/faostat/2021-06-17/faostat_rl
-  data://garden/explorers/2021/food_explorer:
-    - data://garden/faostat/2021-03-18/faostat_qcl
-    - data://garden/faostat/2021-04-09/faostat_fbsc
-    - data://garden/owid/latest/key_indicators
+  # TODO: uncomment!!! this is commented only to avoid running this step when I run command
```
Marigold, line 45: Is there a reason why pattern matching is applied to dependencies as well? This is what I get
```
(.venv) ➜  etl git:(enhance-wpp-population) etl data://garden/owid/latest/key_indicators --dry-run
Running 3 steps:
1. data://garden/owid/latest/key_indicators...
2. data://garden/owid/latest/population_density...
3. data://garden/explorers/2021/food_explorer...
```
there's no easy way how to run only `data://garden/owid/latest/key_indicators` step which makes it a bit hard to debug

dag.yml
```diff
@@ -1,10 +1,18 @@
 steps:
   data://garden/owid/latest/covid:
     - etag://raw.githubusercontent.com/owid/covid-19-data/master/public/data/owid-covid-data.csv
+  data://garden/owid/latest/population_density:
+    - data://garden/owid/latest/key_indicators
```
Marigold, line 5: Would it make sense to move `population_density` to `key_indicators` (possibly together with `land_area`)?

dag.yml
```diff
@@ -1,10 +1,18 @@
...
+    - data://meadow/open_numbers/latest/open_numbers__world_development_indicators
+    - data://reference
   data://garden/owid/latest/key_indicators:
     - data://meadow/hyde/2017/baseline
     - data://meadow/gapminder/2019-12-10/population
+    - data://meadow/wpp/2019/standard_projections
     - data://reference
+  # TODO: we probably don't need this as we harmonize countries in key_indicators
```
Marigold, line 13: Do we want to do this intermediate garden step? (I'd say so, but I wanted to make sure we're getting the same outputs first)

etl/steps/data/garden/owid/latest/key_indicators/table_population.py
```diff
@@ -33,32 +36,116 @@ def make_table() -> Table:
...
+        "Share of world population": "world_pop_share",
+    }
+    # transform it into new shape
+    old_df = (
+        pd.read_csv(DIR_PATH / "output" / "Population (Gapminder, HYDE & UN).csv")
+        .rename(columns=COLS_MAP)[COLS_MAP.values()]
+        .set_index(["country", "year"])
+    )
+
+    # TODO: South America is not yet ready, fix this after https://github.com/owid/etl/issues/92
```
Marigold, line 76: South America & Latina America is blocked by https://github.com/owid/etl/issues/92.

etl/steps/data/garden/wpp/2019/standard_projections.ipynb
```diff
@@ -0,0 +1,185 @@
+{
```
Marigold, line 1: WIP, might be just deleted or moved to a .py file.

Marigold:
Couple of notes / questions:
1. In order to be consistent with [continents definition in importers](https://github.com/owid/importers/blob/master/population/input/continents-according-to-our-world-in-data.csv) I had to move Turkey and Cyprus to Asia in `countries_regions.csv` file. This is just a temporary solution (or not?) before we figure out how `countries_regions.csv` is built
2. Population density and land area are built separately, but I think they could easily be part of `key_indicators` table. What do you think? (We can always merge them later)
3. There are some validation functions to make sure we get similar numbers as in importers, these will be removed before merging together with `input` and `output` folders
4. (Population density is already in world bank dataset, but I assume calculating it ourselves is more robust)

larsyencken:
Responding to your questions:

**On countries and regions:** we want to be the same as our variable 123 in Grapher. If we check the chart preview there, we see that Turkey is in Asia but Cyprus is in Europe (ignoring that half of Cyprus is considered illegally occupied by Turkey). https://owid.cloud/admin/variables/123

@edomt please pipe up if you want to make a different data-editorial decision, though if you do then I think we'd want to change the grapher 123 variable as well, that gets used in a ton of visualisations.

**On `key_indicators`:** we could totally just keep them in the key indicators module. At the moment the `__init__.py` file in that module looks for every `table_*.py` file and runs it. Perhaps we could just do something simpler and make a manual list, then we can put population density to run after population.

**On recalculating this ourselves:** since we're calculating population ourselves, and going back far in history, then we also want to be calculating our own population density that matches it

dag.yml
```diff
@@ -1,10 +1,19 @@
...
+    - data://garden/owid/latest/key_indicators
+    - data://meadow/open_numbers/latest/open_numbers__world_development_indicators
+    - data://reference
   data://garden/owid/latest/key_indicators:
     - data://meadow/hyde/2017/baseline
     - data://meadow/gapminder/2019-12-10/population
+    # NOTE: it would be better to use /garden/wpp/2019/standard_projections, but it uses different harmonization than gapminder and hyde.
+    # in the future we should build `key_indicators` only from garden datasets
+    - data://meadow/wpp/2019/standard_projections
     - data://reference
```
larsyencken, line 14: I realised now that all three of these tables would ideally have been harmonized into Garden before being combined here. I'll have a quick go at Hyde baseline now.

Marigold:
@larsyencken I've addressed your feedback and it's ready for review again. It depends on PR for proper [definition of continents](https://github.com/owid/etl/pull/105), but otherwise is good to go. (`inputs`, `outputs` and other validation functions will go away before merging)

etl/steps/data/garden/owid/latest/key_indicators/table_population.py
```diff
@@ -9,97 +9,103 @@
...
+        -(
+            (df.country.isin(has_gapminder_data))
+            & (df.year >= 1800)
+            & (df.year <= 1949)
+            & (df.source != "gapminder")
+        )
+    ]
+
+    # Test if all countries have only one row per year
+    _assert_unique(df, subset=["country", "year"])
```
larsyencken, line 67: These kinds of basic checks are 👌

larsyencken (approved):
Generally looks good, some tiny suggestions then ship it! :shipit: 

If you could rebase away the validation files when you're done, that'd be awesome, so that they never appear in the git history.

After this, I think we should start adding metadata to the tables.

etl/data_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+        .rename(columns={"continent": "country"})
+    )
+
+    world = (
+        df.drop(["country"], axis=1)
+        .groupby("year")
+        .sum()
+        .reset_index()
+        .assign(country="World")
+    )
```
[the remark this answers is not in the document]
Marigold: Eh, totally missed that. Filtering only to countries with existing continents now. Added a test to be sure.

Marigold:
Fixed everything, thanks for feedback.

> After this, I think we should start adding metadata to the tables.

I'm going to add metadata in a new PR, this one is already a mess.

PR #93, opened 2022-02-21
WIP: Data/population density grapher

Marigold:


Marigold:
@larsyencken would you mind doing a pre-review for population density, please? I'm not clear on some open questions so it's not polished yet. Your opinion would be much appreciated before I move forward.

(grapher part is in a separate PR because it depends on https://github.com/owid/etl/pull/79)

PR #96, opened 2022-02-23
fix: include walden metadata in data freshness check

### LINKED ISSUE #95: Include checksum of walden JSON file in source checksum
## Problem

When the ETL decides whether a dataset is up to date, it bases that decision off of a `source_checksum` of a dependency. For Walden snapshots, we use the `md5` field of the JSON file.

However, if you build everything, then change metadata in the Walden JSON file, it will not be detected and the data will not rebuild

## Expected behaviour

Any change to the metadata in a Walden file should trigger a rebuild of any downstream datasets.

## Proposes solution

Checksum the JSON file itself as a dependency, not just the data file it references.

larsyencken:
Resolves #95.

When you decide if data from Walden has changed, consider not only the data file but the metadata itself. This way, if you change the dataset name or description, it will ripple through the ETL and be copied as necessary to downstream datasets.

### DIFF
--- a/etl/steps/__init__.py
+++ b/etl/steps/__init__.py
@@ -403,10 +403,6 @@ class WaldenStep(Step):
 
         checksum = hashlib.md5(",".join(inputs).encode("utf8")).hexdigest()
 
-        if not checksum:
-            raise Exception(
-                f"no md5 checksum available for walden dataset: {self.path}"
-            )
         return checksum
 
     @property
--- a/etl/steps/__init__.py
+++ b/etl/steps/__init__.py
@@ -391,8 +391,13 @@ class WaldenStep(Step):
         return True
 
     def checksum_output(self) -> str:
-        inputs: List[str] = [
+        if not self._walden_dataset.md5:
+            raise Exception(f"walden dataset is missing checksum: {self}")
+
+        inputs = [
+            # the contents of the dataset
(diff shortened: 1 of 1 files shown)


larsyencken:
@pabloarosado This is the issue we came across today. It's only a small thing, but since our current pattern is to copy metadata from dataset to dataset across transformations, it's important to recalculate them if the metadata changes.

At the last step before Grapher, you get a chance to hand-write new metadata and get it perfect, so as mentioned it's still ok to just do best-effort metadata here.

etl/steps/__init__.py
```diff
@@ -391,7 +391,18 @@ def has_existing_data(self) -> bool:
...
+        inputs = [
+            # the contents of the dataset
+            self._walden_dataset.md5,
+            # the metadata describing the dataset
+            files.checksum_file(self._walden_dataset.index_path),
+        ]
+
+        checksum = hashlib.md5(",".join(inputs).encode("utf8")).hexdigest()
+
         if not checksum:
```
Marigold, line 406: Is this part needed anymore?

Marigold (approved):
looks good!

etl/steps/__init__.py
larsyencken: Good catch, deleted it now!

PR #99, opened 2022-02-25
data(hyde): harmonize Hyde baseline into garden

larsyencken:
The Hyde baseline is used for population, but only gets harmonized at the last minute.

This change introduces a harmonized version to Garden directly.


etl/steps/data/garden/hyde/2017/baseline.mapping.json
```diff
@@ -0,0 +1,238 @@
...
+  "Svalbard & Jan Mayen Isl.": "Svalbard and Jan Mayen",
+  "Syrian Arab Rep.": "Syria",
+  "TFYR Macedonia": "North Macedonia",
+  "Trin. & Tobago": "Trinidad and Tobago",
+  "Turks & Caicos": "Turks and Caicos Islands",
+  "U.Arab Emirates": "United Arab Emirates",
+  "U.R. Tanzania": "Tanzania",
+  "U.S. Virgin Is.": "United States Virgin Islands",
+  "Wallis & Futuna": "Wallis and Futuna",
+  "Zaire/ Congo , Dem. Rep.": "Democratic Republic of Congo"
```
larsyencken, line 237: Everything after Zimbabwe in this file is manually mapped and should get review.

Marigold: Looks good (despite my poor geographic knowledge).

etl/steps/data/garden/hyde/2017/baseline.ipynb
```diff
@@ -0,0 +1,224 @@
+{
```
Marigold, line 1: Would it make sense to include assertion that we mapped every country on line
```
population["country"] = population.country.apply(lambda c: country_mapping[c])
```
(i.e. adding check for null values)? No need to do it right now, might be useful to have a helper function for that.

etl/scripts/countries-regions/06-add-continents.ipynb
```diff
@@ -13,7 +13,7 @@
   },
```
Marigold, line 13: Are there any actual changes to this notebook or were you just checking it out and then it got cleaned with `jupyter-notebook-cleanup`? When I compare scripts vs notebooks, a big advantage of notebooks is committing outputs into the repository (where I can easily check what does it do without having to rerun it myself). Without those outputs, I'd rather use python scripts (with the advantage of being easily reviewable and more straightforward to debug). What do you think about making `jupyter-notebook-cleanup` optional?

Marigold (approved):

etl/steps/data/garden/hyde/2017/baseline.ipynb
larsyencken: Ideally, we'd have a one-liner for the entire country-mapping-only harmonisation. But we can have a few ones like this that are simple for now.

etl/scripts/countries-regions/06-add-continents.ipynb
larsyencken: I removed the code `OWID_KOR` for Korea since there is another one for "United Korea" that we want to prefer to use instead.

PR #105, opened 2022-02-28
data(countries_regions): accurate continents

Marigold:
Changes to `countries_regions.csv` to reflect what we have in our production grapher database

* rename Latin America to South America
* move Turkey to Asia and Cyprus to Europe
* add notebook `etl/scripts/countries-regions/validate_continents.ipynb` to check consistency

I didn't make any changes to `etl/scripts/countries-regions/*.ipynb` notebooks except of replacing `%%mysql` magic with `pd.read_sql` (I couldn't find what library is used there)


danyx23 (commented):
Question for understanding - the validate_contintents notebook checks if there are differences between the continent assignment that is in contries_regions.csv and the database and the basic outcome is the list in the second to last cell - correct?

And then you manually changed certain entries in countries_regions.csv? As I see it, now all differences that came up in validate_continents were addressed but also some additional things, right?

If that is true then it would be good to record this somewhere (in the notebook?). 

As a minor nit: I think short variable names like q or df that are very common are fine, but cf and gf I found confusing. With autocomplete in editors etc I think for such cases it is much nicer to the reader to use longer variable names.

Marigold:
My bad, updated markdown cells in the notebook with answers to your questions.

> And then you manually changed certain entries in countries_regions.csv? As I see it, now all differences that came up in validate_continents were addressed but also some additional things, right?

I was hesitant to make large changes at first, but have now decided to make `countries_regions.csv` exactly like variable 123. So I moved a few other mini-regions to proper continents. There are no differences now.

> As a minor nit: I think short variable names like q or df that are very common are fine, but cf and gf I found confusing. With autocomplete in editors etc I think for such cases it is much nicer to the reader to use longer variable names.

You're right, fixed that. (I was prototyping a lot recently and was naming every dataframe variable `[a-z]f`... need to get rid of that habit)

danyx23:
@Marigold if you push your changes I'll take a final look, sounds all great to me!

Marigold:
@danyx23 just pushed, writing that commit message took a long time :).

larsyencken (approved):
Double-checked the changes now, they look good to me too! :shipit:

larsyencken:
The Jupyter MySQL magic that was used in the earlier script is this one: https://gist.github.com/larsyencken/935452e59246a58c647dd3953de2ab07

PR #109, opened 2022-03-02
fix(food_explorer): remove continents which are already in the population table

Marigold:
New version of population table already contains continents, it is not necessary to compute them again in food explorer.


larsyencken:
Hey! Mid-review on this, I noticed that the data changed, but that doesn't mean the change was wrong. I'm going to QA it properly tomorrow and if the changes are all expected ones we can merge it.

Marigold:
> I'm going to QA it properly tomorrow and if the changes are all expected ones we can merge it.

I haven't compared the actual outputs before & after. I assume this is not the last time we do something like this, any ideas what should be the process here? E.g. adding validation notebook alongside data

Marigold:
(Forgot to mention that I've validated old and new food explorer dataset and all differences are according to our expectations from moving Turkey to Asia.)

PR #112, opened 2022-03-04
enhance(population): add metadata to population, land area and population density

Marigold:
I wrote a simple script for "backporting" metadata from grapher to ETL structure. First I find a chart like https://ourworldindata.org/grapher/population-density and then by its slug find variable, dataset and source. Then I structure them to fit `etl` Dataset - Table - Variable framework.

As you can see, grapher is not very consistent and for instance sources are all over the place. Couple of questions / ideas:

1. `Table` does not have `sources` field, but to me it seems like a better place than keeping sources on variable (maybe we could have sources for all `Dataset` - `Table` - `Variable` and always choose a common denominator)
2. Perhaps `annotations.yaml` would be a better place for metadata? [like this one](https://github.com/owid/etl/blob/master/etl/steps/grapher/who/2021-07-01/annotations.yml). What do you think?
3. When you're changing e.g. description, it's much easier to just update it in grapher than creating a commit (and possibly even a PR). Are we sure users would be willing to do that? Is there an alternative? (or maybe this is not a problem at all?)

cc @larsyencken just in case you have opinion on these. If not I can follow my judgement and make ready for real review.

### DIFF
--- a/etl/steps/data/garden/owid/latest/key_indicators/table_population.py
+++ b/etl/steps/data/garden/owid/latest/key_indicators/table_population.py
@@ -14,7 +14,7 @@ from typing import cast, List
 
 import pandas as pd
 
-from owid.catalog import Dataset, Table
+from owid.catalog import Dataset, Table, Source
 from etl.paths import DATA_DIR
 from etl import data_helpers
 
@@ -23,8 +23,6 @@ GAPMINDER = DATA_DIR / "garden/gapminder/2019-12-10/population"
 HYDE = DATA_DIR / "garden/hyde/2017/baseline"
 WB_INCOME = DATA_DIR / "garden/wb/2021-07-01/wb_income"
 REFERENCE = DATA_DIR / "reference"
-
-COUNTRY_MAPPING = Path(__file__).with_suffix(".mapping.csv")
 
 DIR_PATH = Path(__file__).parent
 
@@ -38,7 +36,35 @@ def make_table() -> Table:
         .pipe(prepare_dataset)
     )
 
+    sources = [
+        Source(
+            name="Gapminder (v6)",
+            url="https://www.gapminder.org/data/documentation/gd003/",
+            date_accessed="October 8, 2021",
+        ),
+        Source(
+            name="UN (2019)",
+            url="https://population.un.org/wpp/Download/Standard/Population/",
+            date_accessed="October 8, 2021",
+        ),
+        Source(
+            name="HYDE (v3.2)",
+            url="https://dataportaal.pbl.nl/downloads/HYDE/",
+            date_accessed="October 8, 2021",
+        ),
+    ]
+
+    # table metadata
     t.metadata.short_name = "population"
+    t.metadata.title = "Population (Gapminder, HYDE & UN)"
+    t.metadata.description = 'Our World in Data builds and maintains a long-run dataset on population by country, region, and for the world, based on three key sources: HYDE, Gapminder, and the UN World Population Prospects. You can find more information on these sources and how our time series is constructed on this page: <a href="https://ourworldindata.org/population-sources">What sources do we rely on for population estimates?</a>'
+    t.metadata.sources = sources
+
+    # variables metadata (variable 72 in grapher)
+    t.population.metadata.title = "Population"
+    t.population.metadata.description = "Population by country, available from 1800 to 2021 based on Gapminder data, HYDE, and UN Population Division (2019) estimates."
+    t.population.metadata.display = {"name": "Population", "includeInTable": True}
+
     return t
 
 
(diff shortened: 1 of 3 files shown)


etl/steps/data/garden/owid/latest/key_indicators/table_population.py
```diff
@@ -36,7 +34,23 @@ def make_table() -> Table:
         .pipe(prepare_dataset)
     )
 
+    source = Source(
+        name="Gapminder (v6), HYDE (v3.2), UN (2019)",
+        description='Our World in Data builds and maintains a long-run dataset on population by country, region, and for the world, based on three key sources: HYDE, Gapminder, and the UN World Population Prospects. You can find more information on these sources and how our time series is constructed on this page: <a href="https://ourworldindata.org/population-sources">What sources do we rely on for population estimates?</a>',
+        url="https://www.gapminder.org/data/documentation/gd003/; https://population.un.org/wpp/Download/Standard/Population/; https://dataportaal.pbl.nl/downloads/HYDE/",
+        date_accessed="October 8, 2021",
+    )
```
larsyencken, line 42: Hmm, interesting. So, in the ETL model, you can have multiple sources, with the idea that as you use different sources you concatenate them in the metadata, ending up with a list of all the sources you actually used. This is better than grapher's model. But it's also worse in that it's missing a single, overall `source_summary` that explains how the different sources were used.

I think this is fine for a backported variable, but we can probably aim for better for data that originates from the ETL.

etl/steps/data/garden/owid/latest/population_density/__init__.py
```diff
@@ -53,21 +53,50 @@ def make_table() -> Table:
...
+
+    source = Source(
+        name="World Bank (2021), Gapminder (v6), HYDE (v3.2), UN (2019)",
+        description='Our World in Data builds and maintains a long-run dataset on population by country, region, and for the world, based on three key sources: HYDE, Gapminder, and the UN World Population Prospects. This combines historical population estimates with median scenario projections to 2100. You can find more information on these sources and how our time series is constructed on this page: <a href="https://ourworldindata.org/population-sources">What sources do we rely on for population estimates?</a>\n\nWe combine this population dataset with the <a href="https://ourworldindata.org/grapher/land-area-km">land area estimates published by the World Bank</a>, to produce a long-run dataset of population density.\n\nIn all sources that we rely on, population estimates and land area estimates are based on today’s geographical borders.',
+        url="https://www.gapminder.org/data/documentation/gd003/; https://population.un.org/wpp/Download/Standard/Population/; https://dataportaal.pbl.nl/downloads/HYDE/; https://datacatalog.worldbank.org/dataset/world-development-indicators",
```
larsyencken, line 68: Likewise here, if it was coming from the ETL we'd want these data sources separately.

larsyencken (commented):
I tend to agree that the metadata might be easier living in a YAML file, though it would be ideal to have it in garden, and not only at the last minute as it gets inserted to grapher.

Perhaps we can solve the "forwards direction" (ETL -> grapher) and the "backwards direction" separately.

The "backwards direction" can be approximate.

The "forwards direction" would be best solved by getting a single dataset we care about upserting to grapher, then experimenting with the metadata mapping until we can see that the upsert gives data managers all the control they need to be precise with the metadata.

larsyencken:
> * `Table` does not have `sources` field, but to me it seems like a better place than keeping sources on variable (maybe we could have sources for all `Dataset` - `Table` - `Variable` and always choose a common denominator)

When writing the metadata model, I didn't find all this very easy to reason through. So it could certainly be that it could be better than how it is.

The motivation for it being at the variable level is that a table might share a `x` and a `x_per_capita` metric, where the second needs more sources than the first. If we do it per table, then we'll would need to do one table per variable in order to not screw up the metadata for grapher. Or, we would need to ignore all the metadata and just rewrite it for grapher, which doesn't feel ideal.

etl/steps/data/garden/owid/latest/key_indicators/table_population.py
Marigold: Oh yeah, list of sources is better for sure. (I thought I left a comment about it somewhere, but apparently not). Fixed.

Marigold:
> I tend to agree that the metadata might be easier living in a YAML file

I _tend to agree_ too, that probably means we don't use it now and postpone the decision :).

> though it would be ideal to have it in garden, and not only at the last minute as it gets inserted to grapher.

👍 

> When writing the metadata model, I didn't find all this very easy to reason through. So it could certainly be that it could be better than how it is.

Thanks for explanation, keeping it on variable level then.

> The "forwards direction" would be best solved by getting a single dataset we care about upserting to grapher, then experimenting with the metadata mapping until we can see that the upsert gives data managers all the control they need to be precise with the metadata.

I think we'll understand it more after we figure out the backporting. Then we can solve the "forwards direction" which is gonna be trickier.

Do you think that this is good to go after fixing your comments? Metadata to grapher & YAML annotations are out of scope of this in my opinion.

Issue #113, opened 2022-03-09
data(sdg): create OWID-curated SDG dataset in `etl`


Marigold:
## Problem

Our SDG Tracker has interesting data that could be reused by others, but we don't have them in a catalog (or other format easily consumable by users).

## Alternatives

SDG metrics are already available [from UN SDG API](https://unstats.un.org/sdgapi/swagger/). Our SDG tracker mostly shows metrics from that API, but in some cases like [Goal 1](https://sdg-tracker.org/no-poverty) our charts often provide longer time-series than the data provided by UN SDG.

Open numbers also [republish UN SDG](https://github.com/open-numbers/ddf--unstats--sdg_indicators), but the data is not curated and hasn't been refreshed for a long time.

## Proposed solution (simple)

Move original UN SDG dataset from https://github.com/owid/importers/tree/master/un_sdg to `etl` and publish it in our catalog. That would enable others to download it easily, but wouldn't be consistent to what we show in https://sdg-tracker.org/. Are we providing any value over UN SDG API by this though?

## Proposed solution (complex)

Create dataset that would be an identical copy of https://sdg-tracker.org/. That means getting together all kinds of data from UN SDG, WHI and other sources and mapping them to SDG Goals (according to [this table](https://docs.google.com/spreadsheets/d/1n0UrpKKS2JVcXSmth_QVYLThlWev6pzEiHP7HaIq9BY/edit#gid=0)). There are more ways how to get all those variables into `etl`:

1. Backport those variables from grapher
2. Migrate UN SDG, WHI and all other sources to `etl` (the "correct" way)
3. ?

Perhaps the best approach would be to start by migrating UN SDG from importers to `etl` and then incrementally replacing the most important variables by our own versions?

cc @spoonerf @larsyencken

Marigold:
Another alternative would be simply backporting [UN SDG from importers](https://github.com/owid/importers/tree/master/un_sdg) into `etl` catalog.

spoonerf:
I think the solution we want would (unfortunately!) be the complex one. I'm not sure how big of a project the back-porting of variables to the grapher is but perhaps it is worth waiting until that happens to do this?

One thing that would be really useful to be able to do within `etl` is to be able to mix variables from different datasets together to create our own 'OWID curated' datasets. I think this project is perhaps an extreme version of that! 

Maybe there's an intermediate step where we just have a step for collecting the variables for one SDG goal?

Marigold:
[...]

PR #114, opened 2022-03-14
WIP: feature(backport): grapher backport to etl

Marigold:
This is the first bare POC for backporting datasets from grapher to ETL. It backports WDI World Bank dataset which contains variables needed for https://github.com/owid/etl/issues/113.

The idea is to have DAG like this
```
data://meadow/grapher/latest/wdi_world_bank:
    - grapher-dataset://World Development Indicators - World Bank (2021.07.30)
```
which checks latest `updateAt` of a dataset (or variables) to decide whether to refresh the dataset or not. `data://` step then pulls the data from grapher and create a regular dataset. We'll need special flag similar to `--grapher` in `etl` to avoid running those steps if we don't have access to DB.

WDI World Bank is backported by iterating over topic codes from their definitions `WDI_CETS.xls` and pulling all the variables from DB. Then we switch from long-format to wide-format and save them as a table into dataset.


Marigold:
Implemented here https://github.com/owid/etl/pull/155

PR #116, opened 2022-03-15
data(countries_regions): update country mappings

pabloarosado:


larsyencken (approved):
They all look good :shipit:

PR #120, opened 2022-03-16
style(harmonize): correct typos and change displayed messages

pabloarosado:
I changed the displayed message in `harmonize`, since I found "(n) skip (s) suggest" sometimes a bit confusing.


larsyencken (approved):
Agreed, that's better!

PR #124, opened 2022-03-17
Add GGDC Maddison Project Database 2020 data to garden

pabloarosado:


etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+DATASET_NAME = "ggdc_maddison_project_database"
+# Original dataset publication date.
+VERSION = "2020-10-01"
+# Column name for GDP in output dataset.
+GDP_COLUMN = "GDP"
+# Column name for GDP per capita in output dataset.
+GDP_PER_CAPITA_COLUMN = "GDP per capita"
+
+
+def load_countries() -> Dict[str, str]:
```
Marigold, line 35: This would make a nice helper function, how about moving it to https://github.com/owid/etl/blob/master/etl/data_helpers.py so that I can reuse it for other projects? :)
It could have path as an input parameter (so that one could use `Path(__file__).parent` instead of specifying the full path)

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+        errors="raise",
+    )[["Country", "Year", "Population", GDP_PER_CAPITA_COLUMN]]
+    # Convert units.
+    data["Population"] = data["Population"] * 1000
+    # Create column for GDP.
+    data[GDP_COLUMN] = data[GDP_PER_CAPITA_COLUMN] * data["Population"]
+
+    # Standardize country names.
+    countries = load_countries()
+    data["Country"] = data["Country"].replace(countries)
```
Marigold, line 88: Do we need some sanity checks for unmapped / null countries or is it not worth it?

Marigold (approved):
Minor nitpicks, feel free to merge as you like

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+    ds.metadata = convert_walden_metadata(walden_ds)
+
+    # Load and process data.
+    df = generate_ggdc_data(data_file=walden_ds.local_path)
+
+    # Create a new table with the processed data.
+    t = Table(df)
+
+    # Assign the same metadata of the walden dataset to this table.
+    t.metadata.short_name = DATASET_NAME
```
larsyencken, line 237: I don't think this is a good naming for this table. Firstly, you might import other data from this database, in which case you will have one dataset but multiple tables. Secondly, it doesn't describe what's in the table, meaning it's not very discoverable by search.

I'd suggest something either `gdp` or `maddison_gdp` for the table short name instead.

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+    t.metadata.title = ds.metadata.title
+    t.metadata.description = ds.metadata.description
+
```
larsyencken, line 240: I'd also set the index here, since that's an important signal as to how the data is arranged.

```suggestion
    t.set_index(['country', 'year'])
    
```

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+        .dropna(how="all", subset=[GDP_COLUMN, GDP_PER_CAPITA_COLUMN])
+        .sort_values(["Country", "Year"])
+        .reset_index(drop=True)
+    )
+
+    # Some rows have spurious zero GDP. Remove them.
+    zero_gdp_rows = combined[GDP_COLUMN] == 0
+    if zero_gdp_rows.any():
+        print(f"WARNING: Removing {zero_gdp_rows.sum()} rows with zero GDP.")
+        combined = combined[~zero_gdp_rows].reset_index(drop=True)
```
larsyencken, line 213: A warning suggests that it's unexpected. Better to QA them and explicitly remove them, and only warn if unexpected zeros are there. Otherwise everyone who builds this data will wonder if something's gone wrong.

larsyencken (changes requested):
Some stylistic feedback, but the data transformations themselves look ok.

I haven't QA'd the data itself.

larsyencken:
> I think the build is failing because it doesn't find the dataset in walden. So after we merge [this other PR](https://github.com/owid/walden/pull/17) I can restart the build and hopefully it will work.

You need to rebase this branch on top of master, then it will pick up the Walden change.

dag.yml
```diff
@@ -70,7 +70,8 @@ steps:
     - walden://wb/2021-07-01/wb_income
   data://garden/wb/2021-07-01/wb_income:
     - data://meadow/wb/2021-07-01/wb_income
-
+  data://garden/ggdc/2020-10-01/ggdc_maddison_project_database:
+    - walden://ggdc/2020-10-01/ggdc_maddison_project_database
```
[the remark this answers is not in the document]
pabloarosado: Hi @larsyencken I was trying to follow the [official name](https://www.rug.nl/ggdc/historicaldevelopment/maddison/releases/maddison-project-database-2020) of the dataset. But I suppose you are right, we can reduce it to `ggdc_maddison` to make our lives easier.

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+    countries : dict
+        Country mappings.
+
+    """
+    # Define path to countries file.
+    garden_dir = STEP_DIR / "data" / "garden" / NAMESPACE / VERSION
+    files = list(garden_dir.glob("*countries.json"))
+    # Ensure there is only one countries file.
+    assert len(files) == 1
+    countries_file = files[0]
```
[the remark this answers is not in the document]
pabloarosado: Nice one, I'll add this in the next commit.

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
pabloarosado: Hi @Marigold for the time being, I've just ensured that all countries are included in the countries.json file.

dag.yml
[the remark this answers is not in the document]
larsyencken: The official name is most important in the user-visible metadata, the short name can be something easier to use and more discoverable for data analysts.

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
pabloarosado: For now I'm going to keep it here, but feel free to move the function to data_helpers if you think that's appropriate (but let's leave it for a different branch).

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
```diff
@@ -0,0 +1,245 @@
...
+    # Convert units.
+    data["Population"] = data["Population"] * 1000
+    # Create column for GDP.
+    data[GDP_COLUMN] = data[GDP_PER_CAPITA_COLUMN] * data["Population"]
+
+    # Standardize country names.
+    countries = load_countries()
+    data["Country"] = data["Country"].replace(countries)
+
+    return cast(pd.DataFrame, data)
```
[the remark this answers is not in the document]
pabloarosado: I was trying to make everything look as close as possible to the final dataset (to be uploaded in grapher). But in my latest commit I have adopted snake_case. I suppose between this and uploading to grapher there would need to be an intermediate step of renaming columns.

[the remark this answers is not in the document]
larsyencken: There's no conflict here. What you see in grapher should be the `title` of these variables, not their `short_name`.

Tables support variable level metadata, so once you've done `t = Table(df)` you can say `t.gdp_per_capita.metadata.title = 'GDP per capita'`, for example.

An early idea we have is that the short name is an internal identifier for a variable, as a new alternative to using an integer database id every time. We haven't taken that idea very far yet though.

etl/steps/data/garden/ggdc/2020-10-01/ggdc_maddison_project_database.py
larsyencken: (I notice now that `population` is there as well. Given that there's some diversity of variables in the same table, it's probably fine to use `ggdc_maddison` as the table name for now.)

larsyencken (approved):

PR #126, opened 2022-03-18
feature(examples): add examples for easier head start

Marigold:
Add a couple of examples to `dag.yml` to have starting templates and ad-hoc prototyping. I created them in a way I found useful for my own work.

It moves `examples` from the top level directory to other steps (there are some minor updates). This simplified the way examples were run and let them use other datasets from `etl`. I'm gonna be updating these examples on the go (e.g. for grapher or backporting).


larsyencken (changes requested):
I'd suggest having the `Catalog().find()` examples use the local catalog instead of the remote one. Other than that, it looks good.

Marigold:
> I'd suggest having the `Catalog().find()` examples use the local catalog instead of the remote one. Other than that, it looks good.

I don't get it, `Catalog().find()` is **walden's catalog**, not owid-catalog-py's, and uses local catalog. Or maybe I misunderstood this completely?

larsyencken:
> I don't get it, `Catalog().find()` is **walden's catalog**, not owid-catalog-py's, and uses local catalog. `LocalCatalog/RemoteCatalog` is in owid-catalog-py and I'm not using it at all (instead I'm using `Dataset(...)` to get local version). Or maybe I misunderstood this completely?

You're totally right, I just got confused. This is good to ship then!

(one day we should rename Walden's API to make it less confusing. Perhaps `Catalog -> Walden` and `Dataset -> Snapshot`?)

larsyencken (approved):

Issue #111, opened 2022-03-03
data: multiple datasets do not follow snake_case naming standards


larsyencken:
As a convention for making data very easy to work with, we want datasets, tables and columns to use `snake_case` variable names.

I'd propose we clean them up, and then modify the `Table` and `Dataset` methods so that they enforce this on save, giving a nice error message if it's violated.

Marigold:
+1 to this, I never know whether to use `-` or `_`, some asserts would certainly help

larsyencken:
This is more important than I thought, since we already generate heaps of datasets that aren't compliant and will only create more.

I tried briefly to load our catalog into `duckdb` and the poor naming conventions became a problem quickly.

I spotted living planet, GHO, and FAO as all having not-ideal naming conventions. Better that we fix it sooner!

Marigold:
On it, my brain barely works on Fridays anyway, so this is an ideal task!

PR #128, opened 2022-03-21
refactor: enforce snake_case in table and variable names

Marigold:
Solves this issue https://github.com/owid/etl/issues/111. It depends on this one https://github.com/owid/owid-catalog-py/pull/25

I've added validation before saving a dataset & adding a table and upserting to grapher (a more elegant version would use `property.setter`, but I didn't want to mess up metadata dataclasses and also allow some flexibility).

Then I recreated an entire catalog making sure all steps use snake_case variable names. I've double checked those notebooks myself to make sure I haven't changed anything else.

Walden datasets **haven't been changed** nor do we enforce under_score for them (is it a priority? or maybe in scope of this one?).

Right now I'm comparing local catalog to the production one to make sure they're identical.


larsyencken (commented):
It looks good, except the changes to the `kernelspec` block in the Jupyter notebooks, which currently break it from running on the command-line.

In a later pass, we might fix the names so that we don't need to use `underscore_table()`, i.e. that tables already have the right column names, but that's not important this time around.

PR #129, opened 2022-03-21
fix(grapher_import): fix type incompatibility when inserting to grapher

Marigold:


etl/grapher_helpers.py
```diff
@@ -105,7 +106,13 @@ def yield_wide_table(table: catalog.Table) -> Iterable[catalog.Table]:
 
     dim_names = [k for k in table.primary_key if k not in ("year", "entity_id")]
 
-    for dims, table_to_yield in table.groupby(dim_names, as_index=False):
+    if dim_names:
+        grouped = table.groupby(dim_names, as_index=False)
+    else:
+        # no dimensions
+        grouped = [([], table)]
```
larsyencken, line 113: This part isn't clear. Do we have tables without dimensions?

(I gave Pablo feedback on his data that it was missing indexed columns, which he fixed)

larsyencken (commented):
The types look fine, just wondering if data should ever be missing dimensions.

etl/grapher_helpers.py
Marigold: Sorry, it should have said
_# a situation when there's only year and entity_id in index with no additional dimensions_

larsyencken: Yeah, that's ok. I was just trying to re-understand what's happening here, and basically it's the situation where there's no dimensions to flatten. The net effect is that the double loop collapses to something like this:

```
if not dim_names:
    yield from [table[[col]] for col in table.columns]
```

larsyencken: Can you also get the slugify method on L122 to use `"__"` instead of `"-"` for the dimension separator, since we're fixing this too?

Then it's good to merge :shipit:

PR #134, opened 2022-03-23
fix(food_explorer): change the way region values are calculated

lucasrodes:
This PR improves our estimates on region (continents, income groups) metric values. We do this by limiting when these are estimated and when not. More details about the heuristic can be found at https://github.com/owid/owid-issues/issues/241#issuecomment-1076618664.

Related to issue https://github.com/owid/owid-issues/issues/241.

### More details

I have been changing the pipeline today and have put the solution in place. I have chosen `T2=20%` and `T1=10%`, meaning:

- **absolute values** will be computed whenever the population_difference is below 20%.
- **per capita** values will only be computed when the population difference is below 10%.

This is only being applied to the following metrics: food supply (kcal, kg, protein, fat), waste in the supply chain, other uses, feed (note that all are from FBS). I can easily add/remove metrics here.

The changes are in this PR: @marcelgerber  

Some notes:
- The file per food has nearly doubled (as a consequence of adding per_capita columns). Some metrics might not need per_capita columns and the other way around, so there is be room for optimisation here.


etl/steps/data/garden/explorers/2021/food_explorer.ipynb
```diff
@@ -1,2290 +1,2530 @@
...
+    "    \"waste_in_supply_chain__tonnes\",\n",
+    "    \"feed__tonnes\",\n",
+    "]\n",
+    "# get removable\n",
+    "remove_all = x.loc[\n",
+    "    (x.population_diff_perc >= t2) & (x.metric.isin(metrics)),\n",
+    "    [\"product\", \"metric\", \"region\", \"year\"],\n",
+    "]\n",
+    "remove_capita = x.loc[\n",
+    "    (x.population_diff_perc >= t2)\n",
```
marcelgerber, line 2045: This line should be dropped right? Otherwise we disregard t1 entirely and only ever remove per-capita values if `>= t2`!

Currently, Oceania still has per-capita metrics in total.csv (wrongly).

marcelgerber (changes requested):

marcelgerber:
> Some metrics might not need per_capita columns and the other way around, so there is be room for optimisation here.

As for that, I have run [a quick analysis over here](https://colab.research.google.com/drive/1-o81q4Jn3szffb4bToswwS3hhTUeIuRl?usp=sharing) as for which columns we need where.
tl;dr, we won't be needing the following columns because they are never exposed in any way:
- `food_available_for_consumption__fat_g_per_day`
- `food_available_for_consumption__kcal_per_day`
- `food_available_for_consumption__kg_per_year`
- `food_available_for_consumption__protein_g_per_day`
- `yield__kg_per_animal__per_capita`
- `yield__tonnes_per_ha__per_capita`

etl/steps/data/garden/explorers/2021/food_explorer.ipynb
marcelgerber: Checked it out, with that line removed it works fine!

lucasrodes: Well spotted! I'm changing it to `x.population_diff_perc < t2`

lucasrodes:
Thanks for looking into the columns that can be removed. Just added a dropping step for them!

marcelgerber:
Uh I gave it another go and it failed running.
The issue is in `In [103]`, `assert cols_missing == [...]`

lucasrodes:
Ow, sorry for that. Totally went unnoticed.

Should be fixed now.

marcelgerber:
Perfect!
I also ran a quick check on the output file sizes, and they are as follows:

| File           | Size    | Size (brotli) |
| :------------- | :------ | :------------ |
| `fruit.csv`      | 3.13 MB | 923 KB        |
| `oilcrops.csv`   | 3.13 MB | 901 KB        |
| `vegetables.csv` | 3.12 MB | 889 KB        |
| `maize.csv`      | 2.95 MB | 853 KB        |
| `wheat.csv`      | 2.87 MB | 835 KB        |
| ...            |         |               |
| `total.csv`      | 904 KB  | 145 KB        |

As always, I'm amazed at how good brotli is, and we're definitely fine with these 🙏

marcelgerber (approved):

marcelgerber:
In another sanity check, I diffed `almonds.csv` and `vegetables.csv` to see if there are any changes in _existing_ regions. And indeed, there were only those changes that were expected, namely that several regions didn't have datapoints for `feed__tonnes` and `other_uses__tonnes` any more.
So all good on that front too!

Issue #76, opened 2022-02-01
faostat: API change breaks faostat_fbsc and prevents the ETL from building


larsyencken:
During ETL of `faostat_fbsc`, an assertion fails on this line:

<img width="1511" alt="Screen Shot 2022-02-01 at 13 27 00" src="https://user-images.githubusercontent.com/36012/151968251-07bcc9ba-771d-46e8-b568-b5b10bd6f440.png">

MarcelGerber:
@lucasrodes and I looked into this at some point.
This is @lucasrodes:

> 1. Although the bulk data comes from walden, the additional meta data (tables meta_item, meta_area, etc.) do not come from Walden. Instead they are added in the meadow step calling an FAO API.
> 2. This is very not-ideal, instead these tables should be also added in the walden step. So we have a complete frozen picture
> 3. However, we did not do this in Walden because a small restructuring of Walden is required and we did not prioritize this. Small refactoring = allowing Walden to have several source data files (currently it only has one).

The FAO updated their dataset at some point, and so some definitions between the data file sourced from walden and the data sourced from the FAO API diverged.

larsyencken:
Ah, the lack of multi-file support strikes! https://github.com/owid/walden/issues/5

lucasrodes:
Thanks for raising it!

I can look into this later this week and fix it for the time being. However, I expect this error to pop up again as FAO updates the metadata files.

larsyencken:
The current workaround is to build the ETL without fao data `.venv/bin/etl --exclude=fao`

larsyencken:
@Marigold Just making a note that we should chat together about how the ETL runs, and about ways to get it running more reliably

PR #138, opened 2022-03-28
fix(food-explorer): improve heuristic for region metrics computation

lucasrodes:
This PR improves the heuristic for metric computation for regions. Details about the heuristic are explained in https://github.com/owid/owid-issues/issues/258.

Additionally, it makes sure that values for `Total` are always shown for all regions.

In the notebook, basically, section 9.5 is new. Note that the order of subsections 9.X has changed (e.g. 9.6 has been moved)


marcelgerber:
Hm, it currently doesn't run

marcelgerber (changes requested):
Hm, it currently fails for me with the following:
```
Exception encountered at "In [91]":
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_2180/3959191247.py in <module>
     16 values_to_remove = fe_bulk_melted_regions.loc[msk, "value"].unique()
     17 if not all(values_to_remove == [0.0, 0.011428571428571429]) or msk.sum() != 1512:
---> 18     raise ValueError("Re-check merge")
     19 # Filter NaNs (controlled)
     20 fe_bulk_melted_regions = fe_bulk_melted_regions[~msk]

ValueError: Re-check merge
```

lucasrodes:
Thanks for the review @marcelgerber.

I just ran this `etl` locally and no error was raised:

```
.venv/bin/etl fbs          
Running 3 steps:
1. data://meadow/faostat/2021-04-09/faostat_fbs...
OK (7s)

2. data://garden/faostat/2021-04-09/faostat_fbsc...
OK (62s)

3. data://garden/explorers/2021/food_explorer...
OK (164s)
```

I am now trying to think where the desync might be occurring, but not sure at the moment

marcelgerber:
Hm, I ran it on `owid-analytics`, and you can try it there too:
```
ssh owid-analytics
cd etl-staging
.venv/bin/etl food_explorer
```

Maybe it's #76 again? Because you have different FAO API data locally than what is on `etl-staging` and my own machine (which are both pretty new installations of the etl)?

lucasrodes:
I slightly relaxed the validation check, the ETL should run now.

There is a mismatch in the output between staging and my local machine. Hence I was not able to detect this error. This is due to some error in my local data files (at `~/.owid/walden`).

marcelgerber (approved):

PR #144, opened 2022-03-30
data(maddison): Production export of the Maddison dataset data

Marigold:
Import Maddison dataset from ETL into grapher. This was done in way to have metadata parity with the old version of Maddison. Here's the [old one](https://staging.owid.cloud/admin/datasets/5219) on staging vs the [new one](https://staging.owid.cloud/admin/datasets/5566) for comparison.

There's one "trick" where we map `Dataset.description` to `Source[0].description` to make it work with the grapher model. It's explicit in the code, but we may decide to make it implicit. Metadata is loaded from YAML file using helper methods, but that one might be also improved in the future.

(There's still [one part](https://github.com/owid/etl/pull/144/files#diff-066f8606fbd0d1ebf10c166179474f38f8c508d55ecdb232b1c81b9530fa13f5R28) I need to improve, but that shouldn't prevent review)

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -10,6 +10,8 @@ from typing import Optional, Dict, Literal, cast, List, Any
 from pydantic import BaseModel
 
 from etl.paths import DATA_DIR
+from etl.db import get_connection
+from etl.db_utils import DBUtils
 
 
 # TODO: remove if it turns out to be useless for real examples
@@ -177,6 +179,13 @@ def country_to_entity_id(
         return cast(pd.Series, entity_id.astype("Int64"))
     else:
         return cast(pd.Series, entity_id.astype(int))
+
+
+def get_or_create_entity(name: str) -> int:
+    cursor = get_connection().cursor()
+    db = DBUtils(cursor)
+    print(name)
+    return db.get_or_create_entity(name)
 
 
 def _unique(x: List[Any]) -> List[Any]:
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -4,14 +4,20 @@ from collections.abc import Iterable
 import yaml
 import slugify
 import warnings
+import logging
 from pathlib import Path
 
-from typing import Optional, Dict, Literal, cast, List, Any
+from typing import Optional, Dict, Literal, cast, List, Any, Set
 from pydantic import BaseModel
 
 from etl.paths import DATA_DIR
-from etl.db import get_connection
+from etl.db import get_connection, get_engine
 from etl.db_utils import DBUtils
+
+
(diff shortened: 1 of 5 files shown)


etl/grapher_helpers.py
```diff
@@ -179,6 +181,13 @@ def country_to_entity_id(
         return cast(pd.Series, entity_id.astype(int))
 
 
+def get_or_create_entity(name: str) -> int:
+    cursor = get_connection().cursor()
+    db = DBUtils(cursor)
+    print(name)
```
larsyencken, line 187: Drop the debug print I think

etl/grapher_helpers.py
```diff
@@ -179,6 +181,13 @@ def country_to_entity_id(
...
+    return db.get_or_create_entity(name)
```
larsyencken, line 188: Not anything you need to change now, but as a comment it seems like we're only ever creating entities here, never getting them. Getting them would mean that our countries/regions is out of sync with the grapher country data.

larsyencken (commented):
Only minor stuff from my end. Code-wise, it's good to go, as long as the data itself looks good too.

Marigold:
@pabloarosado I've merged the core changes needed for this PR, but nothing has changed in this PR. Would you mind approving this (or re-reviewing if you want)?

pabloarosado:
Hi @Marigold thanks. Is the new dataset anywhere, so that I can compare it to the old one? I guess [this dataset](https://staging.owid.cloud/admin/datasets/5566) is no longer the latest version. Otherwise I can fetch the branch and generate it myself.

Marigold:
> Hi @Marigold thanks. Is the new dataset anywhere, so that I can compare it to the old one? I guess [this dataset](https://staging.owid.cloud/admin/datasets/5566) is no longer the latest version. Otherwise I can fetch the branch and generate it myself.

It is the latest version, the data hasn't changed since you last compared them.

pabloarosado (approved):
Thanks @Marigold ! The data, metadata and pipeline look good. Just two comments (that should not affect this PR, but rather future pipelines):
* We could considering establishing some guidelines on what metadata should be for meadow, and what for garden or grapher. In this case, I think it is a bit arbitrary. Most metadata could be from the beginning in meadow (except the note about Tanzania, which is relevant only for >= garden steps).
* It's not clear what should happen next time we update this dataset. Should we duplicate the garden/ggdc/2020-10-01/ folder and its content, and replace the date "2020-10-01" everywhere? It seems a bit redundant, although it's not too worrying, given that this will happen at most once per year for this dataset.

Marigold:
@pabloarosado good questions, created an issue for that https://github.com/owid/etl/issues/158 to get @larsyencken opinion.

Issue #141, opened 2022-03-29
Do a production export of the OWID population dataset


larsyencken:

larsyencken:
- Depends on #142

PR #145, opened 2022-03-30
data(population): population to grapher

Marigold:
Export population density dataset to grapher. This is not currently used in production by any other datasets, but it simplifies metadata for population with YAML files and it could be useful in the future. (And I could finally close the [original issue](https://github.com/owid/etl/issues/141))

### DIFF
--- a/dag.yml
+++ b/dag.yml
@@ -80,6 +80,8 @@ steps:
     - data://meadow/worldbank_wdi/2022-05-26/wdi
   grapher://worldbank_wdi/2022-05-26/wdi:
     - data://garden/worldbank_wdi/2022-05-26/wdi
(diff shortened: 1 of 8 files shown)


danyx23 (commented):
It looks like this move to the YAML has dropped the display settings? E.g. the displayInTable has repercussions for what Grapher displays in the table view and I think for such an important variable we should make sure that this doesn't change anything for the worse. Should we add it to the YAML again? Or was this a concious decision to drop it?

Marigold:
@danyx23 good catch, that was totally unconscious. Either that happened when I was rebasing or I might have simply missed it.

danyx23 (approved):
Looks good! I haven't run it again but I think it should be fine 😬 😉

PR #146, opened 2022-03-30
enhance(grapher): align metadata for grapher and ETL

Marigold:
PR for owid-catalog-py https://github.com/owid/owid-catalog-py/pull/28

The mapping between ETL and grapher metadata is not obvious at all. The grapher data model is quite outdated with some things that are not making much sense such as:

1. Even though dataset can have multiple sources, only the first one is shown - as a workaround we join all sources to a single one before inserting to grapher (this happens for `population density`)
2. Admin UI for dataset weirdly maps fields, e.g. `source.additionalInfo` is displayed as `Dataset description` in the UI, `unit` and `shortUnit` of variables are not mapped at all, but instead we show `display.unit` and `display.shortUnit`, etc.

Nevertheless, once you get used to these it's straightforward to write a script that gets your data into grapher in a way you want by running it against staging and checking it in Admin UI.

This was tested on two cases - [maddison to grapher](https://github.com/owid/etl/pull/144) and population to grapher. Both were created to be at parity with our current versions in grapher.

### DIFF
--- a/etl/db_utils.py
+++ b/etl/db_utils.py
@@ -1,3 +1,5 @@
+"""This module was copied from https://github.com/owid/importers/blob/master/db_utils.py"""
+
 import json
 from typing import Any, Dict, Iterator, List, Optional, Tuple, cast
 from unidecode import unidecode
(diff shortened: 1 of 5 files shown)


etl/db_utils.py
```diff
@@ -237,6 +239,10 @@ def upsert_namespace(self, name: str, description: str) -> int:
     def upsert_source(self, name: str, description: str, dataset_id: int) -> int:
         # There is no UNIQUE key constraint we can rely on to prevent duplicates
         # so we have to do a SELECT before INSERT...
+        # TODO: changing source description and rerunning grapher inserts a new source
+        # without deleting the old one. So we end up with multiple sources in grapher DB.
+        # We could delete & create them on every dataset update, but we risk losing manual
+        # changes to those sources.
```
larsyencken, line 245: Is the source information usually greyed out in the admin for things that come from the data team, locked from being edited? If so, we could do the same here

larsyencken (approved):
Looks good! :shipit:

etl/db_utils.py
Marigold: Yes, that's fortunately what happens. I've updated the note and also created an issues for cleaning up objects like Sources in DB after rerunning grapher multiple times https://github.com/owid/etl/issues/147.

PR #149, opened 2022-03-30
chore(jupytext)

lucasrodes:
In some pipelines, especially complex ones, it makes sense to use modules (`.py`) instead of notebooks, but it would be great to still be able to leverage the power of Jupyter Notebooks.

[jupytext](https://github.com/mwouts/jupytext) is a nice way to interact with plain text (i.e. `.py`) using Jupyter Notebooks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ wikipedia = "^1.4.0"
 pandas = "^1.4.0"
 numpy = "^1.22.1"
 pydantic = "^1.9.0"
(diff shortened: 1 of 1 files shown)


larsyencken:
It looks like it's already there as a dev dependency. You can try it out with `make lab`, then browse to a python file, then you need to right-click on it and open it with jupytext

lucasrodes:
Totally missed that. Thanks, @larsyencken

PR #155, opened 2022-04-04
feature(backport): Perform a once-off bulk backport of the full dataset

Marigold:
Add command `bulk_backport` that calls `backport` for backporting all datasets that are used in at least one chart and are not private to Walden. Next, flag `--backport` creates `Backport` steps for all backported datasets found in Walden. That step aligns grapher metadata with ETL metadata and yields a wide-table (~~or long-table in rare cases~~) with all columns from the backported dataset.

I've tested it on two cases - creating World Bank Indicators from backported data and for getting SDG indicators into ETL catalog.

Datasets & tables are currently named like `dataset_5438_global_health_observatory__world_health_organization__2021_12`. We have [an issue](https://github.com/owid/etl/issues/167) for making them prettier.

### Long vs wide tables

~~The initial plan was to convert every dataset to a wide `Table`, but it turned out there are a few datasets for which wide format wouldn't be effective (creating extremely sparse >4GB dataframes). I tried a few things - sparse dataframes, hardcoded steps for those datasets, better compression, ... but none of them worked well so I ended up keeping the table in long format and added helper function `create_wide_table` to convert them on-demand after pre-filtering (useful for SDG dataset). If this works for us we can wrap this functionality in a new `LongTable` class with `to_wide_table` method.~~

Running [some benchmarks](https://github.com/owid/notebooks/blob/main/MojmirVinkler/2022-04-27%20Wide%20vs%20long%20format.ipynb) showed that long format is much less efficient than wide format. There are a few exception (extremely sparse datasets) that were autogenerated, but rather than supporting another format I've just split it into tables with 1000 variables. In the future we can revisit those datasets and split them logically into tables in the generating script.

### Removed content-hash

Since field `dataEditedAt` is reliable, there's no need for calculating content hash of all values. Instead, we only calculate checksum of all metadata (which includes `dataEditedAt`).

### Performance

Checksum comparison takes ~1s per dataset (we have 800 datasets in total), this could be speeded up by async / threading, but it's ok for now IMHO.

### DIFF
--- a/backport/backport.py
+++ b/backport/backport.py
@@ -43,7 +43,7 @@ def _walden_values_metadata(ds: GrapherDatasetModel, short_name: str) -> WaldenD
         name=ds.name,
         date_accessed=dt.datetime.utcnow().isoformat(),
         description=ds.description,
-        source_name="Our World in Data",
+        source_name="Our World in Data catalog backport",
         url=f"https://owid.cloud/admin/datasets/{ds.id}",
         publication_date="latest",
         file_extension="feather",
@@ -228,7 +228,9 @@ def backport(
     lg.info("backport.loading_values", variables=variable_ids)
     df = _load_values(engine, variable_ids)
     lg.info("backport.upload_values", size=len(df))
-    _upload_values_to_walden(df, _walden_values_metadata(ds, short_name), dry_run, upload)
+    _upload_values_to_walden(
+        df, _walden_values_metadata(ds, short_name), dry_run, upload
+    )
 
     lg.info("backport.finished")
 
--- a/backport/backport.py
+++ b/backport/backport.py
@@ -118,25 +118,26 @@ def _load_config(
 
 
 def _upload_config_to_walden(
-    config: GrapherConfig, meta: WaldenDataset, dry_run: bool
+    config: GrapherConfig, meta: WaldenDataset, dry_run: bool, upload: bool
 ) -> None:
     with tempfile.NamedTemporaryFile(mode="w") as f:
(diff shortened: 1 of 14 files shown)


backport/backport.py
```diff
@@ -14,53 +14,59 @@
...
-    :param origin_md5: MD5 hash of data values in grapher used to decide whether to recompute
-        or not in the next run
     """
     return WaldenDataset(
         namespace=WALDEN_NAMESPACE,
         short_name=f"{short_name}_values",
         name=ds.name,
+        date_accessed=dt.datetime.utcnow().isoformat(),
         description=ds.description,
         source_name="Our World in Data",
```
danyx23, line 46: I think we should changes this to "Our World in Data catalog backport" or something like that to make it clear that we are not the original source and one should inspect the sources of the config json file to understand the upstream sources

danyx23 (approved):
Looks good! I think we should store something else in walden for the source but that is my only issue for now. In the future we might want to do a bit more to make the very large datasets more digestable as you mention in a comment but this can be done later on.

backport/backport.py
Marigold: Good idea, fixed.

Marigold:
> Looks good! I think we should store something else in walden for the source but that is my only issue for now. In the future we might want to do a bit more to make the very large datasets more digestable as you mention in a comment but this can be done later on.

Thanks!!! Gonna merge this and rebuild everything on analytics server... and then do more **smaller** PRs for backporting. Can't wait.

Issue #106, opened 2022-03-02
Group datasets into "channels" when publishing


larsyencken:
## Problem

Right now, we have a pattern where every dataset we care about will have two versions in the ETL, an unharmonized one in `meadow` and a harmonized one in `garden`. So when you search for a dataset, it is not clear at first glance which one you should use.

Secondly, the open numbers datasets are so many that they overwhelm the search results.

## Proposed solution

- We group datasets into _channels_ based on their folder inside the `data/` directory
    - We should have `garden`, `meadow` and `open_numbers` channels initially
- A dataset doesn't know which channel it came from, but it is stored as a categorical variable in the published index
- When you search the catalog, you search `"garden"` by default; but you can also select any other channel, or `"all"`
    - `catalog.channels` should be a list of the options

With this solution, you should be able find data quickly, and default to the best version of the data that we have.

## Steps

* [x] Move open_numbers out of meadow and into its own folder (and update scripts)
* [ ] Move reference into garden (and update scripts)
* [ ] Index and publish channels individually
* [ ] Add channels and lazy loading to owid-catalog-py

Marigold:
All is pretty clear, a couple of questions
1. should we move `open_numbers` from `meadow` into its own `open_numbers` folder? (so that the folder structure aligns with channel definitions) We'd end up with channels: `garden`, `meadow`, `open_numbers`, `examples` (and possibly `backported`). _I think yes_
2. `reference` datasets have their own folder, I'd load them by default together with `garden`

I guess that lazy loading of channels and better logging to show why is it taking so long are in scope of this.

larsyencken:
> 1. should we move `open_numbers` from `meadow` into its own `open_numbers` folder? (so that the folder structure aligns with channel definitions) We'd end up with channels: `garden`, `meadow`, `open_numbers`, `examples` (and possibly `backported`). _I think yes_

I agree, open numbers deserves its own whole channel.

> 2. `reference` datasets have their own folder, I'd load them by default together with `garden`

I'm happy with this too, though I increasingly think we should end up making it just another standard step in garden that depends on an initial walden snapshot, rather than something checked-in to git.

> I guess that lazy loading of channels and better logging to show why is it taking so long are in scope of this.

Seems like it would make sense. I'd probably break it into multiple steps. Something like:

- Move `open_numbers` out of meadow and into its own folder (and update scripts)
- Move `reference` into `garden` (and update scripts)
- Index and publish channels individually
[...]

PR #165, opened 2022-04-19
refactor(reference): move reference dataset into garden

Marigold:
Part of https://github.com/owid/etl/issues/106

Move reference dataset into garden channel, but keep a symlink to the original `data://folder` to not break scripts in the importers repo.

If we go with this we should probably change the path to reference dataset in other repos as well.


danyx23 (commented):
This one removes the csv files from github, right? This could be a good idea but I think maybe this is something to discuss with Lars, Pablo and Lucas in Barcelona?

Marigold:
> This one removes the csv files from github, right? This could be a good idea but I think maybe this is something to discuss with Lars, Pablo and Lucas in Barcelona?

@danyx23 thanks for spotting this, I messed up subfolders in `.gitignore`. Those `.csv` files are just moved from `data/reference` to `data/garden/reference` (and symlinked to the original folder just in case). I fixed that, not it should work.

data/reference
```diff
@@ -0,0 +1 @@
+/Users/mojmir/projects/etl/data/garden/reference
```
danyx23, line 1: Is this file supposed to be here?

danyx23:
@Marigold  Could you rebase this on top of master? I tried running this on the key_indicators and got non-deterministic runs but I think this is due to the issue fixed in #175

data/reference
Marigold: That's a symlink to the new location with the data. I thought it would be safer to keep it there just in case, but perhaps I should just get rid of it and make anyone who could be using this update it? (I don't think there's anyone)

Marigold:
@danyx23 rebased (and tested `etl key_indicators` locally)

data/reference
danyx23: But it seems to be recorded as an absolute path, right? I.e. this won't work on anybody else's machine. Or am I missing something?

danyx23 (approved):
The only thing that is off I think is the symlink dir (/data/reference) - other than that it looks good to me and can be merged!

data/reference
Marigold: Ouch... you're right, I totally missed this.

PR #166, opened 2022-04-20
feature(catalog): split catalog into channels

Marigold:
Core PR in [owid-catalog-py](https://github.com/owid/owid-catalog-py/pull/35)

Split our catalog into multiple channels to make finding garden datasets easier and more efficient. By default we only read `garden` channel which is really small (17 datasets with ~30kb index). Other channels like `open_numbers` or `backport` have much larger indexes (>MBs) and take longer time to reindex & publish.

### DIFF
--- a/etl/publish.py
+++ b/etl/publish.py
@@ -14,7 +14,7 @@ import boto3
 from botocore.client import ClientError
 import pandas as pd
 
-from owid.catalog import LocalCatalog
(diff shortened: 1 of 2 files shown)


danyx23 (approved):
I didn't do an in-depth review on this but conceptually and looking over the code on github I think this makes a lot of sense!

PR #172, opened 2022-04-28
enhance(dag): include all dependencies of the planned subgraph

Marigold:
After computing the subgraph, include their dependencies as well. This needed a couple of performance improvements to make this reasonably fast - caching walden catalog and checking steps dirtiness in parallel (requests from `GithubStep` were slowing it down).

### DIFF
--- a/etl/steps/__init__.py
+++ b/etl/steps/__init__.py
@@ -72,6 +72,10 @@ def to_dependency_order(
         # cut the graph to just the listed steps and the things that
         # then depend on them (transitive closure)
         subgraph = filter_to_subgraph(graph, includes)
+
+        # make sure we also include all dependencies of the subgraph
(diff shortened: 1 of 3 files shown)


pabloarosado (approved):

etl/steps/__init__.py
```diff
@@ -109,6 +113,32 @@ def reverse_graph(graph: Graph) -> Graph:
...
+        nodes |= subgraph[node]
+
+    return dict(subgraph)
+
+
+def graph_nodes(graph: Graph) -> Set[str]:
+    all_steps = set(graph)
+    for children in graph.values():
+        all_steps.update(children)
+    return all_steps
```
larsyencken, line 139: Nice fix!

It looks like `graph_nodes()` is never used, you can kill it.

Marigold: My bad, gonna fix that in future PR.

PR #183, opened 2022-05-09
Benchmark feather file compression methods

wcox13:
Compare the compressed size and read performance for 1) feather files compressed with the native "zstd" compression, and 2) feather files compressing using gzip. From the `etl` directory, you can run it with `python benchmarking_tmp.py`. You can also fiddle with the constants at the top to change the test dataframe or number of iterations.

Sample results:

```
Dataframe len: 21854
Size of zstd feather: 186,626 bytes
Size of gzipped raw feather: 182,668 bytes
Average read time (zstd): 0.00196s
Average read time (uncompressed + gz): 0.01198s
```

### DIFF
--- a/benchmarking_tmp.py
+++ b/benchmarking_tmp.py
@@ -6,9 +6,8 @@ import timeit
 
 from tempfile import TemporaryDirectory as TempDir
 
-ITERATIONS = 100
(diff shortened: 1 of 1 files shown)


wcox13:
(Closing since this isn't intended to be merged)

Issue #186, opened 2022-05-12
Make it possible to add external DAG files `dag.yml`


Marigold:
Our `dag.yml` currently looks something like this

```
steps:
  data://garden/owid/latest/covid:
    - etag://raw.githubusercontent.com/owid/covid-19-data/master/public/data/owid-covid-data.csv
  ...
  data://open_numbers/open_numbers/latest/bp__energy:
    - github://open-numbers/ddf--bp--energy
  ... way too many open_numbers steps that make this hard to read...
```
an idea is to move all `open_numbers` steps into their own `dag_open_numbers.yml` and reference them from the main `dag.yml` file using `include:` stanza, i.e.

```
steps:
  data://garden/owid/latest/covid:
    - etag://raw.githubusercontent.com/owid/covid-19-data/master/public/data/owid-covid-data.csv
  ...

include:
  - dag_open_numbers.yml
```

with `dag_open_numbers.yml` looking like this

```
steps:
  data://open_numbers/open_numbers/latest/bp__energy:
    - github://open-numbers/ddf--bp--energy
  ...
```

wcox13:
Thanks @Marigold - just to make sure I'm understanding, the idea here is to modify the `load_dag` function of `etl/steps/__init__.py`, so that we also load into the dag/dict any sub-files listed with the `include` header?

wcox13:
Also is it fine to support one level of `include` for now or should it be like a BFS (support `include` within the included files as well)?

wcox13:
Opened https://github.com/owid/etl/pull/190 @Marigold

Marigold:
Sorry for late reply, but you got it right.

> Also is it fine to support one level of include for now or should it be like a BFS (support include within the included files as well)?

I don't think we'd use it, but supporting it should be easy (more in your PR).

pabloarosado:
Thanks @Marigold I like it very much! I see that dependences work well even if they are defined inside a sub-dag, so that's perfect.
Just one suggestion: Wouldn't it be better to keep sub-dags inside a folder? Otherwise, the root of the etl will become crowded with dag_*.yml files.

pabloarosado:
Hey @Marigold, I just tested that having a 'dag_files' folder with the sub-dags, and then in the include section cite 'dag_files/dag_faostat.yml' works well. At least I did a `etl --dry-run faostat` and it found all the steps well.

Marigold:
Good idea, `dag_files` folder is definitely better. Please move open numbers there too.

[...]

PR #190, opened 2022-05-22
Make it possible to include additional DAG files in `dag.yml` (#186)

wcox13:
See test data files for an example of how it works.


tests/data/dag.yml
```diff
@@ -0,0 +1,7 @@
+
+steps:
+  data://garden/wpp/2019/standard_projections:
+    - data://meadow/wpp/2019/standard_projections
```
wcox13, line 4: Unfortunately, we have to use real steps here to avoid `no recipe for executing step`. The other tests use the entire real `dag.yml` though, so this is at least a slight improvement :)

tests/test_etl.py
```diff
@@ -28,7 +29,17 @@ def test_all_data_steps_have_code():
...
 
 
-def get_all_steps() -> List[Step]:
-    dag = load_dag()
+def test_sub_dag_import():
+    # Ensure that sub-dag is imported from separate file
+    sub_dag_step = "general_files"
+    assert [
+        s
+        for s in get_all_steps(filename="tests/data/dag.yml")
```
Marigold, line 37: Nitpick - to make our lives simpler I'd only test `load_dag` here. That way you don't need real steps as you won't be calling `compile_steps` and you only test the function of interest.

Marigold (approved):
Only minor things, code in general looks good! Use your judgement whether to implement my suggestions or not and then we can ship it!

tests/test_etl.py
wcox13: Good call. That also allows me to make the test data generic because I don't have to worry about finding a recipe to execute the step.

etl/steps/__init__.py
```diff
@@ -97,11 +97,33 @@ def to_dependency_order(
...
+def _load_dag(filename: Union[str, Path], prev_dag: Dict[str, Any]):
+    """
+    Recursive helper to 1) load a dag itself, and 2) load any sub-dags
+    included in the dag via 'include' statements
+    """
+    dag_yml = _load_dag_yaml(filename)
+    curr_dag = _parse_dag_yaml(dag_yml)
+    curr_dag.update(prev_dag)
+
+    if "include" in dag_yml:
```
Marigold, line 112: Nitpick - this could be written as one-liner
```
for sub_dag_filename in dag_yml.get("include", []):
```

Marigold, line 112: Nitpick - this could be written as one-liner
```
for sub_dag_filename in dag_yml.get("include", []):
```

Marigold:
@wcox13 do you have permissions to merge (and squash) this yourself?

Marigold:
@wcox13 do you have permissions to merge (and squash) this yourself?

etl/steps/__init__.py
wcox13: 👍 Good point

PR #198, opened 2022-05-25
Create folder for sub-dag files

pabloarosado:
To avoid cluttering the root of the repository with sub-dag files (`dag_open_numbers.yml`, `dag_faostat.yml`, etc.), I created a folder called `dag_files/`, and moved the open-numbers sub-dag file inside that folder.


Marigold (approved):

larsyencken (approved):
Good move :shipit:

Issue #94, opened 2022-02-22
Propagate source checksums all the way into grapher


larsyencken:
## Problem

We would like to know which data in grapher is out of date compared to the ETL, without re-running an upsert to grapher. This gives us important information about what's changed, information that's especially important if we believe that upserts to grapher will always be done manually. For example, if we recalculate our population dataset, we may wish to have an indicator which datasets that depend on it need updating in Grapher.

## Proposed solution

One solution is to propagate the `source_checksum` mechanism that the ETL uses, and to maintain in MySQL a mapping of which checksum is the latest used by a dataset. With such a mechanism, it could be cheap to check what data in MySQL is out of date, and list that to the user.

Marigold:
Just a note for my future self -  all updates to grapher are now done manually, so this is not such a big issue, but if we start pushing to grapher automatically then we absolutely have to do this since every push to grapher takes ~10s (because of `data_values`).

larsyencken:
More context from our chat:

- The ETL is "smart" and only rebuilds what needs rebuilding
- But the upsert to grapher is not smart right now, it always upserts everything
- The smarts are based on checksums
  - You never need to rebuild something if the checksums haven't changed (see `checksum_input()`, `checksum_output()`, `is_dirty()` in `etl/steps/__init__.py`)
- This ticket would make it smart, by adding checksums of the data to MySQL
  - That way, the next time you go to upsert, you can see that nothing has changed, and skip it

Open ideas:

- Could we continuously deploy data from the ETL to grapher, not make it so manual?
  - Probably we can, and should, if the ETL is stable enough and used often

Open questions:

- When you upsert over the top of an existing dataset, are variable ids preserved? (charts will refer to data by `variableId`, so once we have a chart, we cannot delete a variable without deleting the chart first)
- Where would the checksum live in the database?
  - Since ETL steps cover one whole dataset at a time, then the checksum would be added to the `datasets` table in MySQL (field to be added via a db migration script in the `grapher` repo)

wcox13:
So here is the rough plan on this:

1) Write a migration in grapher to add a `latestChecksum`. I think it's best for this to be non-null, defaulting to zero or something, since we should always have a checksum
2) Update `db_utils#upsert_dataset` to write the additional column, which is easily accessed in that scope with `dataset.metadata.source_checksum`
[...]

PR #208, opened 2022-06-06
Write source checksum to new db column

wcox13:
#94 

When a dataset is loaded into grapher, write the source checksum. This allows us to compare with the checksums in ETL to determine whether grapher is out-of-date.

Prerequisite:
- [ ] Migration to add `sourceChecksum` db column

### DIFF
--- a/etl/db_utils.py
+++ b/etl/db_utils.py
@@ -166,25 +166,27 @@ class DBUtils:
         namespace: str,
         user_id: int,
         tag_id: Optional[int] = None,
+        source_checksum: Optional[str] = None,
         description: str = "This is a dataset imported by the automated fetcher",
     ) -> int:
         operation = self.upsert_one(
             """
             INSERT INTO datasets
-                (name, description, namespace, createdAt, createdByUserId, updatedAt, metadataEditedAt, metadataEditedByUserId, dataEditedAt, dataEditedByUserId)
+                (name, description, namespace, sourceChecksum, createdAt, createdByUserId, updatedAt, metadataEditedAt, metadataEditedByUserId, dataEditedAt, dataEditedByUserId)
             VALUES
-                (%s, %s, %s, NOW(), %s, NOW(), NOW(), %s, NOW(), %s)
+                (%s, %s, %s, %s, NOW(), %s, NOW(), NOW(), %s, NOW(), %s)
             ON DUPLICATE KEY UPDATE
                 name = VALUES(name),
                 description = VALUES(description),
                 namespace = VALUES(namespace),
+                sourceChecksum = VALUES(sourceChecksum),
                 updatedAt = VALUES(updatedAt),
                 metadataEditedAt = VALUES(metadataEditedAt),
                 metadataEditedByUserId = VALUES(metadataEditedByUserId),
                 dataEditedAt = VALUES(dataEditedAt),
                 dataEditedByUserId = VALUES(dataEditedByUserId)
         """,
-            [name, description, namespace, user_id, user_id, user_id],
+            [name, description, namespace, source_checksum, user_id, user_id, user_id],
         )
         (v,) = self.fetch_one(
             """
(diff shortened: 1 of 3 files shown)


wcox13:
@Marigold sorry not sure I understand, a few questions:

1. 

> source_checksum you are inserting to DB is actually the checksum of DataStep from data://garden/ggdc/2020-10-01/ggdc_maddison. If you change code for grapher/ggdc/2020-10-01/ggdc_maddison.py you should get a different checksum

In `DataStep#run`, we set `dataset.metadata.source_checksum = self.checksum_input()`. As you mention, the logic of the `checksum_input` method already accounts for changes to the code files via `for f in self._step_files():  checksums[f] = files.checksum_file(f)`. So what is the problem with using `dataset.metadata.source_checksum` to access the source checksum?

wcox13:
2. 

> Check out how def checksum_input methods in steps/__init__.py work. First try it yourself on with maddison dataset and then add some unit tests to verify it really works (unless you reuse methods that are already tested).

You're suggesting to add new unit tests for `checksum_input`? It looks like there is some coverage for `DataStep` but not a 1:1 unit test of the `checksum_input` functionality

wcox13:
3.

> The second thing is skipping inserts to DB if source checksum matches checksum in DB (i.e. all sources and code are still the same so upsert wouldn't change anything). For this check def is_dirty methods.

Why do we need additional logic for this if we already have `ON DUPLICATE KEY UPDATE` in the SQL statement? Seems like we already have the behavior of "update the checksum whenever it changes"

Marigold:
1. `grapher://*` steps use [GrapherStep](https://github.com/owid/etl/blob/master/etl/steps/__init__.py#L252) class, not `DataStep` and that checksum is missing there. Maybe we just have to copy that line `dataset.metadata.source_checksum = self.checksum_input()`? (Not 100% if this is enough)
2. Up to you. It depends on how you implement 1. If it is simple enough then it probably doesn't need testing
3. If checksums are identical, we want to avoid unnecessary upserts. One reason is performance (insertion to `data_values` could be expensive) and second one is that we'd be wrongly updating timestamps (which could trigger other processes). As [Lars said](https://github.com/owid/etl/issues/94#issuecomment-1148735944), this basically means just implementing `is_dirty` method for `GrapherStep`

Marigold:
> But does this matter in practice, since we run the DataStep before grapher step, and save the checksum_input() result in dataset.metadata.source_checksum?

It does for two reasons - `checksum_input()` computes checksum for **dependencies** and **step files**. `GrapherStep` could have more dependencies than just `DataStep` (even though that's not the case now) and **step files** are all files in grapher step the directory including the script itself. So if you change logic in the grapher script checksum would update. Otherwise your step wouldn't get rebuilt if you make changes (assuming we have 3. implemented).

> I think this is already broken then - in the db, dataset.updatedAt changes to the current timestamp every time I run etl maddison --grapher. Is that wrong behavior?

Once you implement that `is_dirty` method and avoid running upserts at all if checksums match, this wouldn't be happening anymore.

wcox13:
@Marigold I see, thanks for the explanation

I still think it's best to open a separate PR for the `is_dirty` check (a real behavior change) vs computing+writing the checksum to db (a "harmless" addition of data). That way if something goes wrong with `is_dirty`, it's not coupled with the db migration and we don't have to potentially revert that back-and-forth

wcox13:
@Marigold this should now use the correct checksum value, let me know if it looks OK. Also just got access to `owid-grapher` and opened a PR for the migration itself (simple): https://github.com/owid/owid-grapher/pull/1490

Will open an additional PR for `is_dirty` check

Marigold (approved):
Looks good, however wait for the grapher migration part. Once it's done I can check that migration got applied automatically and we can merge this one.

marcelgerber:
> Once it's done I can check that migration got applied automatically and we can merge this one.

@Marigold The migration has run through already!

PR #209, opened 2022-06-07
enhance(backport): add entity_id and entity_code to dataframe index

Marigold:
Add `entity_id` and `entity_code` to index for backported datasets. This will be useful for the API.

Alternatives are:

* keep it as column instead of having it in multiindex
* keeping the `entity_name -> entity_id` mapping in metadata
* `entity_name` as categorical variable with category codes being entity ids

Keeping it all in multiindex is probably the best though.


danyx23 (approved):
LGTM

PR #213, opened 2022-06-13
fix(backport): use correct flag for private datasets

Marigold:
`after_run` method for labelling datasets private wasn't called for backported data, this fixes that and adds test. (Note: we didn't publish any private backported datasets since `--private` flag is not yet turned on our [analytics server](https://github.com/owid/analytics/pull/44))


danyx23 (approved):
I didn't run it but it looks good

PR #219, opened 2022-06-15
enhance(backport): prune inactive datasets from walden

Marigold:
If dataset is archived or deleted from DB, it will still be saved in walden index (local index) and walden S3 bucket. This PR adds `bulk_backport --prune` command that deletes those datasets. This is meant to be run from production server and commited to walden as soon as the backport finishes (so that catalog in S3 and walden commits are out-of-sync as short time as possible).

### DIFF
--- a/backport/bulk_backport.py
+++ b/backport/bulk_backport.py
@@ -1,13 +1,17 @@
 import re
+from typing import Any, cast
 
 import click
 import pandas as pd
 import structlog
 from owid.catalog.utils import underscore
+from owid.walden import Catalog as WaldenCatalog
+from sqlalchemy.engine import Engine
 
 from etl.db import get_engine
 from etl.steps import load_dag
 
+from . import utils
 from .backport import backport
 
 log = structlog.get_logger()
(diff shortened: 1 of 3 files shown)


backport/bulk_backport.py
```diff
@@ -38,11 +42,58 @@
...
+        backport(
+            dataset_id=ds_row.id,
+            short_name=ds_row.short_name,
+            dry_run=dry_run,
+            upload=upload,
+            force=force,
+        )
+
+    if prune:
+        assert not dataset_ids, "Pruning cannot be used together with dataset-ids"
```
danyx23, line 88: good thinking to add this assert :)

backport/bulk_backport.py
```diff
@@ -65,44 +116,40 @@ def bulk_backport(
...
-    log.info("bulk_backport.finished")
+    log.info("bulk_backport.delete", n=len(datasets_to_delete))
+
+    for ds in datasets_to_delete:
+        log.info("bulk_backport.delete_dataset", short_name=ds.short_name)
+
+        # delete it from local and remote catalog
+        if not dry_run:
+            ds.delete()
+            ds.delete_from_remote()
```
danyx23, line 152: Should we make this delete_from_remote configurable of safeguard it? I'm not sure how often someone will run this but if you have walden write access configured and are testing something you might not be aware that prune deletes datasets not just in your local catalog but also the remote one. So maybe either add a --prune-remote flag or add a safety check for an env var that is by default only set on prod?

danyx23 (commented):
I have one minor concern about pruning remote datasets - otherwise it looks good (didn't run it though, party for remote datasets concerns :) )

backport/bulk_backport.py
Marigold: Makes a lot of sense, better be safe than sorry. Added the `--prune-remote` flag.

Marigold:
Fixed, @danyx23 would you mind taking a second look?

Marigold:
@danyx23 could you please take a look? I've only added `--prune-remote` as you've requested.

danyx23 (approved):
Looks great, thanks fro adding this!

Issue #218, opened 2022-06-15
Zombie data


Marigold:
## Problem

ETL doesn't clean up zombie data from `etl/data` folder for which we don't have a recipe anymore. `reindex` command walks through the entire `data` folder and can't tell if data is zombie or not, so it reindex everything and it ends up in our catalog. This is especially "problematic" for backported datasets that get archived or deleted.

## Proposed solution

Script `prune` that would essentially run `etl --dry-run --force` to get all living datasets, compare it with `etl/data` and prune the folder.

Another alternative would be pruning datasets older than N days and let `etl` regenerate them at the next run (could be expensive operation).

@larsyencken what do you think? Is there a better alternative? Is this something we have to solve or can we just keep those zombie datasets lying around?

larsyencken:
Ah, I was thinking about this the other day. If I work through it in my head:

- Inside a dataset, this shouldn't be a problem, as long as `Dataset.create_empty()` is used so that we fully overwrite the old dataset
- If a dataset is in the DAG, then we know running `etl` will bring it up to date, so it should definitely be in the catalog
- If a dataset is not in the DAG, then it is zombie, and we should prune it

Yep, I come to the same conclusion as you. We should walk the datasets in the local catalog and delete any which aren't in our whitelist.

larsyencken:
I wonder _when_ it would be best to do this. Perhaps the bare `etl` command should be changed to `etl --full` or something like that, and we only garbage collect on a full run. Thoughts?

Marigold:
> I wonder when it would be best to do this. Perhaps the bare etl command should be changed to etl --full or something like that, and we only garbage collect on a full run. Thoughts?

That's how I do it [in walden](https://github.com/owid/etl/pull/219) - there's an optional `--prune` command that takes care of garbage collection and is run at the end. (It might be as well a separate command or [group](https://click.palletsprojects.com/en/8.1.x/commands/#group-invocation-without-command), we'll see what looks best in the end)

PR #222, opened 2022-06-16
feat(prune): command for pruning datasets without steps

Marigold:
Backlink https://github.com/owid/etl/issues/218

Add `prune` command to `etl` that cleans up dataset folders without proper step in `dag.yml`. This is especially useful for backporting as some datasets might get archived or deleted.


etl/prune.py
```diff
@@ -0,0 +1,83 @@
...
+    dag_dirs = dag_datasets_dirs(dag_path)
+    data_dirs = local_datasets_dirs(data_dir)
+
+    to_delete = set(data_dirs) - set(dag_dirs)
+
+    for path in to_delete:
+        log.info("prune.delete", path=data_dir / path)
+        if not dry_run:
+            shutil.rmtree(data_dir / path)
+
```
larsyencken, line 56:
```suggestion

    if to_delete:
        # the index on disk is out of date, reindex
        LocalCatalog(data_dir, channels=CHANNEL.__args__).reindex()

```

larsyencken (approved):
It looks good to ship, just one suggested addition to avoid surprises, which is to always reindex after pruning if anything changed.

PR #227, opened 2022-06-17
Fix/food explorer v1

lucasrodes:
There were some issues with region aggregates, specifically in Europe and Oceania.

This was happening because changes in https://github.com/owid/etl/commit/bca95983e7bd6c8a276371d92fa44e89cf4b6c6b were not being reflected in this pipeline.


Marigold (approved):
Looks good, I couldn't find any more issues with Turkey in ETL repo, hope this is the last one.

PR #235, opened 2022-06-21
optimize per capita for categorical variables

Marigold:


etl/steps/open_numbers.py
```diff
@@ -67,6 +67,13 @@ def run(dest_dir: str) -> None:
                 print("  ERROR: skipping")
                 continue
 
+            except KeyError as e:
+                if e.args == ('type',):
+                    print("  ERROR: skipping due to KeyError('type')")
+                    continue
+                else:
+                    raise e
+
```
pabloarosado, line 76: Was this change here by accident?

PR #236, opened 2022-06-21
enhance(examples): improve examples and getting started

Marigold:
Previously our examples needed to download 5GB Hyde dataset which is not a great way to get started. This is fixed and examples are cleaned up to be useful for newcomers.


danyx23:
I think this is a good start! I have a bunch of larger restructuring ideas for the docs but I would build these on top of what you did here. Good to merge from my side.

danyx23 (approved):

PR #244, opened 2022-06-22
Refactor dependency sorting/filtering for DAG

wcox13:


etl/steps/__init__.py
```diff
@@ -73,22 +72,13 @@ def compile_steps(
 def to_dependency_order(
     dag: DAG, includes: List[str], excludes: List[str]
 ) -> List[str]:
-    # reverse the graph so that dependencies point to their dependents
-    graph = reverse_graph(dag)
```
wcox13, line 77: We had some methods expecting an input A->B to mean "A depends on B", while for others A->B meant "B depends on A". I think it's clearer now - I've flipped things around in a few places so that we always use the same directionality as the original `dag` passed in (namely, A->B means "A depends on B")

etl/steps/__init__.py
```diff
@@ -73,22 +72,13 @@ def compile_steps(
...
-
-    # make sure dependencies are built before running the things that depend on them
-    in_order = topological_sort(subgraph)
+    """
+    Organize the steps in dependency order with a topological sort. In other words,
+    the resulting list of steps is a valid ordering of steps such that no step is run
+    before the steps it depends on. Note: this ordering is not necessarily unique.
+    """
+    subgraph = filter_to_subgraph(dag, includes) if includes else dag
+    in_order = list(graphlib.TopologicalSorter(subgraph).static_order())
```
wcox13, line 81: Inlining this, since it's shorter now without all the reversing + is the whole point of the `to_dependency_order` method

etl/steps/__init__.py
```diff
@@ -98,6 +88,52 @@ def to_dependency_order(
...
+        s for s in all_steps if any(re.findall(pattern, s) for pattern in includes)
+    }
+
+    if incl_forward:
+        # Reverse the graph to find all nodes dependent on included nodes (forward deps)
+        forward_deps = set(traverse(reverse_graph(graph), included))
+        included = included.union(forward_deps)
+
+    # Now traverse the other way to find all dependencies of included nodes (backward deps)
+    return traverse(graph, included)
```
wcox13, line 116: Let me know if I missed something, but I'm pretty sure from my unit tests that the 3 separate BFS's + parent_frontier/child_frontier stuff isn't necessary. We can get the same result by simply BFS'ing in each direction, and we can also make the BFS generic, so I did that here

tests/test_steps.py
```diff
@@ -70,12 +70,18 @@ def temporary_step() -> Iterator[str]:
...
+def test_dependency_ordering():
     "Check that a dependency will be scheduled to run before things that need it."
     dag = {"a": ["b", "c"], "b": ["c"]}
     assert to_dependency_order(dag, [], []) == ["c", "b", "a"]
 
 
+def test_dependency_filtering():
+    dag = {"e": {"a"}, "c": {"b", "d"}, "b": {"a"}}
+    assert filter_to_subgraph(dag, ["b"]) == { "d": set(), "c": {"b", "d"}, "b": {"a"}, "a": set()}
+    assert filter_to_subgraph(dag, ["b"], incl_forward=False) == {"b": {"a"}, "a": set()}
```
wcox13, line 82: This verifies the new default behavior, coming in later PR

tests/test_steps.py
```diff
@@ -70,12 +70,18 @@ def temporary_step() -> Iterator[str]:
...
 
-def test_topological_sort():
+def test_dependency_ordering():
     "Check that a dependency will be scheduled to run before things that need it."
     dag = {"a": ["b", "c"], "b": ["c"]}
     assert to_dependency_order(dag, [], []) == ["c", "b", "a"]
 
 
+def test_dependency_filtering():
+    dag = {"e": {"a"}, "c": {"b", "d"}, "b": {"a"}}
```
wcox13, line 80: Same DAG as @larsyencken 's diagrams (super helpful by the way!) - just opposite direction convention (see comment above on reversing)

etl/steps/__init__.py
wcox13: (In addition to the unit test, I also tested that `etl data://garden/wb/2021-07-01/wb_income --dry-run` seems to generate the same 28 steps before and after my change)

wcox13:
I have a couple lint issues but I just put all these comments down... going to fix those after people read the comments

Marigold (approved):
Awesome! I agree this needed refactor, we were piling up new functionality on the go.

etl/steps/__init__.py
```diff
@@ -98,6 +88,50 @@ def to_dependency_order(
...
+
+
+def traverse(graph: Graph, nodes: Set[str]) -> Graph:
```
larsyencken, line 117: The terminology `traverse` is too neutral for a DAG, where you are only traversing in one direction.

If the graph is in `node -> dependencies` direction, then this would be better called `ancestor_subgraph()`

etl/steps/__init__.py
```diff
@@ -98,6 +88,50 @@ def to_dependency_order(
...
 
 
+def filter_to_subgraph(
+    graph: Graph, includes: Iterable[str], incl_forward: bool = True
+) -> Graph:
+    """
+    Filter the full graph to only the included nodes, and all their dependencies.
+
+    If the incl_forward flag is true, also include "forward dependencies" (dependents),
+    and their own (backward) dependencies.
```
larsyencken, line 98: Perhaps "ancestors" and "descendants" is good terminology here

larsyencken (approved):
A nice improvement! Minor suggestions on terms.

etl/steps/__init__.py
wcox13: 👍  you're right that it's not a full traversal - but also it's used in both logical directions if you look in the above code, and is meant to be general in that way (to avoid rewriting the same logic in the other direction like we were doing before). 

Maybe just `bfs` or `find_reachable` then?

PR #245, opened 2022-06-23
enhance(food_explorer): turn old food explorer into jupytext

Marigold:
Convert old food explorer into jupytext to make it easier to maintain and avoid weird errors with too many open files. It was converted to `.py` in VSCode and then slightly modified by adding function `def run(dest_dir: str) -> None:` at the end (just like in `jupytext_example.py` example). I've compared outputs of the old notebook file and jupytext and they match.

I had to add quite a few type hints to comply with mypy (it's not trivial to just exclude an entire file) and `print` commands are being sent to stdout when running the ETL which makes it a bit messy, but all in all it has been pretty straightforward.

Do we want to merge this @lucasrodes @pabloarosado or should we rather keep it as notebook and wait for the new food explorer?


pabloarosado (approved):

pabloarosado:
Hi @Marigold , if you have checked that each of the output files is identical to the old one, then I think having the code in `.py` would be helpful. So, yes, I'd say we should merge.

PR #249, opened 2022-06-24
chore: add template dot env file

pabloarosado:
Add template .env file to repos.


larsyencken:
Sorry for the double-up @pabloarosado !

PR #254, opened 2022-06-27
feature(fasttrack): automatically trigger backport on updates in DB

Marigold:
Fastrack script is meant to backport updated datasets as soon as possible. It listens to grapher DB for updates and when it detects a change it will:

1. Run `backport` to get the backported dataset to **local** Walden catalog (without uploading to S3)
2. Run ETL on the backported dataset
3. Publish new ETL catalog to S3

All these steps have been optimized to run in a few seconds (except of huge datasets) and make them available through [data-api](https://github.com/owid/data-api).

### DIFF
--- a/backport/fasttrack.py
+++ b/backport/fasttrack.py
@@ -54,12 +54,10 @@ def fasttrack(
     engine = get_engine()
 
     while True:
-        # TODO: ORM might be nicer
         q = """
         select
             id as dataset_id,
             name as dataset_name,
-            -- NOTE: updatedAt might be unnecessary
             GREATEST(updatedAt, metadataEditedAt, dataEditedAt) as latest_timestamp
         from datasets
         -- this assumes there are no ties if we are processing a lot of datasets
--- a/backport/fasttrack.py
+++ b/backport/fasttrack.py
@@ -1,13 +1,16 @@
 import concurrent.futures
 import datetime as dt
 import time
+from typing import cast
 
 import click
 import pandas as pd
 import structlog
 from owid.catalog.utils import underscore
 from owid.walden import CATALOG as WALDEN_CATALOG
+from sqlalchemy.engine import Engine
 
+from etl import config
 from etl.command import main as etl
 from etl.db import get_engine
 from etl.publish import publish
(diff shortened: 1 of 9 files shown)


backport/fasttrack.py
```diff
@@ -0,0 +1,140 @@
...
+            continue
+        else:
+            log.info("fasttrack.processing", dt_start=dt_start, datasets=len(df))
+
+        # use latest timestamp of processed datasets as start for next batch
+        dt_start = df.latest_timestamp.max().to_pydatetime()
+
+        # run backport to walden in parallel
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            executor.map(
```
larsyencken, line 85:
```suggestion
            # add it to local walden if missing or checksums are out of date; on prod, a cron job will commit
            # walden changes to Git every 6 hours
            executor.map(
```

backport/fasttrack.py
```diff
@@ -0,0 +1,140 @@
...
+                lambda r: process_dataset(
+                    r.dataset_id, r.dataset_name, dry_run=dry_run, force=force
+                ),
+                df.itertuples(),
+            )
+
+        # refresh walden catalog manually
```
larsyencken, line 92:
```suggestion
        # refresh local walden catalog manually
```

backport/fasttrack.py
```diff
@@ -0,0 +1,140 @@
...
+        # reindex and publish catalog
+        reindex(
+            channel=["backport"],
+            include=r"|".join([f"dataset_{ds_id}_" for ds_id in df.dataset_id]),
+        )
+        log.info("fasttrack.end", dt_start=dt_start)
+        publish(
+            dry_run=dry_run,
+            private=True,
+            bucket="owid-catalog-staging",
```
larsyencken, line 116: Should this bucket setting be global or from `.env` ?

larsyencken (approved):
Small suggestions for readability, and one bucket setting to fix, otherwise good to go

backport/fasttrack.py
Marigold: I'm gonna add CLI arg for now to be consistent with `publish`, but will change both to take it from `.env` instead in future PR.

PR #256, opened 2022-06-28
data(countries-regions): add former Sudan and Gaza Strip to countries…

pabloarosado:
Add "Sudan (former)" and "Gaza Strip" to countries-regions dataset.


Marigold (approved):
Just a warning that this isn't gonna be reflected in the old food explorer - it uses its own regions definitions (we've had a problem with that recently https://github.com/owid/etl/pull/227/files).

PR #271, opened 2022-07-05
feature(snapshots): add Snapshots as an alternative to Walden

Marigold:
Experimental replacement of Walden by [DVC](https://dvc.org/) demonstrated on Maddison GDP dataset.

`dvc` is essentially `git` for large data that supports similar set of commands.

Comparison with Walden:
* `dvc push` is pretty much the same as `walden.add_to_catalog`
* walden has special objects `Dataset` and `Catalog` that impose structure and rules on the files, `dvc` does not use that (is that a pro or con?)
* walden compares md5 of local files and index in git, `dvc` compares it against S3
* TBD

## Update

I polished this a bit, added support for private snapshots and its own walkthrough snapshot. Highlights

* `Snapshot` object wrapped around metadata & the file, it works just like `owid.catalog.Dataset` which feels nice
* snapshots are saved to s3://owid-catalog/snapshots (or s3://owid-catalog/snapshots-private)
* hyde & ggdc were converted to snapshots
* run `walkthrough snapshot --dummy-data` to generate snapshot template


Marigold:
We might not use this in the short term, but since we're revisiting data architecture and playing with git-lfs, I'd keep it open for some more time.

etl/steps/__init__.py
```diff
@@ -492,6 +508,47 @@ def _walden_dataset(self) -> WaldenDataset:
...
+
+        # if path is in dvc status output, it's dirty
+        return self._dvc_path in cache["dvc_status"]
+
+    def has_existing_data(self) -> bool:
+        return True
+
+    def checksum_output(self) -> str:
+        with open(self._dvc_path, "r") as f:
+            return yaml.safe_load(f)["outs"][0]["md5"]
```
larsyencken, line 537: The checksum of this step should be of the combination of the data and the metadata, which is what Walden steps do in `checksum_output()`.

Probably we should stuff metadata into the `.dvc` file itself and just return the checksum of that file here, instead of reporting the checksum of the binary file that DVC stored.

snapshots/ggdc/2020-10-01/ggdc_maddison/.gitignore
```diff
@@ -0,0 +1 @@
+/ggdc_maddison.xlsx
```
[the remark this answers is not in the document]
Marigold: Added them both to top-level gitignore. (Just to be clear - this `.gitignore` gets created automatically by `dvc add snapshots/ggdc/2020-10-01/ggdc_maddison/ggdc_maddison.xlsx`. Although as you say, if it is already ignored in top-level `.gitignore` it won't get created)

etl/steps/__init__.py
Marigold: > Probably we should stuff metadata into the .dvc file itself and just return the checksum of that file here, instead of reporting the checksum of the binary file that DVC stored.

Yeah, good idea. I was a bit worried about having metadata in `.dvc` file, but it feels good after some refactoring.

pabloarosado (approved):
It looks good! I'll try it out now with a new dataset.
Some minor comments and considerations:
* For cases where the data is fetched for example from an API, it would be convenient if there was a method to ingest a dataframe directly (which would create a file in the right place), instead of having to have additional code in the ingest script to create a local file.
* I think the data should be inside `data/snapshot` instead of being next to the data step code.
* There are a few instances of "Walden" in the code that should be removed.

Marigold:
> I think the data should be inside data/snapshot instead of being next to the data step code.

Good idea, fixed that. Class `Snapshot(...)` now accepts `namespace/version/filename.ext` instead of full path.

> There are a few instances of "Walden" in the code that should be removed.

I'm gonna clean it up after we go all in on Snapshots.

> For cases where the data is fetched for example from an API, it would be convenient if there was a method to ingest a dataframe directly (which would create a file in the right place), instead of having to have additional code in the ingest script to create a local file.

That'd be cool. I'll rewrite your function `add_to_catalog` to work with snapshots.

Marigold:
@pabloarosado I've also added `add_snapshot` helper function and tested it on [aviation safety network](https://github.com/owid/etl/pull/271/commits/658fdd966788b27cec5711385ec9cb7cff595a44). I hope this clarifies how can it work with dataframes.

larsyencken (approved):
Good to merge :shipit: 

### What should be the workflow for metadata?

- Currently
  1. Make a placeholder `snapshot/<filename>.dvc` file with metadata
  2. Make an ingest script `snapshot/<filename>.py` that materialises the placeholder
  3. Run the ingest script, which updates-in-place `snapshot/<filename>.dvc`
- Good stuff
  - Only specify metadata once
  - You can re-run the ingest script over and over and get new snapshots
- Bad stuff
  - Updating in place: not clean separation of input and output

### Where should the .dvc file live?

- Now it lives in `snapshots/<filename>.dvc`, next to the ingest script
  - Downside is that it's not next to the data file
- Other option is `data/snapshots/<filename>.dvc`, next to the data file
  - Downside of this is that `data/` is so far a temporary folder (excl. countries and regions)

PR #274, opened 2022-07-06
WIP: enhance(etl): option for filtering warnings

Marigold:
Warnings from steps end up in stderr and show up as errors in our monitoring (we use `slacktee` to send all stderr to slack). Since some pandas warnings are hard to fix, it's easier to just filter them in production. Only `FutureWarning` gets silenced at the moment (because of faostat steps), but `SettingWithCopyWarning` might be a next one (because of un sdg where this was a real pain).

An alternative would be replacing `stderr -> slacktee` by catching exceptions in code and sending them to Slack with a package or service (Bugsnag, Sentry, Datadog...).

EDIT: hmm, still thinking through whether this is a good idea or not

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -8,6 +8,7 @@ import os
 import time
 import sys
 import re
+import warnings
 from pathlib import Path
 
 import click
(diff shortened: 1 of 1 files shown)


Marigold:
I was worried that we might start getting a lot of warnings after pandas update (which frequently happens) and fixing all of them would be painful. Anyway, we don't have that problem now and we've fixed those warnings, so closing for now.

PR #275, opened 2022-07-06
feature(errors): add bugsnag for error handling

Marigold:
Improvement over slacktee error tracking with better tracebacks and possibility to add metadata to exceptions. Bugsnag has additional features that could make our life easier.


larsyencken (approved):
Sounds good, let's try it!

PR #280, opened 2022-07-11
Add compare tool

danyx23:
This PR implements the first version of a compare CLI that allows easy comparison of either arbitrary dataframes (feather, csv, parquet) or etl files (in that case against the version currently uploaded in the production catalog). Much of the logic for the diffing lives in the data-utils-py repo and is managed in this PR: https://github.com/owid/data-utils-py/pull/29 . 

This PR currently includes a temporary copy of the new HighlevelDiff class for development. It will be removed here and imported from data-utils-py when the PR over there is merged.

### DIFF
--- a/etl/compare.py
+++ b/etl/compare.py
@@ -1,0 +1,251 @@
+#
+#  compare.py
+#
+
+from pathlib import Path
+import click
+from rich_click import RichCommand, RichGroup
+from rich import print
+from typing import cast
+import pandas as pd
+from etl import tempcompare
+from owid import catalog
+
+
+@click.group(cls=RichGroup)
+def cli():
+    """Compare two dataframes, both structurally and the values.
+
+    This tool loads two dataframes, either from the local ETL and the remote catalog
+    or just from two different files. It compares the columns, index columns and index values (row indices) as
+    sets between the two dataframes and outputs the differences. Finally it compares the values of the overlapping
+    columns and rows with the given threshold values for absolute and relative tolerance.
+
+    The exit code is 0 if the dataframes are equal, 1 if there is an error loading the dataframes, 2 if the dataframes
+    are structurally equal but are otherwise different, 3 if the dataframes have different structure and/or different values.
+    """
+    pass
+
+
+def diff_print_and_exit(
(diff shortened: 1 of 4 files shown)


danyx23:
As I mentioned above, most of the diffing logic is done in the other PR. If you can, please focus on the cli itself in this one. To run it, do
```bash
poetry install
poetry shell
compare dataframes DF1 DF2 
# or
compare etl_catalog CHANNEL NAMESPACE DATASET TABLE
```

pabloarosado:
Hi Daniel, thanks a lot for this! The CLI looks very neat. I can't go through the PR now in detail, but I have some general preliminary comments:
* When you have many columns with diffs, the formatting gets a bit messy:
Screen Shot 2022-07-11 at 14 01 50
* Is your plan to move this back to data-utils? I think it makes sense to keep the tool there, instead of having it split between the two repos.

etl/compare.py
```diff
@@ -0,0 +1,273 @@
...
+        "local",
+        absolute_tolerance,
+        relative_tolerance,
+        show_values,
+        show_shared,
+        truncate_lists_at,
+    )
+
+
+def load_table(path_str: str) -> catalog.Table:
```
Marigold, line 148: (There's a [similar method](https://github.com/owid/owid-catalog-py/blob/master/owid/catalog/catalogs.py#L329) that could be used instead, but it'd probably look weirder than your method)

etl/compare.py
```diff
@@ -0,0 +1,273 @@
...
+    help="Show the structural overlap of the two dataframes (shared columns, index columns and index values).",
+)
+@click.option(
+    "--truncate-lists-at",
+    default=20,
+    show_default=True,
+    help="Print truncated lists if they are longer than the given length.",
+)
+def etl_catalog(
+    channel: str,
```
Marigold, line 100: (For me it would be more natural to write it as a path / URI `channel/namespace/dataset/table` instead of having them separated)

Marigold (approved):
Look good! Couple minor comments (feel free to ignore). I'm gonna try it on real datasets when I have a chance, but that shouldn't prevent merging this.

PR #287, opened 2022-07-13
data(countries_regions): update country name

gilmourj:
Fixes typo in name for Saint Barthélemy


lucasrodes:
Thanks for the fix, @gilmourj!

@danyx23, do you know if this is safe to merge or is there any other project that relies on the field `name`?

PR #291, opened 2022-07-17
Raise error when no steps are found for a given input

wcox13:
Now, `etl asdf` throws an error (since there are no matching steps), instead of saying `All datasets are up to date!`

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -180,6 +180,11 @@ def run_dag(
 
     steps = compile_steps(dag, includes, excludes, downstream=downstream, only=only)
 
+    if not steps:
+        raise ValueError(
+            "No steps matched the given input. Check spelling or consult `etl --help` for more options"
+        )
+
     if not force:
         print("Detecting which steps need rebuilding...")
         steps = select_dirty_steps(steps, workers)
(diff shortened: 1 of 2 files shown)


etl/command.py
```diff
@@ -180,6 +180,11 @@ def run_dag(
 
     steps = compile_steps(dag, includes, excludes, downstream=downstream, only=only)
 
+    if not steps:
+        raise ValueError(
```
wcox13, line 184: I think this should be an error rather than a warning right? The "all datasets up-to-date" case is handled below - if there are no steps here, that means invalid input

etl/command.py
```diff
@@ -180,6 +180,11 @@ def run_dag(
 
     steps = compile_steps(dag, includes, excludes, downstream=downstream, only=only)
 
+    if not steps:
```
wcox13, line 183: the other option: check this in the `filter_to_subgraph` method. there, we could verify that every step in `includes` actually matches something in the dag. that would allow us to throw a more specific error - ie in the case that someone passed multiple things to include, we could tell them which one specifically didn't match.

i think the simple way is enough for now though?

Marigold: Right, error makes more sense.

Marigold (approved):
Simple yet useful!

etl/command.py
Marigold: Simple is better, if someone is running multiple steps then she's an advanced user and knows what to except.

PR #293, opened 2022-07-19
Improve harmonizer

pabloarosado:
* Add "institution" as an optional argument.
* When suggesting a new country name, let "Country (Institution)" always be the zeroth option (if "institution" argument was given).
* When suggesting a new country name, let user type a new country name that does not exist in the countries-regions dataset.

### DIFF
--- a/etl/harmonize.py
+++ b/etl/harmonize.py
@@ -23,7 +23,9 @@ from etl.paths import REFERENCE_DATASET
 @click.argument("column")
 @click.argument("output_file")
 @click.argument("institution", required=False)
-def harmonize(data_file: str, column: str, output_file: str, institution: Optional[str] = None) -> None:
+def harmonize(
+    data_file: str, column: str, output_file: str, institution: Optional[str] = None
+) -> None:
     """
     Given a data file in feather or CSV format, and the name of the column representing
     country or region names, interactively generate a JSON mapping from the given names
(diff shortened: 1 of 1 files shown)


etl/harmonize.py
```diff
@@ -22,7 +22,10 @@
 @click.argument("data_file")
 @click.argument("column")
 @click.argument("output_file")
-def harmonize(data_file: str, column: str, output_file: str) -> None:
+@click.argument("institution", required=False)
```
Marigold, line 25: Have you considered using `@click.option` instead? Citing [click.argument](https://click.palletsprojects.com/en/8.1.x/api/#click.argument)

> This is to follow the general convention of Unix tools of using arguments for only the most necessary things, and to document them in the command help text by referring to them by name.

Using `click.option` would also let you add `help=` clarifying what does it do.

Marigold (approved):
Added one minor note, otherwise LGTM.

etl/harmonize.py
pabloarosado: I think I prefer the current option. If you see the help message, it looks like:
```
Usage: harmonize [OPTIONS] DATA_FILE COLUMN OUTPUT_FILE [INSTITUTION]
...

Options:
  --help  Show this message and exit.
``` 
which is clearer than when using `click.option("-i", "--institution")`:
```
Usage: harmonize [OPTIONS] DATA_FILE COLUMN OUTPUT_FILE
...
Options:
  -i TEXT  Optionally add institution name
  --help   Show this message and exit.
```

PR #295, opened 2022-07-19
feature(walkthrough): walkthrough for creating grapher step

Marigold:
Adding `walkthrough grapher` step. It contains only the basic template for the script. The weirdest thing is having to change your `.env` to test it out on your local / staging / production. We should revisit that if it turns out to be a bottleneck.


pabloarosado (approved):
I'd suggest to remove all instances of ggdc names and URLs. Using "dummy" is probably a better idea.

Marigold:
> I'd suggest to remove all instances of ggdc names and URLs. Using "dummy" is probably a better idea.

Hmm, that's probably a good idea. I wanted to have real dataset used in placeholders, but perhaps it just confuses users.

PR #296, opened 2022-07-20
data: UN WPP meadow and garden steps

lucasrodes:


lucasrodes:
Pending merge from https://github.com/owid/walden/pull/56 so the pipeline can be tested.

Dataset version has changed from `2022` to `2022-07-11`

PR #297, opened 2022-07-20
Add shift data steps

pabloarosado:
Added Shift data to etl, ensuring region aggregates are well constructed.


Marigold (approved):
Neat! Your work on historical regions looks great, that's gonna be very useful for other datasets too.

pabloarosado:
> Neat! Your work on historical regions looks great, that's gonna be very useful for other datasets too.

Thanks. On every new dataset, I keep changing the logic a bit. Hopefully, after several iterations, we'll figure out a generic way to get this done and transfer it to datautils.

PR #298, opened 2022-07-21
fix(steps): run notebook with proper working directory

Marigold:
Previously we weren't able to import modules from the same directory as notebook (e.g. `from shared import ...`). This sets working directory correctly.


pabloarosado (approved):
Great, this is useful. But does it also work with `.py` files, in other words, will `etl` run well when importing something like `from shared import ...`?

pabloarosado (approved):
This looks great!

PR #308, opened 2022-07-25
enhance(open_numbers): parallelise table imports

larsyencken:
Several open_numbers datasets are large, and importing them can take a long time when done table by table.

This PR parallelises each open_numbers dataset import, whilst leaving the overall process serial.

## Benchmarks

On a 10-core M1 Max, running `rm -f data/open_numbers; etl open_numbers`

Before: `2352.86s user 130.22s system 99% cpu 41:35.25 total`
After: `4143.78s user 554.04s system 654% cpu 11:57.23 total` (3.5x faster)

### DIFF
--- a/etl/steps/open_numbers.py
+++ b/etl/steps/open_numbers.py
@@ -56,10 +56,15 @@ def run(dest_dir: str) -> None:
     # copy tables one by one
     with ProcessPoolExecutor() as executor:
         for short_name, resources in resource_map.items():
-            executor.submit(add_resource, ds, short_name, resources)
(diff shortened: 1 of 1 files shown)


larsyencken:
Good catch! When I bench it now, `concurrent.futures.ProcessPoolExecutor` is waaaay slower than `multiprocessing.Pool`, probably because of batching behaviour.

I'll check the final numbers, and will only merge this if it's substantially faster than the single-threaded version.

larsyencken:
@Marigold It occurred to me that now that we might run into memory issues on the analytics server now after this change. If we do, we can revert it.

PR #312, opened 2022-07-26
refactor(shift): Rename Shift datasets and step files

pabloarosado:
To avoid redundancy, I changed the names of the Shift datasets and step files from `shift/VERSION/shift_dataset_short_name` to `shift/VERSION/dataset_short_name`.
I have done the same thing for all other energy datasets, but in a different branch.


larsyencken (approved):
Good move, I also think it's better generally not to repeat the namespace in the table name.

Tests pass, the dataset builds cleanly for me, ship it! :shipit:

PR #320, opened 2022-07-27
Add fossil fuel production dataset

pabloarosado:
Main changes:
* The BP's statistical review 2022 dataset is still processed by [importers](https://github.com/owid/importers/tree/master/bp_statreview), however, I have created an additional garden step in etl that ensures the region aggregates are well constructed. This implies that [this dataset](https://owid.cloud/admin/datasets/5650) will be replaced by another one that has the right OWID definitions of regions.
* I also added a garden step for the BP statistical review 2021. However, this is only used by the 2022 step, to fill missing data.
* [The energy mix dataset](https://owid.cloud/admin/datasets/5657) will also be replaced by a new dataset (with some minor improvements).
* I have created a garden dataset for Shift data, but this will not be used in grapher (at least for now).
* [The fossil fuel production dataset](https://owid.cloud/admin/datasets/5568) will be replaced by an updated version.
* I have added some grapher helper functions (and at the end I discovered that one of them is kind of a duplicate, although they do slightly different things). @Marigold could you please review this?

@HannahRitchie could you please review the `*.meta.yml` files? They contain the descriptions that will appear in the SOURCES tabs. Let me know if you want to make any changes. And note that the description for the BP statistical review has been copied to all datasets that use it, so you don't have to review that part every time (and if you want to make changes, I'll ensure they are changed consistently in all datasets).

Speaking of which, @Marigold, we could automatically concatenate the source's dataset descriptions, to avoid copy/pasting the same descriptions every time, right? (I'm not going to refactor this branch again, but maybe for the next datasets I could do it this way, if you think that makes sense).


etl/grapher_helpers.py
```diff
@@ -286,3 +287,143 @@ def join_sources(sources: List[catalog.meta.Source]) -> catalog.meta.Source:
...
+
+        setattr(default_source, attribute, combined_value)
+
+    # Remove other sources and keep only the default one.
+    metadata.sources = [default_source]
+
+    return metadata
+
+
+def adapt_dataset_metadata_for_grapher(
```
Marigold, line 362: This function is awesome, gonna add it to walkthrough.

etl/grapher_helpers.py
```diff
@@ -286,3 +287,143 @@ def join_sources(sources: List[catalog.meta.Source]) -> catalog.meta.Source:
             meta[key] = sep.join(keys)
 
     return catalog.meta.Source(**meta)
+
+
+# TODO: The following function and join_sources are redundant. Merge the functionality of both into one function.
+def combine_metadata_sources(metadata: catalog.DatasetMeta) -> catalog.DatasetMeta:
```
Marigold, line 293: Can you then remove `join_sources`, replace its use in `population_density` and find/replace all its mentions?

etl/grapher_helpers.py
```diff
@@ -286,3 +287,143 @@ def join_sources(sources: List[catalog.meta.Source]) -> catalog.meta.Source:
...
+        Adapted table, ready to be inserted into grapher.
+
+    """
+    table = deepcopy(table)
+    # Grapher needs a column entity id, that is constructed based on the unique entity names in the database.
+    table["entity_id"] = country_to_entity_id(table[country_col], create_entities=True)
+    table = table.drop(columns=[country_col]).rename(columns={year_col: "year"})
+    table = table.set_index(["entity_id", "year"])
+
+    # Ensure the default source of each column includes the description of the table (since that is the description that
```
Marigold, line 417: This is already done in [grapher_import](https://github.com/owid/etl/blob/master/etl/grapher_import.py#L190), but it's gonna be better to do it explicitly with your function and raise a warning if source is missing. I'll create a PR after you merge this.

etl/steps/data/garden/bp/2022-07-11/statistical_review.py
```diff
@@ -0,0 +1,281 @@
...
+    dataset.metadata.update_from_yaml(METADATA_FILE_PATH)
+    # Create new dataset in garden.
+    dataset.save()
+
+    # Add table to the dataset.
+    table.metadata.title = dataset.metadata.title
+    table.metadata.description = dataset.metadata.description
+    table.metadata.dataset = dataset.metadata
+    table.metadata.short_name = dataset.metadata.short_name
+    table.metadata.primary_key = list(table.index.names)
```
Marigold, line 280: This is done automatically.

Marigold (approved):
Sweet helpers for grapher! You should totally let others know about them.

etl/grapher_helpers.py
pabloarosado: ok, I'm already working in a different sub-branch, I'll do it there.

pabloarosado:
Hi @HannahRitchie I will merge this PR and update the datasets. Once they are live I'll let you know, so you can more easily review the descriptions, titles, etc.

PR #321, opened 2022-07-27
fix(steps): run steps in isolated environment

### LINKED ISSUE #319: Importing the wrong modules when executing various steps
I have the following structure:
-`bp/2022-07-14/shared.py`
-`bp/2022-07-14/statistical_review.py`
-`energy/2022-07-20/shared.py`
-`energy/2022-07-20/fossil_fuel_production.py`
And, inside `shared.py` I have some common paths within that particular version (for example, I have a `__file__` that lets me get the path to that specific version). If I run the steps `statistical_review` and `fossil_fuel_production` separately, there is no problem. But if etl runs the two steps one after the other, the second step fails. It seems that the `shared` that it's importing in the second step is the one of the previous step. I think this may be related to [this other small issue](https://github.com/owid/etl/pull/298) we discussed the other day.

Marigold:
Fixes #319

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -3,8 +3,11 @@
 #  etl
 #
 
-from contextlib import contextmanager
+import sys
 import tempfile
+from collections.abc import Generator
+from contextlib import contextmanager
+from pathlib import Path
 from typing import Any, Iterator, List, cast
 
 import requests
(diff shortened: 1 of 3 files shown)


pabloarosado (approved):

Marigold:
> Would it even be worth using a separate process for isolation, or do you think it's not worth it?

That's a good option, too. I honestly don't know whether there are any major pros or cons of each approach. I tend to shy away from processes unless necessary (it could complicate debugging, leave zombie processes, etc.). Do you think I should rewrite it to use process or should we go with this naive approach and only do it if we find other issues?

larsyencken:
> That's a good option, too. I honestly don't know whether there are any major pros or cons of each approach. I tend to shy away from processes unless necessary (it could complicate debugging, leave zombie processes, etc.). Do you think I should rewrite it to use process or should we go with this naive approach and only do it if we find other issues?

Let's start here, but if we find even one other isolation issue, we should just switch to processes I think.

larsyencken (approved):

PR #323, opened 2022-07-27
fix(grapher): fix dimensions when yielding long or wide table to grapher

Marigold:
Helper functions for yielding tables for grapher don't work well with dimensions. They don't support titles and had bugs in creating short names.

This PR adds the following functionality:
* Title (= variable name in grapher) is now created automatically from the original title and dimensions. For instance variable `deaths` with dimensions `age group` and `sex` would be converted into `deaths - age group: 10-18 - sex: male`. It also supports new argument `dim_titles` for custom dimension naming, e.g. `Deaths - Age group: 10-18 - Sex: male` in this case
* Short names are created similarly to title, though they're not used for anything in grapher yet
* Adds test coverage for yielding wide and long tables (see unit tests for toy examples)

`yield_long_table` groups table by variable and then yields `yield_wide_table` for every such group.

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -171,7 +171,7 @@ def yield_wide_table(
                 )
                 tab[short_name].metadata.title = title_with_dims
             else:
-                title_with_dims = None  # type: ignore
+                title_with_dims = None
 
             log.info(
                 "yield_wide_table.create_variable",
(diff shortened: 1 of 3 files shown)


larsyencken:
https://github.com/owid/etl/issues/348

larsyencken (approved):
Looked over this with Lucas, thanks so much for adding the tests. Looks good!

PR #324, opened 2022-07-27
Adding in some functions to match variables and suggest chart revisions

spoonerf:
Borrowed from importers, two functions built by Pablo and Bobbie to match variable names with an older dataset and then to use the resulting json of old and new variable names to pass into the Chart Revision Suggester.

Not sure if I've located the files with these functions in the best place or if there are plans elsewhere in the etl to deal with this part of the import process.


etl/chart_revision_suggester.py
```diff
@@ -0,0 +1,567 @@
+# type: ignore
```
Marigold, line 1: Might be useful to reference the original file and say it was copied from there.

etl/chart_revision_suggester.py
```diff
@@ -0,0 +1,567 @@
...
+from dotenv import load_dotenv
+import pandas as pd
+from tqdm import tqdm
+from pymysql.err import IntegrityError
+from etl.db_utils import DBUtils
+from etl.db import get_connection
+from etl.grapher_helpers import IntRange
+
+
+load_dotenv()
```
Marigold, line 20: You can replace this and envs below by `from etl.config import DEBUG, USER_ID`.

etl/match_variables_from_two_versions_of_a_dataset.py
```diff
@@ -0,0 +1,461 @@
...
+manually. This script is a CLI tool that may help in either scenario.
+
+"""
+
+import argparse
+import json
+import os
+from typing import Callable, Any
+import pandas as pd
+from fuzzywuzzy import fuzz
```
Marigold, line 16: This will have to be added to `etl` with `poetry add fuzzywuzzy`.

etl/steps/grapher/un_sdg/2022-07-07/output/variable_replacements.json
```diff
@@ -0,0 +1,297 @@
+{
```
Marigold, line 1: (Just a reminder that you'll have to run it against production DB, running it against local DB or staging would produce different variable ids).

Marigold (approved):
Looks good. Both chart revision and variable matcher could use some slight refactoring (adding type hints & simplifying DB connections), but that's for a different PR.

PR #326, opened 2022-07-28
refactor(sdg): remove backported SDG indicators that we have ETL version

Marigold:
We now have ETL version of SDG indicators which is better than backported version.

(Dataset `wdi__2022_05_26` came from ETL and is not being imported any more, so SDG stopped working anyway)


spoonerf:
It would be good to keep the backported version of the SDG Tracker as it has different variables in than the `un_sdg` import. 

The [SDG Tracker](https://sdg-tracker.org/) variables are made up of a bunch different datasets and are a lot more curated than what is in the `un_sdg` bulk import. 

We would ideally be able to have a download of just the data used in the SDG Tracker and for users to be able to download this separately, in a similar sense to the [CO2 repo](https://sdg-tracker.org/).

Would it be okay with you to keep it even though it's not currently working? Sorry the update of the SDG Tracker has been on the back-burner a bit this year but this backport of all the data used in the SDG Tracker is still a valuable thing to have access to.

Marigold:
@spoonerf thanks, that makes sense. I'll fix it then in a different PR.

Issue #329, opened 2022-07-28
Add metadata for variables to `data://garden/worldbank_wdi/2022-05-26/wdi`


Marigold:
## Problem

Garden step of `worldbank_wdi/2022-05-26/wdi` is missing variables metadata. These are loaded from grapher DB dynamically in its [grapher step](https://github.com/owid/etl/blob/master/etl/steps/grapher/worldbank_wdi/2022-05-26/wdi.py#L124). It would be better to generate a static file with all metadata with a script and add it already in garden step.

This is an issue because `data://garden/sdg/latest/sdg` step depends on that dataset, however the only way how to get its metadata is to get it from its backported version (which shouldn't be backported in the first place).

## Solution

* [ ] generate static `.meta.yml` file and load it in grapher step
* [ ] use `garden/worldbank_wdi/2022-05-26/wdi` in `data://garden/sdg/latest/sdg` instead of its backported version

Marigold:
@bnjmacdonald do you think this is something you'd have capacity to do? If not I'll be happy to do it for you, just wanted to get a permission first.

bnjmacdonald:
Yup, sounds like a pretty small chunk of work and I'll already be wrapping up some loose ends on the WDI bulk update this data team cycle anyways, so I will add it to my list of todos this cycle.

If you need it done earlier (i.e. before end of cycle on 28 August), let me know (or feel free to tackle it yourself).

Marigold:
No rush, it's not a blocker. Thanks!

PR #330, opened 2022-07-28
fix(sdg): backport WDI dataset that comes originally from ETL

Marigold:
This problem breaks ETL and started happening after we added checksum to WDI dataset which then stopped being imported.

It's a bit of a hotfix before we [do it properly](https://github.com/owid/etl/issues/329).


larsyencken (approved):
Looks like it should work, let's not leave it in too long though!

Issue #325, opened 2022-07-28
Refactor chart revision tool and variable matcher


Marigold:
[Chart revision suggester](https://github.com/owid/etl/blob/master/etl/chart_revision_suggester.py) and [variable matcher](https://github.com/owid/etl/blob/master/etl/match_variables_from_two_versions_of_a_dataset.py) could be refactored a bit (they were pretty much copied from importers repo).

Suggestions:
* add type hints to chart revision suggester
* simplify DB connection with [context managers](https://github.com/owid/etl/blob/master/etl/grapher_import.py#L343)
* integrate [VariableMatcher](VariableMatcher)?

pabloarosado:
I just tried the script `match_*` and it works well (although the output file argument should be mandatory, which wasn't the case in importers).
However, the `chart_revision_suggester` script does not work (at least) because:
* `simplejson` is not installed.
* `pymysql` are not installed. I would avoid installing `pymysql`, since, in importers, this was always yielding a warning (because pandas prefers `sqlalchemy`). I guess this script should use the default etl connection, as you suggested.
* `USER_ID` in importers used to be taken from the environment variables, but in `etl` it does not exist (at least it's not defined in the `.env.example` file).

PR #335, opened 2022-07-29
refactor(chart_revision_suggester): add typing and fix issues

Marigold:
Refactor chart revision tool based on @pabloarosado [suggestions](https://github.com/owid/etl/issues/325#issuecomment-1198192076). I haven't changed its logic in any way, just fixed the necessary stuff to be able to run it. I couldn't find any real example to test it on, so if you see any issues, please don't be afraid to change whatever you want.

(Also sorted some imports automatically with isort, let me know if it's annoying noise in the PR).


pabloarosado (approved):

pabloarosado:
> (Also sorted some imports automatically with isort, let me know if it's annoying noise in the PR).

I started using an "optimize imports" option in pycharm, but it treats owid-catalog as a local import, whereas isort seems to treat it as a third party import. Which one is correct? Given that it's a submodule it is something in between.

PR #344, opened 2022-08-01
Multi-format support including parquet

### LINKED ISSUE #332: Co-publish data in multiple formats (csv, feather, parquet)
## Background

We want to make it easy to browse our data and get it in a variety of formats. Today the ETL publishes the dataset it generates through the `publish` command. Each ETL step itself decides on what format to use at the moment the table is added to the dataset: [`Dataset.add(table)`](https://github.com/owid/owid-catalog-py/blob/574dc31457ca0d5c6790446d2ed8e164143e259e/owid/catalog/datasets.py#L65). Current options are `csv` (for explorers) or `feather`.

## Proposal

We should pick a set of formats (csv, feather, parquet) and generate the data in _all those formats_, rather than picking one.

```mermaid
graph LR
etl[ETL step] --> memory[in-memory]
memory --> csv
memory --> feather
memory --> parquet
csv --> s3
feather --> s3
parquet --> s3
```

### Upsides

- Consistency between all our published datasets, no need to remember what format the data was in
- Faster development and prototyping for us, since we can always find a format that a new client can use
  - New API can more easily consume parquet files

### Downsides

- BIG on-disk footprint, especially due to CSV files (~10x? 5G -> 50G?)
- Slower ETL due to extra format generation
- Slower `publish` step due to more files to sync to S3

## Scope

- [x] https://github.com/owid/owid-catalog-py/pull/53
- [ ] https://github.com/owid/etl/pull/344
- [ ] https://github.com/owid/owid-catalog-py/pull/55

larsyencken:
Fixes #332

This change enables support for co-publishing in `feather` and `parquet` formats. The `parquet` format includes the full table metadata inside the Parquet schema. At this time, `csv` was disabled by default since it would 50x the catalog size.

The motivation for this was to support quick (or even, lazy) DuckDB indexing for the API.

## Updates

- Code forcing an output format is updated, in particular `reference/countries_regions` and `**/food_explorer` (`ds.add(t, format='csv') --> ds.add(t, formats=['csv'])`).
- Channels are reindexed in parquet as well as feather
- Publishing co-publishes parquet indexes

## After merging

This PR only causes newly rebuilt datasets will get published this way. A full forced rebuild of the ETL is recommended to generate the first batch of `parquet` files.

### DIFF
--- a/etl/harmonize.py
+++ b/etl/harmonize.py
@@ -286,7 +286,7 @@ def save_alias(name: str, alias: str) -> None:
 
     # pack up and save
     rc.loc[rc.name == name, "aliases"] = json.dumps(sorted(aliases))
-    ref.add(rc, "csv")
+    ref.add(rc, formats=["csv"])
 
 
 def print_mapping(region: str, name: str) -> None:
--- a/etl/publish.py
+++ b/etl/publish.py
@@ -14,9 +14,11 @@ import concurrent.futures
 import click
 import boto3
 from botocore.client import ClientError
-import pandas as pd
+from owid.catalog.catalogs import INDEX_FORMATS
+from owid.catalog.datasets import FileFormat
 
 from owid.catalog import LocalCatalog, CHANNEL
+from owid.catalog.helpers import read_frame
 
 from etl import config, files
 from etl.paths import DATA_DIR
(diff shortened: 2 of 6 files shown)


Marigold (approved):
Looks good!

PR #347, opened 2022-08-01
Improved docs

danyx23:
This is a first stab at explaining a bit more what our ETL is all about. Still TBD how we should split the documentation between this and the readme.

Still to be added are three getting started document:
* setup, prerequisites and guides for debugging setup issues
* rebuilding specific datasets and making small modifications/bugfixes
* adding a new dataset from scratch (this has gotten much easier thanks to @Marigold's wizard tool)


Marigold:
Wizard! That's the word I've been looking for. I never liked "walkthrough"...

Marigold (approved):
I think it's great, but we should make sure it's reachable from somewhere. I can link there from the wizard, but perhaps it deserves a space of its own? (e.g. in `README`? or at least a link from there?)

larsyencken (approved):
Nice! I would simply merge this in then make new PRs for improvements from here.

danyx23:
@larsyencken this is my stab at integrating my writing with yours. I basically opted to make your version of the intro the history doc, streamlined my version a bit and added some mermeaid graphs. All of this writing can be edited/chopped of course. Let me know what you think, if you think this is decent we could merge and continue from there.

Issue #348, opened 2022-08-01
VariableMeta source description mix up


spoonerf:
Link to relevant dataset: https://staging.owid.cloud/admin/datasets/5628 & branch: https://github.com/owid/etl/blob/sdg_fix/etl/steps/grapher/un_sdg/2022-07-07/un_sdg.py#L93

I'm having a strange issue with the Source description in VariableMeta. 

For each variable I want to create a link to the relevant metadata pdf and this follows a more or less predictable pattern where the indicator is in the url path. As far as I can tell my code is creating the correct urls for the relevant variables.
![Screenshot 2022-08-01 at 16 47 22](https://user-images.githubusercontent.com/10774728/182188535-17ce24af-77e6-4f91-a57b-83c1cee5a96b.png)

But when I check on staging a handful of the urls are incorrect - for example the below _should_ link to https://unstats.un.org/sdgs/metadata/files/Metadata-17-13-02.pdf but it is https://unstats.un.org/sdgs/metadata/files/Metadata-17-13-01.pdf.

![Screenshot 2022-08-01 at 16 44 16](https://user-images.githubusercontent.com/10774728/182187879-77f5297d-2a93-4eb4-a0f1-403793fbd6e6.png)

spoonerf:
Is it possible that something is going on within `yield_long_table()`? As far as I can tell the Source description metadata is correct before the table is yielded.

Marigold:
@spoonerf taking a look. Function `yield_long_table` is [awaiting some fixes](https://github.com/owid/etl/pull/323), so it's possible it has something to do with it.

Marigold:
The problem is that if multiple different sources have the same name (i.e. World Bank), only a single source will be created in a database. `source.name` works more as a `short_name` in that it's not shown anywhere in grapher, but is used as an identifier for upserting.

We should therefore have unique `source.name` for different sources. Here's a [PR that validates it](https://github.com/owid/etl/compare/348-variablemeta-source-description-mix-up?expand=1).

PR #353, opened 2022-08-02
WIP: enhance(steps): check that different sources have different names

Marigold:
Validation for https://github.com/owid/etl/issues/348

When iterating over tables from `get_grapher_tables` we keep a dictionary of `source.name -> source hash` and check that there aren't two sources with the same name, but different contents. This is done with generator (let me know if this is overkill, but the code was cleaner). If this was the case, sources would be overwritten (over-upserted) and user wouldn't even notice.


Marigold:
I've realised this could happen for titles and short_names too (especially when someone is using `yield_long_table`, this would have prevented bug with dimensions), so added too as a prevention.

Marigold:
Might not be worth the extra code, such edge case should be quite rare.

PR #354, opened 2022-08-02
UN SDG: Making garden tables unique and ensuring source name is unique

spoonerf:
* Previously used just the series code as the Garden table short name but have changed this to `indicator_seriesCode` as some series codes are shared across indicators.
* Ensuring Source.name is unique so that Source.description can vary when Data Publisher's source is the same.


spoonerf:
Moved the source description into the variable description as this doesn't face the same constraints.

etl/steps/grapher/un_sdg/2022-07-07/output/variable_replacements.json
```diff
@@ -1,297 +1,155 @@
...
-    "417077": "431304",
-    "417079": "429400",
-    "417082": "429370",
-    "417100": "430041",
-    "417159": "430042",
-    "417272": "430121",
-    "417273": "430122",
-    "417421": "431708"
-}
+    "414079": "456880",
```
Marigold, line 2: (The list got shorter, just checking whether this is expected)

spoonerf: Ah yeh I think that's okay - the variable matching and chart revisions is still a manual step. I wonder if we should add it as a dag step or if that is overkill, I don't think we would want it run every time but maybe we don't want manual bits of code hanging around in the etl?

Marigold: Hmm, interesting idea. Maybe not exactly a dag step, but some helper command would be helpful.

I really like your `match_vars_crs.py` script for simplicity though. If we decide to automate it, I'd just steal it anyway.

spoonerf: Yeh maybe an optional flag would be cool! okay will merge this now :)

PR #355, opened 2022-08-02
Add ember steps

pabloarosado:
Add meadow and garden steps for Ember's Global Electricity Review 2022 and Ember's European Electricity Review 2022.


etl/steps/data/garden/ember/2022-08-01/european_electricity_review.py
```diff
@@ -0,0 +1,276 @@
...
+
+def run(dest_dir: str) -> None:
+    #
+    # Load data.
+    #
+    # Read dataset from meadow.
+    ds_meadow = catalog.Dataset(MEADOW_DATASET_PATH)
+
+    # Load countries-regions table (required to convert country codes to country names in net flows table).
+    countries_regions = catalog.find(
```
Marigold, line 232: It think it's better to get data with `countries_regions = Dataset(DATA_DIR / 'garden/owid/reference')['countries_regions']` just like meadow when you can. It's not an issue here with countries regions, but it is fetching data from remote catalog and users might expect it to fetch data from local catalog.

This was a long version of saying you shouldn't do this for meadow versions (or upstream deps of this step) and that it's ok for countries regions, although getting it directly is better.

(You can obviously keep it like this)

Marigold (approved):
You're a machine!

Marigold:
Those type checks have nothing to do with your changes, feel free to merge this even if CI is failing on them.

etl/steps/data/garden/ember/2022-08-01/european_electricity_review.py
pabloarosado: Done

Issue #314, opened 2022-07-26
Update grapher export and backports to use new short name and version fields


larsyencken:
## Scope

- [ ] Use dataset and variable `short_name` fields from grapher during backport when non-null
- [ ] Push `short_name` and `version` fields to the Grapher model during grapher export

## Why

- For backports: we want to encourage authors to use readable handles for datasets and column names, which will make API usage of this data better
- For ETL-to-Grapher: we want grapher to know how to uniquely address a variable and a dataset back into the ETL (and thus the API)

larsyencken:
Currently blocked on getting the fields into the Grapher data model

PR #359, opened 2022-08-03
feature(grapher): update grapher export and backports to use new short name and version fields

Marigold:
Issue https://github.com/owid/etl/issues/314

[Grapher DB migration](https://github.com/owid/owid-grapher/pull/1566)

### Datasets

Sync attributes
* `Dataset.short_name` -> `datasets.shortName`
* `Dataset.title` -> `datasets.name`
* `Dataset.version` -> `datasets.version`

Migration adds unique constraints on `(dataset, shortName, version)` and updates `shortName` and `version` of existing datasets. That means rerunning grapher steps won't create new datasets, but only update `dataset.name` with `Dataset.title`

### Variables

Sync attributes
* `VariableMeta.short_name` -> `variables.shortName`
* `VariableMeta.title` -> `variables.name`

Migration adds unique constraints on `(dataset, shortName)`, but there's also an existing `(dataset, name)` constraint. So running grapher steps would only update `variables.shortName` and don't create new variables.

### Steps updates

I've tested this on all grapher steps we have. Some had to be slightly updated to adhere to new standards. I've also made it compulsory that every variable has exactly one source. It was handled implicitly, but now that we have `adapt_table_for_grapher` function it's better to do it explicitly.

### TODO

* [ ] let everyone know we're changing schema and they shouldn't be running grapher on production / staging
* [ ] apply migration on staging
* [ ] run all grapher steps on staging and compare staging data with production (make sure variables and datasets were upserted, nothing new was created)
* [ ] apply migration on production
* [ ] merge this PR
* [ ] let everyone know about applying migration locally and about stricter requirements

### DIFF
--- a/etl/db_utils.py
+++ b/etl/db_utils.py
@@ -163,7 +163,9 @@ class DBUtils:
 
     def upsert_dataset(
         self,
-        name: str,
+        short_name: str,
+        title: str,
+        version: str,
         namespace: str,
         user_id: int,
         tag_id: Optional[int] = None,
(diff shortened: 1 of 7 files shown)


larsyencken (approved):
This one looks good to go as well, after the grapher schema is updated you should :shipit: 

Then I think the data team will need an update.

Awesome to see grapher and ETL moving towards closer alignment in their data model!

PR #365, opened 2022-08-04
chore: add "make deploy" helper command

larsyencken:
The deploy command does a full ETL run on production using the analytics script `ops/scripts/etl-prod.sh`. The script helpfully acquires a global lock, so you don't have to worry about any race conditions with production runs.


Marigold (approved):
Looks good! It's a bit weird to have etl-ops in analytics repo (totally my fault) and script for running it here, but it doesn't matter in practice.

larsyencken:
@Marigold Should we move it across to the `etl` repo in an `ops/` folder?

PR #368, opened 2022-08-04
feature(helpers): add Naming object

Marigold:
@pabloarosado do you think such syntactic sugar could be useful or am I trying to simplify something that's already simple enough? I find writing paths to meadow dataset and other files quite annoying.

(This is not a final version, just want to get initial feedback)

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -5,9 +5,13 @@
 
 import tempfile
 from contextlib import contextmanager
+from pathlib import Path
 from typing import Any, Iterator, List, cast
 
 import requests
+from owid import catalog, walden
+
+from etl import paths
 
 
 @contextmanager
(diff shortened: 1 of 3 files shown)


etl/helpers.py
```diff
@@ -50,3 +54,38 @@ def _get_github_branches(org: str, repo: str) -> List[Any]:
...
+    def version(self) -> str:
+        return self.f.parent.name
+
+    @property
+    def short_name(self) -> str:
+        return self.f.with_suffix("").name
+
+    @property
+    def country_mapping_path(self) -> Path:
+        return self.f.parent / (self.short_name + ".country_mapping.json")
```
pabloarosado, line 81: Didn't we decide to call it `.countries.json` for simplicity?

etl/helpers.py
```diff
@@ -50,3 +54,38 @@ def _get_github_branches(org: str, repo: str) -> List[Any]:
...
+
+    @property
+    def metadata_path(self) -> Path:
+        return self.f.parent / (self.short_name + ".meta.yml")
+
+    @property
+    def meadow_dataset(self) -> catalog.Dataset:
+        return catalog.Dataset(
+            paths.DATA_DIR / f"meadow/{self.namespace}/{self.version}/{self.short_name}"
+        )
```
pabloarosado, line 91: This is a reasonable assumption that will cover most cases, but in principle the meadow version of a dataset can be different from the garden version.

etl/steps/data/garden/shift/2022-07-18/fossil_fuel_production.py
```diff
@@ -12,18 +11,13 @@
...
-NAMESPACE = "shift"
-DATASET_SHORT_NAME = "fossil_fuel_production"
 
-VERSION = Path(__file__).parent.name
-COUNTRY_MAPPING_PATH = (
-    Path(__file__).parent / f"{DATASET_SHORT_NAME}.country_mapping.json"
-)
-METADATA_PATH = Path(__file__).parent / f"{DATASET_SHORT_NAME}.meta.yml"
+N = Naming(__file__)
+
```
pabloarosado, line 20: Maybe it's better to call it something more descriptive, maybe `NAMES`?

pabloarosado (approved):

etl/helpers.py
Marigold: Oh, you're right. That's exactly why I need something like this - so that I don't have to remember our conventions.

etl/helpers.py
Marigold: For sure, these are all just recommendations that might help if you're a newcomer (or don't want to think about it), but you're free to use full path.

etl/steps/data/garden/shift/2022-07-18/fossil_fuel_production.py
Marigold: Good idea, renamed to `Names`.

Marigold:
Thanks @pabloarosado! So do you think it's worth giving a try? If we find no one is using it, we can just abandon it.

pabloarosado:
Hi @Marigold yes, let's give it a go!

PR #371, opened 2022-08-05
Add electricity mix

pabloarosado:
* Create a Combined Electricity Review (combination of Global and European Electricity Review by Ember).
* Create the Electricity Mix dataset (combination of BP's processed Statistical Review and Ember's Combined Electricity Review).


Marigold (approved):
Code looks fine and I was able to run it locally. Can't get better than that.

PR #373, opened 2022-08-05
data: add OECD Regional Wellbeing dataset to meadow

larsyencken:
The regional wellbeing dataset is a series of indicators meant to summarise wellbeing and reported at the subnational level. The purpose of bringing them in to meadow is purely to have some subnational data to play with.

This PR loads them in and converts a long form data frame to a wide data frame (one column per indicator), then stuffs the extra columns into the metadata for those indicators, simplifying the dataset.

The region codes are special OECD codes, which we would still need to map to ISO codes if we wanted to be able to standardise them more.


etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
```diff
@@ -0,0 +1,122 @@
...
+df.columns = [
+    "region_code",
+    "region_name",
+    "ind_code",
+    "ind_name",
+    "year",
+    "unit_code",
+    "unit_name",
+    "value",
+]
```
pabloarosado, line 38: I'd suggest using `df = df.rename(columns={'old_column': 'new_column', ...})` instead of renaming `df.columns = []`, which assumes that columns are in a certain order.

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
```diff
@@ -0,0 +1,122 @@
...
+
+df["ind_code"] = df.ind_code.apply(lambda s: s.lower())
+
```
pabloarosado, line 41: It may be faster and clearer to do `df["ind_code"] = df["ind_code"].str.lower()`

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
```diff
@@ -0,0 +1,122 @@
...
+df.head()
+
+# ## Add metadata
+
+t = Table(df)
+
+for ind_code in t.columns:
+    meta = t[ind_code].metadata
+    meta.short_name = ind_code.lower()
```
pabloarosado, line 99: Instead of `ind_code.lower()` you could use `underscore(ind_code)` from `catalog.utils`.

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+# ## Save the dataset
+
+
+def run(dest_dir: str) -> None:
+    ds_meta = convert_walden_metadata(walden_ds)
+    ds = Dataset.create_empty(dest_dir, ds_meta)
+    ds.add(t)
+    ds.add(regions)
```
pabloarosado, line 122: Don't you need to do `ds.save()` to save the dataset?

pabloarosado (approved):
Looks good. I made some suggestions, feel free to disregard them and merge.

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
larsyencken: Good catch! I wonder if a context manager wouldn't be a good pattern for this:

```python
with Dataset.create_empty(...) as ds:
    ...
    # autosave when the manager closes
```

pabloarosado: Yes, I think that would be good, there have been already a few cases where we forgot the `ds.save()`.

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
larsyencken: Done

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
larsyencken: Skipped this since the indicators are all coded with all-caps and underscores

etl/steps/data/meadow/oecd/2016-06-01/regional_wellbeing.py
larsyencken: Added it now

PR #377, opened 2022-08-10
Shortening the table variable names as in some cases they were over 255 chars

spoonerf:
I switched from using the variable name to the variable code which worked in most cases, there is still at least one v long dimension name so I use only the first 255 characters in the variable name.


spoonerf:
Hey @larsyencken just a tiny PR on shortening the Table variable names but thought I would pass it by you before merging!

larsyencken:
Cheers, just taking a look!

PR #378, opened 2022-08-10
enhance: generate CSVs for backported data

larsyencken:
We would like to be able to use backported variables in explorers. Currently we do this in one of two ways. We either:

- Create an ETL step for the explorer that generates CSVs
- Link to a CSV from the auto-generated `owid-datasets` repo that Grapher publishes

This PR adds a third way, by publishing all backported data in CSV format.


larsyencken:
On second thought this isn't a sensible approach, we'll stick with either using grapher charts or explicit explorer steps.

PR #379, opened 2022-08-10
docs: init submodule before installing packages

danielgavrilov:
I ran into this and someone else did too – when setting up for the first time, we need to init the submodules before attempting to install all dependencies with poetry.

I'm not sure if this should instead point to a `make` step, since that initializes submodules:

https://github.com/owid/etl/blob/a9e96f2b6f5d0f97c1b054a69aa9525b79d4d024/Makefile#L38-L41


lucasrodes:
Slightly re-organized the README

PR #381, opened 2022-08-10
fix(slug): separate variable name and dimensions by __

lucasrodes:
Separating by `__` (double underscore) the dimensions so that they can be differentiated from metric/dimension names.


lucasrodes:
This is solved by #323

PR #384, opened 2022-08-12
Add cait data steps

pabloarosado:
Add meadow/garden/grapher steps for CAIT dataset on greenhouse gas emissions by sector. I created one table for each gas inside the dataset.

@larsyencken you can see that I created one grapher step for each table, please let me know if/how this could/should be achieved with just one grapher step, thanks.


etl/steps/data/garden/cait/2022-08-10/ghg_emissions_by_sector.py
```diff
@@ -0,0 +1,229 @@
...
+
+import pandas as pd
+from owid import catalog
+from owid.datautils import dataframes, geo
+
+from etl.paths import DATA_DIR
+from shared import CURRENT_DIR
+
+# Details for dataset to export.
+DATASET_SHORT_NAME = "ghg_emissions_by_sector"
```
Marigold, line 11: (I'm not suggesting that you use this, it's just a syntactic sugar. So totally up to you. You'd have to rebase to use it)

You could give `Names` helper object a try and replace these constants by
```
from etl.helpers import Names
N = Names(__file__)
```
and then use `N.country_mapping_path` in code instead of `COUNTRY_MAPPING_PATH `

pabloarosado: Thanks, I think I'll start using Names for the next one.

etl/steps/grapher/cait/2022-08-10/ch4_emissions.py
```diff
@@ -0,0 +1,48 @@
...
+        column for column in table.columns if column not in ["country", "year"]
+    ]
+    for column in variables:
+        if table[column].metadata.unit == "million tonnes":
+            table[column].metadata.unit = "tonnes"
+            table[column].metadata.short_unit = "t"
+            table[column].metadata.display["conversionFactor"] = 1e6
+    table = gh.adapt_table_for_grapher(table)
+
+    yield from gh.yield_wide_table(table, na_action="drop")
```
Marigold, line 48: Instead of having 4 separate grapher datasets, you could have a single dataset with 4 tables. The dataset name would have to common to all of them (e.g. "Emissions by sector (CAIT, 2022)" instead of "Methane emissions by sector (CAIT, 2022)"), but that doesn't really matter since dataset name [isn't shown in chart anywhere anyway](https://files.slack.com/files-pri/T3DR40MTQ-F03RU6K2LUE/etl-to-grapher.png).

Code would then look like this
```
yield from gh.yield_wide_table(table_ch4, na_action="drop")
yield from gh.yield_wide_table(table_co2, na_action="drop")
yield from gh.yield_wide_table(table_n2o, na_action="drop")
```

Marigold (approved):
Grapher tables could be potentially combined into a single dataset, but multiple datasets are also fine.

etl/steps/grapher/cait/2022-08-10/ch4_emissions.py
larsyencken: Ah, yes, thanks for spelling it out Mojmir! This is what I was trying to nudge you towards last week, but nowhere near as clearly

pabloarosado: Ah right, I didn't know how to achieve this. Next time I'll do it this way.

PR #386, opened 2022-08-12
enhance: match_variables command

lucasrodes:
Improvements on the match variables script

- Prettify the menu with variable options (using `click` and colouring)
- Changed `argparse` for `click` to manage executable
- Created command `etl-match-variables` to run the script
- Removed unused code
- Moved database-specific functions to `etl.db`
- Added argument `--max-suggestions`, to select an arbitrary number of new variables to be displayed

**Test the code**
1. Install the library: `poetry install`
2. Run `etl-match-variables --help`

The library might need a patch upgrade (0.1.1)

### DIFF
--- a/etl/db.py
+++ b/etl/db.py
@@ -2,7 +2,9 @@ import traceback
 from collections.abc import Generator
 from contextlib import contextmanager
 from urllib.parse import quote
+from typing import Any
 
+import pandas as pd
 import MySQLdb
 import structlog
 from sqlalchemy import create_engine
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -16,6 +16,7 @@ fasttrack = 'backport.fasttrack:fasttrack'
 walkthrough = 'walkthrough.cli:cli'
 run_python_step = 'etl.run_python_step:main'
 compare = 'etl.compare:cli'
+etl-match-variables = 'etl.match_variables_from_two_versions_of_a_dataset:main_cli'
```
pabloarosado, line 19: Should we use a shorter name, at least for the etl command? Maybe just `match_variables`.

etl/db.py
```diff
@@ -56,3 +59,73 @@ def open_db() -> Generator[DBUtils, None, None]:
             cursor.close()
         if connection:
             connection.close()
+
+
+def get_dataset_id(db_conn: MySQLdb.Connection, dataset_name: str) -> Any:
```
pabloarosado, line 64: I'm not sure if the `db` module is the right place for these functions. Maybe the `db_utils` makes more sense?

pabloarosado (approved):

pyproject.toml
lucasrodes: Regarding the name for the command, there tare two things:

1. I have a weak preference for separator `-` (dash) over `_` (underscore)
2. Prefixing it with `etl` may ensure that there is no collision with any other tool that one may have installed on their computer.

Other commands have no prefix, and I think we could consider adding some prefix to these, but this might be something to be discussed later on...

etl/db.py
lucasrodes: I thought I should place them in the same module where function `get_connection` lives. 

I am not super sure I understand the difference between these two modules, though (`db` and `db_utils`). I discarded `db_utils` because I saw it was dedicated to the object `DBUtils`.

PR #390, opened 2022-08-15
fix(open_numbers): fix mixed type

Marigold:
Fix failing column `age_group_1year` and make sure we raise an error instead of a warning that's hard to debug.

### DIFF
--- a/etl/steps/open_numbers.py
+++ b/etl/steps/open_numbers.py
@@ -8,21 +8,21 @@ Convert repositories with data in DDF format to OWID's Dataset and Table
 format.
 """
 
+import datetime as dt
+import hashlib
+import tempfile
+import warnings
 from multiprocessing import Pool
 from pathlib import Path
 from typing import Dict, List, Tuple, cast
-import hashlib
-from owid.catalog.meta import Source
-from owid.catalog import utils
-import datetime as dt
-import tempfile
 
-import pandas as pd
-from etl.git import GithubRepo
 import frictionless
+import pandas as pd
 from frictionless.exception import FrictionlessException
+from owid.catalog import Dataset, Table, utils
+from owid.catalog.meta import Source
 
-from owid.catalog import Dataset, Table
+from etl.git import GithubRepo
 
 
 def run(dest_dir: str) -> None:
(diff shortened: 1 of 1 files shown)


etl/steps/open_numbers.py
```diff
@@ -128,7 +128,13 @@ def load_and_combine(
...
 
         f.flush()
-        df = pd.read_csv(f.name)
+
+        # raise warnings as exceptions since they would cause errors later on anyway
+        # if you see `pandas.errors.DtypeWarning: Columns (2) have mixed types.` make
+        # sure you set the proper type in `dtype`
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            df = pd.read_csv(f.name, dtype={"age_group_1year": "str"})
```
larsyencken, line 137: This file processes a hundred datasets. I don't think patching a random column is a good approach.

A more general one:

- This situation only seems to happen with `load_and_combine()` because we're using low memory settings
- We can loop over the columns and use `pd.api.types.infer_dtype(col)`, checking for responses with the prefix "mixed-"
- For those columns
  - Try `pd.to_numeric(col)` first, catch the ValueError if it fails
  - Fallback to `col.astype('str')`

larsyencken (changes requested):

etl/steps/open_numbers.py
Marigold: Ok, I'll do that. I actually had that solution first, but I didn't want to unnecessarily complicate the code and make open numbers even slower.

etl/steps/open_numbers.py
```diff
@@ -128,7 +128,20 @@ def load_and_combine(
...
-        df = pd.read_csv(f.name)
+
+        # ignore mixed type warnings
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", pd.errors.DtypeWarning)
+            df = pd.read_csv(f.name)
+
+    # fix mixed types of object columns
+    for col in df.select_dtypes(object).columns:
+        if pd.api.types.infer_dtype(df[col]).startswith("mixed"):
```
larsyencken, line 139:
```suggestion
        if col.dtype.name == 'object' and pd.api.types.infer_dtype(df[col]).startswith("mixed"):
```

larsyencken (approved):
Works like a charm 👍

Issue #294, opened 2022-07-19
Walkthrough - improvements


Marigold:
Suggested improvements of walkthrough from this [Slack thread](https://owid.slack.com/archives/C0193RW5E2J/p1658143999235099):

**Initial list of issues:**
* [ ] And I'd say that, before creating those files (especially the *.py), it should check if a file already exists. Otherwise it's currently overwriting it.
* [x] Why having a validate script? I think it would be better if those validations were already at the end of the meadow step. Additionally, if you think a separate validate script is better, I’d suggest to use .py instead of .ipynb.
* [ ] add buttons for a few of the functions to execute stuff - this way you could stay a bit more in the tool itself and only switch over to the terminal or VS code (which might also be opened with a button?
* [ ] I think ingest script should be expected to work not just once, but also for future releases of the data. So I think that it would be better to not hardcode dates in the metadata file. We could have some kind of tags for years and dates?
* [x] I get a `Dataset not found in Walden, did you upload it?` However the walden index file already exists in the submodule and the file is already in s3, so does it need to be merged to master?
  * could not reproduce
* [x] And a final note: there are many ways to misspell "walkthrough". Maybe calling it "guide" will be appreciated by our future selves?

@pabloarosado:
* [x] "Afghanistan" should not be by default in the mapping, because a dataset may not include it (I just found one example). Maybe it should be just empty.
* [x] Although I like the name "Excludistan", I guess it should not be there, and the list of excluded countries should also be empty (otherwise after a while the etl may end up having many instances of "Excludistan").
* [x] We already have a `geo.harmonize_countries` function in datautils. Maybe that should be used by default (we know that we have to improve it, but it is still better than the default code you have added to the garden steps).

@Marigold:
* [x] I've noticed that the walkthrough script creates a countries file named institution.country_mapping.json, but shouldn't it be institution.countries.json? I guess it doesn't really matter, but I though the latter was sort of accepted
* [x] I'd suggest to remove all instances of ggdc names and URLs. Using "dummy" is probably a better idea.
* [x] question about the .meta.yml file generated in the garden step. I see that the template contains some fields which I believe to be used in grapher only. Shouldn’t this be in the grapher step instead? Field `entityAnnotationsMap` for instance
* [x] add variable matching to grapher walkthrough, see [this file for an inspiration](https://github.com/owid/etl/pull/324/files#diff-cbf288226b30062782ecd9b21a622abf2e040d70235d158ea8dbba685a5ea542)
  * #690
  * #706
[...]

PR #391, opened 2022-08-15
Enhance: improve walkthrough

Marigold:
Fixes a couple of minor things from [suggested improvements](https://github.com/owid/etl/issues/294). I haven't implemented them all as they don't seem to be priority at the moment.


dag.yml
```diff
@@ -118,6 +118,8 @@ steps:
     - data://garden/cait/2022-08-10/ghg_emissions_by_sector
 
   # Autogenerated steps by walkthrough
+
```
pabloarosado, line 121: I was thinking that, just to keep the dag cleaner, and to avoid conflicts, maybe the walkthrough steps could be in their own dag in dag_files.

walkthrough/garden_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.countries_exclude.json
```diff
@@ -0,0 +1,2 @@
+[
```
pabloarosado, line 1: Maybe instead of `*.countries_exclude.json`, the file could be called `*.excluded_countries.json`.

walkthrough/garden_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
```diff
@@ -1,28 +1,20 @@
...
+from owid.datautils import geo
 from structlog import get_logger
 
+from etl.helpers import Names
 from etl.paths import DATA_DIR
 
 log = get_logger()
 
+# naming conventions
+N = Names(__file__)
```
pabloarosado, line 16: `__file__` does not work when working on a jupyter notebook, which is annoying. What I usually do to work around this is to create a `shared.py` file that contains the `__file__` (and possibly other definitions, like log). I don't know if you like this idea or if there is a better work around. For me it's like having an `__init__.py` on every module.

walkthrough/garden_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
```diff
@@ -36,48 +28,40 @@ def run(dest_dir: str) -> None:
...
-    if countries.isnull().any():
-        missing_countries = [
-            x for x in df["country"].drop_duplicates() if x not in country_mapping
-        ]
+    df = df.loc[~df.country.isin(excluded_countries)]
+
+    log.info("{{cookiecutter.short_name}}.harmonize_countries")
+    unharmonized_countries = df["country"]
+    df = geo.harmonize_countries(df=df, countries_file=str(N.country_mapping_path))
+
```
pabloarosado, line 38: We should work on `geo.harmonize_countries` to accept `excluded_countries` as an argument (and several other necessary improvements). For the time being, this boilerplate code is something I would move to `shared.py`. In this case, the goal of moving things to `shared.py` is to make the code inside `run` as simple and understandable as possible.

walkthrough/grapher_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
```diff
@@ -3,40 +3,26 @@
...
 
-    # grapher needs a column entity id, that is constructed based on the unique entity names in the database
-    table["entity_id"] = gh.country_to_entity_id(table["country"])
+    # convert `country` into `entity_id` and set indexes for `yield_wide_table`
+    table = gh.adapt_table_for_grapher(table)
 
-    # use entity_id and year as indexes in grapher
-    table = table.set_index(["entity_id", "year"])
+    # optionally set dimensions
+    # table = table.set_index(["dim1", "dim2"], append=True)
```
pabloarosado, line 25: Is it actually necessary to set the index? If not, then it should not even be an option. The only reason I see for this is to verify that the index does not have duplicates (using `verify_integrity=True`, but this is something that should have been done in garden).
On the other hand, if it is necessary to add index, then it could also be done by `adapt_table_for_grapher` (maybe adding arguments `dim1='country'` and `dim2='year'`).

walkthrough/meadow.md
```diff
@@ -10,3 +10,5 @@ Here's a summary of this walkthrough, you don't have to manually execute anythin
 2. **Add the new meadow step to the dag**, including its dependencies.
 
 3. **Run `make test` in `etl`** and  ensure the step runs well. To run the step: `etl data://meadow/example_institution/YYYY-MM-DD/example_dataset`
+
+`meadow` is for _syntactic harmonisation_, meaning getting something into the right format. The idea is that instead of sitting on the dataset that you're working on for a long time without sharing it, you could spend a few hours and get it into a form where others can already reuse it. Out of this you get more, smaller pull-requests. It's also important to have this form for debugging, so that we can compare changes we make later to the "original data".
```
pabloarosado, line 14: I don't know if the sentence `The idea ...` is very realistic. So far I have never pushed just the meadow steps, because I often need to rethink them (e.g. add or remove raw data, or create one or several tables) while working on the garden steps. I would say that the goal of `meadow` is to do the minimum possible work on the original data, to convert it into one (or more than one) table.

walkthrough/meadow_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
```diff
@@ -1,15 +1,18 @@
...
 from structlog import get_logger
 
+from etl.helpers import Names
 from etl.paths import DATA_DIR, REFERENCE_DATASET
 from etl.steps.data.converters import convert_walden_metadata
 
 log = get_logger()
+
+# naming conventions
+N = Names(__file__)
```
pabloarosado, line 14: The same suggestion about having a `shared.py` module with all the boilerplate code applies to meadow. Eventually, when we have all the tools we need in either `etl.helpers` or `datautils` there should be almost nothing in `shared.py`.

pabloarosado (approved):
All good. I made some suggestions, but feel free to ignore them. Overall one of my goals when writing data steps is to make the content of `run` as understandable and concise as possible (e.g. `load data`, `process data`, `save outputs`). I always keep in mind that the code is not just for my future self, but also to whoever wants to easily understand how a dataset was built (even if they don't understand python). So I would avoid having a lot of boilerplate code like loading country mapping files and applying them, and instead have a function `harmonize_countries` and a comment declaring the intention of the code.

dag.yml
Marigold: Great idea, fixed.

walkthrough/garden_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
Marigold: Eh, that's right. `shared.py` is a really good solution, but I don't want to complicate walkthrough even more (also given that you're the only one using notebooks this way).

Alternative is to use ugly
```python
from pathlib import Path
N = Names(globals().get('__file__', (Path() / "dummy.py").resolve()))
```
which falls back to working directory + hardcoded module name if `__file__` doesn't exist.

walkthrough/garden_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
Marigold: > make the code inside run as simple and understandable as possible.

Yeah, you're right. I moved the code to functions and left only high-level calls in `run` method. I didn't go as far as creating `shared.py`, but I might reconsider that in the future.

walkthrough/grapher_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
Marigold: Not sure if we're talking about the same thing. I've changed these lines to
```python
# optionally set additional dimensions
# table = table.set_index(["sex", "income_group"], append=True)
```

If you have more these dimensions in multiindex, `yield_wide_table` would create variables grouped by these dimensions (e.g. variables `gdp__sex_male` and `gdp__sex_female`)

walkthrough/meadow.md
Marigold: That's actually Lars's sentence :D. I shamelessly stole it from one of our discussions on github. I guess it's up to everyone's workflow. Ideally we'd be creating small incremental PRs and get them reviewed continuously, but honestly that adds so much overhead that it's easier to have one monstrous PR (and someone who's willing to review it). I do the same...

walkthrough/meadow_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
Marigold: > Eventually, when we have all the tools we need in either etl.helpers or datautils there should be almost nothing in shared.py.

I like the idea that what we put in `shared.py` is meant to be eventually moved to a shared space (and unit tested etc.). I'm not adding it _yet_, but will think about it before next walkthrough improvements.

PR #395, opened 2022-08-17
Remove unused steps and notebooks related to FAOSTAT

pabloarosado:
Removed notebooks and data steps that were never used (as far as I know):
* etl/steps/data/explorers/owid/2021/income.ipynb - This seems to be in git by mistake, isn't it @lucasrodes ?
* the meadow and garden steps for faostat_rl were created only as part of a tutorial about using etl, if I remember correctly. But they have never been used. I think we should delete them, since they haven't been properly analysed and therefore shouldn't be in garden.
I updated the dag appropriately to remove the steps above.


larsyencken (approved):
They don't lead to either grapher or an explorer, we should feel free to prune them 👍

PR #398, opened 2022-08-19
Enhance(walkthrough): commit dummy dataset

Marigold:
* Commit dummy dataset for easier experimentation
* Add warnings for archived datasets or namespaces
* Fix some bugs in walkthrough


pabloarosado (approved):
My only suggestion would be to avoid real information in dummy: For example, instead of citing "Maddison ..." it could just be "Template dataset..." or something like that. Otherwise we may, by mistake, have a reference to the GGDC in a dataset that has nothing to do with it. Also, having such a complete country mapping could lead to confusion, and a data manager could believe that they already have harmonized countries.
But we could change these things later.

Marigold:
Thanks, not sure about that one yet (=lazy to change everything), but I've added it to the list of [walkthrough improvements](https://github.com/owid/etl/issues/294).

PR #401, opened 2022-08-22
feature(isort): add isort to formatting

Marigold:
Sort imports automatically with `isort` in `make format` command.


pabloarosado (approved):

larsyencken (approved):
Looks good. Do the same with the vendored repos too, no need to do them as PRs.

PR #402, opened 2022-08-22
enhance(mypy): typecheck steps in parallel

Marigold:
Typechecking steps serially takes ~42s. Running them all in parallel takes under 12s.


larsyencken (approved):
Good optimisation, cuts typechecking steps to just a few seconds for me. 👍 

(Side note: I rebased to get rid of vendor changes, you can probably make your life easier by only checking in the file you changed instead of the whole folder.)

PR #403, opened 2022-08-22
refactor(black): change black line length to 120

Marigold:
Change black line length to 120. This will make our step scripts more concise (pandas commands can be long), but on the other hand reading core codebase might be a bit harder. We could also wait for steps - ETL split into two repositories and apply 120 line length only to the steps repo (that split is however not a priority at the moment so who knows when it gets done...).

I haven't changed line length of walden and catalog repositories as I don't think it's needed there. What do you think, is consistency worth it?

@pabloarosado hurry with approval before @larsyencken gets back from vacation!


pabloarosado:
Thanks for moving this, @Marigold. I definitely prefer the 120 characters :-).
But is it necessary to edit all files accordingly? I would have thought that we could leave old files with their 88 characters limitation, and new ones would be allowed to have no more than 120 characters.
This PR is too aggressive. Firstly, it may create complicated conflicts with my current branch on faostat scripts. And secondly, it's going to mess up the history of line changes (it helps checking git blame to see the last changes in the code). So, can we allow for 120 characters without messing up the old code? If not, then I'd appreciate waiting until I merge my current branch (and we should possibly check with others).

Marigold:
@pabloarosado apparently there's a way how to fix git blame with `.git-blame-ignore-revs` file (added it there).

> I would have thought that we could leave old files with their 88 characters limitation, and new ones would be allowed to have no more than 120 characters.

As far as I know, that's not possible (unless you enumerate all those files for black). We're gonna end up with merge conflicts anyway. Maybe we should postpone it until we decide to split the repo into steps and core code? I don't feel great about setting 120 lines for core code given it's not necessary (we could use two different line lengths in this repo, but that would end up as a mess). Let's perhaps hear what @larsyencken thinks.

larsyencken:
I think we should more or less have one standard for the team that we use for all repos. I do think that means using black to reformat everything at the time we set the new standard.

I do a lot of terminal based editing, so this change is expected to be worse for me but better for everyone else. On that basis we've already agreed to do it. Provided you both think it's better, you should merge it, then do the same with all our other Python repos.

larsyencken:
Good to double check on this one. Once we agree here, no need for review on the PRs for other repos I think 🙂

larsyencken (approved):
I tried this branch out in `neovim`, we actually have pretty few extra long lines anyway, so I've come to feel it's not a big deal. Please feel free to rebase and then merge.

pabloarosado (approved):

PR #409, opened 2022-08-24
fix(grapher_helper): fix inf check

Marigold:
Silly mistake, this was causing assertion errors for categorical variables.

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Literal, Optional, Set, cast
 
+import numpy as np
 import pandas as pd
 import slugify
 import structlog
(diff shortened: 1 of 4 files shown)


etl/grapher_helpers.py
```diff
@@ -499,3 +500,8 @@ def from_values(xs: List[int]) -> "IntRange":
 
     def to_values(self) -> list[int]:
         return [self.min, self.max]
+
+
+def contains_inf(s: pd.Series) -> bool:
+    """Check if a series contains infinity."""
+    return pd.api.types.is_numeric_dtype(s.dtype) and np.isinf(s).any()  # type: ignore
```
pabloarosado, line 507: Isn't the `np.isinf(s)` going to raise a TypeError if the series is not numeric? If not, all good.

pabloarosado (approved):

etl/grapher_helpers.py
Marigold: It won't get past the first condition. If it's not numeric then the second part after `and` won't be evaluated.

(Another way would be to catch `TypeError`)

PR #410, opened 2022-08-25
Update faostat data in grapher

pabloarosado:
* When creating new FAOSTAT steps, ensure each step loads the correct version (the one that was latest when the step was created, instead of the step that is currently latest). Versions are now specified in a separate file (one for meadow, one for garden, and one for grapher).
* Fix known issue of `faostat_scl` having a different type of faostat index.
* Fix data values given as upper bounds (e.g. `<0.1`), or with commas (e.g. `1,234.56`).
* Ensure variable short names are not too long.
* Improve custom names of datasets, items and elements.
* Add a new file that specifies known outliers.
* In `faostat_food_explorer`, specify the element codes that need to be considered.


etl/steps/data/garden/faostat/2022-05-17/shared.py
```diff
@@ -1725,7 +1754,18 @@ def optimize_table_dtypes(table: catalog.Table) -> catalog.Table:
         if c in table.columns
     }
 
-    return catalog.frames.repack_frame(table, dtypes=dtypes)
+    # Store variables metadata before optimizing table dtypes (otherwise they will be lost).
```
Marigold, line 1757: (This is probably something that can happen implicitly in `repack_frame` function, but keep it as it is)

etl/scripts/faostat/create_new_steps.py
```diff
@@ -795,6 +798,58 @@ def apply_custom_rules_to_list_of_steps_to_create(
     return step_names
 
 
+def create_file_with_latest_versions(dag_steps: Dict[str, Set[str]]) -> None:
+    """Create a csv file in the new version folder, that keeps track of all relevant channel-dataset-versions for the
+    current update.
+
+    This file of versions is necessary because, in the future, when running old steps, they shouldn't simply load the
```
Marigold, line 805: This is good solution, but I'm wondering whether managing versions with git somehow would be doable (and simplify this). If we wanted to recreate an old version we'd checkout old commit and run it. There'd be many complications for sure, but perhaps it's something we should consider thinking about?

(Nothing to do in this PR, just dumping my thoughts)

Marigold (approved):
Looks good!

PR #416, opened 2022-08-29
refactor(grapher): use ORM for grapher inserts

Marigold:
Blocked by https://github.com/owid/etl/pull/417

Move upsert methods from `DBUtils` to ORM and simplify code for grapher inserts. Remove unused methods from `DBUtils` and keep only stuff that is still used elsewhere.

I could have replaced DBUtils with ORM everywhere, but we wouldn't gain much from typechecks for e.g. deletes. Using raw SQL in those cases looks simpler.

Variable upserts are slightly slower (10-20%), but it's not a big deal since upserts to `data_values` are the bottleneck.

`DBUtils` had no unit test coverage. I didn't add any for ORM yet, but I'm going through all grapher datasets and comparing old vs new approach on staging. If I find anything problematic worth testing I'm gonna add it there.


etl/grapher_import.py
```diff
@@ -67,62 +67,80 @@ def upsert_dataset(dataset: catalog.Dataset, namespace: str, sources: List[catal
...
+        ds = gm.Dataset(
+            shortName=dataset.metadata.short_name,
+            name=dataset.metadata.title,
+            version=dataset.metadata.version,
+            namespace=namespace,
+            metadataEditedByUserId=user_id,
+            dataEditedByUserId=user_id,
+            createdByUserId=user_id,
             description=dataset.metadata.description or "",
         )
```
larsyencken, line 97: Nit: Hide this in `gm.Dataset.from_catalog_dataset(ds, user_id)`?

etl/grapher_import.py
```diff
@@ -67,62 +67,80 @@ def upsert_dataset(dataset: catalog.Dataset, namespace: str, sources: List[catal
...
-            if ns_is_archived:
-                log.warning("upsert_dataset.namespace_is_archived", namespace=ns_name)
+        ns = gm.Namespace(name=namespace, description="")
+        ns = ns.upsert(session)
+        if ns.isArchived:
+            log.warning("upsert_dataset.namespace_is_archived", namespace=ns.name)
 
         log.info(
             "upsert_dataset.upsert_dataset.start",
             short_name=dataset.metadata.short_name,
```
larsyencken, line 85: Nit: promote `short_name` to it's own variable at the top for readability:

```
short_name = dataset.metadata.short_name
```

etl/grapher_import.py
```diff
@@ -67,62 +67,80 @@ def upsert_dataset(dataset: catalog.Dataset, namespace: str, sources: List[catal
...
+            createdByUserId=user_id,
             description=dataset.metadata.description or "",
         )
+        ds = ds.upsert(session)
+        assert ds.id
+        if ds.isArchived:
+            log.warning(
+                "upsert_dataset.dataset_is_archived",
+                id=ds.id,
+                short_name=ds.shortName,
```
larsyencken, line 104: Nit: could use the same `short_name` var suggested above

etl/grapher_import.py
```diff
@@ -67,62 +67,80 @@ def upsert_dataset(dataset: catalog.Dataset, namespace: str, sources: List[catal
...
             short_name=dataset.metadata.short_name,
-            id=dataset_id,
+            id=ds.id,
         )
 
         source_ids: Dict[str, int] = dict()
         for source in sources:
-            source_ids[source.name] = _upsert_source_to_db(db, source, dataset_id)
+            assert source.name
+            source_ids[source.name] = _upsert_source_to_db(session, source, ds.id)
```
larsyencken, line 116: Consider moving the `_upsert_source_to_db` logic to the grapher model:

```python
source_ids[source.name] = gm.Source.from_catalog_source(source, ds.id).upsert(session)
```

etl/grapher_import.py
```diff
@@ -198,66 +216,54 @@ def upsert_table(table: catalog.Table, dataset_upsert_result: DatasetUpsertResul
             # Not exists, upsert it
             # NOTE: this could be quite inefficient as we upsert source for every variable
             #   optimize this if this turns out to be a bottleneck
-            source_id = _upsert_source_to_db(db, source, dataset_upsert_result.dataset_id)
+            source_id = _upsert_source_to_db(session, source, dataset_upsert_result.dataset_id)
```
larsyencken, line 219: see above suggestion

etl/grapher_import.py
```diff
@@ -198,66 +216,54 @@ def upsert_table(table: catalog.Table, dataset_upsert_result: DatasetUpsertResul
...
             description=table[column_name].metadata.description,
-            code=None,
             unit=table[column_name].metadata.unit,
-            short_unit=table[column_name].metadata.short_unit,
+            shortUnit=table[column_name].metadata.short_unit,
             timespan=timespan,
             coverage="",
             display=table[column_name].metadata.display,
-            original_metadata=None,
         )
```
larsyencken, line 232: Suggest moving the logic to the grapher model:

```python
variable = gm.Variable.from_catalog_table(table, column, timespan)
```

larsyencken (commented):
No correctness issues spotted, mainly stylistic feedback, suggesting to push more methods into grapher models for readability.

Happy to discuss, and if we have good reasons not to, you can merge.

backport/backport.py
```diff
@@ -75,14 +75,15 @@ def backport(
     with Session(engine) as session:
         lg.info("backport.loading_dataset")
         ds = gm.Dataset.load_dataset(session, dataset_id)
+        assert ds.id
```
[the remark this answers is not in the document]
Marigold: It's not possible, but type checker doesn't know that... so I can either ignore it or assert that it's there. Any preference?

backport/backport.py
```diff
@@ -75,14 +75,15 @@ def backport(
...
         lg.info("backport.loading_variables")
         vars = gm.Dataset.load_variables_for_dataset(session, dataset_id)
 
         variable_ids = [v.id for v in vars]
 
         # get sources for dataset and all variables
         lg.info("backport.loading_sources")
-        sources = gm.Source.load_sources(session, dataset_id=ds.id, variable_ids=variable_ids)
+        sources = gm.Source.load_sources(session, dataset_id=ds.id, variable_ids=variable_ids)  # type: ignore
```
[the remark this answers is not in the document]
Marigold: > Interesting that the models need optional primary keys

Yeah, it's annoying, but if the primary key is autoincremented then you don't have it until you insert it to DB.

etl/grapher_import.py
Marigold: Way nicer. I had to keep there some boilerplate code with thread locks (which will go away soon), but it's still much better.

etl/grapher_import.py
Marigold: Moved it there (in a slightly different way).

Marigold:
@larsyencken thanks for review, super useful. In summary:
* I moved those functions to ORM objects wherever possible
* removed unused models from ORM
* tested parity on several existing grapher steps

This PR "depends" on thread lock from [this PR](https://github.com/owid/etl/pull/422) that should probably get merged first.

larsyencken (approved):
Nice! Looks good :shipit:

PR #423, opened 2022-08-31
Feature/variable mapping translate

lucasrodes:
This PR adds a new command tool to the library: `etl-translate-varmap`.

The aim of this tool is to ease the process of obtaining the variable mapping JSON file when working with different environments. The background story is that variable IDs are usually different depending on the environment (local, staging, production). Hence, if we work locally and obtain the variable matching this one is useless once we switch to another environment (e.g. staging or production).

Usage is as follows:

```
$ etl-translate-varmap -e1 <env-file-1> -e2 <env-file-2> -m1 <mapping-file-1> -m2 <mapping-file-2>
```

Where there are two environments (1 and 2), and given that we have the mappings in environment 1, we want to obtain the equivalent for environment 2. Arguments are:

- `<env-file-1>`: File with environment variables required to connect to database (environment 1). Format as in [.env.example](https://github.com/owid/etl/blob/master/.env.example),
- `<env-file-2>`: File with environment variables required to connect to database (environment 2). Format as in [.env.example](https://github.com/owid/etl/blob/master/.env.example),
- `<mapping-file-1>`: File with variable mappings (previously generated, see command `etl-match-variables`).
- `<mapping-file-2>`: New generated file with variable mappings based on the variable IDs from environment 2.

The new mapping is stored as `<mapping-file-2>`.


etl/variable_mapping_translate.py
```diff
@@ -0,0 +1,291 @@
...
+    pandas.DataFrame :
+        Dataframe with variable fields: `id`, `name` and `dataset_name`.
+    """
+    query = f"""
+        select variables.id id, variables.name name, datasets.name dataset_name
+        from variables
+        left join datasets on variables.datasetId=datasets.id
+        where variables.id in {variable_ids};
+    """
+    df: pd.DataFrame = pd.read_sql_query(_format_query(query), conn)
```
Marigold, line 206: You could use `params=` like this instead of `_format_query`

```
query = f"""
    select variables.id id, variables.name name, datasets.name dataset_name
    from variables
    left join datasets on variables.datasetId=datasets.id
    where variables.id in %(variable_ids)s;
"""
df: pd.DataFrame = pd.read_sql_query(query, conn, params={"variable_ids": variable_ids})
```

using `params` prevents SQL injection and might be easier to use (though honestly I don't think we'll be injecting ourselves...).

Marigold (approved):
Looks good! I always thought that variable ids would be identical in local / staging / production (since they're clones of production), but you're right it might not be always the case.

etl/variable_mapping_translate.py
lucasrodes: Thanks, I'm modifying query calls!

PR #424, opened 2022-08-31
enhance(.env): allow custom .env file

Marigold:
Currently when I run grapher steps on staging on production I change my envs in `.env` file and then change it back. This could be dangerous if I forget to do so and then test against production envs in `.env`. Running `ENV=.env.production etl ...` instead is much more flagrant.

(I don't think that `dotenv` supports setting env file from environment or does it?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ data/*
 data/garden/*
 !data/garden/reference/
 .env
+.env.*
 *.dot
 *.pdf
 pyrightconfig.json
--- a/etl/config.py
+++ b/etl/config.py
@@ -12,7 +12,9 @@ from os import environ as env
 import bugsnag
 from dotenv import load_dotenv
 
-load_dotenv()
+ENV_FILE = env.get("ENV", ".env")
+
+load_dotenv(ENV_FILE)
 
 DEBUG = env.get("DEBUG") == "True"
 


larsyencken (approved):
I read over its API docs, doesn't look like there's any existing convention here. If you want, feel free to shorten `ENV_FILE` to `ENV`. Sounds like in your case you want to operate without a default `.env` file entirely, so that you're always explicitly choosing.

larsyencken:
https://github.com/theskumar/python-dotenv

Marigold:
> If you want, feel free to shorten ENV_FILE to ENV

It uses `ENV` environment name (`ENV_FILE` is just a variable name).

> Sounds like in your case you want to operate without a default .env file entirely, so that you're always explicitly choosing.

All people have only local DB credentials in their `.env`, so using it by default is ok I think (and users wouldn't notice). It's enough if we're explicitly choosing `ENV=.env.prod`.

lucasrodes:
Maybe I am a bit off track here, but I have been using `dotenv_values` to load variables from a `.env` file as a dictionary:

```
from dotenv import dotenv_values
config = dotenv_values(path)
```

Marigold:
@lucasrodes both `load_dotenv` and `dotenv_values` are ok, this PR only adds possibility to load a different env file with e.g. `ENV=.env.prod etl ...`

PR #427, opened 2022-09-01
data: rename wpp/2019/standard_projections to un/2019/un_wpp

larsyencken:
We recently imported a fresh version of UN World Population Prospects. The new and old version currently don't use the same naming conventions or namespace. This PR brings them into alignment, by renaming the old version to match the new.

- Rename to match naming scheme of new 2022 import by Lucas
- Convert ipynb file to Jupytext python script


lucasrodes (changes requested):
@larsyencken While https://github.com/owid/owid-issues/issues/465 is not solved, should we change the line https://github.com/owid/etl/blob/1b99bcf0368f11a434af7e38815aa49ea0eabbd8/etl/steps/data/garden/owid/latest/key_indicators/table_population.py#L21 to use the new path `garden/un/2019/un_wpp`?

larsyencken:
Thanks @lucasrodes ! I found one more spot where the reference was broken, and made sure to test rebuilding it now with an empty `data/` dir, which seems to work.

lucasrodes (approved):
Thanks @larsyencken !

PR #431, opened 2022-09-02
Small enhancements to World Bank - World Development Indicators bulk update

### LINKED ISSUE #329: Add metadata for variables to `data://garden/worldbank_wdi/2022-05-26/wdi`

bnjmacdonald:
Adds a few small enhancements to the World Bank - World Development Indicators (worldbank_wdi) bulk update:

1. Adds a handful of owid-maintained metrics to the `garden` step; and
2. Moves variable meta creation from `grapher` step to `garden` step.

closes #329 , owid/owid-issues#409

@Marigold is this sufficient to close #329 ? I've only added a small kludge in `sdg.py` to retrieve metadata from garden (instead of backport) for world bank WDI specifically.


Marigold (approved):
Thanks Bobbie, this is enough to close the issue. In the future it'd be good if running `data://garden/worldbank_wdi/2022-05-26/wdi` wouldn't have to connect to MySQL so that anyone can run it (i.e. writing a script that'd grab data from MySQL once and save them as static file in the repo). Luckily no one needs that at the moment so we can defer it.

larsyencken:
Hey @bnjmacdonald , just saw the grapher step accessing MySQL now, which breaks an important design contract for the system. Raised an issue for it now: https://github.com/owid/etl/issues/442

If the connection is to get some once-off metadata from grapher, then the best approach would be to make and check in a JSON file with the metadata from a manual query.

Issue #432, opened 2022-09-02
Bug: inserting `source` freezes if `dataset` doesn't exists


Marigold:
## Problem

Running grapher for new dataset freezes and errors with

```
(MySQLdb.OperationalError) (1205, 'Lock wait timeout exceeded; try restarting transaction')
```

## Solution

Commit dataset before upserting sources.

PR #433, opened 2022-09-02
fix(grapher_import): commit dataset after upsert to unfreeze sources

Marigold:
Fix https://github.com/owid/etl/issues/432. There was a missing commit and opening new connection within an existing connection was a dumb idea. Add missing commits and only use single connection.


Marigold:
@larsyencken could you please take a look at this if you have time? Other PRs start depending on this.

larsyencken (approved):
Yeah, looks less confusing than holding many connections 👍 

Ship it! :shipit:

PR #436, opened 2022-09-02
refactor(grapher): migrate all grapher steps to variable views

Marigold:
~~Blocked by: https://github.com/owid/owid-grapher/pull/1649~~

Migrate all grapher steps to `DataStep` in `grapher` channel. After we merge this we won't need the old `GrapherStep` any more (cleanup will be done in subsequent PR). I haven't changed logic in grapher steps, only rewrote them to use the `DataStep` structure (I've had to slightly modify datasets that were using `yield_long_table` function, but I've compared them and they should produce the same output in grapher).

Note that I left all grapher steps in `steps/grapher` and only created symlink to `steps/data/grapher` to make it easier to review.

## Future Cleanup
* [ ] remove `GrapherStep`
* [ ] move all grapher steps from to `steps/grapher` to `steps/data/grapher` and remove symlink
* [ ] change path in walkthrough garden
* [ ] rename all `grapher-new://` to `grapher://`
* [ ] run `adapt_table_for_grapher` and `adapt_dataset_metadata_for_grapher` by default for all grapher steps
* [ ] make `yield_wide_table` and yield_long_table` methods internal


Marigold:
@larsyencken to deal with grabbing `entity_ids` from MySQL when building the grapher channel, I've added `--upsert` flag to ETL. It works like this now:

1. No flag builds all channels but grapher
2. `--grapher` builds also grapher channel (as it might need MySQL for entity ids)
3. `--grapher --upsert` builds grapher channel and **upserts** it into DB

I believe this is cleaner for now. Ideally our grapher steps wouldn't depend on MySQL (we'd move `country -> entity_id` to upsert step), then we could get rid of the `--grapher` flag. I plan to do it soon, but there's an intermediary cleaning step that needs to be done first.

Do you have better ideas or can we go with this for now?

etl/command.py
```diff
@@ -80,12 +87,17 @@ def main_cli(
 ) -> None:
     _update_open_file_limit()
 
+    if upsert and not grapher:
+        click.echo("ERROR: --upsert also requires --grapher")
+        sys.exit(1)
```
larsyencken, line 92: What if instead you just say:

```python
# enable grapher when called with --upsert
grapher = grapher or upsert
```

Then the usages become shorter too, just `etl --upsert ...`

etl/grapher_helpers.py
```diff
@@ -263,6 +267,36 @@ def yield_long_table(
...
 
 
+def long_to_wide_tables(
+    table: catalog.Table,
+    annot: Optional[Annotation] = None,
+) -> Iterable[catalog.Table]:
+    """Yield wide tables from long table with the following columns:
+    - variable: short variable name (needs to be underscored)
+    - value: variable value
+    - meta: either VariableMeta object or null in every row
```
larsyencken, line 277: In the description, it's ambiguous whether the wide table or the long table has this column structure.

```
Takes as input a long table, which must have columns: ...

Yields one or more wide tables from it, which have one column per variable.
```

etl/run_python_step.py
```diff
@@ -21,7 +21,7 @@ def main(uri: str, dest_dir: str, ipdb: Optional[bool]) -> None:
     Import and run a specific step of the ETL. Meant to be ran as
     a subprocess by the main `etl` command.
     """
-    if not uri.startswith("data://") and not uri.startswith("data-private://"):
+    if not uri.startswith("data://") and not uri.startswith("data-private://") and not uri.startswith("upsert://"):
```
larsyencken, line 24: I can't see any reference to `upsert://` steps using this path. Were you experimenting?

etl/steps/grapher/bp/2022-07-14/statistical_review.py
```diff
@@ -12,16 +10,10 @@
...
 
-
-def get_grapher_tables(dataset: catalog.Dataset) -> Iterable[catalog.Table]:
     # There is only one table in the dataset, with the same name as the dataset.
-    table = dataset[dataset.table_names[0]].reset_index().drop(columns=["country_code"])
-    table = gh.adapt_table_for_grapher(table)
-
-    yield from gh.yield_wide_table(table, na_action="drop")
+    table = garden_dataset[garden_dataset.table_names[0]].reset_index().drop(columns=["country_code"])
+    dataset.add(gh.adapt_table_for_grapher(table))
```
larsyencken, line 19: Missing `dataset.save()` at the end, to update table metadata?

Perhaps we should add a `__del__()` method to `Dataset` that does this automatically, since it's not obvious that it's expected or needed.

etl/steps/grapher/cait/2022-08-10/all_ghg_emissions.py
```diff
@@ -27,11 +26,9 @@ def get_grapher_dataset() -> catalog.Dataset:
 
     dataset.metadata = gh.adapt_dataset_metadata_for_grapher(dataset.metadata)
 
-    return dataset
-
+    dataset.save()
```
larsyencken, line 29: Move to the end of the file

etl/steps/grapher/cait/2022-08-10/co2_emissions.py
```diff
@@ -27,11 +26,9 @@ def get_grapher_dataset() -> catalog.Dataset:
 
     dataset.metadata = gh.adapt_dataset_metadata_for_grapher(dataset.metadata)
 
-    return dataset
-
+    dataset.save()
```
larsyencken, line 29: Move to the end of the file (it updates metadata on every table)

etl/steps/grapher/dummy/2020-01-01/dummy.py
```diff
@@ -8,22 +6,16 @@
...
     table = gh.adapt_table_for_grapher(table)
 
     # optionally set additional dimensions
     # table = table.set_index(["sex", "income_group"], append=True)
 
-    # convert table into grapher format
-    # if you data is in long format, use gh.yield_long_table
-    yield from gh.yield_wide_table(table, na_action="drop")
+    # if you data is in long format, check gh.long_to_wide_tables
+    dataset.add(table)
```
larsyencken, line 21: Missing `dataset.save()`

etl/steps/grapher/faostat/2022-05-17/faostat_ei.py
```diff
@@ -1,7 +1,11 @@
...
+from .shared import catalog, get_grapher_dataset_from_file_name, get_grapher_table
 
 
-def get_grapher_dataset() -> catalog.Dataset:
-    return get_grapher_dataset_from_file_name(__file__)
+def run(dest_dir: str) -> None:
+    garden_dataset = get_grapher_dataset_from_file_name(__file__)
+
+    dataset = catalog.Dataset.create_empty(dest_dir, garden_dataset.metadata)
+    dataset.save()
```
larsyencken, line 9: Move to after table add

etl/steps/grapher/faostat/2022-05-17/faostat_emn.py
```diff
@@ -1,7 +1,11 @@
...
+from .shared import catalog, get_grapher_dataset_from_file_name, get_grapher_table
 
 
-def get_grapher_dataset() -> catalog.Dataset:
-    return get_grapher_dataset_from_file_name(__file__)
+def run(dest_dir: str) -> None:
+    garden_dataset = get_grapher_dataset_from_file_name(__file__)
+
+    dataset = catalog.Dataset.create_empty(dest_dir, garden_dataset.metadata)
+    dataset.save()
```
larsyencken, line 9: As above

etl/steps/grapher/faostat/2022-05-17/faostat_esb.py
```diff
@@ -1,7 +1,11 @@
...
+from .shared import catalog, get_grapher_dataset_from_file_name, get_grapher_table
 
 
-def get_grapher_dataset() -> catalog.Dataset:
-    return get_grapher_dataset_from_file_name(__file__)
+def run(dest_dir: str) -> None:
+    garden_dataset = get_grapher_dataset_from_file_name(__file__)
+
+    dataset = catalog.Dataset.create_empty(dest_dir, garden_dataset.metadata)
+    dataset.save()
```
larsyencken, line 9: ...

walkthrough/grapher_cookiecutter/{{cookiecutter.directory_name}}/{{cookiecutter.short_name}}.py
```diff
@@ -8,22 +6,16 @@
...
     table = gh.adapt_table_for_grapher(table)
 
     # optionally set additional dimensions
     # table = table.set_index(["sex", "income_group"], append=True)
 
-    # convert table into grapher format
-    # if you data is in long format, use gh.yield_long_table
-    yield from gh.yield_wide_table(table, na_action="drop")
+    # if you data is in long format, check gh.long_to_wide_tables
+    dataset.add(table)
```
larsyencken, line 21: `dataset.save()`

larsyencken (changes requested):
- The overall approach is good, with minor suggestion to automatically set `--grapher` when `--upsert` is set
- It's slightly weird to resolve the entities to ids in the ETL, since they're useless outside of grapher; my preference would be to keep entity names and add codes
  - The entity name and code is enough for baking
  - It might even be possible to skip MySQL and just use enumerated entity names as entity ids at bake time, in which case there'd be no extra cost to skipping ids completely here (@danyx23, thoughts ?) The code is what's needed to display on a map
 - Basically every step modified had `dataset.save()` either missing or in the wrong place. I wonder if we shouldn't make that easier or more automatic.

etl/grapher_helpers.py
```diff
@@ -219,19 +219,7 @@ def _slugify_column_and_dimensions(column: str, dims: List[str], dim_names: List
...
-) -> Iterable[catalog.Table]:
-    """Yield from long table with the following columns:
-    - variable: short variable name (needs to be underscored)
-    - value: variable value
-    - meta: either VariableMeta object or null in every row
-
-    :param dim_titles: Custom names to use for the dimensions, if not provided, the default names will be used.
-        Dimension title will be used to create variable name, e.g. `Deaths - Age: 10-18` instead of `Deaths - age: 10-18`
-    """
+def _assert_long_table(table: catalog.Table) -> None:
```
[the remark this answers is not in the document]
Marigold: Oooh, didn't know that! Would you mind if I move them down after I merge those hanging PRs?

etl/run_python_step.py
Marigold: These are dynamically generated from datasets in the grapher channel.

etl/steps/grapher/bp/2022-07-14/statistical_review.py
Marigold: I saw the `.save` being called right after creating a dataset in one of the very first examples in ETL (I think it was hyde) and it sticked with me. I thought it has a special reason, but it doesn't. I also find calling `save` at the end more intuitive.

Since this is being used in pretty much all recipes we have, can I create a separate issue & PR for it? Table metadata is updated correctly right now (unless someone changes dataset metadata after calling save, but that's not happening).

larsyencken (approved):
After discussion, good go! Just a typecheck error to fix before merging:

```
.venv/bin/mypy --exclude='etl/steps|walkthrough/.*_cookiecutter'  etl backport walkthrough tests
etl/prune.py:73: error: Unexpected keyword argument "upsert" for "construct_dag"
etl/command.py:171: note: "construct_dag" defined here
```

etl/grapher_helpers.py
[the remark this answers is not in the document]
larsyencken: Totally fine 👍

etl/steps/grapher/bp/2022-07-14/statistical_review.py
larsyencken: Yeah, a separate PR is a good idea 👍

Issue #437, opened 2022-09-05
Bug: misaligned mutliindex in SDG dataset


Marigold:
Recent update to WDI datasets led to `ValueError: Length of names must match number of levels in MultiIndex.` error. There's a misalignment in indexes now from two different datasets that causes it.

PR #438, opened 2022-09-05
fix(sdg): fix multiindex misalignment

Marigold:
Fix for https://github.com/owid/etl/issues/437

### DIFF
--- a/etl/steps/data/garden/sdg/latest/sdg.py
+++ b/etl/steps/data/garden/sdg/latest/sdg.py
@@ -63,6 +63,18 @@ def _indicator_prefix(name: str, indicator: str) -> str:
     return f"indicator_{indicator.replace('.', '_').lower()}__{name}"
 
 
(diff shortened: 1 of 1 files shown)


larsyencken (approved):
It works, and it all looks good.

Something that would have been good in the description is any change to the actual downstream data after this.

It looks to me like it should be changing the index to `(country, year)` for each indicator. But I loaded the old data, and it already seemed to be `(country,  year, ...)` in each of the SDG garden files. So I can't say I fully understand what's going on.

Feel free to merge though 😅

Marigold:
@larsyencken sorry, did it in a rush. Post-mortem description

> Something that would have been good in the description is any change to the actual downstream data after this.

There's no downstream data... yet.

> So I can't say I fully understand what's going on.

Might have been cached files or something like that. No idea. I haven't investigated since as I mentioned, it has no downstream data.

PR #439, opened 2022-09-05
chore: add --ipdb option to enter debugger on failure

larsyencken:
This non-essential flag makes it easier to check where something is going wrong. If you run the `etl` command but add `--ipdb`, it will enter the debugger on failure, regardless of whether you are in normal ETL code or within a Python ETL step.

- Adds a global variable to `etl.config` indicating whether the debugger should be triggered on exceptions
- Adds `--ipdb` options to both the `etl` command and the `run_python_step` command

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -95,6 +95,7 @@ def main_cli(
     )
 
     if ipdb:
+        config.IPDB_ENABLED = True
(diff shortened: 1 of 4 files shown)


Marigold (approved):
Works great.

Issue #442, opened 2022-09-05
WDI garden step connects to grapher DB


larsyencken:
## Problem

- The general public can no longer build our `garden` dataset, since it requires a MySQL database for the WDI step for garden.
- Our `garden` build also became non-reproducible, since it is no longer a strict function of its inputs, but depends on whatever the contents of the grapher DB is
  - In general, even `grapher://` steps shouldn't use data from the DB as input to the dataset, they should only connect in order to do simple upserts of datasets

## Proposed solutions

- Delay any grapher data until a grapher step; OR
- Do a once-off metadata query to grapher, and check it in as a JSON file as part of the recipe for this dataset

## Full error

```
Detecting which steps need rebuilding...
Running 2 steps:
1. data://garden/worldbank_wdi/2022-05-26/wdi...
/Users/lars/Documents/owid/etl/etl/steps/data/garden/worldbank_wdi/2022-05-26/wdi.py:54: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  df["country"] = countries
2022-09-05 11:40:33,711 - [bugsnag] WARNING - No API key configured, couldn't notify
Traceback (most recent call last):
  File "/Users/lars/Documents/owid/etl/.venv/bin/run_python_step", line 6, in <module>
    sys.exit(main())
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/lars/Documents/owid/etl/etl/run_python_step.py", line 26, in main
    _import_and_run(path, dest_dir)
  File "/Users/lars/Documents/owid/etl/etl/run_python_step.py", line 44, in _import_and_run
[...]

PR #443, opened 2022-09-05
fix(wdi): save static copy of grapher metadata rather than dynamically fetching it

Marigold:
Fix for https://github.com/owid/etl/issues/442

Instead of fetching metadata dynamically we save it as feather file and load it at runtime. Minimal changes were made to `VariableMatcher`, it was pretty much just moved to its own file.

I've also considered using `.csv` for metadata, but it would have to be compressed to have small size. Feather seemed like an easier option in the end.


pabloarosado (commented):
I see a 1.11MB feather file inside the steps folder, was that intentional?

Marigold:
It contains metadata about variables ("backported" from grapher DB). Previously it was fetched dynamically from MySQL which wasn't ideal (others wouldn't be able to replicate it). An alternative is to have it in Walden S3 and get the data from there. Given that the file is only 1MB it's in my opinion not worth the hassle.

pabloarosado:
> It contains metadata about variables ("backported" from grapher DB). Previously it was fetched dynamically from MySQL which wasn't ideal (others wouldn't be able to replicate it). An alternative is to have it in Walden S3 and get the data from there. Given that the file is only 1MB it's in my opinion not worth the hassle.

I see. But wouldn't it be better to actually load the backported grapher dataset, and work on it locally? Otherwise, this garden step depends on a grapher dataset (if I have understood correctly). If the grapher dataset changes over time, this step will return a different dataset, or fail.

larsyencken:
It's only 1MB, but it's 1MB forever. The entire repo's only 40MB now, including all history, so it's still quite light.

I'm also wondering whether this is going to be a pattern for steps, so I want to understand how it was supposed to play out, to avoid future binary blobs in the ETL repo. I guess this metadata is also in a backported dataset, so it could be copied from there instead?

Marigold:
Reduced file size to 600kb. Whitelisting only reduced it by 100kb and wasn't really worth it. Also moved everything to `wdi` folder to make it part of checksum.

larsyencken (approved):
Thanks Mojmir, fix linting and then it's good to ship

Issue #445, opened 2022-09-06
Bug: SDG indicator 15.9.1 series ER_BDY_ABT2NP has duplicate values after conversion


Marigold:
It's gonna be easiest to explain it on an example. Here's dataframe that goes into step `data://garden/un_sdg/2022-07-07/un_sdg`

<img width="819" alt="image" src="https://user-images.githubusercontent.com/1550888/188632746-6787b283-e963-47aa-99fb-febf4e8b5bfb.png">

I've included only France and World on purpose. There's a [custom code](https://github.com/owid/etl/blob/master/etl/steps/data/garden/un_sdg/2022-07-07/un_sdg.py#L327) that handles specifically this indicator. Since there's only one `level_status` (excluding `No breakdown`) with value 1 for **countries**, the custom code correctly converts the variable into string with value `National target reflecting ABT2  exists and pro...`.

This works well for countries, but it doesn't work for regions such as World in the example. We then end up with World having 6 distinct values for year 2021, variable ER_BDY_ABT2NP and entity World. That is neither allowed in grapher nor makes much sense.

@spoonerf what do you think would be the easiest fix? Can I just exclude regions or should I create a special variable for that? (Might be less trivial)

spoonerf:
Hey @Marigold

Ah good spot! For this one I think it is best to just ignore the regions as it's kind of meaningless for them anyway, for our purposes at least. 

Let me know if I can do anything to help!

PR #446, opened 2022-09-06
Bug: SDG indicator 15.9.1 series ER_BDY_ABT2NP has duplicate values after conversion

Marigold:
Fix for https://github.com/owid/etl/issues/445


spoonerf (approved):
Looks great! Thanks for spotting and fixing it!

PR #452, opened 2022-09-08
enhance(compare): preview samples and fix minor bugs

Marigold:
* Improve error messages when table is not found
* Add `--debug` flag for easier development
* Fix a couple of issues that popped up when running `compare etl-catalog garden faostat faostat_ek faostat_ek$ --show-values`


danyx23 (approved):
Looks great - thanks for improving this tool!

PR #454, opened 2022-09-08
Add smil energy dataset

pabloarosado:
Add meadow and garden steps for Smil global primary energy dataset.

### DIFF
--- a/etl/steps/data/garden/smil/2017-01-01/global_primary_energy.py
+++ b/etl/steps/data/garden/smil/2017-01-01/global_primary_energy.py
@@ -1,0 +1,44 @@
+import pandas as pd
+from owid.catalog import Dataset, Table
+from owid.catalog.utils import underscore_table
+from structlog import get_logger
+
+from etl.helpers import Names
+from etl.paths import DATA_DIR
+
+log = get_logger()
+
+# naming conventions
+N = Names(__file__)
+
+
+def run(dest_dir: str) -> None:
+    log.info("global_primary_energy.start")
+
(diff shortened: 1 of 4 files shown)


etl/steps/data/garden/smil/2017-01-01/global_primary_energy.py
```diff
@@ -0,0 +1,44 @@
...
+
+# naming conventions
+N = Names(__file__)
+
+
+def run(dest_dir: str) -> None:
+    log.info("global_primary_energy.start")
+
+    # Load dataset from meadow.
+    ds_meadow = Dataset(DATA_DIR / "meadow/smil/2017-01-01/global_primary_energy")
```
Marigold, line 19: This could be replaced by `N.meadow_dataset` if you decide to keep `Names` object.

Marigold (approved):
Looks good! (by the way the data is only till 1960, but I assume that's expected)

etl/steps/data/garden/smil/2017-01-01/global_primary_energy.py
pabloarosado: Oh, I thought N would only contain names, I find it a bit weird that it contains the whole dataset.

Marigold: It is a bit weird, but useful syntactic sugar. (`Names` object doesn't keep the dataset as an attribute, it's just a helper function that ended up in that class)

PR #459, opened 2022-09-12
feature(compare): add subcommand for grapher comparison

Marigold:
Add command for comparing grapher metadata. For instance

```
compare  --show-values grapher ggdc 2020-10-01 ggdc_maddison__2020_10_01 --data-values
```

The last commit refactors code by moving shared arguments into parent command. That DRYs the code, but those arguments must be put after `compare` command and are not shown in `compare grapher --help`. I'm not married to it and totally fine with leaving them as they are right now.

### DIFF
--- a/etl/compare.py
+++ b/etl/compare.py
@@ -251,6 +251,18 @@ def grapher(
         **ctx.obj,
     )
 
+    remote_sources_df = read_sources_from_db(remote_env, namespace, version, dataset)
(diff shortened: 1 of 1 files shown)


Marigold:
@danyx23 would you mind taking a quick look, please? I'm starting to use this a lot and it'd be cool to have this merged before any further improvements.

danyx23 (approved):
Nice! Sorry for the late review!

PR #465, opened 2022-09-13
enhance(population): use un wpp 2022

lucasrodes:
Switch to UN WPP 2022 revision population for data starting from 1950:

New countries (only UN data, i.e. > 1950 year):
- Kosovo
- Guernsey
- Jersey

Deleted:
- Channel Islands (now Guernsey and Jersey)
- Melanesia, Polynesia (not countries)


larsyencken (changes requested):
Looks good. You just missed updating `dag.yaml` for `data://garden/owid/latest/key_indicators` to depend on the new WPP version.

PR #471, opened 2022-09-15
refactor(grapher steps): make adapt functions implicit

Marigold:
Move functions `adapt_dataset_metadata_for_grapher` and `adapt_table_for_grapher` to the step that upserts data into DB. Remove those functions from all existing steps.

`country_to_entity_id` has been modified to work with country codes.

This moves creating entities, which need access to MySQL, to the upsert step. In the future we'd want all steps from channels meadow - garden - grapher to work without access to MySQL. There are still a few steps that use `country_to_entity_id` in the grapher channel, but that'll be fixed in the next PR.

### DIFF
--- a/etl/steps/grapher/faostat/2022-05-17/shared.py
+++ b/etl/steps/grapher/faostat/2022-05-17/shared.py
@@ -7,7 +7,6 @@ from pathlib import Path
 import pandas as pd
 from owid import catalog
 
-from etl import grapher_helpers as gh
 from etl.paths import DATA_DIR, STEP_DIR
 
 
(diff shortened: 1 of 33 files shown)


etl/steps/grapher/faostat/2022-05-17/shared.py
```diff
@@ -43,9 +42,6 @@ def get_grapher_dataset_from_file_name(file_path: str) -> catalog.Dataset:
     # Load latest garden dataset.
     dataset = catalog.Dataset(garden_data_dir)
 
-    # Adapt dataset metadata to grapher requirements.
-    dataset.metadata = gh.adapt_dataset_metadata_for_grapher(dataset.metadata)
```
pabloarosado, line 47: I'm a bit confused by this case. Wouldn't the new dataset have no metadata if we remove this line?

pabloarosado (approved):

etl/steps/grapher/faostat/2022-05-17/shared.py
Marigold: It loads the metadata of garden dataset on the line above
```
dataset = catalog.Dataset(garden_data_dir)
```
this PR changes makes `grapher://` run `adapt_dataset_metadata_for_grapher` by default for all steps right before upserting to database.

(I've double checked this with `compare --show-values grapher faostat 2022-05-17 faostat_fs --data-values`)

pabloarosado: Ah fantastic, thank you!

PR #476, opened 2022-09-15
enhance(grapher_helpers): combine sources of variables

Marigold:
If table contains variable with multiple sources, combine them into a single source ready for upsert to grapher.

I've also slightly refactored `combine_metadata_sources` as `adapt_*` methods will be soon applied to all datasets by default. I haven't changed the original logic, but to be honest I find it quite confusing how we make description fallbacks to dataset description or table description. I couldn't think of a better solution though.

@pabloarosado are you fine with the existing fallback logic? If yes then I'd say we go ahead and keep it like that until we either find it problematic or design a better schema for sources.

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -368,27 +368,25 @@ def _unique(x: List[Any]) -> List[Any]:
     return list(dict.fromkeys(x))
 
 
-def combine_metadata_sources(metadata: catalog.DatasetMeta) -> catalog.DatasetMeta:
-    """Combine each of the attributes in the sources of a dataset's metadata, and assign them to the first source, since
+def combine_metadata_sources(sources: List[catalog.Source]) -> catalog.Source:
+    """Combine each of the attributes in the sources and assign them to the first source, since
     that is the only source that grapher will read.
 
     Parameters
     ----------
-    metadata : catalog.DatasetMeta
-        Dataset metadata.
+    sources : List[Source]
+        List of sources to combine.
 
     Returns
     -------
-    metadata : catalog.DatasetMeta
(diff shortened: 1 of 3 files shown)


pabloarosado:
Hi @Marigold I think our decisions at this point depend on what happens later in grapher. It's hard for me to foresee the final outcome in a chart after these changes, so I'm going to write here two conditions that (I think) should be fulfilled:
1. From the visitor's perspective: Charts must have, for each variable, both the variable descriptions and the grapher-dataset description (of each variable) in the SOURCES tab*.
2. From the data manager's perspective: I think it makes sense that, if you don't explicitly modify anything, the metadata will combine the metadata of all sources. But if you want to replace something (e.g. the description) you can just modify that field in the corresponding metadata yaml file. Therefore falling back to the previous data step makes sense to me.

Do you agree with these conditions? If so, we could test with some dummy datasets in local grapher to see if they still work. And if so, it's all good to go.

*There is one caveat with condition 1: When you have a chart using multiple variables from the same grapher-dataset, you end up with the same description multiple times in SOURCES, like in [this chart](https://ourworldindata.org/grapher/share-electricity-source-facet?country=~GBR). I don't know what the best solution is. I guess keeping the description is still good, but we could collapse the descriptions to avoid them being so overwhelming.

pabloarosado (approved):

Marigold:
@pabloarosado merged the original fallback heuristics and added combining sources for variables. It works ok (for key indicators for instance), but those magic fallbacks to dataset / table descriptions are confusing. Perhaps it'd be better to do this magic with anchors in yaml file instead of having implicit rules, but I'm leaving it for further discussion.

PR #477, opened 2022-09-16
refactor(gbd): use wide instead of long format

Marigold:
Convert GBD into wide format to improve performance and simplify code.


Marigold:
@spoonerf feel free to merge or close this as you wish.

PR #494, opened 2022-09-21
feature: new command tool

lucasrodes:
Command tool to submit the new variable mapping to Grapher to generate the appropriate chart revisions.

Its usage is:

```
~ etl-chart-suggester --help
Usage: etl-chart-suggester [OPTIONS]

  Using the <mapping-file> JSON file mapping variable IDs from an old dataset
  to a new dataset, it creates a bunch of suggested chart revisions in
  Grapher.

  It connects to Grapher based on the environment file found in the project's
  root directory `path/to/etl/.env`.

Options:
  -f, --mapping-file TEXT  Path to variable mapping json file.  [required]
  -n, --namespace TEXT     Namespace of the dataset.  [required]
  -v, --version TEXT       Version of the dataset.  [required]
  -d, --dataset-name TEXT  Name of the dataset.  [required]
  --help                   Show this message and exit.
```

Currently, it is using a hacky solution. Basically it temporarily copies the `mapping-file` to `etl/steps/grapher/<namespace>/<version>/config/` under the filename `variable_replacements.json` and runs [ChartRevisionSuggester](https://github.com/owid/etl/blob/c78edaee01add1089d3d21cd19b673547bb00ce3/etl/chart_revision_suggester.py#L22). Then all files are rolled back to their versions prior to command execution.

### DIFF
--- a/etl/chart_revision_suggester.py
+++ b/etl/chart_revision_suggester.py
@@ -5,15 +5,15 @@ It connects to Grapher based on the environment file found in the project's root
 """
 # The original script was originally from the owid/importers repo: https://github.com/owid/importers/blob/master/standard_importer/chart_revision_suggester.py
 
-import click
 import os
 import re
+import shutil
+import time
 import traceback
 from copy import deepcopy
 from typing import Any, Dict, Iterable, List, Literal, Optional, Tuple, Union
-import time
-import shutil
 
+import click
 import pandas as pd
 import simplejson as json
 import structlog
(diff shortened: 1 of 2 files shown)


etl/chart_revision_suggester.py
```diff
@@ -1,11 +1,19 @@
+"""
+Using the <mapping-file> JSON file mapping variable IDs from an old dataset to a new dataset, it creates a bunch of suggested chart revisions in Grapher.
+
+It connects to Grapher based on the environment file found in the project's root directory `path/to/etl/.env`.
```
Marigold, line 4: FYI you can use `ENV=.env.staging etl-chart-suggester ...` or `ENV=.env.production etl ...` assuming your staging / prod envs are in file `.env.staging`, ... This seemed less prone to errors than having a single `.env` and modifying its contents.

Marigold (approved):
Neat!

PR #495, opened 2022-09-21
Join sources with semicolon

pabloarosado:
* Improve the logic of how to join metadata sources (for example, use `;` without additional spaces when joining source names).
* Fix some bad source names in the electricity mix dataset.


etl/steps/data/garden/energy/2022-08-03/electricity_mix.meta.yml
```diff
@@ -69,13 +69,13 @@ dataset:
...
     -
-      name: Our World in Data based on Ember's Global Electricity Review (2022).
+      name: Our World in Data based on Ember's Global Electricity Review (2022)
       published_by: Ember
       publication_year: 2022
       date_accessed: 2022-08-01
       url: https://ember-climate.org/data/data-catalogue/yearly-electricity-data/
     -
-      name: Our World in Data based on Ember's European Electricity Review (2022).
+      name: Our World in Data based on Ember's European Electricity Review (2022)
```
lucasrodes, line 78: Minor thing, but probably this does not belong in this PR?

lucasrodes (approved):

etl/steps/data/garden/energy/2022-08-03/electricity_mix.meta.yml
Marigold: I think we're doing worse things than this :D.

Marigold (approved):

etl/steps/data/garden/energy/2022-08-03/electricity_mix.meta.yml
pabloarosado: The thing is that I randomly picked this file to use as a test for the main changes of this PR :-).

PR #497, opened 2022-09-22
chore: Update datautils

pabloarosado:
Update datautils version to v0.4.10-alpha.


Marigold:
Thanks! I was just about to do it myself.

PR #498, opened 2022-09-22
fix(gbd): reduce memory peak usage

Marigold:
Meadow gbd step needs too much memory - apparently because `drop_duplicates` is such a memory intensive function. I've tried several approaches, but the only one that worked was processing feather file in chunks (one metric per chunk) and cleaning them separately.

This was really painful process, one more reason to get a bigger machine (cc @larsyencken).


spoonerf (approved):

larsyencken:
Ah, clever! You read in independent partitions, and do the cleaning on each one separately before merging them. Nice work!!! 🎉

Issue #499, opened 2022-09-22
Unable to run grapher step for electricity mix dataset (sources related?)


pabloarosado:
Step `grapher://grapher/energy/2022-08-03/electricity_mix` fails. The error starts with:
```
MySQLdb.DataError: (1406, "Data too long for column 'name' at row 1")
```
And in the end it has:
```
sqlalchemy.exc.DataError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(MySQLdb.DataError) (1406, "Data too long for column 'name' at row 1")
[SQL: INSERT INTO sources (description, `createdAt`, `updatedAt`, name, `datasetId`) VALUES (%s, %s, %s, %s, %s)]
[parameters: ('{"link": "https://www.bp.com/en/global/corporate/energy-economics/statistical-review-of-world-energy.html ; https://ember-climate.org/data/data-catal ... (444671 characters truncated) ...  Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember; Ember", "dataPublisherSource": null}', datetime.datetime(2022, 9, 22, 15, 53, 10, 825422), datetime.datetime(2022, 9, 22, 15, 53, 10, 825423), "Our World in Data based on BP Statistical Review of World Energy (2022); Our World in Data based on Ember's Global Electricity Review (2022); Our Wor ... (6493 characters truncated) ... ew (2022); Our World in Data based on Ember's Global Electricity Review (2022); Our World in Data based on Ember's European Electricity Review (2022)", 5678)]
(Background on this error at: https://sqlalche.me/e/14/9h9h)
```
I haven't looked into it, but there I see the source name repeated many times, so, maybe there's a problem when combining sources, which we have modified recently?

larsyencken:
Yeah, maybe we need to be deduplicating already here. It doesn't look like a great end-user experience to have all the double-ups.

Marigold:
@pabloarosado it's a bug, `combine_metadata_sources` was modifying the input sources. Fix https://github.com/owid/etl/pull/500

PR #500, opened 2022-09-22
fix(combine_sources): do not modify original source

Marigold:
Fix for https://github.com/owid/etl/issues/499


etl/grapher_helpers.py
```diff
@@ -383,7 +383,7 @@ def combine_metadata_sources(sources: List[catalog.Source]) -> catalog.Source:
     assert len(sources) >= 1, "Dataset needs to have at least one source in metadata."
 
     # Define the 'default_source', which will be the one where all sources' attributes are combined.
-    default_source = sources[0]
+    default_source = copy.deepcopy(sources[0])
```
pabloarosado, line 386: Argh, of course! :'-D

pabloarosado (approved):
Thanks a lot for finding and fixing it!

PR #501, opened 2022-09-23
chore: limit steps to 16G of memory on Linux

larsyencken:
Use `prlimit` on Linux to limit steps to 16G of virtual memory.

We want the ETL to run correctly on machines with 16GB of memory, of which not all can be dedicated to the ETL. We also want to be able to potentially parallelise steps in future.

By setting this limit, we force expensive steps to break predictably on Linux rather than have them sometimes work and sometimes not depending on other factors.

## Why 16GB?

16GB is a memory size that's increasingly the default for data work on laptops and reasonably priced cloud servers, so that's the default we choose here.

Conveniently, all our current work fits into 16GB of memory, so merging this constraint will only cause future memory-intensive work to fail rather than existing work.

### DIFF
--- a/etl/config.py
+++ b/etl/config.py
@@ -41,6 +41,10 @@ IPDB_ENABLED = False
 # NOTE: this will soon be deprecated after we get rid of data_values
 GRAPHER_INSERT_WORKERS = int(env.get("GRAPHER_WORKERS", 10))
 
+# forbid any individual step from consuming more than this much memory
+# (only enforced on Linux)
+MAX_VIRTUAL_MEMORY_LINUX = 10 * 2**30  # 10 GB
+
 
 def enable_bugsnag() -> None:
     BUGSNAG_API_KEY = env.get("BUGSNAG_API_KEY")
--- a/etl/config.py
+++ b/etl/config.py
@@ -43,7 +43,7 @@ GRAPHER_INSERT_WORKERS = int(env.get("GRAPHER_WORKERS", 10))
 
 # forbid any individual step from consuming more than this much memory
 # (only enforced on Linux)
-MAX_VIRTUAL_MEMORY_LINUX = 10 * 2**30  # 10 GB
+MAX_VIRTUAL_MEMORY_LINUX = 16 * 2**30  # 16 GB
 
 
 def enable_bugsnag() -> None:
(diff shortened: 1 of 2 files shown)


larsyencken:
Related to: #498

Marigold:
This is cool! At least we'll see the processes failing locally... However, I'm not sure if our current steps are gonna make it. `faostat` and `ihme_gbd` datasets are at the very limit and there's no more low hanging optimisation fruit there. Have you tried rebuilding ETL locally with this? (or perhaps we can try on staging, but need to include `grapher` step first)

We need more memory... or swap.

larsyencken:
⌛ Currently running the full ETL from scratch on `analytics` to confirm that all our existing steps work with this limit.

## On MacOS

Could not find a working tool to limit memory usage from the command-line. In particular:

- Neither `ulimit` or `/usr/bin/ulimit` seem to allow setting the a virtual memory limit on MacOS
- `resources.setrlimit()` does not seem to honour the `RLIMIT_RSS` flag on MacOS (see also `man getrlimit`), it's just ignored for compatibility.

Any other ideas here would be awesome.

larsyencken:
@Marigold If we have some steps that are mega expensive, I'd rather we find a way to spool intermediate results to temp files on disk rather than expect that we can use a huge amount of memory. Otherwise we'll end up with a system that we can't run on normal laptops.

Marigold:
@larsyencken huge memory is something different, we should definitely optimise these steps, but when the process uses <20gb memory it can easily run on laptops with 16gb mem with swap. That's what has been happening - users optimise their steps to the point where they can run it locally and then push them. If our server had 32gb we'd have enough margin to run most of these steps without needing to overoptimize.

> Any other ideas here would be awesome.

This is just an idea, but could we in `run_python_step` run tracemalloc and then check peak memory afterwards and raise a warning / error / log if it's over >10gb? That would work cross-platform

```python
tracemalloc.start()
module.run(dest_dir)
_, peak_mem = tracemalloc.get_traced_memory()
tracemalloc.stop()
if peak_mem > xxx:
  raise ...
```

(I haven't tried it, so it might not work. I'm not an expert on memory profiling)

larsyencken:
Rebased and updated this today. I'm running a manual BuildKite build to see what currently breaks.

larsyencken:
@Marigold I did some more tests today, and we get one thing breaking if 14GB is the limit, but nothing breaking at 16GB.

I'd prefer to merge this and roll forward with it rather than have no limit have it bite us later on.

Issue #466, opened 2022-09-14
Step checksum does not include `shared.py` module


Marigold:
When deciding whether DataStep is dirty, we calculate checksum of its [step files](https://github.com/owid/etl/blob/master/etl/steps/__init__.py#L347). There are two situations that can happen:

1. If there's a folder `[short_name]/*` then we calculate checksum of all its files
2. If there's a module `[short_name].py` then we calculate checksum of all files in the same directory with prefix `[short_name].`

This is relatively unknown thing and users assume checksum just magically works for **all files**. We heavily use `shared.py` module that often contains code shared between datasets with different names (like [here](https://github.com/owid/etl/tree/master/etl/steps/data/garden/energy/2022-09-09), but we use it all over the place). Since `shared.py` doesn't make it to step files either way, code changes won't be reflected in the checksum.

What would be the best way to fix this? I can think of:

1. Adding all `shared.*` files to step files (hardcoding)
2. Move everything into `[short_name]/*` directories (but then you cannot share `shared.py` between datasets)
3. Adding **all** files into step files
4. Dynamically get all dependencies of `[short_name].py` and include them as step files (no idea how hard can it be)

@larsyencken any elegant solutions?

larsyencken:
Hmm... my rough preferences in order:

1. (less magic) Symlink `shared.py --> ../shared.py` for these cases, then they'll be picked up in the checksum
2. (more magic) Convention that `../shared*.py` is also included in the checksum, when available

Marigold:
I'm gonna implement 1. as a hotfix and will be closely watching incoming PRs for these types of mistakes. Even though 2. is less magic, using symlinks might look like more magic to users.

Marigold:
As an another heuristic I think it'd be useful to also checksum files matching `short_name*` and not just `short_name.*` as people are using modules like `[short_name]_tools.py`.

larsyencken:
> As an another heuristic I think it'd be useful to also checksum files matching `short_name*` and not just `short_name.*` as people are using modules like `[short_name]_tools.py`.

It's more vulnerable to overlap though, like `population.py` and `population_by_age.py`, or some similar situations. 

It's honestly hard to get implicit detection right. Perhaps we should be encouraging more folders.

larsyencken:
I'd put them in folders and symlink them both in, rather than come up with fancy rules.

Marigold:
[...]

PR #502, opened 2022-09-23
enhance(steps): add shared modules to checksum

Marigold:
Partial fix for https://github.com/owid/etl/issues/466

## Problem

If you make any changes to `shared.py` (e.g. in `faostat/`), its checksum won't change and running `etl faostat_fs` is gonna tell you `All datasets up to date!` even though its recipe has changed.

Locally you can get around this by using `--force --only`, but once you push it to production, etl on our server won't rebuild the dataset.

ETL only looks for changes in `[dataset].py` or if you use directory `[dataset]/*` it looks for all changes in that directory (this would be best, but it would mean creating many redundant directories e.g. for faostat).

## Solution

In addition to files above, look for changes in all `shared*` files (`shared.py`).

We could also got with directories and have symlinks for shared files, but that might be an overkill compared to this heuristic.

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -23,8 +23,6 @@ config.enable_bugsnag()
 
 WALDEN_NAMESPACE = os.environ.get("WALDEN_NAMESPACE", "backport")
 
-THREADPOOL_WORKERS = 5
-
 # if the number of open files allowed is less than this, increase it
 LIMIT_NOFILE = 5000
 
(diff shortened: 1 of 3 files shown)


pabloarosado (approved):
Nice. I think often you may also just make a small change in `shared.py` that will affect only one of the datasets in the folder, and you don't want to execute all steps. But for that case you can use the `--only` flag. In general I think it makes sense to declare all steps dirty if a shared module has changed.

Marigold:
Yeah, it's a tradeoff. Though it's better to have up to date datasets in our S3 catalog than outdated.

PR #506, opened 2022-09-23
Fix issue when loading population

pabloarosado:
Several garden steps were loading the `population` table by using
```
catalog.find("population", namespace="owid", dataset="key_indicators").load()
```
But this was failing because `catalog.find` can't distinguish between `population` and `population_density`.

This PR replaces that line with:
```
catalog.Dataset(DATA_DIR / "garden/owid/latest/key_indicators/")["population"]
```
to fix the problem. And it also adds `key_indicators` as a dependency to all affected steps (the dependency should have been there anyway).


larsyencken (approved):
Approach is solid, and the steps seem to run cleanly :shipit:

PR #511, opened 2022-09-26
Add long term uk electricity data

pabloarosado:
Add data steps for long-term uk electricity data.


Marigold (approved):
Looks good, I was able to run all steps locally and variables in admin look ok too.

Issue #257, opened 2022-06-28
Backported data is intermittently inconsistent due to updating in-place


larsyencken:
## Problem

When a backported dataset is updated, it overwrites the old object in S3. This invalidates all previous walden versions, which no longer point to data with a valid checksum.

## Proposed solution

Overwrite the walden record, but not the S3 object. Instead, treat S3 as an append-only store and use the has in the S3 path.

## Example

```
977. walden://backport/latest/dataset_5085_google_mobility_trends__2020_values...
Traceback (most recent call last):
  File "/Users/lars/Documents/owid/etl/.venv/bin/etl", line 5, in <module>
    main()
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/lars/Documents/owid/etl/.venv/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/lars/Documents/owid/etl/etl/command.py", line 93, in main
    run_dag(
  File "/Users/lars/Documents/owid/etl/etl/command.py", line 159, in run_dag
    time_taken = timed_run(lambda: step.run())
  File "/Users/lars/Documents/owid/etl/etl/command.py", line 166, in timed_run
    f()
  File "/Users/lars/Documents/owid/etl/etl/command.py", line 159, in <lambda>
    time_taken = timed_run(lambda: step.run())
  File "/Users/lars/Documents/owid/etl/etl/steps/__init__.py", line 440, in run
    self._walden_dataset.ensure_downloaded(quiet=True)
  File "/Users/lars/Documents/owid/etl/vendor/walden/owid/walden/catalog.py", line 196, in ensure_downloaded
    files.download(url, filename, expected_md5=self.md5, quiet=quiet)
  File "/Users/lars/Documents/owid/etl/vendor/walden/owid/walden/files.py", line 90, in download
    raise ChecksumDoesNotMatch(f"for file downloaded from {url}")
[...]

PR #513, opened 2022-09-26
draft: use checksum in walden backports instead of latest

larsyencken:
See: https://github.com/owid/etl/issues/257

Currently when backporting, we overwrite previous versions of the dataset. This changes the checksum, which basically means that every current copy of the ETL anywhere on any machine now is currently invalid, and will fail to handle this backport step without an explicit code update.

This breaks a pretty big contract of the ETL, namely that it is repeatable.

**Principle:** the ETL should not break because of a side-effect happening on a remote machine; the previous and next versions of the ETL should live side-by-side and both work

This change moves us from overwriting the previous backport to appending it, by using the md5 of the data instead of "latest" in the path.

It doesn't yet solve the problem of how we deal with having multiple copies around.

### DIFF
--- a/backport/backport.py
+++ b/backport/backport.py
@@ -16,7 +16,7 @@ from etl import config
 from etl import grapher_model as gm
 from etl.backport_helpers import GrapherConfig
 from etl.db import get_engine
-from etl.files import checksum_str
+from etl.files import checksum_str, checksum_file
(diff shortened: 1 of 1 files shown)


larsyencken:
Closing this one as stale.

PR #519, opened 2022-09-26
data(internet): add dataset on Internet metrics

lucasrodes:
Garden and Grapher steps to generate OWID custom Internet dataset. 

Used in charts like this one: https://ourworldindata.org/grapher/number-of-internet-users-by-country


dag.yml
```diff
@@ -159,6 +159,13 @@ steps:
   data://grapher/un/2022-07-11/un_wpp:
     - data://garden/un/2022-07-11/un_wpp
 
+  # Internet
+  data://garden/owid/latest/internet:
```
pabloarosado, line 163: Why should this dataset be under "owid"?

pabloarosado (approved):

dag.yml
lucasrodes: So far I got the impression that namespaces always were entities/orgs. So in this case we are the ones creating the dataset...

What do you propose? Having a namespace like 'technology'?

PR #528, opened 2022-09-27
fix(etl): minor improvements

Marigold:
Couple minor fixes / improvements.

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -340,11 +340,7 @@ def country_to_entity_id(
     if entity_id.isnull().any() and fill_from_db:
         ix = entity_id.isnull()
         db_entities = _get_entities_from_db(set(country[ix]), by=by)
-        # NOTE: this is hotfix, check out entity_id type
-        if entity_id.isnull().all():
-            entity_id = country.map(db_entities)
-        else:
-            entity_id[ix] = country[ix].map(db_entities)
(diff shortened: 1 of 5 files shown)


etl/steps/data/garden/un/2022-07-11/population.py
```diff
@@ -122,7 +122,7 @@ def _add_metric_sexratio(df: pd.DataFrame) -> pd.DataFrame:
 
 
 def _add_metric_population(df: pd.DataFrame) -> Tuple[pd.DataFrame, pd.DataFrame]:
-    df_p = df[df.metric == "population"]
+    df_p = df.loc[df.metric == "population"]
```
pabloarosado, line 125: Does `.loc` make any difference here?

pabloarosado (approved):

etl/steps/data/garden/un/2022-07-11/population.py
Marigold: It does the same thing, but `.loc` works better with type hints (I'm getting warning in VSCode otherwise and wrongly inferred `pd.Series` type)

pabloarosado: Oh interesting. Yeah, I suppose using `.loc` is better practice than not using it, although in reality (unless doing assignments) it doesn't make a difference.

Issue #527, opened 2022-09-27
ETL seems to be rebuilding far too often


larsyencken:
## Problem

Anecdotally, it seems like the ETL is rebuilding too much, far too often.

## Things we could check

- [ ] Do we get the same data checksums building on Linux and MacOS
- [ ] If we rebuild several hours later, are only steps with new data needing updates

larsyencken:
cc @Marigold

PR #530, opened 2022-09-27
fix(files): ignore __pycache__ when walking files

Marigold:
Fix for https://github.com/owid/etl/issues/527

There was a bug that didn't ignore `__pycache__` which is why our checksums were different on different machines. `files.walk` is being used for [calculating checksums](https://github.com/owid/etl/blob/master/etl/steps/__init__.py#L361).


larsyencken (approved):
Woo! Good find 👑

Issue #534, opened 2022-09-29
etl-translate-varmap fails when having a dictionary of only one variable


pabloarosado:
I was executing etl-translate-varmap to translate the variables of a dataset from staging to live, with:
```
etl-translate-varmap -e1 ./.env.staging -e2 ./.env.live -m1 /Users/prosado/Desktop/staging_mapping.json -m2 /Users/prosado/Desktop/live_mapping.json
```
This is what I got:
```
  File "/Users/prosado/Documents/owid/repos/etl/etl/variable_mapping_translate.py", line 83, in main_cli
    mapping_2 = var_translator.translate()
  File "/Users/prosado/Documents/owid/repos/etl/etl/variable_mapping_translate.py", line 111, in translate
    mapping_2 = variable_mapping_translate(eng_1, eng_2, self.mapping_1)
  File "/Users/prosado/Documents/owid/repos/etl/etl/variable_mapping_translate.py", line 166, in variable_mapping_translate
    mapping_new = build_mapping_from_df(sql_2, df)
  File "/Users/prosado/Documents/owid/repos/etl/etl/variable_mapping_translate.py", line 252, in build_mapping_from_df
    dix: Dict[str, str] = df.set_index("id_old")["id_new"].squeeze().to_dict()
AttributeError: 'str' object has no attribute 'to_dict'
```
My intuition is that the squeeze returns a different type of object if the array has 1 or more elements. So, while
```
pd.Series([1, 2]).squeeze().to_dict()
```
works,
```
pd.Series([1]).squeeze().to_dict()
```
fails.

lucasrodes:
Should be fixed now

pabloarosado:
Thanks a lot @lucasrodes!

PR #536, opened 2022-09-29
fix(cmd): unnecessary squeeze (solves #534)

lucasrodes:
There was an unnecessary call to `pd.Series.squeeze`. 

Most of the time, it would squeeze a Series of length > 1, and would not complain (as it leaves the object unchanged). However, if the Series was from length 1, it would return a scalar, and subsequent operations would fail. An example of this is #534.


lucasrodes:
@Marigold I was not aware of these stubs for pandas! Looks very helpful in `etl` and possibly other repos too, due to our extensive usage of pandas.

PR #533, opened 2022-09-28
Adding the WHO global health estimates dataset

spoonerf:
It includes all available diseases/injuries but only for the largest aggregate: total age group and both sexes

PR #538, opened 2022-09-29
fix(grapher): move country_to_entity_id from grapher steps and keep only country column

Marigold:
Grapher steps should work without MySQL, since `country_to_entity_id` is trying to create entities in MySQL, we have to get rid of it in grapher steps and leave just country there.

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -11,6 +11,7 @@ import structlog
 import yaml
 from owid import catalog
 from owid.catalog.utils import underscore
+from owid.datautils import dataframes
 from pydantic import BaseModel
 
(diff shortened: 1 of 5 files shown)


pabloarosado:
Maybe this should wait until @spoonerf merges [this other PR](https://github.com/owid/etl/pull/533) (or vice versa).

spoonerf:
> Maybe this should wait until @spoonerf merges [this other PR](https://github.com/owid/etl/pull/533) (or vice versa).

merged!

larsyencken (approved):
Nice! Confirmed successful `data://grapher` channel build from scratch without MySQL now.

:shipit:

PR #542, opened 2022-10-01
Adding full Global Health Estimates data to etl

spoonerf:
The previous upload https://owid.cloud/admin/datasets/5785 did not have the full dataset. We can probably delete the original one from grapher.


data/garden/reference/countries_regions.csv
```diff
@@ -2,7 +2,7 @@ code,name,iso_alpha2,iso_alpha3,imf_code,cow_letter,cow_code,unctad_code,marc_co
 AFG,Afghanistan,AF,AFG,512.0,AFG,700.0,AFG,AF,AFGN,AFG,AFG,http://www.wikidata.org/entity/Q889,Afghanistan,15.0,562.0,,
```
pabloarosado, line 2: I don't know how this happened, but I don't think we should add country codes as aliases. The aliases are for country names that are called different things (e.g. "United States", "United States of America").

etl/steps/data/garden/who/2022-09-30/ghe.countries.json
```diff
@@ -0,0 +1,185 @@
...
+  "VCT": "Saint Vincent and the Grenadines",
+  "VEN": "Venezuela",
+  "VNM": "Vietnam",
+  "VUT": "Vanuatu",
+  "WSM": "Samoa",
+  "YEM": "Yemen",
+  "ZAF": "South Africa",
+  "ZMB": "Zambia",
+  "ZWE": "Zimbabwe"
+}
```
pabloarosado, line 185: If a dataset provides ISO country codes, we don't need to harmonize names. The countries_regions dataset already has those ISO codes, which you could use directly to get the harmonized country names.

etl/steps/data/garden/who/2022-09-30/ghe.py
```diff
@@ -0,0 +1,67 @@
...
+
+# naming conventions
+N = Names(__file__)
+
+
+def run(dest_dir: str) -> None:
+    log.info("ghe.start")
+
+    # read dataset from meadow
+    ds_meadow = Dataset(DATA_DIR / "meadow/who/2022-09-30/ghe")
```
pabloarosado, line 20: I think you can replace this line with
`ds_meadow = N.meadow_dataset`

etl/steps/data/garden/who/2022-09-30/ghe.py
```diff
@@ -0,0 +1,67 @@
...
+    tb_meadow = ds_meadow["ghe"]
+
+    df = pd.DataFrame(tb_meadow)
+
+    log.info("ghe.harmonize_countries")
+    df = harmonize_countries(df)
+
+    ds_garden = Dataset.create_empty(dest_dir)
+    ds_garden.metadata = ds_meadow.metadata
```
pabloarosado, line 29: These two lines can be replaced by one:
```

ds_garden = Dataset.create_empty(dest_dir, metadata=ds_meadow.metadata)
```

pabloarosado (changes requested):

etl/steps/data/garden/who/2022-09-30/ghe.countries.json
spoonerf: Thanks @pabloarosado - this has been resolve now.

etl/steps/data/garden/who/2022-09-30/ghe.py
```diff
@@ -0,0 +1,65 @@
...
+    ds_garden.metadata.update_from_yaml(N.metadata_path)
+    tb_garden.update_metadata_from_yaml(N.metadata_path, "ghe")
+
+    ds_garden.add(tb_garden)
+    ds_garden.save()
+
+    log.info("ghe.end")
+
+
+def harmonize_countries(df: pd.DataFrame) -> pd.DataFrame:
```
pabloarosado, line 43: You can remove this function now, if it's not used.

dag_files/dag_walkthrough.yml
```diff
@@ -13,3 +13,9 @@ steps:
   - data://garden/un/2022-07-11/un_wpp
   data://grapher/who/2022-07-17/who_vaccination:
   - data://garden/who/2022-07-17/who_vaccination
+  data://meadow/who/2022-09-30/ghe:
+  - walden://who/2022-09-30/ghe
+  data://garden/who/2022-09-30/ghe:
+  - data://meadow/who/2022-09-30/ghe
+  data://grapher/who/2022-09-30/ghe:
+  - data://garden/who/2022-09-30/ghe
```
pabloarosado, line 21: Here I suppose the `data://garden/reference/`dataset should be included as a dependency of the garden step, since the country_code_to_country function uses it.

pabloarosado (approved):

PR #544, opened 2022-10-03
feat(internet): add continent data

lucasrodes:
Adds continent data to the dataset.

some context: https://github.com/owid/owid-issues/issues/551#issuecomment-1265773009

### DIFF
--- a/etl/data_helpers.py
+++ b/etl/data_helpers.py
@@ -6,6 +6,8 @@ from owid.catalog import Dataset
 
 from etl.paths import REFERENCE_DATASET
 
+REGIONS = ["World", "Asia", "Africa", "North America", "South America", "Europe", "Oceania"]
+
 
 def calculate_region_sums(df: pd.DataFrame) -> pd.DataFrame:
     """
@@ -16,19 +18,7 @@ def calculate_region_sums(df: pd.DataFrame) -> pd.DataFrame:
     assert {"year", "country"} < set(df.columns)
     df = cast(
         pd.DataFrame,
-        df[
-            -df.country.isin(
-                [
-                    "North America",
-                    "South America",
-                    "Europe",
-                    "Africa",
-                    "Asia",
-                    "Oceania",
-                    "World",
-                ]
-            )
-        ],
+        df[-df.country.isin(REGIONS)],
     )
 
     countries = Dataset(REFERENCE_DATASET)["countries_regions"]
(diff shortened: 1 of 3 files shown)


etl/data_helpers.py
```diff
@@ -6,6 +6,8 @@
 
 from etl.paths import REFERENCE_DATASET
 
+REGIONS = ["World", "Asia", "Africa", "North America", "South America", "Europe", "Oceania"]
+
 
 def calculate_region_sums(df: pd.DataFrame) -> pd.DataFrame:
```
pabloarosado, line 12: This function has a few issues:
1. Firstly, even if you specify a list of `REGIONS`, in the end it's going to include in the output dataframe as many regions as there are in the `countries_regions` table of the `reference` dataset. For example, even if `High-income countries` is not specified in `REGIONS` it will be added to your data.
2. I don't think it's a good idea to create the `World` aggregate. The reason why we create our own aggregates is because there may be different definitions of continents. But there is only one definition of the World. So, unless your dataset does not have data for `World` (which is unusual) I'd avoid creating our own aggregate.
We should remove this function and use the one in datautils (which also needs to be improved). Up to you if you want to use it in this case, but I just wanted to let you know these caveats.

pabloarosado (approved):
All good, I just made a comment about the use of `calculate_region_sums` but you can decide if it affects your dataset or not. Additionally, if you are under the hood using the `countries_regions` dataset, I suppose you should add the `reference` dataset to the list of garden dependencies in the dag.

lucasrodes:
Hi @pabloarosado, thanks for reviewing! I just changed the code so it uses `owid.datautils.geo.add_region_aggregates` instead.

etl/steps/data/garden/technology/2022/internet/__init__.py
```diff
@@ -44,6 +69,39 @@ def make_users_table() -> Table:
...
+        "South America",
+        "Africa",
+        "Oceania",
+        "High-income countries",
+        "Low-income countries",
+        "Lower-middle-income countries",
+        "Upper-middle-income countries",
+    ]
+    for region in regions:
+        table = geo.add_region_aggregates(df=table, region=region, aggregations={"num_internet_users": sum})
```
pabloarosado, line 91: At the moment, by default, this function is very strict (and I think we should make it less strict). It checks if all relevant countries have data, and skips aggregates if it finds too many nans. The least strict settings would be:
`countries_that_must_have_data=[],
num_allowed_nans_per_year=None, frac_allowed_nans_per_year=0.99`
This will not care if important countries (in terms of population) have no data, and it imposes that at least 1% of the data for a given year must be not-nan.

pabloarosado (approved):

PR #553, opened 2022-10-04
fix(dag): reference to reference dataset

lucasrodes:
I just added a reference to `data://garden/reference` for those garden datasets that may require it.

@pabloarosado realised that whenever we use `owid.datautils.geo.add_region_aggregates` function, a link to the reference dataset should be added to the corresponding ETL step. This is because this function makes use of its table `countries_regions`.


pabloarosado (approved):
Thank you!

PR #559, opened 2022-10-06
feature(sync_backport): command for syncing backported variables with MySQL

Marigold:
Part of https://github.com/owid/owid-grapher/issues/1669

After backporting dataset and publishing to our catalog, update `variables.catalogPath` and `variables.shortName` in MySQL. With these fields grapher can fetch data from catalog instead of `data_values` table.

There's an argument for removing data from `data_values` table, but I guess we shouldn't turn it on yet.

## Effect

Running it on staging DB (= [all variables used in charts](https://github.com/owid/owid-grapher/blob/master/db/exportChartData.ts)) would reduce number of rows in `data_values` from 55M to 9M. Those 9M values are **private** datasets that we don't export (around 7M out of those 9M are COVID datasets).

In production we have much larger `data_values` that also contain variables not used in any charts. The problem is that **we don't currently backport them** and therefore have no way how to get them out of data_values. Should we start backporting them?

## Private datasets

Either we keep private datasets in `data_values` or we start backporting them and use S3 credentials when accessing them (e.g. during baking). Would this work for COVID which is being updated almost in real-time? (COVID makes majority of the remaining data)

## Issues with renaming

If someone renames a variable in admin, its `shortName` and `catalogPath` will still use the old path. Even though names will be out of sync, it will point to the correct variable (with the same variable id). However, after we backport the renamed dataset / variable, we'll overwrite it in the catalog and `catalogPath` will point to the non-existing data until the next run of `sync_backport`. This gap will happen between `publish` and `sync_backport` and could be small if we run them right after each other.

This would be easier if datasets went through ETL instead of fasttrack or if we used variable ids and dataset ids instead of short names from backport.

## TODO

* [ ] once merged, run against production and make sure we don't have any out-of-sync variables
* [ ] make variables.shortName `varchar(750)`
* [ ] stop upserting data from ETL grapher step
* [ ] remove all ETL upserted data from `data_values` that have valid `catalogPath`


Marigold:
This depends on parquet support which has been parked. Reopening in case we move forward with parquet.

larsyencken:
Just a reminder that stale bot will keep archiving things unless you tag them as `pinned`.

larsyencken:
Converted this to a draft, since it's not currently pending review, and it's blocked on Parquet support in Grapher

PR #563, opened 2022-10-10
fix(key-indicators): fix field name

lucasrodes:
Fixes bug due to typo introduced in #560


larsyencken (approved):
Thanks for the fix Lucas! I'll merge it now so that the ETL builds cleanly again.

lucasrodes:
thanks Lars! had to rush out and couldn't connect.

PR #564, opened 2022-10-11
Update world carbon pricing

pabloarosado:
* Duplicated meadow/garden/grapher steps to have a new version.
* Change the `version` field where appropriate.
* Removed a patch now that an issue with the source data has been fixed.

@Marigold I'm not able to run the grapher step on my local grapher, because MySQL claims it's a duplicate:
```
MySQLdb.IntegrityError: (1062, "Duplicate entry 'World carbon pricing for any sector-rff' for key 'datasets.datasets_name_namespace_d3d60d22_uniq'")
```
In fact, it seems to be replacing the previous version in grapher, which is not the expected behaviour. Is this a bug or have I made a mistake somewhere? Thank you very much!


Marigold:
> In fact, it seems to be replacing the previous version in grapher, which is not the expected behaviour. Is this a bug or have I made a mistake somewhere? Thank you very much!

It's not your fault. The problem is that `namespace-shortName-version` hasn't been adopted by admin yet and it still uses only `namespace-name`. There's a unique constraint on `namespace-name` which is the error you're seeing. We could get rid of it, but the problem is that version is not displayed in https://staging.owid.cloud/admin/datasets yet so you'd end up with two datasets with the same name.

For now I'd just rename it to `World carbon pricing for any sector (2022-10-11)` and the old one to `World carbon pricing for any sector (2022-09-14)`. Rerunning the old dataset will rename it, but otherwise keep it as it is. You should be able to upsert the second one without any problems.

Marigold (approved):

Marigold:
(Make sure to merge walden part and get buildkite green before merging)

pabloarosado:
Thanks a lot @Marigold ! I changed the name of the new grapher datasets and the problem was solved.

Issue #552, opened 2022-10-04
walden doesn't invalidate local cache


Marigold:
## Problem

If we upload a corrected walden dataset to S3, local cache in `~/.owid/walden` won't be invalidated. This happened for `un/2022-07-11/un_wpp` where we uploaded a corrected version with more data, but our live server was still using old `~/.owid/walden/un/2022-07-11/un_wpp.zip` (and we ended up with missing data in meadow step even though it worked locally).

## Solution

1. Compare walden file checksum against index in [WaldenStep.is_dirty](https://github.com/owid/etl/blob/master/etl/steps/__init__.py#L447) (computing checksums could have performance implications)
2. Use [checksums in S3 walden paths](https://github.com/owid/etl/pull/513)
3. Use checksums in local walden cache `~/.owid/walden`

(I've manually refreshed `~/.owid/walden/un/2022-07-11/un_wpp.zip` to make it work. This bug didn't affect production since grapher upload was run locally)

PR #565, opened 2022-10-12
fix(WaldenStep): compare checksums in is_dirty

Marigold:
Fix for https://github.com/owid/etl/issues/552

[PR in walden](https://github.com/owid/walden/pull/91)

`WaldenStep` should also compare checksums before using file from cache. However, this is computationally expensive so we use runtime cache with filename as a key. (Using mtime of files plus filename would be also possible, but is it necessary over one run of ETL?)

### DIFF
--- a/etl/files.py
+++ b/etl/files.py
@@ -4,14 +4,16 @@
 
 import hashlib
 from pathlib import Path
-from typing import List, Set, Union
+from threading import Lock
+from typing import Dict, List, Set, Union
+
+# runtime cache, we need locks because we usually run it in threads
+cache_md5: Dict[str, str] = {}
+cache_md5_locks: Dict[str, Lock] = {}
 
 
-def checksum_file(filename: Union[str, Path]) -> str:
-    "Return the md5 hex digest of the file contents."
-    if isinstance(filename, Path):
-        filename = filename.as_posix()
-
+def checksum_file_nocache(filename: Union[str, Path]) -> str:
+    "Return the md5 hex digest of the file without using cache."
     chunk_size = 2**20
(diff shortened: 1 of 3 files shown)


etl/files.py
```diff
@@ -4,23 +4,37 @@
...
+        with cache_md5_locks[filename]:
+            chunk_size = 2**20
+            _hash = hashlib.md5()
+            with open(filename, "rb") as istream:
+                chunk = istream.read(chunk_size)
+                while chunk:
+                    _hash.update(chunk)
+                    chunk = istream.read(chunk_size)
+
+            md5 = _hash.hexdigest()
```
larsyencken, line 33: Move this into a single higher level method `checksum_file_nocache()` that you can call here.

larsyencken (commented):
The approach looks correct. In terms of the caching strategy, it seems we commit to doing one md5 calculation for every file in walden per full ETL run. Are we actually doing it multiple times at the moment?

If we are, then it would mean that WaldenStep doesn't cache the `is_dirty()` check, or else that we have multiple copies of each WaldenStep in memory at one time. Do you know which it is?

Marigold:
> Are we actually doing it multiple times at the moment?
> If we are, then it would mean that WaldenStep doesn't cache the is_dirty() check

We're not caching `is_dirty` for any steps, so md5 can be recalculated several times during one ETL run. If we have DAG `WaldenStep -> DataStep` and run ETL on it, we run `is_dirty` once for `WaldenStep` and then when calculating dirtiness of `DataStep` dependencies.

This wasn't an issue before because we were only checking whether walden path exists.

Did it answer the question?

PR #566, opened 2022-10-12
Limit RFF data until 2020, since after that the data is inc…

pabloarosado:
Limit RFF data until 2020, since after that the data is incomplete.


spoonerf (approved):
Looks good to me!

PR #567, opened 2022-10-13
Adding IHME child mortality data to grapher

spoonerf:


etl/steps/grapher/dummy/2020/dummy.py
```diff
@@ -0,0 +1,19 @@
+from owid import catalog
```
pabloarosado, line 1: Was this file accidentally added?

pabloarosado (approved):

etl/steps/grapher/dummy/2020/dummy.py
spoonerf: It was!

spoonerf: Have deleted it now @pabloarosado

PR #569, opened 2022-10-14
Fixing decimal places in the WHO GHE csv download

spoonerf:


spoonerf:
Hi @larsyencken - I was wondering if you could help me out with this if you have time. I'm pretty sure I am rounding all of the variables in this dataset to either 0 or 2 d.p. and am reinforcing this in the display sections of the meta.yml. However, when I download the csv of a variable (e.g. [this one](https://owid.cloud/admin/variables/527011)), it is giving many, many more decimal places!

I'm not sure if I'm doing something wrong?

Marigold:
@spoonerf taking a look, I don't think it's your fault. The reason it's doing that is likely because repacking lowers precision for floats and it somehow gets messed up on grapher upserts.

spoonerf:
Hi @Marigold - think this is okay to merge now but thought I should pass it by you first!

Marigold (approved):
Looks good

PR #575, opened 2022-10-18
Add aviation data

pabloarosado:
Add meadow, garden, and grapher steps for data on aviation accidents and fatalities.


etl/steps/data/garden/aviation_safety_network/2022-10-14/aviation_statistics.meta.yml
```diff
@@ -0,0 +1,73 @@
...
+      publication_year: 2022
+      date_accessed: 2022-05-26
+
+tables:
+  aviation_statistics:
+    variables:
+      accidents_per_million_flights:
+        title: Fatal accidents per million commercial flights
+        short_unit: 
+        unit: accidents
```
spoonerf, line 33: if the variable is a rate I tend to include this in the unit - so I would have 'accidents per million flights'. I think this is more informative as a hover on the map.

spoonerf: Same for other rate units below.

spoonerf (approved):

etl/steps/data/garden/aviation_safety_network/2022-10-14/aviation_statistics.meta.yml
pabloarosado: Thanks. [In this case](https://ourworldindata.org/grapher/air-passengers-per-fatality), there is only global data, for a line chart, and it looks a bit busy with something like "accidents per million flights". I'll see what's best.

PR #582, opened 2022-10-24
fix(mypy): fix mypy error when importing shared.py module

Marigold:
Fix for error `etl/steps/data/meadow/irena/2022-10-20/renewable_power_generation_costs.py:15: error: Module "shared" has no attribute "CURRENT_DIR"`


pabloarosado (approved):
Thanks!

PR #583, opened 2022-10-24
WIP: feature(publish): add option to publish to R2

Marigold:
Add option to publish to Cloudflare's R2 with `publish -c grapher --private --r2`.

R2 promises free bandwidth fees (which are minimal for us on DigitalOcean Spaces though). The main problem with R2 is that **it's not any faster than Spaces S3** (apparently they're not globally distributing content yet) and we're getting 500ms+ latency on all requests (from our staging server in London). Currently it doesn't do anything extra compared to our current S3 + Cloudflare caching solution. The idea was to try this for range requests, but its latency is a dealbreaker.


Marigold:
All is R2 now.

PR #587, opened 2022-10-25
data(wb_gender): add wb gender statistics to meadow

lucasrodes:
Adding Meadow dataset for World Bank Gender Statistics.

### DIFF
--- a/etl/steps/data/garden/rff/2022-10-11/world_carbon_pricing.py
+++ b/etl/steps/data/garden/rff/2022-10-11/world_carbon_pricing.py
@@ -270,7 +270,7 @@ def run(dest_dir: str) -> None:
     df_subnational = pd.DataFrame(tb_meadow_subnational)
 
     # Load dictionary mapping sub-national jurisdictions to their countries.
-    country_members = io.local.load_json(COUNTRY_MEMBERS_FILE)
+    country_members = io.load_json(COUNTRY_MEMBERS_FILE)
 
     #
     # Process data.
--- a/etl/steps/data/garden/un/2022-07-11/un_wpp/deaths.py
+++ b/etl/steps/data/garden/un/2022-07-11/un_wpp/deaths.py
@@ -91,6 +91,8 @@ def add_age_groups(df: pd.DataFrame) -> pd.DataFrame:
         as_index=False,
         observed=True,
     ).sum()
+    # 100+ age group
+    df_100 = df[df.age == "100+"].copy()
     # Merge all age groups
-    df = pd.concat([df_0, df_1_4, df_5, df_10], ignore_index=True)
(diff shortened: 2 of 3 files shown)


etl/steps/data/meadow/wb/2022/wb_gender.py
```diff
@@ -0,0 +1,258 @@
...
+    metadata_filename = "Gender_StatsSeries.csv"
+
+    def __init__(self) -> None:
+        log.info("Loading walden dataset...")
+        self.walden_ds = walden.Catalog().find_one("wb", "2022", "wb_gender")
+
+    @property
+    def columns_relevant(self) -> List[str]:
+        """Get list of relevant columns from data file."""
+        return list(self.columns_data_rename.keys()) + [str(i) for i in range(1960, 2022)]
```
pabloarosado, line 34: I'd recommend defining these kinds of parameters as global variables at the top, to be aware that they need to be changed by the next update.

etl/steps/data/meadow/wb/2022/wb_gender.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @property
+    def coluns_not_relevant(self) -> List[str]:
```
pabloarosado, line 37: Typo.

etl/steps/data/meadow/wb/2022/wb_gender.py
```diff
@@ -0,0 +1,258 @@
...
+    for t in tables:
+        table = Table(t[0])
+        table.metadata.short_name = t[1]
+        ds.add(utils.underscore_table(table))
+    return ds
+
+
+def run(dest_dir: str) -> None:
+    """Run pipeline."""
+    WBGenderMeadowStep().run(dest_dir)
```
pabloarosado, line 258: I'd personally prefer to have the `run` function in a data step as explicit as possible. So, if anyone wants to quickly have an overview of what's done to the data, they should be able to check the content of `run` and get a sense of it. So, at the very least, I'd recommend moving the content of the `run` method of `WBGenderMeadowStep` into the global `run` function. This way you can still see:
```
# Load data and metadata
df, metadata = self.load_data_from_walden()
...
```
Feel free to ignore my suggestion if you disagree.

pabloarosado (approved):

etl/steps/data/meadow/wb/2022/wb_gender.py
lucasrodes: makes sense, thanks!

etl/steps/data/meadow/wb/2022/wb_gender.py
lucasrodes: makes sense!

lucasrodes:
thanks for your review, @pabloarosado!

PR #588, opened 2022-10-26
enhance(walkthrough): various improvements

Marigold:
Implemented some ideas from https://github.com/owid/etl/issues/294

* CLI `etl-metadata-export` that creates a `*.meta.yml` file from a dataset, this is recommended in walkthrough, but could be used outside it too (@pabloarosado is this helpful? Any ideas how to make it more useful?)
* Add `version` field to walden
* Rearrange fields to be in order `namespace-version-short name`
* Ingestion scripts generated in `institution/version` folder

Does it make sense to generate `*.meta.yml` for meadow or should we only encourage to do it on garden level?

### DIFF
--- a/etl/metadata_export.py
+++ b/etl/metadata_export.py
@@ -1,0 +1,104 @@
+import os
+from typing import Any, Dict, cast
+
+import click
+import yaml
+from owid.catalog import Dataset
+
+
+@click.command(help=__doc__)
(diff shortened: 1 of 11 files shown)


pabloarosado (approved):
Thanks a lot! It looks good. I will need to try it out a few times to see how or whether it can be improved.

PR #590, opened 2022-10-26
Add irena patents data

pabloarosado:
Add meadow, garden and grapher steps for IRENA data on patents in renewable energy technologies.


etl/steps/data/garden/irena/2022-10-25/renewable_energy_patents.meta.yml
```diff
@@ -0,0 +1,159 @@
...
+      url: https://www.irena.org/Statistics/View-Data-by-Topic/Innovation-and-Technology/Patents-Evolution
+      date_accessed: 2022-10-25
+
+tables:
+  renewable_energy_patents:
+    variables:
+      patents:
+        title: Patents
+        short_unit: patents
+        unit: patent
```
spoonerf, line 21: I would leave the `short_unit` blank as otherwise it shows up on the map legend.

etl/steps/data/garden/irena/2022-10-25/renewable_energy_patents.meta.yml
```diff
@@ -0,0 +1,159 @@
...
+        description: Number of patents.
```
spoonerf, line 22: I would add the number of decimal of places option and set it as 0 as the default in grapher is 2.

spoonerf (approved):
Looks good to me, just a couple of metadata issues.

etl/steps/data/garden/irena/2022-10-25/renewable_energy_patents.meta.yml
pabloarosado: Thanks, good suggestion!

PR #593, opened 2022-10-27
enhance(wb_gender)

lucasrodes:
Modified code, so it is functional programming. This way, the data flow might be easier to follow.


etl/steps/data/meadow/wb/2022/wb_gender.py
```diff
@@ -32,172 +32,175 @@
...
+        - year: Year (starting from 1960 til 2021).
+        - value: Value for the variable in the given location and year.
+
+    Also runs checks on the data to see that it is consistent with the metadata.
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Data dataframe
+    df : pd.DataFrame
```
pabloarosado, line 100: It should be metadata.

etl/steps/data/meadow/wb/2022/wb_gender.py
```diff
@@ -32,172 +32,175 @@
...
+
+
+def format_metadata(df: pd.DataFrame) -> pd.DataFrame:
+    """Format metadata."""
+    # Drop columns
+    log.info("Reshaping dataframe: Preserving only relevant columns in metadata...")
+    df = df.drop(columns=df.filter(regex="Unnamed").columns)
+    # Rename columns
+    log.info("Reshaping dataframe: Renaming columns in metadata...")
+    df.columns = [m.replace(" ", "_").lower() for m in df.columns]
```
pabloarosado, line 132: Just in case you don't know, there is a `catalog.utils.underscore` function to do this. There is also a function `catalog.utils.underscore_table` but (annoyingly) it only works on tables (we should create one that works on dataframes).

etl/steps/data/garden/wb/2022/wb_gender/wb_gender.countries.json
```diff
@@ -0,0 +1,225 @@
...
+  "Namibia": "Namibia",
+  "Nauru": "Nauru",
+  "Nepal": "Nepal",
+  "Netherlands": "Netherlands",
+  "New Caledonia": "New Caledonia",
+  "New Zealand": "New Zealand",
+  "Nicaragua": "Nicaragua",
+  "Niger": "Niger",
+  "Nigeria": "Nigeria",
+  "North America": "North America",
```
pabloarosado, line 147: World Bank's definition of North America is different from OWID's definition. I suppose the mapping should be `"North America": "North America (WB)"`. The same may apply to other regions (although not to income groups).

pabloarosado (approved):

etl/steps/data/meadow/wb/2022/wb_gender.py
lucasrodes: thanks

PR #594, opened 2022-10-28
fix(structure): Remove grapher symlink

Marigold:
Grapher steps have been rewritten as a normal data steps in `grapher` channel, but we've kept them in the original folder and instead created a symlink. This PR moves steps from `etl/steps/grapher` to `etl/steps/data/grapher` and removes the symlink.

@pabloarosado is this worth it? For instance if you have any uncommited work in that folder, it might be a pain. It's just a small technical debt and we could live with it I guess.


pabloarosado (approved):
I think it's a good idea to do this, to avoid future confusion. I don't have any unmerged branches that could be affected. However, I'd suggest writing on some slack to ensure the others are also fine with these changes.

pabloarosado:
I think [this FAOSTAT script](https://github.com/owid/etl/blob/master/etl/scripts/faostat/create_new_steps.py) may also need to be checked. It has some logic on how to create new data/grapher steps. Do you want to check it out? Otherwise please create an issue so that we remember about this and I'll do it. Thanks.
Could there be other scripts affected by this?

Marigold:
@pabloarosado I'll check those scripts myself and merge this on Monday if everything looks good. Don't want to end up with dysfunctional pipeline on Friday...

lucasrodes:
I implemented a new grapher step in this PR: https://github.com/owid/etl/pull/595. You might want to merge mine first. I am happy to merge mine with master later on otherwise.

Marigold:
@lucasrodes thanks for heads up, ping me when you merge your PR and I'll rebase this one.

lucasrodes:
@Marigold just merged https://github.com/owid/etl/pull/595

PR #595, opened 2022-10-28
feature(wb_gender): add garden and grapher steps

lucasrodes:
## Done

**Meadow**
- Changed versioning from year to date
- Propagate country metadata
- docstring fixes

**Garden**
- Implemented step

**Grapher**
- Implemented step


etl/steps/data/garden/wb/2022-10-29/wb_gender/wb_gender.countries.json
```diff
@@ -219,7 +218,16 @@
...
+  "Upper middle income": "Upper-middle-income countries",
+  "North America": "North America (WB)",
+  "Latin America & Caribbean": "Latin America and Caribbean (WB)",
+  "Europe & Central Asia": "Europe and Central Asia (WB)",
+  "East Asia & Pacific": "East Asia and Pacific (WB)",
+  "South Asia": "South Asia (WB)",
+  "Middle East & North Africa": "Middle East and North Africa (WB)",
+  "Sub-Saharan Africa": "Sub-Saharan Africa (WB)",
+  "European Union": "European Union",
+  "OECD members": "OECD members"
```
pabloarosado, line 232: Shouldn't this be `OECD members (WB)`, given that it's not an official OWID region?

etl/steps/data/meadow/wb/2022-10-29/wb_gender/__init__.py
```diff
@@ -236,13 +262,15 @@ def add_tables_to_ds(ds: Dataset, df: pd.DataFrame, metadata: pd.DataFrame) -> D
 def run(dest_dir: str) -> None:
     """Run pipeline."""
     # Load data and metadata
-    walden_ds = walden.Catalog().find_one("wb", "2022", "wb_gender")
-    df, metadata = load_data_from_walden(walden_ds)
-    # Format metadata
-    metadata = metadata.pipe(format_metadata)
+    walden_ds = walden.Catalog().find_one("wb", "2022-10-29", "wb_gender")
```
pabloarosado, line 265: I'd define the version at the top of the file, to be clear that this is a parameter that will need to change on the next update.

etl/steps/grapher/wb/2022-10-29/wb_gender/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+log = get_logger()
+
+DATASET_GARDEN = DATA_DIR / "garden/wb/2022-10-29/wb_gender"
+
+
+def process_table(table: catalog.Table) -> catalog.Table:
+    log.info("Processing table...")
+    # Reset index
+    table = table.reset_index()
+    # Harmonize country names
```
pabloarosado, line 19: I suppose this comment is misplaced. You could actually do everything in one line: `table = table.reset_index().set_index(["country", "year"])`.

pabloarosado (approved):
All good, I just made some minor comments and suggestions. One thing I'm thinking is: wouldn't it make sense to create the long table and the wide table (both with their appropriate metadata) in the garden step? (And then the grapher step would simply load the wide table). Otherwise, it seems that you create a table of metadata in garden that is not really used, but it's necessary in grapher to create the wide table (if I understood the logic properly). You don't need to make these changes, if it's too much hassle.

etl/steps/data/garden/wb/2022-10-29/wb_gender/wb_gender.countries.json
```diff
@@ -219,7 +218,16 @@
...
-  "Upper middle income": "Upper-middle-income countries"
+  "Upper middle income": "Upper-middle-income countries",
+  "North America": "North America (WB)",
+  "Latin America & Caribbean": "Latin America and Caribbean (WB)",
+  "Europe & Central Asia": "Europe and Central Asia (WB)",
+  "East Asia & Pacific": "East Asia and Pacific (WB)",
+  "South Asia": "South Asia (WB)",
+  "Middle East & North Africa": "Middle East and North Africa (WB)",
+  "Sub-Saharan Africa": "Sub-Saharan Africa (WB)",
+  "European Union": "European Union",
```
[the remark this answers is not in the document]
lucasrodes: It probably makes sense `European Union (WB)`, as it is unclear what old data refers to ("European Union (27)?", "European Union (26)?", etc.) .

etl/steps/data/garden/wb/2022-10-29/wb_gender/wb_gender.countries.json
lucasrodes: Same as in European Union. Unclear if the same entity definition propagates back in time.

etl/steps/data/garden/wb/2022-10-29/wb_gender/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+    # extract units
+    unit_1 = table_metadata["indicator_name"].str.extract(r"\((% [^\)]*)\)")  # (%...)
+    unit_12 = table_metadata["indicator_name"].str.extract(r"\((%), tertiary\)")  # (%, tertiary)
+    unit_2 = table_metadata["indicator_name"].str.extract(r"\(([^\)]*\$)\)")  # (...$)
+    unit_3 = table_metadata["indicator_name"].str.extract(r"\(([^\)]*\%)\)")  # (...%)
+    unit_4 = table_metadata["indicator_name"].str.extract(r"[Ee]xpected ([Yy]ears)")  # Expected years...
+    unit_4[0] = unit_4[0].apply(lambda x: x.lower() if not pd.isnull(x) else np.nan)
+    unit_5 = table_metadata["indicator_name"].str.extract(r"\((per[^%)]*)\)")  # (per...)
+    unit_6 = table_metadata["indicator_name"].str.extract(r"\((?:calendar )?(days)\)")  # (days)
+    unit_7 = table_metadata["indicator_name"].str.extract(r"\((liters)[^\)]*\)")  # (liters...)
```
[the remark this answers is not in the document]
lucasrodes: done, thanks

lucasrodes:
thanks for your review, @pabloarosado! I added most of the changes that you proposed.

> wouldn't it make sense to create the long table and the wide table (both with their appropriate metadata) in the garden step? 

I see the point, but I don't think is a good idea in this case. Pivoting the long table into wide format would lead to ~1000 columns and a _very_ sparse matrix. While I can understand your reasoning, I don't think having the wide format here is necessary: It adds redundancy, and the table would be quite "ugly" (sparse and super wide).

> Otherwise, it seems that you create a table of metadata in Garden that is not used, but it's necessary in Grapher to create the wide table (if I understood the logic properly). You don't need to make these changes if it's too much hassle.

Yes, that's the way I thought it was best to propagate the metadata so Grapher can use it to build variable metadata. Ideally, I think this metadata should live as `DatasetMeta`, `TableMeta` and `VariableMeta` in Garden, but the current ETL structure does not allow this in long tables. A long table can only have variable metadata for its columns (and not for values that a certain column takes, as is the case in long table format).

While not ideal, I still think that wide format here is not great.

PR #596, opened 2022-10-31
fix(dag): replace non-existing wb/2022/wb_gender version by wb/2022-10-29/wb_gender

Marigold:
Hotfix for non-existing `walden://wb/2022/wb_gender`. @lucasrodes do you remember what happened here? Did you manually remove the old 2022 version from walden or was it something else that could have done it?

### DIFF
--- a/dag.yml
+++ b/dag.yml
@@ -48,7 +48,7 @@ steps:
   data://meadow/wb/2021-07-01/wb_income:
     - walden://wb/2021-07-01/wb_income
   data://meadow/wb/2022/wb_gender:
-    - walden://wb/2022/wb_gender
+    - walden://wb/2022-10-29/wb_gender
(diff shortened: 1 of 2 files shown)


pabloarosado (commented):
I think this PR is not necessary, since this is taken care of in [this other PR](https://github.com/owid/etl/pull/595). If @lucasrodes thinks that his PR may take a bit longer to be merged, the current one could be merged first to fix the current issue with walden.

larsyencken:
In this case, I disagree. We should commit to having a continuous, clean ETL run, and to fixing it within the hour if the build is failing. Otherwise, everyone else's work will get polluted with a spurious CI and confusing CI failure.

larsyencken (approved):
I double checked that `wb_gender` builds cleanly after this change, ship it :shipit:

PR #597, opened 2022-10-31
fix(sdg): update SDG dependencies

Marigold:
SDG tracker charts had their dependencies updated. This PR refreshes those dependencies.


pabloarosado (approved):

larsyencken (commented):
It still seems to fail the build in CI with:

```
KeyError: 'no match for dataset None/None/dataset_5821_gender_statistics__world_bank__2022_10_29_values'
```

It's also unclear to how the build started failing here. Did we archive a dataset in Grapher that we were previously relying on?

Marigold:
> It's also unclear to how the build started failing here. Did we archive a dataset in Grapher that we were previously relying on?

I forgot to commit one index file to walden. Fixed & merging.

PR #599, opened 2022-10-31
feature(pyright): replace mypy with pyright

Marigold:
Replace `mypy` type checker with `pyright` and fix all type problems that popped up. Pyright runs in 12s (single core) while mypy runs 150s without cache / ~60s with cache (using multiple processes). Unlike mypy, pyright can handle modules with the same names (`shared.py`) and doesn't need `typecheck_steps.sh` script that checks all the steps individually in parallel. It's also compatible with VSCode. Mypy has also recently started raising non-determinstic errors (likely caused by running it in parallel).

Pyright uses `node`, but installing it with `pip install pyright` automagically installs node in virtualenv if you don't have it globally, so we don't have to update our ops scripts or write special instructions.


larsyencken (approved):
From my perspective, the use of a PyPI package to install `pyright` removes nearly all of the objections. I'm happy for us to move forward, knowing that we can always change it if it gives us problems in the future.

larsyencken:
I moved the Todos into the parent issue, and unlinked this one from the parent so that the parent will remain open when this is closed.

larsyencken:
I checked now that on a fresh Linux install without node, it successfully installs node and does not require any new Ubuntu packages 👍

PR #602, opened 2022-11-01
Adding postnatal care coverage variable from World Bank Health Statistics

spoonerf:
This is only a single variable, it is not part of the World Development Indicators so we have had to import it separately. It would be useful to have this for the health explorer.


spoonerf:
HI @Marigold - the buildkite test seems to be failing on an odd point. It says I don't have an `.aws/config` file set up, I do locally but I'm not sure if there is a problem elsewhere?

Marigold:
> HI @Marigold - the buildkite test seems to be failing on an odd point. It says I don't have an .aws/config file set up, I do locally but I'm not sure if there is a problem elsewhere?

Walden got uploaded as [private dataset](https://github.com/owid/walden/blob/master/owid/walden/index/postnatal_care/2022-09-19/postnatal_care.json#L12). Have you run the script again after changing the `public` argument and **committed it to walden**? (Ping me if you did and it's not working)

spoonerf:
@Marigold Ah damn, I changed it on DigitalOcean but not in the json, thanks for the spot! Fixing here https://github.com/owid/walden/pull/121

Marigold (approved):
Looks good after CI passes.

PR #603, opened 2022-11-01
enhance(docs): improve quickstart docs and fix some WSL bugs

Marigold:


paarriagadap:
Looks good to me. I would include two extra things in the docs:
This step for WSL users: include/edit` /etc/wsl.conf` and add
```
[interop]
appendWindowsPath = false
```
I used `sudo vi /etc/wsl.conf` for that, copied and typed `:qw`

The other thing was updating this
`sudo apt install default-libmysqlclient-dev build-essential`, because I needed them even after running `sudo apt install python3.9-dev mysql-client`

paarriagadap (approved):

paarriagadap (approved):

PR #609, opened 2022-11-03
data(hmd): life tables

lucasrodes:
Adding meadow step for "Life Tables" by Human Mortality Database.

The source data provides a zip which contains 6 folders. Each folder contains a TXT file per country (not great). This step generates a dataset with 6 tables, one for each folder. Each table contains the data from all TXT files.


Marigold:
(Just a quick tip if you don't want to go nuts from walden conflicts - instead of adding files using e.g. `git add -A :/` add them with `git add -- . ':!vendor'`)

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
+import os
```
pabloarosado, line 1: The name of the step is not very informative, what about `life_tables` instead? I'd also suggest having a small docstring for the file (I think that's at the moment the only source of generic documentation for data steps).

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+    ds.metadata = convert_walden_metadata(walden_ds)
+    ds.metadata.version = VERSION_MEADOW
+    return ds
+
+
+def create_and_add_tables_to_dataset(local_file: str, ds: Dataset, walden_ds: WaldenCatalog) -> Dataset:
+    """Create and add tables to dataset.
+
+    This method creates tables for all the folders found once `local_file` is uncompressed. Then,
+    it cleans and ads them to the dataset `ds`. It uses the metadata from `walden_ds` to create the tables.
```
pabloarosado, line 71: Typo: "ads".

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+
+def init_meadow_dataset(dest_dir: str, walden_ds: WaldenCatalog) -> Dataset:
+    """Initialize meadow dataset."""
+    ds = Dataset.create_empty(dest_dir)
+    ds.metadata = convert_walden_metadata(walden_ds)
+    ds.metadata.version = VERSION_MEADOW
+    return ds
+
+
+def create_and_add_tables_to_dataset(local_file: str, ds: Dataset, walden_ds: WaldenCatalog) -> Dataset:
```
pabloarosado, line 67: I don't think that's the right type for `walden_ds`, it should be `owid.walden.catalog.Dataset`. Maybe you'd need to import `from owid.walden.catalog import Dataset as WaldenDataset` and then the type would be `WaldenDataset`. If this is correct, then change the docstring too.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        # Decompress files
+        decompress_file(local_file, tmp_dir)
+
+        # Check available files
+        _sanity_check_files(tmp_dir)
+
+        # Load data
+        for i in [1, 5]:
+            for j in [1, 5, 10]:
```
pabloarosado, line 96: Wouldn't it be better to simply loop over all folders in the decompressed folder? I'm thinking that in the next update these names and numbers may change.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+
+def _sanity_check_files(path: str) -> None:
+    """Checks that all required files are present once zip is uncompressed."""
+    files_found = sorted(os.listdir(path))
+    assert (
+        files_found == FILES_EXPECTED
+    ), f"Files found are not the ones expected! Check that {FILES_EXPECTED} are actually there!"
+
+
+def make_table(input_folder: str, table_name: str, walden_ds: WaldenCatalog) -> catalog.Table:
```
pabloarosado, line 116: Again, check the type of `walden_ds`.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+        if match is not None:
+            country, _, table_str = match.group(1, 2, 3)
+        else:
+            raise ValueError("No match found!")
+        # Build country df
+        df_ = _make_df_country(country, table_str)
+        dfs.append(df_)
+    # Concatenate all country dfs
+    df = pd.concat(dfs, ignore_index=True)
+    return df  # type: ignore
```
pabloarosado, line 184: Instead of ignoring type checks, you can do `return cast(pd.DataFrame, df)` (and add `from typing import cast`).

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+    dfs = []
+    # Load each file
+    for f in files:
+        with open(f, "r") as f:
+            text = f.read()
+        # Get relevant fields
+        match = re.search(regex, text)
+        if match is not None:
+            country, _, table_str = match.group(1, 2, 3)
+        else:
```
pabloarosado, line 177: This process seems quite sensitive to changes in the source on the next update, it would be better to list all relevant fields and ensure you fetch all of them, but if you think this procedure is robust enough, then leave it like that.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+    df = _clean_correct_missing_data(df)
+    # Set dtypes
+    df = _clean_set_dtypes_df(df)
+    return df
+
+
+def _clean_rename_columns_df(df: pd.DataFrame) -> pd.DataFrame:
+    """Rename columns."""
+    return df.rename(
+        columns={
```
pabloarosado, line 236: Column names is another thing that is likely to change in the next update. What I usually do is to define a global `COLUMNS` at the beginning of the script. And the same for dtypes.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+        "year": "category",
+    }
+    # Numeric
+    cols_numeric = [col_name for col_name, dtype in dtypes.items() if dtype in ["Int64", "Float64"]]
+    for col in cols_numeric:
+        df[col] = pd.to_numeric(df[col])
+    return df.astype(dtypes)
+
+
+def df_to_table(walden_ds: WaldenCatalog, table_name: str, df: pd.DataFrame) -> catalog.Table:
```
pabloarosado, line 298: Again, check `walden_ds` type.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
```diff
@@ -0,0 +1,322 @@
...
+        Dataframe.
+
+    Returns
+    -------
+    catalog.Table
+        Table created from dataframe, walden metadata and table name.
+    """
+    # create table with metadata from dataframe
+    table_metadata = TableMeta(
+        short_name=f"agexyear_{table_name}",
```
pabloarosado, line 317: Maybe think of a different name for the tables?

pabloarosado (approved):
Looks good. I made some suggestions, feel free to disregard them.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
lucasrodes: There is a previous step `_sanity_check_files` which ensures that all files are present. I think that should suffice.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
lucasrodes: I improved the description of the table so the user can have a better understanding of what the table contains.

etl/steps/data/meadow/hmd/2022-11-03/hmd_lt.py
lucasrodes: I have improved the error message in case this fails, so we can know what has failed.

PR #611, opened 2022-11-03
data(zijdeman2015): add life expectancy estimates

lucasrodes:
Adds data from Walden to Meadow.

Source is [Life Expectancy at Birth (Total)](https://datasets.iisg.amsterdam/dataset.xhtml?persistentId=hdl:10622/LKYT53), by Zijdeman et al.


dag_files/dag_walkthrough.yml
```diff
@@ -1,28 +1,30 @@
 # Autogenerated steps by walkthrough
```
pabloarosado, line 1: I think we should stop using `dag_walkthrough.yml` and put dag entries in more meaningful places. Maybe `dag_papers.yml` for the current one? Otherwise in the old `dag.yml`.

etl/steps/data/meadow/papers/2022-11-03/zijdeman_et_al_2015/__init__.py
```diff
@@ -0,0 +1,87 @@
+from typing import Tuple
```
pabloarosado, line 1: Why create an `__init__.py` file? You could just have a file `zijdeman_et_al_2015.py`.

pabloarosado (approved):

etl/steps/data/meadow/papers/2022-11-03/zijdeman_et_al_2015/__init__.py
lucasrodes: From my point of view, I see this as a more organized organization of folders of files. I lean towards _always_ creating a dedicated folder for the step.

This way, I know that whatever I find within a folder is for one and only one step. The other way around, I may have a mixture of files and folders in the same directory. I find that more confusing.

pabloarosado: OK, I agree. Usually it's useful to keep folders by version because all code is related in one way or another, so you can share modules. But in this case, where it's separate papers, it may not make sense. That said, it's unlikely that you'll have two papers datasets on the same day.

dag_files/dag_walkthrough.yml
lucasrodes: Done! thanks

PR #612, opened 2022-11-04
data(riley2005): meadow step

lucasrodes:
Add meadow step for Riley (2005) Life Expectancy estimates.

- Walden data comes as a PDF, extracting its data is complex and considering the dataset size (<60 rows), I manually transcribed the data from the PDF to a CSV. I ran some checks on the PDF to ensure the data is as expected.
- I would appreciate if we could double-check that the transcription makes sense. (more details at https://github.com/owid/owid-issues/issues/565#issuecomment-1302819114)
- This pipeline adds one dependency: PyPDF2, as we need to extract some text from a PDF.


etl/steps/data/meadow/papers/2022-11-04/riley_2005/riley_2005.data.csv
```diff
@@ -0,0 +1,56 @@
...
+1990,Europe,74.2
+2001,Europe,76.8
+1885,Asia,27.5
+1900,Asia,28
+1913,Asia,28.1
+1950,Asia,41.6
+1973,Asia,57.5
+1990,Asia,64.5
+2001,Asia,67.1
+1920,Africa,26.4
```
pabloarosado, line 20: Shouldn't this be 1925? Given that you are taking the average year in all other intervals, if I understand correctly.

etl/steps/data/meadow/papers/2022-11-04/riley_2005/riley_2005.text_expected.txt
```diff
@@ -0,0 +1,28 @@
+TABLE 1 Estimates of life expectancy before the beginning of the health transition, by region, and of regional and global
```
pabloarosado, line 1: I totally see that it would be hard to parse the table programmatically, and it's not worth it. But I don't understand the point of this file, if you are not extracting data from it. It would suffice to check the md5 of the file. Maybe I'm missing something.

pabloarosado (approved):
Apart from a few small things, looks good.

etl/steps/data/meadow/papers/2022-11-04/riley_2005/riley_2005.text_expected.txt
lucasrodes: That makes sense, thanks!

etl/steps/data/meadow/papers/2022-11-04/riley_2005/riley_2005.data.csv
lucasrodes: That is totally correct, thanks pablo!

lucasrodes:
Thanks, @pabloarosado !

PR #613, opened 2022-11-04
fix(ember): Update links to Ember Global Electricity Review

pabloarosado:
Ember has changed their URL's, so this PR updates all mentions of their Global Electricity Review.

I haven't changed Walden, I suppose it makes sense to keep the old snapshot as it was at the time, and on the next release I will adapt the ingest script and metadata.


lucasrodes (approved):
LGTM

PR #614, opened 2022-11-04
fix(walkthrough): missing channel

Marigold:


lucasrodes (approved):
LGTM! [The reported bug](https://owid.slack.com/archives/C46U9LXRR/p1667501198390589) is no longer appearing on my end. Thanks, @Marigold

PR #615, opened 2022-11-04
fix(steps): proper private dependencies for grapher

Marigold:
Enable pushing to grapher from private datasets. This supports DAGs such as

```
grapher://grapher/gcp/2022-11-11/global_carbon_budget
  - data-private://grapher/gcp/2022-11-11/global_carbon_budget
```

that are run with `etl ... --private --grapher`. It shouldn't raise validation error any more. `grapher-private://` step got also removed as it didn't make any sense (grapher database is always _private_).

@pabloarosado can you merge it yourself if you approve it?


pabloarosado (approved):
Thanks, it looks good!

PR #620, opened 2022-11-11
Update owid co2 dataset

pabloarosado:
Update OWID CO2 dataset to include the latest Global Carbon Budget release, and add land-use change emissions variables to the OWID CO2 dataset.

NOTE: The previous version of the OWID CO2 dataset will be removed from the dag, given that it depends on a backported dataset (and the dataset will be archived soon to avoid confusion). However, the code of that outdated data step can safely stay in ETL (without being executed). @Marigold please let me know if you think this could cause any issues, thanks.

Before merging this PR, the following steps need to be completed:
* [x] Merge https://github.com/owid/walden/pull/127
* [x] Merge https://github.com/owid/etl/pull/619


Marigold (approved):
> However, the code of that outdated data step can safely stay in ETL (without being executed). @Marigold please let me know if you think this could cause any issues, thanks.

It's fine.

One less backport in dependencies, way to go!

Issue #618, opened 2022-11-11
Tracking issue: private datasets


pabloarosado:
## Overview

This issue tracks all reported problems when working with private datasets.

## Specific issues

_Ideally we convert them to individual issues, and close the ones we can't confirm._

- [ ] #678
- [ ] `data-private://grapher` does not seem to work
- [ ] `etl --private` writes to the Grapher DB even when `--grapher` is not specified
- [ ] `etl --private --grapher` fails with an assertion error

## Original issue

<details>
<summary>Reported by @pabloarosado  </summary>

Working in private mode (and then changing to public) went smoothly except for a few issues (that I listed below).

In general, the only differences between working with private and public datasets are:
1. In the dag, private steps need to start with `walden-private://` or `data-private://` instead of `walden://` or `data://`.
2. When running the ETL steps, the flag `--private` needs to be included.

So, when the dataset becomes public, one simply needs to remove the instances of `-private` in the dag, and re-run the steps.

However, for the grapher step I encountered some issues (probably interconnected, I haven't had time to look into this in detail):
1. Instead of simply having `data-private://grapher` I had to use `grapher://grapher/...` (and then, when making the dataset public, I changed it to `data://grapher`). Ideally, I suppose `data-private://grapher` should work, analogous to other data steps.
2. When still working in private mode, running `etl ... --private --grapher` failed on [this line](https://github.com/owid/etl/blob/da0a30c22e0579e919415c6ca84637b50d607e58/etl/steps/__init__.py#L509), so I simply commented out the assertion `assert path == dependencies[0].path` and then it worked. This issue happens only in private mode.
4. When running `etl ... --private`, ETL writes to grapher database (even if the `--grapher` flag was not used).
5. While in private mode, I added some variables in the grapher step, but they were not showing up in the grapher admin (also, in my grapher data channel the dataset did not exist). After I made it public, the grapher data step worked as usual (and the additional variables were properly created).
6. There were some other problems with the approval tool in local grapher, as discussed on [this slack thread](https://owid.slack.com/archives/C46U9LXRR/p1668070983642789).

</details>

Marigold:
I've added [private datasets](https://github.com/owid/etl/pull/621) support in walkthrough and tested it, but weirdly I couldn't replicate any the issues above. Since they're hard to replicate, I'm not gonna dive into them, but rather wait for our next private dataset and see if they reappear (should be much easier to debug then).

pabloarosado:
[...]

PR #621, opened 2022-11-15
feature(walkthrough): support private datasets

Marigold:
Add support for private datasets in walkthrough by adding checkbox "Is private" to all steps and change instructions to basically reflect steps described by Pablo [here](https://github.com/owid/etl/issues/618).


pabloarosado (approved):
Looks good!

PR #623, opened 2022-11-22
Improve emissions data

pabloarosado:
Minor issues solved in the GCB dataset:
* Fixed wrong unit in variable metadata.
* Removed rows of spurious zeros (for example, in the original data, total emissions were zero when all individual sources of emissions were nan).
  * Then, in the grapher step, ensured that all countries have data for all years (so that stacked area charts work well).
* Included original region aggregates that were given in the raw data files, for comparison.
* Added sanity checks.

### DIFF
--- a/etl/steps/data/garden/gcp/2022-11-11/global_carbon_budget.meta.yml
+++ b/etl/steps/data/garden/gcp/2022-11-11/global_carbon_budget.meta.yml
@@ -287,8 +287,8 @@ tables:
         description: "Annual production-based emissions of carbon dioxide (CO₂), including land-use change, measured in tonnes per person. This is based on territorial emissions, which do not account for emissions embedded in traded goods."
       emissions_total_including_land_use_change_per_gdp:
         title: "Annual CO₂ emissions including land-use change per GDP"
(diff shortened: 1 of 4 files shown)


etl/steps/data/garden/gcp/2022-11-11/global_carbon_budget.meta.yml
```diff
@@ -287,8 +287,8 @@ tables:
         description: "Annual production-based emissions of carbon dioxide (CO₂), including land-use change, measured in tonnes per person. This is based on territorial emissions, which do not account for emissions embedded in traded goods."
       emissions_total_including_land_use_change_per_gdp:
         title: "Annual CO₂ emissions including land-use change per GDP"
-        unit: "tonnes"
-        short_unit: "t"
+        unit: "kilograms per $PPP"
```
bnjmacdonald, line 290: should "$PPP" be "international $" here and elsewhere, for consistency with how we refer to international dollars elsewhere on the site? e.g. https://ourworldindata.org/grapher/gdp-per-capita-worldbank (https://owid.cloud/admin/variables/446390)

PR #627, opened 2022-11-23
data(gcp): Change units of GDP to international-$

pabloarosado:
Changed "$PPP" -> "international-$", for consistency.


larsyencken:
I rebased this off `master` to get CI passing cleanly as it should

PR #633, opened 2022-11-25
Getting UN IGME data into etl

spoonerf:


spoonerf:
Hi @Marigold - I think I messed up something by editing the source name in walden and I can't figure out how to fix it :( 

The failing point here is that in the garden step: `ds_meadow = Dataset(DATA_DIR / "meadow/un/2021-12-20/un_igme")` is failing, I think because of the source name change. Any tips on how to fix would be ace!

Marigold:
@spoonerf taking a look

Marigold:
@spoonerf this is totally our fault. You're actually the first one to use License which is why it failed. Could you [review it](https://github.com/owid/owid-catalog-py/pull/70), please?

One other thing - License only has fields `name` and `url`, `version` is not available and would be thrown away (working on better validation right now so that you get a warning next time). Can you somehow put version into name or should we consider adding version to license?

spoonerf:
> @spoonerf this is totally our fault. You're actually the first one to use License which is why it failed. Could you [review it](https://github.com/owid/owid-catalog-py/pull/70), please?
> 
> One other thing - License only has fields `name` and `url`, `version` is not available and would be thrown away (working on better validation right now so that you get a warning next time). Can you somehow put version into name or should we consider adding version to license?

Ack sorry `version` doesn't need to be in license, my bad. I will put it in `dataset` instead.

etl/steps/data/garden/un/2021-12-20/un_igme.py
```diff
@@ -0,0 +1,240 @@
...
+
+    ds_garden = Dataset.create_empty(dest_dir)
+    ds_garden.metadata = ds_meadow.metadata
+
+    tb_garden = Table(dfc)
+    tb_garden.metadata = tb_meadow.metadata
+    # Flatten the indexed columns
+    tb_garden.columns = ["__".join(col).strip() for col in tb_garden.columns.values]
+    # Create one table per column and auto-populate the metadata
+    for col in tb_garden.columns:
```
Marigold, line 60: Instead of creating columns yourself, you could move some columns to dimensions and things like title will be generated automatically. Here's a sample code

```
tb_garden = Table(dfc, metadata=tb_meadow.metadata)
tb_garden.columns.names = ["metric", "sex", "indicator", "unit"]

# move sex and metric to dimensions
tb_garden = tb_garden.stack(level="metric").stack(level="sex")

# add units to variable metadata
units = tb_garden.columns.get_level_values("unit")
tb_garden.columns = tb_garden.columns.droplevel(level="unit")

for col, unit in zip(tb_garden.columns, units):
    tb_garden[col].metadata.unit = unit
```

spoonerf: Thanks - this is much nicer!

PR #634, opened 2022-11-25
WIP: Feature/snapshots separate dvc

Marigold:
@larsyencken separating dvc and yaml doesn't feel much different from having them in the same file.

What I'm not sure about is the interface:

```
snap = Snapshot("ggdc/2020-10-01/ggdc_maddison.xlsx")
# download from source, could be also fetching from API and saving dataframe to `snap.path`
snap.download_from_source()
# might change this to `snap.add`
snap.dvc_add(upload=upload)
```

I've tried many different approaches (loading SnapshotMeta first, using absolute paths to files, etc.) and this one looked the most intuitive and the most similar to `catalog.Dataset` and `catalog.DatasetMeta`. Can you think of a better one?


Marigold:
Closing this spike. Separating .dvc and metadata haven't changed much. I was able to get backporting working with previous version of Snapshots quite easily, so no need for this PR.

PR #636, opened 2022-11-28
fix: add EU-27 population

lucasrodes:


pabloarosado (commented):
The change in `key_indicators` looks good. But I don't understand why you included the `life_tables` changes in this PR, it seems to be incomplete (for example, the metadata file is the one walkthrough creates by default).

lucasrodes:
@pabloarosado ー Thanks for reviewing. I accidentally added files from another branch.

pabloarosado (approved):

PR #637, opened 2022-11-28
data(hmd): add life tables (garden)

lucasrodes:
Adds `life_tables` dataset to Garden, which will help obtain the long run variable for life expectancy at 10 years old.


pabloarosado (changes requested):
Similar to the other PR, the metadata is not complete. If this is because you are importing it from meadow, I'd suggest to remove the metadata yaml here (or at least remove the mentions to GGDC generated by walkthrough, which is confusing). Apart from that, it looks good.

Marigold:
> I'd suggest to remove the metadata yaml here (or at least remove the mentions to GGDC generated by walkthrough, which is confusing).

@pabloarosado this proves your point that we shouldn't be using GGDC-related information in walkthrough. You think it makes sense to remove it from walkthrough and only use dummy names?

lucasrodes:
Thanks for your feedback, @pabloarosado!

I have added relevant metadata.

etl/steps/data/garden/hmd/2022-11-04/life_tables.py
```diff
@@ -0,0 +1,97 @@
...
+
+# naming conventions
+N = Names(__file__)
+
+
+def run(dest_dir: str) -> None:
+    log.info("life_tables.start")
+
+    # read dataset from meadow
+    ds_meadow = Dataset(DATA_DIR / "meadow/hmd/2022-11-04/life_tables")
```
pabloarosado, line 23: For the future: Something I usually do is to define versions at the top of the script, so that, on the next update, it's clear what needs to be manually changed. In this case, it would be `MEADOW_VERSION="2022-11-04`.

etl/steps/data/garden/hmd/2022-11-04/life_tables.py
```diff
@@ -0,0 +1,97 @@
...
+        log.info(f"life_tables:{table_name}.start")
+        tb_garden = make_table(ds_meadow, table_name)
+        ds_garden.add(tb_garden)
+        ds_garden.save()
+        log.info(f"life_tables:{table_name}.end")
+
+    log.info("life_tables.end")
+
+
+def make_table(ds_meadow, table_name: str) -> Table:
```
pabloarosado, line 43: You forgot to define the type of `ds_meadow` (I'm surprise linting didn't catch it).

etl/steps/data/garden/hmd/2022-11-04/life_tables.py
```diff
@@ -0,0 +1,97 @@
...
+    tb_garden = underscore_table(Table(df))
+    tb_garden.metadata = tb_meadow.metadata
+
+    # Edit variables
+    for col in tb_garden.columns:
+        tb_garden[col].metadata = tb_meadow[col].metadata
+
+    # Edit table
+    tb_garden.update_metadata_from_yaml(N.metadata_path, table_name)
+    tb_garden = tb_garden.set_index(["country", "year", "age"])
```
pabloarosado, line 61: I'd suggest using `verify_integrity=True`, unless you know that there can be duplicates for some reason. Just as an easy additional sanity check.

pabloarosado (approved):
Looks good. Feel free to disregard my comments.

lucasrodes:
Thanks for your feedback, @pabloarosado. Very useful

PR #640, opened 2022-11-28
data(zijdeman2015): garden dataset

lucasrodes:
Add Garden step bringing data from Zijdeman et al. (2015) to Garden. This dataset will help create the long-run variable `life expectancy at birth`.


pabloarosado (changes requested):
The metadata is not complete. If that's because you are importing it from meadow (I haven't checked), maybe it would be better to remove the metadata yaml file. Currently it shows the default walkthrough example based on GGDC (which can be a bit confusing).
Also (optionally), if no excluded countries, maybe you could remove that file and part of the code.

etl/steps/data/garden/papers/2022-11-03/zijdeman_et_al_2015/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+from structlog import get_logger
+
+from etl.helpers import Names
+from etl.paths import DATA_DIR
+
+log = get_logger()
+
+# naming conventions
+N = Names(__file__)
+COUNTRY_MAPPING_PATH = N.directory / "countries.json"
```
Marigold, line 14: You could use `N.country_mapping_path` directly. Same for `N.excluded_countries_path`

lucasrodes: Does not work with this file structure ー It builds paths such as `etl/steps/data/garden/papers/2022-11-03/zijdeman_et_al_2015/__init__.countries.json`

etl/steps/data/garden/papers/2022-11-03/zijdeman_et_al_2015/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+    tb_garden.metadata = tb_meadow.metadata
+    for col in tb_garden.columns:
+        tb_garden[col].metadata = tb_meadow[col].metadata
+
+    # Clean countries
+    tb_garden = format_columns(tb_garden)
+    tb_garden = clean_countries(tb_garden)
+
+    # Set index
+    tb_garden = tb_garden.set_index(["country", "year"])
```
pabloarosado, line 55: Consider using `set_index(["country", "year"], verify_integrity=True)`, and optionally followed by `.sort_index()` just to make it extra neat.

pabloarosado (approved):
Looks good.

PR #641, opened 2022-11-28
Migrate backporting from walden to snapshots

Marigold:
Migrate backports from Walden to Snapshots. This migration slightly simplifies the logic and prevents checksum conflicts that have been occasionally happening.

Related PR in analytics https://github.com/owid/analytics/pull/64 (this will be further tested).

## How to review

Backporting logic hasn't changed and snapshots have only simplified things (e.g. we're not using special `origin_md5` field on walden dataset anymore because it's already part of snapshot). We're still generating one JSON for config and one feather for data, only the way we access them has changed (used to be `WaldenDataset` now it is `SnapshotMeta` and `Snapshot` objects). Additionally, I've added `workers` argument to backporting to make it faster (checking all backports used to take >1min which was annoying).

Files in `snapshots/backport/latest` were generated by `bulk_backport` and contain all backports needed by ETL. No need to review them.

To test it backporting you can run `bulk_backport -d 1892` and to test ETL you can run `etl data://garden/country_profile/2022/overview --dry-run` (and see `snapshot://` steps there).

### DIFF
--- a/.dvcignore
+++ b/.dvcignore
@@ -4,9 +4,8 @@
 walkthrough/
 etl/
 data/
-!data/snapshots/
 tests/
-backport/
 vendor/
 .cachedir/
 .venv/
+!data/snapshots/
--- a/backport/backport.py
+++ b/backport/backport.py
@@ -1,14 +1,9 @@
 import datetime as dt
-import os
-import tempfile
-from typing import List, cast
+from typing import List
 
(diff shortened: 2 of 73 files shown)


pabloarosado (approved):
I've quickly checked the code and executed a bulk_backport and worked well.

PR #643, opened 2022-11-29
data(riley2005): change continent names

lucasrodes:
Use custom continent names and add continent definitions in the metadata description field. See this comment for more details https://github.com/owid/owid-issues/issues/565#issuecomment-1330267905.


etl/steps/data/garden/papers/2022-11-04/riley_2005/meta.yml
```diff
@@ -2,31 +2,37 @@ dataset:
   namespace: papers
   short_name: riley_2005
   title: Estimates of Regional and Global Life Expectancy, 1800-2001 - Riley (2005)
-  description: "Historians and demographers have gone to considerable trouble to reconstruct\
+  description:
+    "Historians and demographers have gone to considerable trouble to reconstruct\
```
pabloarosado, line 6: Are you sure these line breaks will be parsed properly in the SOURCES tab?

pabloarosado (approved):
Looks good, although I'm not sure if the line breaks (e.g. \n) are going to work in the SOURCES tab. I suppose that you would need an HTML <br> tag to have line breaks. But maybe both work.

etl/steps/data/garden/papers/2022-11-04/riley_2005/meta.yml
lucasrodes: This is the format I got from running the function `etl-metadata-export`. I am not sure ー I'll remove this just to ensure it works.

PR #662, opened 2022-12-05
data(life-exp): add region estimates

lucasrodes:
I have added a function which adds region estimates using population-weighted sum. That is, for each region and year, we compute the population-weighted sum of all its member countries' life expectancy values.

However, this is currently not in use, as it might not be the best methodology. This might change in the future.

This PR only introduces some code optimisations and assigns the same name to continents coming from the UN and Riley (2005) after discussing this with Ed. We think this is fine, given the time difference between both sources.


lucasrodes:
approving after discussing with Ed

PR #665, opened 2022-12-07
enhance(hmd): new version

lucasrodes:
This PR adds a new version of HMD (2022-12-07). Major changes include:

- Most recent data from HMD (they published data on 2022-11-29). Minor updates.
- Using Snapshot instead of Walden. There, we copy all the HMD data (and not only life tables as before).
- Dataset now includes data broken down by sex (both, female and male).

Note that this dataset (including previous versions) is not ported into Grapher. Instead, it is used by other garden datasets (long-run life expectancy), which might be ported to Grapher.

Previous version, lived in `garden/hmd/2022-11-04/life_tables` and `meadow/hmd/2022-11-04/life_tables`

### DIFF
--- a/etl/steps/data/garden/hmd/2022-12-07/life_tables.py
+++ b/etl/steps/data/garden/hmd/2022-12-07/life_tables.py
@@ -56,7 +56,7 @@ def run(dest_dir: str) -> None:
 def make_table(ds_meadow: Dataset, tables_old_names: List[str], table_new_name: str) -> Table:
     """Create a Garden table from multiple Meadow tables.
 
-    In Meadow, we have a table per agexyear and sex. In Garden, brin the sex dimension in the table as
+    In Meadow, we have a table per age-year and sex. In Garden, bring the sex dimension in the table as
     a column in the index. That is, tables "both_1x1", "female_1x1", "male_1x1" are combined into one table
     called "period_1x1".
     """
@@ -82,9 +82,8 @@ def combine_sex_tables(ds_meadow: Dataset, table_names: List[str]) -> pd.DataFra
     dfs = []
     for table_name in table_names:
         tb_meadow = ds_meadow[table_name]
-        # Country management
         df = pd.DataFrame(tb_meadow)
-        # Sex
+        # assign sex
         if table_name.startswith("both_"):
             df = df.assign(sex="all")
         elif table_name.startswith("female_"):
@@ -95,7 +94,7 @@ def combine_sex_tables(ds_meadow: Dataset, table_names: List[str]) -> pd.DataFra
             raise ValueError(f"Unknown sex for table {table_name}!")
         dfs.append(df)
     df = pd.concat(dfs, ignore_index=True)
-    return df  # type: ignore
+    return cast(pd.DataFrame, df)
 
 
 def clean_countries(df: pd.DataFrame) -> pd.DataFrame:
--- a/etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
+++ b/etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
@@ -36,6 +36,7 @@ from io import StringIO
 from pathlib import Path
 from typing import List, Tuple, cast
 
+import numpy as np
 import pandas as pd
 from owid import catalog
 from owid.catalog import Dataset, Table, TableMeta
@@ -108,17 +109,39 @@ COLUMNS_RENAME = {
 }
 # Column dtypes
 DTYPES = {
-    "central_death_rate": "Float64",
-    "probability_of_death": "Float64",
-    "avg_survival_length": "Float64",
-    "num_survivors": "Int64",
-    "num_deaths": "Int64",
-    "num_person_years_lived": "Int64",
-    "num_person_years_remaining": "Int64",
-    "life_expectancy": "Float64",
+    "central_death_rate": "float64",
+    "probability_of_death": "float64",
+    "avg_survival_length": "float64",
+    "num_survivors": "float64",
+    "num_deaths": "float64",
+    "num_person_years_lived": "float64",
+    "num_person_years_remaining": "float64",
+    "life_expectancy": "float64",
     "age": "category",
     "country": "category",
     "year": "category",
+}
+# missing data: this are the countries where data is missing for a specific table.
+# these checks are done in function _clean_correct_missing_data.
+MISSING_DATA_COUNTRIES = {
+    "bltper_1x1": ["Belgium"],
+    "bltper_1x5": ["Belgium"],
+    "bltper_1x10": [],
+    "bltper_5x1": ["Belgium"],
+    "bltper_5x5": ["Belgium"],
+    "bltper_5x10": [],
+    "fltper_1x1": ["Belgium"],
+    "fltper_1x5": ["Belgium"],
+    "fltper_1x10": [],
+    "fltper_5x1": ["Belgium"],
+    "fltper_5x5": ["Belgium"],
+    "fltper_5x10": [],
+    "mltper_1x1": ["Belgium"],
+    "mltper_1x5": ["Belgium"],
+    "mltper_1x10": [],
+    "mltper_5x1": ["Belgium"],
+    "mltper_5x5": ["Belgium"],
+    "mltper_5x10": [],
 }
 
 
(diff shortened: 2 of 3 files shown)


etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,400 @@
...
+    "ax": "avg_survival_length",
+    "lx": "num_survivors",
+    "dx": "num_deaths",
+    "Lx": "num_person_years_lived",
+    "Tx": "num_person_years_remaining",
+    "ex": "life_expectancy",
+}
+# Column dtypes
+DTYPES = {
+    "central_death_rate": "Float64",
```
pabloarosado, line 111: I would try to avoid Float64, given that there is [an open bug](https://github.com/pandas-dev/pandas/issues/49818) that is absolutely terrifying, which makes methods like isnull, dropna, and so on, useless in certain circumstances. But if you have checked that in this case you don't have any problem, the go ahead.

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,400 @@
...
+
+    Returns
+    -------
+    catalog.Table
+        Table with data.
+    """
+    # List all files (each file corresponds to a country roughly)
+    files = glob(os.path.join(input_folder, "*.txt"))
+    log.info(f"hmd_lt: looking for available files in {input_folder}: {len(files)} found.")
+    # # Create df
```
pabloarosado, line 233: Double comment (same a few lines below).

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,400 @@
...
+    # df to table
+    table = df_to_table(df, age, year, folder["sex"])
+    # # underscore all table columns
+    table = underscore_table(table)
+    return table
+
+
+def _age_year(input_folder: str) -> Tuple[str, str]:
+    ageyear = input_folder.split("_")[-1]
+    age, year = ageyear.split("x")
```
pabloarosado, line 248: You could easily check here that age and year are numbers.

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,400 @@
...
+    table = table.strip()
+    # Remove spacing after newline
+    table = re.sub(r"\n\s+", "\n", table)
+    # Replace spacing with tabs
+    table = re.sub(r"[^\S\r\n]+", "\t", table)
+    # Build df
+    df = pd.read_csv(StringIO(table), sep="\t")
+    # Assign country
+    df = df.assign(Country=country)
+    # # Filter columns
```
pabloarosado, line 300: You can remove these comments.

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,400 @@
...
+    # Correct missing data
+    df = _clean_correct_missing_data(df)
+    # Set dtypes
+    df = _clean_set_dtypes_df(df)
+    return df
+
+
+def _clean_rename_columns_df(df: pd.DataFrame) -> pd.DataFrame:
+    """Rename columns."""
+    return df.rename(columns=COLUMNS_RENAME)
```
pabloarosado, line 336: An additional check here would be to add `errors='raise'` to ensure that all columns to be renamed are found.

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,400 @@
...
+    # Find missing data
+    rows_missing = df[df.central_death_rate == "."]
+    num_rows_missing = len(rows_missing) / len(df)
+    countries_missing_data = rows_missing.country.unique()
+    # Run checks
+    assert num_rows_missing < num_rows_missing_expected, (
+        f"More missing data than expected was found! {round(num_rows_missing*100, 2)} rows missing, but"
+        f" {round(num_rows_missing_expected*100,2)} were expected."
+    )
+    assert len(countries_missing_data) <= countries_missing_data_expected, (
```
pabloarosado, line 353: If you have identified exactly what the missing data is, a safer option would be to actually list, e.g., the country that has missing data, on the top of the script. Then assert that the data is indeed missing (in case this changes on next update), and that there is no additional data missing.

snapshots/hmd/2022-12-07/hmd.py
```diff
@@ -0,0 +1,41 @@
+"""Get HMD full dataset into Snapshots.
```
pabloarosado, line 1: I think we should add something like this as template in walkthrough, for the cases when the file needs to be added manually @Marigold .

snapshots/hmd/2022-12-07/hmd.py
```diff
@@ -0,0 +1,41 @@
...
+    - Obtain the publication date in the section "Previous Versions" or hardcoded in the name of the downloaded file.
+"""
+
+from pathlib import Path
+
+import click
+
+from etl.snapshot import Snapshot
+
+CURRENT_DIR = Path(__file__).parent
```
pabloarosado, line 16: I think you can remove the previous line.

snapshots/hmd/2022-12-07/hmd.py
```diff
@@ -0,0 +1,41 @@
...
+    # Ensure destination folder exists.
+    snap.path.parent.mkdir(exist_ok=True, parents=True)
+    # Copy local data file to snapshots data folder.
+    snap.path.write_bytes(Path(path_to_file).read_bytes())
+    # Add file to DVC and upload to S3.
+    snap.dvc_add(upload=upload)
+
+
+if __name__ == "__main__":
+    print(CURRENT_DIR)
```
pabloarosado, line 40: I think you can remove the previous line.

etl/steps/data/garden/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,136 @@
...
+        ds_garden.save()
+        log.info(f"life_tables: {table_new}.end")
+
+    log.info("life_tables.end")
+
+
+def make_table(ds_meadow: Dataset, tables_old_names: List[str], table_new_name: str) -> Table:
+    """Create a Garden table from multiple Meadow tables.
+
+    In Meadow, we have a table per agexyear and sex. In Garden, brin the sex dimension in the table as
```
pabloarosado, line 59: agexyear -> age-year
bring -> bring

etl/steps/data/garden/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,136 @@
...
+
+    return tb_garden
+
+
+def combine_sex_tables(ds_meadow: Dataset, table_names: List[str]) -> pd.DataFrame:
+    """Combine meadow tables."""
+    dfs = []
+    for table_name in table_names:
+        tb_meadow = ds_meadow[table_name]
+        # Country management
```
pabloarosado, line 85: What does "Country management mean?"

etl/steps/data/garden/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,136 @@
...
+            df = df.assign(sex="all")
+        elif table_name.startswith("female_"):
+            df = df.assign(sex="female")
+        elif table_name.startswith("male_"):
+            df = df.assign(sex="male")
+        else:
+            raise ValueError(f"Unknown sex for table {table_name}!")
+        dfs.append(df)
+    df = pd.concat(dfs, ignore_index=True)
+    return df  # type: ignore
```
pabloarosado, line 98: Instead of ignoring type check, you can `cast(pd.DataFrame, df)`.

etl/steps/data/garden/hmd/2022-12-07/life_tables.py
```diff
@@ -0,0 +1,136 @@
...
+def exclude_countries(df: pd.DataFrame) -> pd.DataFrame:
+    """Exclude countries."""
+    excluded_countries = load_excluded_countries()
+    return cast(pd.DataFrame, df.loc[~df.country.isin(excluded_countries)])
+
+
+def harmonize_countries(df: pd.DataFrame) -> pd.DataFrame:
+    """Country harmonization."""
+    unharmonized_countries = df["country"]
+    df = geo.harmonize_countries(df=df, countries_file=str(N.country_mapping_path))
```
pabloarosado, line 125: Even if we know that harmonize_countries needs some additional work, an easy win would be to add the option to pass the path to the excluded countries file. Having that we could spare having to define exclude_countries, and harmonize_countries every time.

pabloarosado (approved):
Nice work (and what an annoying input data format).

etl/steps/data/garden/hmd/2022-12-07/life_tables.py
lucasrodes: Good idea, moved this to https://github.com/owid/owid-datautils-py/issues/55

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
lucasrodes: Thanks!

etl/steps/data/meadow/hmd/2022-12-07/life_tables.py
lucasrodes: good idea!

lucasrodes:
Thanks, @pabloarosado, for your thoughtful feedback!

PR #669, opened 2022-12-09
Create Penn World Table: final steps

paarriagadap:
`grapher` step and `garden` modifications


etl/steps/data/grapher/ggdc/2022-11-28/penn_world_table.py
```diff
@@ -0,0 +1,41 @@
...
+
+    table = N.garden_dataset["penn_world_table"]
+
+    # %%
+    # Specify sheet id and sheet (tab) name for a Google Sheet with the variables to keep
+    sheet_id = "1gbk8lBc4ZTjzE94pG8vgFX1Ta5baIQhpdD158GhPJsc"
+    sheet_name = "variable_metadata"
+
+    # Read in variable metadata as dataframe
+    url = f"https://docs.google.com/spreadsheets/d/{sheet_id}/gviz/tq?tqx=out:csv&sheet={sheet_name}"
```
Marigold, line 20: In ETL we try to avoid external sources to make it as replayable as possible and consistent in case someone changes something in the spreadsheet. Before we merge this we should copy the spreadsheet to the repo (as either csv or we could convert it to our YAML format).

paarriagadap: Ah, OK, let me see the alternative. Now it is only to filter the list of variables, so I can add something like a json file

etl/steps/data/grapher/ggdc/2022-11-28/penn_world_table.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def run(dest_dir: str) -> None:
+    dataset = catalog.Dataset.create_empty(dest_dir, N.garden_dataset.metadata)
+
+    table = N.garden_dataset["penn_world_table"]
+
+    # %%
+    # Read in variables to keep file
+    df_vars_to_keep = pd.read_csv(VARS_TO_KEEP_FILE)
```
Marigold, line 18: Why don't we send all variables to grapher? It doesn't really matter if have unused variables there.

(In case you really don't want to send them, it might be easier to just add list of variables as a constant to the top of the file rather than reading them from CSV. It's only 32 variables which is totally fine.)

paarriagadap: There are very specific variables that are useful to have in garden for analysis but not for plotting. In that `vars_to_keep` file there are also the current variable IDs. I will check the `etl-match-variables` function to create the final format that I need.

Marigold: We don't have to plot all variables we have in grapher dataset. It's quite common to have a dataset with 100 variables, but only one is plotted.

paarriagadap: Actually there are several variables in the dataset currently not used in any chart. It's just we don't want the noise of variables that are useless, like some price levels or several variables which their values are like *Extrapolated (0), benchmark (1) or interpolated (2)*. We can change our mind of course, and that's why they're in `garden`

paarriagadap:
I have an error here because I replace the name of the source and I changed this bit in meta.py to replace:
`def update_from_yaml(
        self, path: Union[Path, str], if_source_exists: Literal["fail", "append", "replace"] = "replace"`
 Should I just merge? I'm not pushing that change.

etl/steps/data/garden/ggdc/2022-11-28/penn_world_table.py
```diff
@@ -22,6 +23,136 @@ def run(dest_dir: str) -> None:
...
+
+    # Multiplying by 1 million to get "people" instead of "millions of people"
+    df["pop"] = df["pop"] * 1000000
+    df["emp"] = df["emp"] * 1000000
+
+    # %% [markdown]
+    # A range of variables are provided as shares (0-1), which we multiply by 100 to express as a percentage.
+
+    # %%
+    df["labsh"] = df["labsh"] * 100
```
Marigold, line 54: You simplify this as:
```
df[["labsh", "irr", ..., "csh_r"]] *= 100
```

etl/steps/data/garden/ggdc/2022-11-28/penn_world_table.py
```diff
@@ -22,6 +23,136 @@ def run(dest_dir: str) -> None:
...
+    df["csh_m"] = df["csh_m"] * 100
+    df["csh_r"] = df["csh_r"] * 100
+
+    # %% [markdown]
+    # ## GDP per capita variables
+    # Penn World Table do not directly provide GDP per capita. We calculate
+    # these by dividing GDP by the population figures they provide (both now multiplied
+    # by 1,000,000).
+    # %%
+    df["rgdpe_pc"] = df["rgdpe"] / df["pop"]
```
Marigold, line 70: Might be better to write it as:

```python
cols_per_capita = ["rgdpe", ..., "rgdpna"]
for col in cols_to_convert_per_capita:
    df[f"{col}_pc"] = df[col] / df["pop"]
```

etl/steps/data/garden/ggdc/2022-11-28/penn_world_table.py
```diff
@@ -22,6 +23,136 @@ def run(dest_dir: str) -> None:
...
+        "Netherlands Antilles",
+        "USSR",
+        "Yugoslavia",
+    ]
+
+    # Create a list of countries available only in the national accounts dataset, with missing v_x and v_m data
+
+    # print(df_na[df_na['country']=="Yugoslavia"])
+
+    # df_na_null_values = df_na.groupby(['country']).all().reset_index().copy()
```
Marigold, line 127: This code should be probably deleted before merging.

Marigold (changes requested):
Should `hmd/2022-12-07/life_tables` be part of this PR? It seems like it got into your PR by some accident.

snapshots/backport/latest/dataset_2904_figure1_10_config.json.dvc
```diff
@@ -1,5 +1,5 @@
 meta:
```
Marigold, line 1: These backports shouldn't be in this PR. Can you rebase on top of master?

Marigold:
> I have an error here because I replaced the name of the source and I changed this bit in meta.py to replace:
> def update_from_yaml(
        self, path: Union[Path, str], if_source_exists: Literal["fail", "append", "replace"] = "replace"
> Should I just merge? I'm not pushing that change.

It fails because in garden step you're calling `ds_garden.metadata.update_from_yaml(N.metadata_path)` which tries to update metadata from `penn_world_table.meta.yml`. However, your meadow step already has a source and it's not clear for `update_from_yaml` if the new source should be replaced or appended. Since the source is identical to meadow, you should just **remove that source** from garden version of `penn_world_table.meta.yml`.

paarriagadap:
> Should `hmd/2022-12-07/life_tables` be part of this PR? It seems like it got into your PR by some accident.

I think it's because of the rebase. I never saw this when I was commiting my things

paarriagadap:
@Marigold So I rebased and now there are some errors about missing commits. While rebasing I skipped some which didn't affect my files

dag_files/dag_walkthrough.yml
```diff
@@ -30,7 +30,9 @@ steps:
   - walden://ggdc/2021-06-18/penn_world_table
   data://meadow/ggdc/2022-11-28/penn_world_table_national_accounts:
   - walden://ggdc/2021-06-18/penn_world_table_national_accounts
+  - data://garden/reference
   data://garden/ggdc/2022-11-28/penn_world_table:
   - data://meadow/ggdc/2022-11-28/penn_world_table
-  data://garden/ggdc/2022-11-28/penn_world_table_national_accounts:
```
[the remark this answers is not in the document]
paarriagadap: I did. I first created a garden step but then I left it until meadow, because I just wanted it to be merged with the main in garden.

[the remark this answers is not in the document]
paarriagadap: So `penn_world_table` goes all the way until grapher and `penn_world_table_national_accounts` stops at meadow

Marigold (approved):
Looks good! Glad to be out of rebase hell :)

PR #671, opened 2022-12-09
enhance(life-expectancy): add americas, document anomalies

lucasrodes:
## Americas
After [a discussion](https://owid.slack.com/archives/C03N1V9KXB9/p1670516968289509) on slack, we are now adding "Americas" to our Life Expectancy dataset. Also, I will be documenting some of the anomalies found in the data.

- Americas: _Riley (2005)_ provides values for this entity, but _UN WPP_ and others don't. Instead, they report "Northern America" and "Latin America and the Caribbean" separately.
  - With @mallikasnyder, we reviewed different methodologies to estimate a region aggregate's life expectancy based on its member regions.
  - From the literature, we could not find one single method but several. Some proposed are birth-weighted average, others population-weighted average, population-at-X-age-weighted average, etc. Some references are https://population.un.org/wpp/Publications/Files/WPP2022_Methodology.pdf (pages 6-7), https://pubmed.ncbi.nlm.nih.gov/12317480/.
  - In the end, @maxroser suggested using a population-weighted approach.  

## Documenting anomalies.
@eoo-owid **[suggested](https://owid.slack.com/archives/C03N1V9KXB9/p1670516968289509)** it would be great if we could document the anomalies found in the data. Especially if coming from combining different sources, but also if the source already presents such anomalies.

I have added a file, `life_expectancy.anomalies.yml`, which lists historical events in the Garden step. 

### Garden
This is added in Garden because this is not necessarily a Grapher thing, but rather a Garden, I would say. The user consuming directly from the catalogue should also have access to this list.

This is now added to the dataset by creating a table called `_anomalies`, which contains one cell with the raw YAML file. This is not ideal and could have a regular table with columns `country`, `year_start`, `year_end`, `event_name` and `link` instead. However, I was not sure about the structure yet.

### Grapher
In Grapher, we read the raw YAML file and created an HTML-rich text added at the bottom of the dataset description.

Usually, this list of anomalies might affect a single variable, so perhaps this list is something we want at the variable level. However, I would say this is good enough with the current constraints.

You can get a preview of how this will look in [staging](https://staging.owid.cloud/admin/charts/64/edit).


etl/steps/data/garden/demography/2022-11-30/life_expectancy.py
```diff
@@ -395,3 +435,50 @@ def add_region_aggregates(frame: pd.DataFrame) -> pd.DataFrame:
...
+    ]
+    assert len(set(df["location"])) == 2, f"Check that all of {locations} are in df"
+    df["location"] = df["location"].replace(REGION_MAPPING)
+
+    # rename columns
+    df = df.rename(columns={"location": "country", "value": "population"})
+    return df
+
+
+def make_anomaly_table() -> Table:
```
Marigold, line 471: Oh, interesting way how to get arbitrary information across steps :). In `VariableMeta` there's a field `additional_info` that we use to carry around extra information (usually for internal purposes). You can put your dictionary there. I'm just not sure where since it lives under variable. Perhaps `life_expectancy_0`? We could also add `additional_info` to Table or Dataset if that would feel more natural.

If you feel that data annotations are close then don't bother rewriting this, we can refactor it once we have proper place for annotations.

etl/steps/data/garden/demography/2022-11-30/life_expectancy.py
```diff
@@ -342,6 +354,34 @@ def merge_dfs(df_wpp: pd.DataFrame, df_hmd: pd.DataFrame, df_zij: pd.DataFrame,
...
+    # add population
+    df = add_population_americas(df)
+    # estimate values for regions
+    # y(country) = weight(country) * metric(country)
+    df["life_expectancy_0"] *= df.population
+    df["life_expectancy_15"] *= df.population
+    df["life_expectancy_65"] *= df.population
+    df["life_expectancy_80"] *= df.population
+    # z(region) = sum{ y(country) } for country in region
+    df = df.groupby("year", as_index=False).sum(numeric_only=True)
```
pabloarosado, line 371: Keep in mind that `groupby().sum()` will treat nans as zeros. I don't know if that could bias the weights. It would be good to do some further sanity checks here.

etl/steps/data/garden/demography/2022-11-30/life_expectancy.py
```diff
@@ -395,3 +435,50 @@ def add_region_aggregates(frame: pd.DataFrame) -> pd.DataFrame:
...
+    # Load anomaly yaml file
+    with open(PATH_ANOMALIES) as f:
+        anomalies = yaml.safe_load(f)
+    # store all yaml's content as a string in a cell in the table
+    df = pd.DataFrame({"anomalies": [str(anomalies)]})
+    tb = Table(df)
+    # add metadata
+    tb.metadata = TableMeta(
+        short_name="_anomalies",
+        description="this table contains the anomalies for the life expectancy data in YAML format.",
```
pabloarosado, line 482: I just want to point out that what you are providing here are not anomalies. They are possible causes of the anomalies. The anomalies themselves would be simply data points (e.g. France 1918). Proving that those strange points are caused by those events would require some complicated causal inference. So, somewhere, you could make that clarification (maybe in the grapher step, where you list the events).

pabloarosado (approved):
Looks good. The only two things I'd suggest that are important are:
1. Checking that nans are not affecting the weighted averages (this could be a separate file, not necessarily here).
2. Clarifying what you mean by anomalies (this could simply be a comment in the grapher step).

etl/steps/data/garden/demography/2022-11-30/life_expectancy.anomalies.yml
```diff
@@ -0,0 +1,274 @@
+Afghanistan:
```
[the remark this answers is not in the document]
lucasrodes: I think the structure of the historical events needs re-thinking. Not only can a single event affect multiple countries, but also multiple metrics or other dimensions.

That is likely the more extensive project. I'll be addressing this soon.

etl/steps/data/garden/demography/2022-11-30/life_expectancy.py
lucasrodes: added an assert to ensure there is no NaN before doing groupby

etl/steps/data/garden/demography/2022-11-30/life_expectancy.py
lucasrodes: changed naming of files to "historical events"

lucasrodes:
@pabloarosado thanks for your feedback, addressed your suggestions.

PR #674, opened 2022-12-12
Fix electricity mix share

pabloarosado:
* Fixed issue on electricity mix dataset: Some electricity sources used to add up to > 100% of the total electricity generation, due to small discrepancies between BP and Ember data. I fixed it by prioritizing Ember data.
* Updated all garden and grapher steps using the electricity mix dataset.
* Added a new variable (electricity as a share of primary energy) to the energy dataset.
* As a bonus, I combined the `shared` modules of various data steps, to avoid redundancy. Some of these functions could be moved to datautils (although having historical countries in our population dataset would be a better solution).

The data steps have been copy/pasted from previous versions, with minor modifications (as mentioned above), so there's no need to check the code thoroughly.


etl/steps/data/garden/energy/2022-12-12/electricity_mix.py
```diff
@@ -0,0 +1,324 @@
...
+            "renewable_generation__twh",
+            "nuclear_generation__twh",
+        ],
+    }
+
+    # Create new columns, by adding up other columns (and allowing for only one nan in each sum).
+    for new_column in aggregates:
+        table_bp[new_column] = pd.DataFrame(table_bp)[aggregates[new_column]].sum(
+            axis=1, min_count=len(aggregates[new_column]) - 1
+        )
```
lucasrodes, line 81: wouldn't the following also work?

```python
table_bp[new_column] = table_bp[aggregates[new_column]].sum(
    axis=1, min_count=len(aggregates[new_column]) - 1
)
```

or perhaps adding a step before the loop with `df_bp = pd.DataFrame(table_bp)` and then

```python
df_bp[new_column] = df_bp[aggregates[new_column]].sum(
    axis=1, min_count=len(aggregates[new_column]) - 1
)
```

etl/steps/data/garden/energy/2022-12-12/electricity_mix.py
```diff
@@ -0,0 +1,324 @@
...
+
+    # Prepare data in a dataframe with a dummy index.
+    df_bp = pd.DataFrame(table_bp).reset_index()
```
lucasrodes, line 84: If `df_bp` has already been defined (see my previous comment above), then you'd simply `df_bp = df_bp.reset_index()`

etl/steps/data/garden/energy/2022-12-12/electricity_mix.py
```diff
@@ -0,0 +1,324 @@
...
+    # In BP data, there is a variable "Geo Biomass Other", which combines all other renewables.
+    # In Ember data, "other rewenables" excludes bioenergy.
+    # To be able to combine both datasets, create a new variable for generation of other renewables including bioenergy.
+    table_ember["other_renewables_including_bioenergy_generation__twh"] = (
+        pd.DataFrame(table_ember)["other_renewables_excluding_bioenergy_generation__twh"]
+        + table_ember["bioenergy_generation__twh"]
+    )
+
+    # Prepare data in a dataframe with a dummy index.
+    df_ember = pd.DataFrame(table_ember).reset_index()
```
lucasrodes, line 134: Idem to my previous two comments for `table_bp`. Perhaps worth defining `df_ember = pd.DataFrame(table_ember)`.

Just for readability purposes.

etl/steps/data/garden/energy/2022-12-12/electricity_mix.py
```diff
@@ -0,0 +1,324 @@
...
+        share variables.
+
+    Returns
+    -------
+    table : catalog.Table
+        Original data in a table format with metadata.
+
+    """
+    # Sort rows and columns conveniently and set an index.
+    combined = combined[sorted(combined.columns)]
```
lucasrodes, line 264: you can use

```python
combined.sort_index(axis=1)
```

etl/steps/data/garden/energy/2022-12-12/owid_energy.py
```diff
@@ -0,0 +1,211 @@
...
+    ds_energy_mix = catalog.Dataset(ENERGY_MIX_DATASET_PATH)
+    ds_fossil_fuels = catalog.Dataset(FOSSIL_FUEL_PRODUCTION_DATASET_PATH)
+    ds_primary_energy = catalog.Dataset(PRIMARY_ENERGY_CONSUMPTION_DATASET_PATH)
+    ds_electricity_mix = catalog.Dataset(ELECTRICITY_MIX_DATASET_PATH)
+
+    # Gather all required tables from all datasets.
+    tb_energy_mix = ds_energy_mix[ds_energy_mix.table_names[0]].reset_index()
+    tb_fossil_fuels = ds_fossil_fuels[ds_fossil_fuels.table_names[0]].reset_index()
+    tb_primary_energy = ds_primary_energy[ds_primary_energy.table_names[0]].reset_index()
+    tb_electricity_mix = ds_electricity_mix[ds_electricity_mix.table_names[0]].reset_index()
```
lucasrodes, line 155: I would rather define the specific names of the tables you expect to load (instead of assuming the first one is the one).

etl/steps/data/garden/energy/2022-12-12/owid_energy.py
```diff
@@ -0,0 +1,211 @@
...
+
+    Returns
+    -------
+    tb_combined : catalog.Table
+        Combined table with metadata.
+
+    """
+    # Merge all tables as a dataframe (without metadata).
+    dfs = [pd.DataFrame(table) for table in tables.values()]
+    df_combined = dataframes.multi_merge(dfs, on=["country", "year"], how="outer")
```
lucasrodes, line 76: yeah! `multi_merge` rocks!

etl/steps/data/garden/energy/2022-12-12/uk_historical_electricity.py
```diff
@@ -0,0 +1,211 @@
...
+    #
+    # Load data.
+    #
+    # Read all required datasets.
+    ds_beis = catalog.Dataset(UK_BEIS_DATASET_PATH)
+    ds_elec = catalog.Dataset(ELECTRICITY_MIX_DATASET_PATH)
+
+    # Gather all required tables from all datasets.
+    tb_beis = ds_beis[ds_beis.table_names[0]]
+    tb_elec = ds_elec[ds_elec.table_names[0]]
```
lucasrodes, line 177: I'd specify table names. Perhaps even as global vars.

etl/steps/data/grapher/energy/2022-12-12/electricity_mix.py
```diff
@@ -0,0 +1,19 @@
...
+# Path to garden dataset to be loaded.
+DATASET_PATH = DATA_DIR / "garden" / "energy" / "2022-12-12" / "electricity_mix"
+
+
+def run(dest_dir: str) -> None:
+    garden_dataset = catalog.Dataset(DATASET_PATH)
+    dataset = catalog.Dataset.create_empty(dest_dir, garden_dataset.metadata)
+
+    # There is only one table in the dataset, with the same name as the dataset.
+    table = garden_dataset[garden_dataset.table_names[0]].reset_index().drop(columns=["population"])
```
lucasrodes, line 17: I'd specify the table name (instead of assuming it is the first one). Also, is there a need to preserve the column `population` after the Garden step? Otherwise, I'd drop it in Garden.

lucasrodes (approved):
Minor comments ー LGTM

etl/steps/data/grapher/energy/2022-12-12/electricity_mix.py
pabloarosado: I don't think it's strictly necessary, but it's convenient to keep population in garden, next to per capita variables.

pabloarosado:
Thanks @lucasrodes! I've implemented all your suggestions.

PR #682, opened 2022-12-15
feat(geo): migrate from owid.datautils.geo

lucasrodes:
## Summary

- All methods from `owid.datautils.geo` are now in `etl.data_helpers.geo`. You can import them as
    ```python
    from etl.data_helpers.geo import *
    ```
- All has been copied word by word, except for how the ETL datasets are loaded. Now, instead of using `RemoteCatalog` I use `owid.catalog.Dataset` and load these from the local disk.
- I have copied the tests from `owid.datautils.geo`.
- I kept the code from former `etl.data_helpers` under `etl.data_helpers.geo_old`. This code is not used anywhere at the moment, but we might want to re-use it in the future.

I'd say that the next steps should be to add some deprecation warnings to `owid.datautils`.

First thoughts @Marigold @pabloarosado ?


etl/data_helpers/__init__.py
```diff
@@ -0,0 +1,4 @@
+from etl.data_helpers.geo_old import calculate_region_sums
```
pabloarosado, line 1: If this function is not used in any data step, I'd suggest to simply remove it. Especially because we should not use a function that, by default, creates its own aggregate for the world. If a dataset contains data for the world (which is common) but we want to create aggregates for our own region definitions, this function can't be used.

etl/data_helpers/geo_old.py
```diff
@@ -1,3 +1,8 @@
+"""
```
pabloarosado, line 1: I'd suggest removing this module if it's not used anywhere. It can be confusing and someone may start using it by mistake.

pabloarosado: Idem with `test_geo_old`.

pabloarosado (approved):
I've suggested removing some unused code. Apart from that, it looks good.

pabloarosado:
> ## Summary
> * All methods from `owid.datautils.geo` are now in `etl.data_helpers.geo`. You can import them as
>   ```python
>   from etl.data_helpers.geo import *
>   ```
> * All has been copied word by word, except for how the ETL datasets are loaded. Now, instead of using `RemoteCatalog` I use `owid.catalog.Dataset` and load these from the local disk.
> * I have copied the tests from `owid.datautils.geo`.
> * I kept the code from former `etl.data_helpers` under `etl.data_helpers.geo_old`. This code is not used anywhere at the moment, but we might want to re-use it in the future.
> 
> I'd say that the next steps should be to add some deprecation warnings to `owid.datautils`.
> 
> First thoughts @Marigold @pabloarosado ?

Yes, plus finding all usages of the old `datautils.geo` and letting them use the new one.

Marigold (approved):
Looks good. This is exactly the situation where having automated check of all rebuilt datasets against production would be beneficial.

lucasrodes:
Created these issues as a follow up:

- #691
- https://github.com/owid/owid-datautils-py/issues/56

PR #683, opened 2022-12-16
enhance(chart-suggester): ergonomics

lucasrodes:
This PR updates the API of command tool `etl-chart-suggester`:

```
etl-chart-suggester --help
Usage: etl-chart-suggester [OPTIONS] MAPPING_FILE

  Generate chart revisions in Grapher using MAPPING_FILE JSON file.
  MAPPING_FILE is a JSON file with old_variable_id -> new_variable_id pairs.
  E.g. {2032: 147395, 2033: 147396, ...}.

  Make sure that you are connected to the database. By default, it connects to
  Grapher based on the environment file found in the project's root directory
  `path/to/etl/.env`.

Options:
  --help  Show this message and exit.
```

That is, you simply have to run `etl-chart-suggester mapping.json`.

## Background
In the past, the usage of `etl-chart-suggester` was a bit confusing and redundant, as we had to provide four arguments:

- JSON file with variable mapping(old -> new)
- Updated dataset namespace
- Updated dataset name
- Updated dataset version

These last three seemed unnecessary. The only reason they were used is to specify a "Reason for suggested revision" in the chart approval tool in the Grapher UI:

image

### ChartRevisionSuggester
Object `ChartRevisionSuggester` has an updated API now:

You can instantiate it directly:

```python
from etl.chart_revision_suggester import ChartRevisionSuggester
mapping = {123: 456, 789: 101}
suggester = ChartRevisionSuggester(mapping)
```

Or indirectly via a mapping file

```python
from etl.chart_revision_suggester import ChartRevisionSuggester
mapping_file = "mapping.json"
suggester = ChartRevisionSuggester.from_dict(mapping_file)
```
where "mapping.json" contains a dictionary like `"{123: 456, 789: 101}"`.

### AdminUI
I found an alternative way to find the reason for the suggested revision. Please refer to function `ChartRevisionSuggester._get_chart_update_reason` in `etl.chart_revision_suggester` for more details. But basically, it checks for each revised chart which variables have been updated and produces a more detailed reason:

image


lucasrodes:
@spoonerf I changed a file from a step you worked on (file `etl/chart_revision_suggester.py`). Could you take a look, please? Unsure if it is still in use. If so, could you check if it runs? Happy to discuss this on slack.

lucasrodes:
After discussing with @spoonerf, we decided to move ahead and delete `etl/steps/data/grapher/un_sdg/2022-07-07/match_vars_crs.py` file, as it is no longer in use.

pabloarosado (approved):
Thanks, this is an improvement.

PR #684, opened 2022-12-16
fix(emdat): Fix bug in natural disasters decadal data after last refa…

pabloarosado:
Fix bug in natural disasters decadal data after last refactor.


Marigold (approved):
Looks good. CI is failing, but it's unrelated and this can be merged.

PR #685, opened 2022-12-16
WIP: migrate walden to snapshots and remove walden from ETL

Marigold:
I was midway through the migration, but then we realised that removing walden isn't worth potential errors from the refactored code.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -64,8 +64,3 @@ jobs:
         run: |
           cd vendor/owid-catalog-py
           make test
-
-      - name: walden tests
-        run: |
-          cd vendor/walden
-          make test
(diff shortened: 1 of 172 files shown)


danyx23:
Did I misunderstand you earlier today? I thought you did still think that this would be worthwhile, it would just be good to have a better change detection in CI?

Marigold:
> Did I misunderstand you earlier today? I thought you did still think that this would be worthwhile, it would just be good to have a better change detection in CI?

I wrote that before our discussion. I agreed it would be worthwhile, though only with proper CI that would make sure nothing has changed.

Marigold:
Having walden in the repo is not hurting us that much. We might reopen this in the future if we decide to migrate.

Issue #525, opened 2022-09-27
Lint for Python files that are not used in any step


larsyencken:
## Problem

It's possible to add scripts to the `steps/` folder which are used in building a step, but not considered part of a step's dependencies.

This happened for UN WPP (#524) and we didn't notice it for a long time. There might be other datasets with this problem too.

## Proposed solution

We should add a lint step that scans the `steps/data/` folder for these cases and fails in CI if there are any spare Python files that are unaccounted for.

larsyencken:
Triaging as "important", since it's a potential correctness issue, and estimating at 1d, since you basically need to create a version of the DAG that extends to the file level.

larsyencken:
Since this is important to catch errors, tagging it as "pinned".

larsyencken:
Examples in the wild: ``etl/steps/data/garden/energy/*/owid_energy_variable_mapping.csv`

larsyencken:
We've lived without it this far, calling it "nice to have"

PR #686, opened 2022-12-16
Load paths from dag

pabloarosado:
In the last tech tea, we mentioned that it would be good to ensure that a data step loads data only from its dependencies. In fact, currently, all versions and paths to datasets used in data steps are hardcoded, which causes that, when duplicating a data step to create an update, one has to manually change versions inside the code, which is error-prone.

Here I propose a solution, which you can understand with an example:
```
paths = PathFinder(__file__)
# Get path to electricity mix dataset (which is one of the dependencies of current data step).
ELECTRICITY_MIX_DATASET_PATH = paths.get_dataset_path(channel="garden", short_name="electricity_mix")

```
This will get the path to the `electricity_mix` dataset, which is one of the dependencies of the current data step. By the way, here, `PathFinder` is actually `Names` (but I renamed it for convenience).

**UPDATE**: After Mojmir's suggestion, I simplified the above, so that a dependency can be loaded directly, without even specifying the channel, as in:
```
paths = PathFinder(__file__)

def run(dest_dir):
    electricity_mix_dataset = paths.load_dependency("electricity_mix")
```
If there is ambiguity (e.g. two dependencies have the same short name), the function raises an error.

Changes:
* Extended `helpers.Names` to be able to load datasets that are listed as dependencies of a particular data step.
* Renamed `Names` -> `PathFinder` (I think `Names` is not very informative), but also defined `Names` for backward compatibility. I'm happy to change it back if you disagree.
* I had to move `get_etag` to avoid circular modular dependencies (and adjusted tests accordingly).
* I modified the code of one of our smallest datasets as an example of how to use `PathFinder`.

Note that I had to load the dag inside etl.helpers, which may not be a good idea. An alternative would be to load it when initialising `PathFinder`.

What do you think?


etl/helpers.py
```diff
@@ -118,3 +142,49 @@ def walden_dataset(self) -> WaldenDataset:
...
+        return self._create_step_name(
+            channel=self.channel, namespace=self.namespace, version=self.version, short_name=self.short_name
+        )
+
+    @property
+    def dependencies(self) -> List[str]:
+        return DAG[self.step]
+
+    def get_dataset_path(
+        self, channel: str, short_name: str, namespace: Optional[str] = None, version: Optional[Union[str, int]] = None
```
Marigold, line 168: Could `channel` be optional? I'd like to get dependency with as little as possible. E.g. `paths.deps["electricity_mix"]` that would give me `Dataset` instance (or Snapshot). I could then use a proper type in code. Something like

```python
def __get__(self) -> Union[Dataset, Snapshot]:
   ...

electricity_mix_ds: Dataset = paths.deps["electricity_mix"]
electricity_snapshot: Snapshot = paths.deps["electricity_mix"]
```

Marigold:
This is really cool! I wanted to do something similar, but by passing dependencies to `run(..., deps)` function. Your solution is essentially the same thing, but simpler and not as invasive as mine. You should coordinate with @danyx23 as he has been prototyping the same thing.

pabloarosado:
Thanks for the suggestions @Marigold! I have implemented them in my last push. Is there any way to be able to load dependency datasets without needing to specify the object type? That's a bit annoying, although not a deal breaker. If you agree with the changes, I'll add tests.

Marigold:
> Is there any way to be able to load dependency datasets without needing to specify the object type?

I don't think so, you'd have to go with something similar to `typed_dag` by @danyx23. Adding `ds_beis: catalog.Dataset` is no big deal though, it improves readability. 

> If you agree with the changes, I'll add tests.

You should discuss this with @danyx23. You've both implemented something really valuable and similar, the difference being that you get error at runtime while @danyx23 at type checking time. Having it typed is nicer, on the other hand managing `typed_dag` file and making sure it is in sync with `dag` could be annoying. Both solutions solve the problem, make some compromise ;).

pabloarosado:
Thanks @Marigold yes, let's talk about it tomorrow during the tech tea.

larsyencken:
Just taking a look. At a high level, I wonder if it shouldn't look like:

```python
def run(dest_dir, deps):
    electricity_mix_dataset = deps["electricity_mix"]
    ...
```

larsyencken (commented):
See comments on overall PR

Marigold:
(@larsyencken that was my first idea, but I liked @pabloarosado solution more because it works for notebooks (well, which we don't use...) and we already use `Names` / `PathFinder` object in almost every step. Anyway, it's just a syntactic sugar, I don't have a strong opinion.)

larsyencken:
> (@larsyencken that was my first idea, but I liked @pabloarosado solution more because it works for notebooks (well, which we don't use...) and we already use `Names` / `PathFinder` object in almost every step. Anyway, it's just a syntactic sugar, I don't have a strong opinion.)

Does it work for notebooks though? They don't have `__file__`, so all solutions here look kinda the same: you need to initialise it manually with a string describing the path (for manual runs). For automatic runs, you override that with a parameter passed in via Papermill.

pabloarosado:
Hi @larsyencken, currently `__file__` has to be manually written within a jupyter notebook (which is very annoying). A workaround I often do is to create a `shared` module that contains `Path(__file__).parent`, which I can then import from the data step without having to manually write the path to the file.
So, having `run(dest_dir, deps)` would not make things much more complicated for notebooks. Therefore I also have no strong opinion on this.

But the idea of `deps["electricity_mix"]` feels a bit incomplete to me. For example:
* How do you import a local csv file that is next to the data step?
* How do you import two dependencies that have the same `short_name`, for example `shift/*/fossil_fuel_production` and `energy/*/fossil_fuel_production`. Also, at some point we may want to load two versions of the same dataset (for example, if we wanted to check how data changes over different versions, which is an interesting use of ETL).

In such cases, it's better to have more control on how to load dependencies. Regarding local files, `PathFinder.load_dependency` could also work to load local files (stored in the same folder as the current data step).

larsyencken:
I think it's quite possible to make either `Dependencies.__getitem__()` or your `load_dependencies()` method flexible enough -- e.g. consider that we're searching with suffixes, and that short names are just a special case of a suffix.

On side-loaded resources, like CSVs, recall that we have an [open correctness issue](https://github.com/owid/etl/issues/525) where you can use ingredients in your ETL step that the ETL doesn't know about. Since it doesn't know about them, they won't be hashed. Since they aren't hashed, you can edit these files and go crazy -- the ETL will never know and will happily not rebuild the dataset.

For example, you can edit `etl/steps/data/garden/energy/*/owid_energy_variable_mapping.csv` without triggering any rebuild of the energy datasets. We should at consider whether checked access to other resources would be worthwhile too.

Let's talk more tomorrow!

pabloarosado:
Hi @larsyencken should I go ahead with the current approach of this PR, or do you think we should implement the `run(dest_dir, deps)` one? I think that the current one is easier to adopt, as it's almost identical to the existing `Names` that we have been using so far. But if you think the alternative is better, I can work on it. Thanks!

etl/helpers.py
```diff
@@ -118,3 +151,110 @@ def walden_dataset(self) -> WaldenDataset:
     @property
     def snapshot_dir(self) -> Path:
         return paths.SNAPSHOTS_DIR / self.namespace / self.version
+
+    @staticmethod
+    def _create_step_name(
+        short_name: str,
+        channel: Optional[str] = None,
```
Marigold, line 158: Could be also 
```
from owid.catalog import CHANNEL
...
    channel: Optional[CHANNEL] = None
```

etl/helpers.py
```diff
@@ -3,17 +3,24 @@
...
+from owid.datautils.common import ExceptionFromDocstring
 from owid.walden import Catalog as WaldenCatalog
 from owid.walden import Dataset as WaldenDataset
 
 from etl import paths
+from etl.snapshot import Snapshot
+from etl.steps import load_dag
+
+# Load ETL dag.
+DAG = load_dag()
```
Marigold, line 23: Loading it on module level is ok I think, we're not making dynamic changes to DAG.

etl/helpers.py
```diff
@@ -118,3 +151,110 @@ def walden_dataset(self) -> WaldenDataset:
...
+
+        return dependency
+
+    def load_dependency(
+        self,
+        short_name: str,
+        channel: Optional[str] = None,
+        namespace: Optional[str] = None,
+        version: Optional[Union[str, int]] = None,
+    ) -> Union[catalog.Dataset, Snapshot, WaldenCatalog]:
```
Marigold, line 237: I'd drop support for Walden and its catalog. Only snapshots need to be supported.

etl/steps/data/garden/energy/2022-12-13/electricity_mix.py
```diff
@@ -10,14 +10,14 @@
 from owid.datautils.dataframes import combine_two_overlapping_dataframes
 from shared import CURRENT_DIR, add_population
 
-from etl.paths import DATA_DIR
+from etl.helpers import PathFinder
+
+# Get relevant paths for current file.
+paths = PathFinder(__file__)
```
Marigold, line 16: Definitely better than `Names`, though `PathFinder.load_dependency` sounds also strange. I can't think of better name though so `PathFinder` wins.

Marigold, line 16: Definitely better than `Names`, though `PathFinder.load_dependency` sounds also strange. I can't think of better name though so `PathFinder` wins.

Marigold (approved):
Looks good!

etl/helpers.py
pabloarosado: Let's keep it while it still exists. Once we get rid of Walden I'll remove all mentions here.

etl/steps/data/garden/energy/2022-12-13/electricity_mix.py
pabloarosado: I'm not a big fan of the naming either, but we can refactor later if we have a better idea.

pabloarosado:
Thanks @Marigold !

PR #687, opened 2022-12-16
sketch: outline what a typed dag could look like

danyx23:


danyx23:
Decided to not do this at this point and instead go with Pablo R's solution that does similar things at runtime

PR #688, opened 2022-12-19
enhance(snapshots): use Cloudflare cdn for accessing public snapshots

Marigold:
ETL users without AWS config (any config, not necessarily ours) couldn't pull snapshots and got weird DVC error. Make DVC pull public files from https://catalog.ourworldindata.org/snapshots instead of S3. Using CDN also speeds up access and reduce bandwidth costs.

### DIFF
--- a/.dvc/config
+++ b/.dvc/config
@@ -4,6 +4,9 @@
     url = s3://owid-catalog/snapshots
     endpointurl = https://nyc3.digitaloceanspaces.com
     acl = public-read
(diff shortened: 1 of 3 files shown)


marcelgerber (approved):
Can confirm that this now works well without AWS config ([as the original thread suggested](https://discuss.dvc.org/t/public-read-only-s3-remote/355/2)).
That's all I can vow for; whether this works fine for uploading files is not my area of expertise :)

PR #690, opened 2022-12-19
feature(walkthrough charts): new step

lucasrodes:
This PR adds a new step to `walkthrough`: charts.

### Context
Suppose a user inserts a new dataset into Grapher, which is an updated version of an old dataset in Grapher. In that case, they will have to update several charts using old variables. That is, we want that charts using variables from the old dataset to use variables from the new one instead.

This is where `walkthrough charts` becomes useful. Given the names of the old and new datasets, it helps create the chart revisions and submits them to our admin "Chart approval" tool.

Once we merge this PR, I'll prepare a short demo video to "walk through" this new step.

### Other minor updates
- Did minor refactoring of `utils` in `walkthrough`: Added new functions, moved functions from `walkthrough.grapher` in there.
- New function in `etl.db` to get all dataset names.
- Improved ergonomics of `etl.match_variables_from_two_versions_of_a_dataset`: e.g. changed module name to `etl.match_variables` and improved internal structure for a better re-usage in both UI and terminal use cases.

### Short intro (video)

https://user-images.githubusercontent.com/18101289/208677467-253b0a66-acce-4d8f-8f07-ff6935801d5c.mp4


etl/db.py
```diff
@@ -61,7 +61,7 @@ def open_db() -> Generator[DBUtils, None, None]:
             connection.close()
 
 
-def get_dataset_id(db_conn: MySQLdb.Connection, dataset_name: str) -> Any:
+def get_dataset_id(db_conn: Union[MySQLdb.Connection, None], dataset_name: str) -> Any:
```
larsyencken, line 64: `Union[X, None]` is better written as `Optional[X]`.

It's also weird to pass in `None` as an argument, better to move the `db_conn` argument to the end and default it to `None`, so that you can just call `get_dataset_id(dataset_name)` and it will work.

etl/db.py
```diff
@@ -79,19 +79,25 @@ def get_dataset_id(db_conn: MySQLdb.Connection, dataset_name: str) -> Any:
...
+
+        assert len(result) == 1, f"Ambiguous or unknown dataset name '{dataset_name}'"
+        dataset_id = result[0][0]
+        return dataset_id
 
-    return dataset_id
+    if db_conn is None:
+        with get_connection() as db_conn:
+            return _get_dataset_id(db_conn, dataset_name)
+    return _get_dataset_id(db_conn, dataset_name)
```
larsyencken, line 101: If you set `db_conn` in the first line, then you won't need an inner method at all. It will look like:

```python
if db_conn is None:
    db_conn = get_connection()

# do stuff with it below
```

etl/db.py
```diff
@@ -127,3 +133,30 @@ def get_variables_in_dataset(db_conn: MySQLdb.Connection, dataset_id: int, only_
         warnings.simplefilter("ignore", UserWarning)
         variables_data = pd.read_sql(query, con=db_conn)
     return variables_data
+
+
+def get_all_datasets(db_conn: Union[MySQLdb.Connection, None] = None, archived: bool = True) -> pd.DataFrame:
```
larsyencken, line 138: Likewise, make `db_conn` an `Optional[MySQLdb.Connection]`

etl/db.py
```diff
@@ -127,3 +133,30 @@ def get_variables_in_dataset(db_conn: MySQLdb.Connection, dataset_id: int, only_
...
+        query = " SELECT namespace, name, id FROM datasets"
+        if not archived:
+            query += " WHERE isArchived = 0"
+        datasets = pd.read_sql(query, con=db_conn)
+        return datasets.sort_values(["name", "namespace"])
+
+    if db_conn is None:
+        with get_connection() as db_conn:
+            return _get_all_dataset(db_conn)
+    return _get_all_dataset(db_conn)
```
larsyencken, line 162: Likewise, restructure this to avoid the need for an inner method.

etl/match_variables_from_two_versions_of_a_dataset.py
```diff
@@ -10,7 +10,7 @@
```
larsyencken, line 10: The module name is overly verbose, I'd shorten it to `match_variables.py`

walkthrough/utils.py
```diff
@@ -134,3 +136,107 @@ def generate_step(cookiecutter_path: Path, data: Dict[str, Any]) -> Path:
...
+                " consult the with ETL team."
+            )
+        )
+
+
+class OWIDEnvType(Enum):
+    LIVE = 1
+    STAGING = 2
+    LOCAL = 3
+    UNKNOWN = -1
```
larsyencken, line 183: It's not clear that using an enum with integer codes is better than strings for this type.

Would it be simpler if it was just a literal?

```python
OWIDEnvType = Literal["live", "staging", "local", "unknown"]
```

That would shorten a bunch of code, and it would still get type-checked properly.

walkthrough/utils.py
```diff
@@ -134,3 +136,107 @@ def generate_step(cookiecutter_path: Path, data: Dict[str, Any]) -> Path:
...
+
+
+class OWIDEnv:
+    env_type_id: Literal[OWIDEnvType.LIVE, OWIDEnvType.STAGING, OWIDEnvType.LOCAL, OWIDEnvType.UNKNOWN]
+
+    def __init__(
+        self,
+        env_type_id: Union[
+            Literal[OWIDEnvType.LIVE, OWIDEnvType.STAGING, OWIDEnvType.LOCAL, OWIDEnvType.UNKNOWN], None
```
larsyencken, line 192: This whole huge type string should just be `Optional[OWIDEnvType]`.

walkthrough/utils.py
```diff
@@ -134,3 +136,107 @@ def generate_step(cookiecutter_path: Path, data: Dict[str, Any]) -> Path:
...
+        ] = None,
+    ):
+        if env_type_id is None:
+            self.env_type_id = self.get_env_type_id_from_config()
+        else:
+            self.env_type_id = env_type_id
+
+    def get_env_type_id_from_config(
+        self,
+    ) -> Literal[OWIDEnvType.LIVE, OWIDEnvType.STAGING, OWIDEnvType.LOCAL, OWIDEnvType.UNKNOWN]:
```
larsyencken, line 202: Likewise, type should not be `Literal[...]` but instead `OWIDEnvType`.

This will work if `OWIDEnvType` is defined as a literal as suggested.

walkthrough/utils.py
```diff
@@ -134,3 +136,107 @@ def generate_step(cookiecutter_path: Path, data: Dict[str, Any]) -> Path:
...
+
+    @property
+    def env_type_label(self) -> str:
+        if self.env_type_id == OWIDEnvType.LIVE:
+            return "live"
+        elif self.env_type_id == OWIDEnvType.STAGING:
+            return "staging"
+        elif self.env_type_id == OWIDEnvType.LOCAL:
+            return "local"
+        return "unknown"
```
larsyencken, line 222: This whole method disappears if you use string literals

walkthrough/utils.py
```diff
@@ -134,3 +136,107 @@ def generate_step(cookiecutter_path: Path, data: Dict[str, Any]) -> Path:
...
+        env_type_id: Union[
+            Literal[OWIDEnvType.LIVE, OWIDEnvType.STAGING, OWIDEnvType.LOCAL, OWIDEnvType.UNKNOWN], None
+        ] = None,
+    ):
+        if env_type_id is None:
+            self.env_type_id = self.get_env_type_id_from_config()
+        else:
+            self.env_type_id = env_type_id
+
+    def get_env_type_id_from_config(
```
larsyencken, line 200: Would rename this `detect_env_type()`, since the config isn't listing the type, instead it's being inferred from magic values of config strings.

larsyencken (changes requested):
Looks cool! It would help review to have a quick example to run, to trial the interface, for now I mainly reviewed it for idiomatic Python style, rather than for overall functionality or correctness.

I would reorder `match_variables_from_two_versions_of_a_dataset.py` to follow normal linear module ordering, so that you can read it from top to bottom, high level to low level. That would mean making `main_cli()` the first method, then `main()`, then reordering from there to make it as linear to read as possible.

etl/db.py
lucasrodes: oh man... I made it so complex

lucasrodes:
Thanks, @larsyencken. Your suggestions made lots of sense; I updated the code accordingly.

What would be the best way to provide a quick example? A video walking you through a use case?

etl/db.py
```diff
@@ -61,24 +61,27 @@ def open_db() -> Generator[DBUtils, None, None]:
             connection.close()
 
 
-def get_dataset_id(db_conn: MySQLdb.Connection, dataset_name: str) -> Any:
+def get_dataset_id(dataset_name: str, db_conn: Optional[MySQLdb.Connection] = None) -> Any:
```
Marigold, line 64: You could also add these functions to `grapher_model.py` (our ORM) as methods of `Dataset` class. Working with datasets might be more idiomatic then. Just letting you know, it's not necessary and you can keep using functions if you find them nicer.

walkthrough/charts.py
```diff
@@ -0,0 +1,479 @@
...
+    SIMILARITY_NAMES,
+    find_mapping_suggestions,
+    preliminary_mapping,
+)
+
+from .utils import OWIDEnv, _check_env, _show_environment
+
+CURRENT_DIR = Path(__file__).parent
+
+ETL_DIR = Path(etl.__file__).parent.parent
```
Marigold, line 30: Nit - we usually use `from etl import paths; paths.BASE_DIR`, but up to you.

walkthrough/charts.py
```diff
@@ -0,0 +1,479 @@
...
+
+log = structlog.get_logger()
+
+
+def app(run_checks: bool, dummy_data: bool) -> None:
+
+    nav = Navigation(run_checks)
+
+    # show instructions
```
Marigold, line 39: Nicely refactored! I have to do the same for the rest of walkthrough.

Marigold (approved):
Code looks good (thanks for refactoring). It'd be more useful to get review from actual users on usability and features. I've never done any chart revisions, so it's hard to tell. I'm good with merging this and incrementally improving as we use it more.

etl/db.py
lucasrodes: yeah ー I think this might make sense. We can re-think what goes into `etl.db` and what to `etl.grapher_model`.

paarriagadap:
Hi! looks fantastic. I find very useful to show (and select) dataset IDs, which is maybe missing from the previous commands.
I wanted to try it on my end on staging, but it failed.
```
An internal error occurred in the application
×
  File "/home/parriagadap/etl/.venv/lib/python3.9/site-packages/pywebio/session/threadbased.py", line 86, in main_task
    target()
  File "/home/parriagadap/etl/walkthrough/cli.py", line 53, in <lambda>
    lambda: phase_func(run_checks=run_checks, dummy_data=dummy_data),
  File "/home/parriagadap/etl/walkthrough/charts.py", line 44, in app
    nav.get_input_params()
  File "/home/parriagadap/etl/walkthrough/charts.py", line 100, in get_input_params
    datasets_available = get_all_datasets(archived=False)
  File "/home/parriagadap/etl/etl/db.py", line 154, in get_all_datasets
    db_conn = get_connection()
  File "/home/parriagadap/etl/etl/db.py", line 22, in get_connection
    return MySQLdb.connect(
  File "/home/parriagadap/etl/.venv/lib/python3.9/site-packages/MySQLdb/__init__.py", line 123, in Connect
    return Connection(*args, **kwargs)
  File "/home/parriagadap/etl/.venv/lib/python3.9/site-packages/MySQLdb/connections.py", line 185, in __init__
    super().__init__(*args, **kwargs2)
MySQLdb.OperationalError: (2003, "Can't connect to MySQL server on '127.0.0.1:3309' (111)")
```

I copied my .env.staging setup to the .env file

PR #698, opened 2022-12-21
fix(packaging): unpin pathspec and upgrade dvc

Marigold:
The problem with pathspec and DVC seems to be fixed. Unpin pathspec and upgrade DVC.


Marigold:
(Had to merge this to unblock my work on datadiff).

PR #699, opened 2022-12-21
feat(compare): Let etl-compare specify version

pabloarosado:
The `compare etl-catalog` tool does not accept a `version` argument, which makes it unusable if there is more than one version of the dataset you want to compare with.

This PR adds an optional `version` argument.

I've also created [this issue](https://github.com/owid/owid-catalog-py/issues/74), given that, ideally, we should use `find_latest` (which does not exist) instead of `find_one`, and assume we always want to compare with the latest version of a dataset.

### DIFF
--- a/etl/compare.py
+++ b/etl/compare.py
@@ -3,7 +3,7 @@
 #
 
 from pathlib import Path
-from typing import Any, Dict, cast
+from typing import Any, Dict, Optional, Union, cast
 from urllib.parse import quote
 
 import click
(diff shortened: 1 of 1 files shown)


Marigold (approved):
(Haven't tested it, but I assume you did)

pabloarosado:
> (Haven't tested it, but I assume you did)

Thanks @Marigold I have tested it to compare a local `statistical_review` with a remote one (for which there are two versions).

PR #704, opened 2022-12-22
Use find latest in etl compare

pabloarosado:
The tool `compare etl-catalog` currently is only able to compare a local dataset with **the same version** of the same dataset in the remote catalog. However, often a user has fixed a small issue in one country/variable/year, creates a new fixed dataset, and wants to ensure that, with respect to the previous dataset remote version, only that specific country/variable/year has changed, and nothing else.

This PR uses `find_latest` instead of `find_one` in `compare etl-catalog` so that a user can compare a new version of a dataset with its previous version in the remote catalog. However, if a user wants to compare a specific version (and not the latest), the optional argument `--version` can also be used.

Caveat: A user may not be aware that they are comparing different versions of a dataset. It would be good to print "Comparing local version XXXX-XX-XX with remote version YYYY-YY-YY".

Ideally, the tool should let the user compare two arbitrary local and remote datasets. For example, you may want to compare datasets of different channels (e.g. what's the difference between garden and meadow?). But for now, this PR already provides some value.


lucasrodes (approved):
Looks good to me!

The only thing I would slightly change is the description of the command (docstring under `cli` function). That's the description that the user gets if they type `compare --help`.

I think that it is not generic enough, as it specifically talks about "dataframes".

pabloarosado:
> Lookgs good to me!
> 
> The only thing I would slightly change is the description of the command (docstring under `cli` function). That's the description that the user gets if they type `compare --help`.
> 
> I think that it is not generic enough, as it specifically talks about "dataframes".

Good point. I have changed it a bit, to be shorter and more generic. The previous docstring was actually pretty much the same as that for `dataframes` only.

PR #705, opened 2022-12-22
chore(datautils): bump to version 0.5.2

lucasrodes:
I was getting this error:

```
Because etl depends on walden (0.1.0) @ file:///Users/lucasrodes/repos/etl/vendor/walden which depends on owid-datautils (0.5.2) @ git+https://github.com/owid/owid-datautils-py.git@v0.5.2-alpha, owid-datautils is required.
So, because etl depends on owid-datautils (0.5.0) @ git+https://github.com/owid/owid-datautils-py.git@v0.5.1-alpha, version solving failed.
```

It seems that `etl` relies on `walden==0.1.0` and `datautils==0.5.1`. But `walden==0.1.0` relies on `datautils==0.5.2`. We might want to change etl dependencies so that it depends on `datautils==0.5.2`, to avoid such errors.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2419,14 +2419,14 @@ files = [
 
 [[package]]
 name = "ipywidgets"
-version = "8.0.3"
+version = "8.0.4"
 description = "Jupyter interactive widgets"
 category = "dev"
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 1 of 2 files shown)


lucasrodes:
accidentally modified other dependencies

PR #711, opened 2022-12-27
Relax dependency constraints

danyx23:
In walden, owid-catalog and owid-datautils-py I relaxed the version constraints because the fixed upper bounds were causing issues.

In this PR I am doing the same on ETL but here I think more discussion is warranted. Are there libraries that we want to fix to a more narrow range for now? I could imagine pandas and numpy to be two good candidates for this.

I also bumped the vendor versions of walden and owid-catalog-py. These are now commits on branches - does this cause an issue when we merge? We should probably merge the changes in those repos first, then update this PR and merge it then when it can point to master of both submodules again.


lucasrodes:
When trying to import one dataset to Grapher I get a `sqlalchemy.exc.NoForeignKeysError` error. This is the same error that is raised by the CI / build (see [logs](https://github.com/owid/etl/actions/runs/3788762508/jobs/6441866738)).

This seems to be because of differences in newer versions of `sqlmodel` library. I tried setting the dependency on this library to `sqlmodel = "^0.0.6"` instead and seems to work.

**Complete error logs:**
```
✗ etl --only grapher://grapher/owid/latest/key_indicators --grapher          
OK (0s)
Running 1 steps:
1. grapher://grapher/owid/latest/key_indicators...
2022-12-27 19:30.02 [info     ] upsert_dataset.verify_namespace namespace=owid
2022-12-27 19:30:02,280 - [bugsnag] WARNING - No API key configured, couldn't notify
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/relationships.py", line 2749, in _determine_joins
    self.primaryjoin = join_condition(
  File "<string>", line 2, in join_condition
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/sql/selectable.py", line 1229, in _join_condition
    raise exc.NoForeignKeysError(
sqlalchemy.exc.NoForeignKeysError: Can't find any foreign key relationships between 'posts' and 'tag'.

The above exception was the direct cause of the following exception:

  File "/home/lucas/repos/etl/.venv/bin/etl", line 6, in <module>
    sys.exit(main_cli())
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/lucas/repos/etl/etl/command.py", line 119, in main_cli
    main(**kwargs)  # type: ignore
  File "/home/lucas/repos/etl/etl/command.py", line 147, in main
    run_dag(
  File "/home/lucas/repos/etl/etl/command.py", line 245, in run_dag
    time_taken = timed_run(lambda: step.run())
  File "/home/lucas/repos/etl/etl/command.py", line 252, in timed_run
    f()
  File "/home/lucas/repos/etl/etl/command.py", line 245, in <lambda>
    time_taken = timed_run(lambda: step.run())
  File "/home/lucas/repos/etl/etl/steps/__init__.py", line 602, in run
    dataset_upsert_results = upsert_dataset(
  File "/home/lucas/repos/etl/etl/grapher_import.py", line 94, in upsert_dataset
    ns = gm.Namespace(name=namespace, description="")
  File "<string>", line 4, in __init__
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/state.py", line 476, in _initialize_instance
    manager.dispatch.init(self, args, kwargs)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/event/attr.py", line 334, in __call__
    fn(*args, **kw)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/mapper.py", line 3672, in _event_on_init
    instrumenting_mapper._check_configure()
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/mapper.py", line 1924, in _check_configure
    _configure_registries({self.registry}, cascade=True)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/mapper.py", line 3483, in _configure_registries
    _do_configure_registries(registries, cascade)
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/mapper.py", line 3522, in _do_configure_registries
    mapper._post_configure_properties()
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/mapper.py", line 1941, in _post_configure_properties
    prop.init()
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/interfaces.py", line 231, in init
    self.do_init()
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/relationships.py", line 2147, in do_init
    self._setup_join_conditions()
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/relationships.py", line 2243, in _setup_join_conditions
    self._join_condition = jc = JoinCondition(
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/relationships.py", line 2638, in __init__
    self._determine_joins()
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/orm/relationships.py", line 2771, in _determine_joins
    util.raise_(
  File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/sqlalchemy/util/compat.py", line 208, in raise_
    raise exception
sqlalchemy.exc.NoForeignKeysError: Could not determine join condition between parent/child tables on relationship Posts.tag - there are no foreign keys linking these tables.  Ensure that referencing columns are associated with a ForeignKey or ForeignKeyConstraint, or specify a 'primaryjoin' expression.
```

pyproject.toml
```diff
@@ -27,66 +27,66 @@ etl-metadata-export = 'etl.metadata_export:cli'
...
-PyPDF2 = "^2.11.1"
-dvc = {extras = ["s3"], version = "^2.38.1"}
-"ruamel.yaml" = "^0.17.21"
-owid-repack = "^0.1.1"
+thefuzz = ">=0.19.0"
+python-Levenshtein = ">=0.12.2"
+regex = ">=2022.1.18"
+wikipedia = ">=1.4.0"
+pandas = ">=1.4.0"
+numpy = ">=1.22.1"
```
pabloarosado, line 48: I don't have a strong opinion on other libraries, but I think I feel more comfortable having pandas constrained. We were even considering pinning it to a particular version. I think it's one that we can manually update every now and then, and be aware of changes.

pabloarosado (approved):
I don't have a strong opinion (except regarding pandas, as I mentioned in my comment). I'd like to know what Mojmir and Lars think.

pyproject.toml
Marigold: Yeah, we've agreed on this a while back and there's a plan to pin it to our current version (which is not 1.4.0, but 1.5.2 I think). We can either do it in this PR or I can do it later separately.

Marigold (approved):
Pinning `sqlmodel==0.0.6` should fix CI, we should also pin `pandas==1.5.2` (since bumping pandas automatically is dangerous).

pyproject.toml
larsyencken: My thinking has shifted a little on reflection.

So far, I don't think we have evidence that they've broken more than they've fixed with new releases.

Our only known problem arose because we began using different Pandas features, not because we upgraded unexpectedly.

Unless we do find evidence that they're doing a bad job upstream, I actually think we should be more bullish about tracking recent Pandas releases.

The 1.5.1 to 1.5.2 comparison has dragged out, since there was tons of spurious changes just because our ETL dataset had changed between the two runs, so now I'm having to run them both from scratch again to compare, which takes a while. I'll see how it goes and maybe update my perspective if we find something there.

Marigold: I still think it's better to do it in a controlled way by opening special PR and checking some kind of automatic validation [like datadiff](https://github.com/owid/etl/pull/694). They quite often deprecate features or add warnings with new versions (like the most recent [flood of warnings](https://github.com/owid/etl/pull/700)), so it's better to check beforehand. (I got also bitten by pandas upgrades in the past. They're doing better job for sure, but I'd still be careful given that everything we have depends on pandas).

> The 1.5.1 to 1.5.2 comparison has dragged out

(By the way we're already on 1.5.2, so it might be better to do 1.5.2 vs 1.5.3 which should be out soon)

Marigold: @larsyencken what's your opinion on pinning it to 1.5.2? We recently had [an issue](https://github.com/owid/etl/issues/733) caused by update from 1.5.1 -> 1.5.2 so I'd rather pin it to 1.5.2 and update it in a controlled fashion.

PR #713, opened 2022-12-27
data(karlinsky): add completion on death reporting

lucasrodes:
This PR adds the data on completion on death reporting, by A. Karlinsky. Further details can be found in https://github.com/owid/owid-issues/issues/815.


lucasrodes:
I am unsure why buildkite is failing. It complains about a call to `pd.read_csv`. couldn't recreate this locally.

pabloarosado:
> I am unsure why buildkite is failing. It complains about a call to `pd.read_csv`. couldn't recreate this locally.

I don't know either why this is failing. I tried locally:
```
with downloaded(PATH_DATASET) as filename:
    df = pd.read_csv(filename, sep=",")
```
And worked well. You can try directly `pd.read_csv(PATH_DATASET)`, and if that works, maybe it's an issue with `etl.helpers.downloaded`.

Also, I left [a comment](https://github.com/owid/owid-issues/issues/815#issuecomment-1366849332) in the original issue, suggesting to use snapshot instead of reading directly from github. We can continue the conversation there or here. And I'll continue the review once that's decided.

lucasrodes:
@pabloarosado Changed the pipeline to use `snapshot`-`meadow` instead of `etag`. Thanks for your feedback!

pabloarosado:
> @pabloarosado Changed the pipeline to use `snapshot`-`meadow` instead of `etag`. Thanks for your feedback!
> 
> The error with `pandas` disappeared when directly reading from the URL without using `etl.helpers.downloaded` 🤷‍♂️

Cool, then it seems to be an issue with `etl.helpers.downloaded`, so feel free to create a separate issue.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
+dataset:
+  sources:
+  #   -
+  #     name: Maddison Project Database 2020 (Bolt and van Zanden, 2020)
+  #     published_by: Bolt, Jutta and Jan Luiten van Zanden (2020), “Maddison style estimates of the evolution of the world economy. A new 2020 update“.
+  #     publisher_source: |
+  #       The Maddison Project Database is based on the work of many researchers that have produced estimates of...
+
```
pabloarosado, line 8: Please remove all this dummy metadata (again, this should be removed from walkthrough at some point).

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
...
+tables:
+  deaths_karlinsky:
```
pabloarosado, line 10: Why not just calling the table `deaths`. It seems unnecessary to include the source in the table's name.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
...
+    title: International Completeness of Death Registration 2015-2019 – Karlinsky (2021)
```
pabloarosado, line 11: Is it necessary to include the date range in the title?

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
...
+    description: >
+      Contains all metrics used in the International Completeness of Death Registration 2015-2019 paper.
+    variables:
+      # Life expectancy
+      reg_deaths:
+        title: Registered deaths
+        unit: "deaths"
+        description: >
+          The number of deaths registered in the vital registration system for the country-year. NA denotes no registered deaths data was obtained.
```
pabloarosado, line 20: Do we actually have verbose "NA" in the data? I guess we shouldn't. And if we don't, then this should be removed from th description.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
...
+      expected_wpp:
+        title: Expected number of deaths (WPP 2019)
```
pabloarosado, line 22: For the sake of future maintenance, I'd suggest removing years in the titles and descriptions of variables. Otherwise, you will need to manually update them. I think they are not necessary, but if they are, then keep them..

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
...
+        title: Expected number of deaths (GHE 2019)
+        unit: deaths
+        description: >
+          The number of deaths expected to occur for the country-year from Global Health Estimates 2019.
+
+      expected_deaths:
+        title: Expected number of deaths
+        unit: deaths
+        description: >
+          The mean of expected_wpp, expected_gbd and expected_ghe.
```
pabloarosado, line 43: Given that these descriptions will end up appearing in grapher (where the user does not see the variable short name, but the title) I'd suggest referring to variables by their title.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      death_comp:
+        title: Estimated completeness of death reporting
+        short_unit: "%"
+        unit: "%"
+        description: >
+          The estimated completness rate, ranges from 0 to 100. Derived as reg_deaths divided by expected_deaths. See paper (https://www.medrxiv.org/content/10.1101/2021.08.12.21261978v1) for details.
```
pabloarosado, line 50: Refer to variables by their title.

snapshots/health/2022-12-28/karlinsky.py
```diff
@@ -0,0 +1,26 @@
+"""This script has been autogenerated by `walkthrough snapshot`."""
```
pabloarosado, line 1: I'd suggest having a more meaningful docstring, describing this snapshot.

snapshots/health/2022-12-28/karlinsky.py
```diff
@@ -0,0 +1,26 @@
...
+
+CURRENT_DIR = pathlib.Path(__file__).parent
+
+
+@click.command()
+@click.option(
+    "--upload/--skip-upload",
+    default=True,
+    type=bool,
+    help="Upload dataset to Walden",
```
pabloarosado, line 17: This should be "Upload snapshot". This neds to be changed in walkthrough.

etl/steps/data/meadow/health/2022-12-28/karlinsky.py
```diff
@@ -0,0 +1,40 @@
...
+    log.info("karlinsky.start")
+
+    # retrieve snapshot
+    snap = Snapshot(SNAPSHOT_DATASET)
+    df = pd.read_csv(snap.path)
+
+    # create new dataset and reuse walden metadata
+    ds = Dataset.create_empty(dest_dir, metadata=convert_snapshot_metadata(snap.metadata))
+    ds.metadata.version = MEADOW_VERSION
+    ds.metadata.date_accessed = dt.date.today().strftime("%Y-%m-%d")
```
pabloarosado, line 29: This should be a fixed date (otherwise it'll change next time the step is run).

etl/steps/data/garden/health/2022-12-28/karlinsky.py
```diff
@@ -0,0 +1,109 @@
...
+from etl.helpers import Names
+from etl.paths import DATA_DIR
+
+log = get_logger()
+
+# naming conventions
+N = Names(__file__)
+MEADOW_DATASET = "meadow/health/2022-12-28/karlinsky"
+TABLE_NAME = "deaths_karlinsky"
+# PATH_DATASET = "https://raw.githubusercontent.com/akarlinsky/death_registration/main/death_reg_final.csv"
```
pabloarosado, line 15: This comment can be removed.

lucasrodes:
@pabloarosado it works for me locally, though. I couldn't recreate it yet.

etl/steps/data/garden/health/2022-12-28/karlinsky.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # ensure absolute values make sense (positive, lower than population)
+    columns_absolute = [col for col in df.columns if col not in columns_perc]
+    df_ = df.reset_index()
+    df_ = geo.add_population_to_dataframe(df_)
+    for col in columns_absolute:
+        x = df_.dropna(subset=[col])
+        assert all(
+            x[col] < 0.2 * x["population"]
+        ), f"{col} contains values that might be too large (comapred to population values)!"
```
pabloarosado, line 109: Typo "comapred"

etl/steps/data/grapher/health/2022-12-28/karlinsky.py
```diff
@@ -0,0 +1,23 @@
...
+
+N = Names(__file__)
+
+
+def run(dest_dir: str) -> None:
+    dataset = catalog.Dataset.create_empty(dest_dir, N.garden_dataset.metadata)
+
+    table = N.garden_dataset["deaths_karlinsky"]
+
+    # optionally set additional dimensions
```
pabloarosado, line 13: Remove unnecessary comments.

pabloarosado (approved):
Looks good, I made some suggestion. The only thing that definitely needs to be changed is the date accessed being dynamically calculated.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
lucasrodes: This is relevant, as the estimations by WPP, GBD and GHE can differ substantially between revisions. The user should know which version is being used. Even more, in this case, where COVID had not yet happened.

Also, we will need to create the metadata file again when we revise the version. With newer versions new columns might even appear. I think this is precisely one of the reasons why `snapshot`->`meadow` makes more sense vs. `etag`

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
lucasrodes: We have NaNs, as these are present ins some of the variables. I think this is fine and also expected, as some country-years are missing data.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
lucasrodes: Yes, that's the title of the dataset and paper:

- repository: https://github.com/akarlinsky/death_registration
- paper: https://www.medrxiv.org/content/10.1101/2021.08.12.21261978v2

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
pabloarosado: Sure, I'm just wondering if they will be seen as "NA", as the description says. I guess they will not, they will simply be missing. If so, then it makes more sense to say that some country-years may not have data because there were no registered deaths.

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
lucasrodes: Makes sense!

lucasrodes:
Thanks, pablo!

etl/steps/data/garden/health/2022-12-28/karlinsky.meta.yml
lucasrodes: I see! That makes sense. To some extent this is the metadata for Garden, so I think talking about `NaN`s is fine. However, this is also what we currently show in Grapher, where `NaN`s are not visible. I think your description is compatible with both channels, thanks!

PR #718, opened 2022-12-28
enhance(walkthrough): typos, use data_helpers

lucasrodes:
This PR solves a typo in `walkthrough meadow` step, which wrongly had the help text as "Upload dataset to Walden" instead of "Upload dataset to Snapshot".

Also, changed python script template for Garden, so it uses `etl.data_helpers.geo` instead of `owid.datautils.geo`.


pabloarosado (approved):
Looks good. Thanks for doing this!

PR #720, opened 2022-12-29
refactor: Remove old, unnecessary FAOSTAT steps from dag

pabloarosado:
Every now and then we have issues with some of the old FAOSTAT steps, which used to be necessary for the global food explorer, but now they are not used anywhere.

This PR removes those old steps from the dag.


lucasrodes (approved):
LGTM!

Issue #721, opened 2022-12-29
etl-chart-suggester fails when trying to write a long 'suggestedReason'


pabloarosado:
When trying to launch the chart revision tool for the statistical review dataset on staging (to move charts from [this old dataset](https://staging.owid.cloud/admin/datasets/5666) to [this new one](https://staging.owid.cloud/admin/datasets/5837)), I get the following error:
```
)
2022-12-29 16:38.39 [error    ] Rolling back changes...
2022-12-29 16:38.39 [error    ] INSERT operation into `suggested_chart_revisions` cancelled. Error: (1406, "Data too long for column 'suggestedReason' at row 1")
2022-12-29 16:38.41 [info     ] 0 of 27 suggested chart revisions inserted.
2022-12-29 16:38.41 [error    ] (1406, "Data too long for column 'suggestedReason' at row 1")
```
It happens for certain variables, not all of them. I think it happens because it attempts to write a very long 'suggestedReason', which is a text that may include the names of many variables, and SQL complains because it's too long. But I haven't looked into it, so the issue may be something else.

lucasrodes:
Can you provide the command line so I can recreate this error?

I recently changed how `suggestedReason` was created, making it potentially longer on average. I was not aware of this "Data too long" type error.

pabloarosado:
I launched `walkthrough charts`, which failed. Then I copied the mapping, and executed the old tool
```
ENV=.env.staging etl-chart-suggester /Users/prosado/Desktop/mapping.json
```
With all variables, I was getting the error. But with subsets of variables, this didn't happen.

lucasrodes:
This is likely because of how we currently create the suggested chart revision reason in https://github.com/owid/etl/blob/36aeeb1c5234baa41becd17244ad1581e6698dc8/etl/chart_revision_suggester.py#L166.

If the generated string is too long (>512 characters), it exceeds the maximum variable length given by the SQL field `suggestedReason` datatype.

Working on a fix.

PR #722, opened 2022-12-30
enhance(walkthrough charts)

lucasrodes:
* Fixes https://github.com/owid/etl/issues/721
* Uses `from rich_click.rich_command.RichCommand` in `etl-chart-suggester`.
* Slight improvement on modularity of `etl.chart_revision_suggester`. Now `walkthrough charts` displays a processing bar instead of a loading circle.

### DIFF
--- a/etl/chart_revision_suggester.py
+++ b/etl/chart_revision_suggester.py
@@ -1,5 +1,4 @@
-"""
-Generate chart revisions in Grapher using MAPPING_FILE JSON file. 
+"""Generate chart revisions in Grapher using MAPPING_FILE JSON file.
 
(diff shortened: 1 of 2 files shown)


lucasrodes:
@pabloarosado would be nice if you could try out this version and see if #721 is now fixed.

I don't expect a code quality review. Module `etl.chart_revision_suggester` should soon undergo a major refactor.

pabloarosado (approved):
Thanks a lot @lucasrodes ! Yes, I've just run it on the same dataset and it works, and looks very user-friendly. I haven't checked the code, though, but you can merge for now, so we can keep using it while improving the tool.

PR #723, opened 2022-12-30
Remove misleading bp regions

pabloarosado:
This PR fixes [this issue](https://github.com/owid/owid-issues/issues/837), and, given that BP Statistical Review dataset needed to be updated (which involves updating many other datasets), I decided to refactor the energy data steps, to make them more consistent. Therefore, now they all share the same functions (in the "shared" module), and some other minor issues have been fixed.

Most files are copy/pasted from their previous versions. They have simply been adapted to more recent ETL changes. So there's no need to review functions in detail. It would be good to review:
* The dag, to check that the new versions make sense.
* Quickly check the metadata files (just dates and names, not detailed revision of variables and descriptions).
* Quickly check the `run` function of the data steps.


Marigold (approved):
I used [datadiff in CI](https://buildkite.com/our-world-in-data/etl-mojmir-ci-diff-test/builds/36#0185724d-a95b-4eb3-a74b-da6d37eda7bf) to do the comparison automatically. You can check it for yourself, but in summary:

- some tables lost metadata (which is totally ok, we don't use them anyway)
- data shape has changed (because we removed countries which is expected)
- some renaming happened

That means everything looks good!

PR #727, opened 2023-01-02
fix: explode dimensions in deterministic order

### LINKED ISSUE #726: SDG grapher step generates a non-deterministic short-name
## Problem

As part of a comparison of Pandas 1.5.1 and 1.5.2 output for the ETL, it appears that many tables in the `data/grapher/un_sdg/2022-07-07/un_sdg` dataset get minor renames like below.

Before: `data/grapher/un_sdg/2022-07-07/un_sdg/_10_3_1__vc_vov_gdsd__no_breakdown_by_disability__age__all_areas__both_sexes.feather`

After: `data/grapher/un_sdg/2022-07-07/un_sdg/_10_3_1__vc_vov_gdsd__all_areas__no_breakdown_by_disability__age__both_sexes.feather`

My interpretation is that the table names are being generated by exploding dimensions, but the explosion is not done in a deterministic order.

## Impact

It's not a big deal for our usage in Grapher, but it makes QA on them more difficult, and it means the short names aren't stable handles for these variables at the moment.

larsyencken:
Fixes #726, where the name of exploded dimensional variables is unstable for the SDG grapher dataset.

### DIFF
--- a/etl/steps/data/grapher/un_sdg/2022-07-07/un_sdg.py
+++ b/etl/steps/data/grapher/un_sdg/2022-07-07/un_sdg.py
@@ -130,7 +130,7 @@ def create_dataframe_with_variable_name(dataset: Dataset, tab: str) -> pd.DataFr
     cols_meta = ["indicator", "seriesdescription", "seriescode"]
     cols = ["indicator", "seriescode"]
     if tab_df.shape[1] > 11:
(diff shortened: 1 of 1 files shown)


Marigold (approved):
I'm pretty sure that fixes it. I couldn't find a different reason.

PR #730, opened 2023-01-03
enhance(etl): improve ETL performance

Marigold:
* stop running `walden tests` now that we use snapshots (takes ~1.5min in CI)
* migrate `un/2022-07-11/un_wpp` to snapshot (speeds up certain `etl` runs by the factor of 10x)

(better to wait for `datadiff` to make sure un_wpp hasn't changed)


pabloarosado (approved):
Looks good, that speed up will be appreciated!

PR #731, opened 2023-01-03
enhance(fasttrack): add fields for snapshot metadata

Marigold:
Add couple fields to the template that will be saved in Snapshot metadata. Fix couple of minor bugs.


pabloarosado (approved):
I think it would be good at some point to add more documentation to this code. Apart from that, I think it's good to go.

PR #732, opened 2023-01-03
Ingest energy investments data

pabloarosado:
Ingest energy investments data by UNEP. The source data is a table (an image) in a pdf file. I manually extracted the data and created a text file at the meadow step.


etl/steps/data/meadow/unep/2023-01-03/renewable_energy_investments.py
```diff
@@ -0,0 +1,53 @@
...
+import pandas as pd
+from owid.catalog import Dataset, Table
+
+from etl.helpers import PathFinder
+from etl.steps.data.converters import convert_snapshot_metadata
+
+# naming conventions
+paths = PathFinder(__file__)
+
+EXTRACTED_DATA_FILE = paths.directory / "extracted_data.csv"
```
Marigold, line 16: If you ever change the data, the step won't be rebuilt because it doesn't use `extracted_data.csv` to compute checksum. It has to have prefix `renewable_energy_investments*` or you have to put both files to `renewable_energy_investments/` folder.

pabloarosado: Thanks. I tried calling it `renewable_energy_investments_data.csv`, but that was not recognised. However, `renewable_energy_investments.data.csv` does work. So it has to be `name_of_step.something`, right?

etl/steps/data/garden/unep/2023-01-03/renewable_energy_investments.py
```diff
@@ -0,0 +1,33 @@
...
+# Get paths and naming conventions.
+paths = PathFinder(__file__)
+
+
+def run(dest_dir: str) -> None:
+    #
+    # Load data.
+    #
+    # Load dataset from meadow.
+    ds_meadow: catalog.Dataset = paths.load_dependency("renewable_energy_investments")
```
Marigold, line 18: So... pretty...

Marigold (approved):
Lovely.

PR #740, opened 2023-01-05
Include pandas version in checksum

larsyencken:
This PR includes the Pandas version in the checksum for any dataset that gets created.

If the Pandas version gets updated for the ETL, then it will force CI to rebuild the datasets, giving us an opportunity to notice build problems and also data differences.


larsyencken:
@Marigold This is a simple version of what we talked about. I'm not sure whether or not you prefer it to a global value that we can increment, but at least it would catch Pandas issues.

Marigold (approved):
Nice, I was originally thinking about global value, but this is more elegant and less work for us.

PR #741, opened 2023-01-05
Homicide datasets used to create long-run homicide OMM

spoonerf:
I'm not sure if `snapshots/homicide/2023-01-03/WHOMortalityDatabase_Deaths_sex_age_a_country_area_year-Violence_22nd December 2022 14_53.csv` should be included in here. It is necessary to manually download this data but not sure if it should be in here?


Marigold:
> I'm not sure if snapshots/homicide/2023-01-03/WHOMortalityDatabase_Deaths_sex_age_a_country_area_year-Violence_22nd December 2022 14_53.csv should be included in here. It is necessary to manually download this data but not sure if it should be in here?

It's better not to commit it, it's enough if you just delete it (or download it to a different location than `etl/` in the script). Scripts look good.

(Getting my foot in the door - I've recently added `etl-datadiff` command that is run automatically as a check for your PR. You can get there through `buildkite/etl-build-public-datasets-branch -> Details -> Datadiff` or go [directly there](https://buildkite.com/our-world-in-data/etl-build-public-datasets-branch/builds/1019#018582d0-eb32-440e-8577-83a54305a850). It shows that some of the new tables have `index` columns which probably shouldn't be there)

image

spoonerf:
> > I'm not sure if snapshots/homicide/2023-01-03/WHOMortalityDatabase_Deaths_sex_age_a_country_area_year-Violence_22nd December 2022 14_53.csv should be included in here. It is necessary to manually download this data but not sure if it should be in here?
> 
> It's better not to commit it, it's enough if you just delete it (or download it to a different location than `etl/` in the script). Scripts look good.
> 
> (Getting my foot in the door - I've recently added `etl-datadiff` command that is run automatically as a check for your PR. You can get there through `buildkite/etl-build-public-datasets-branch -> Details -> Datadiff` or go [directly there](https://buildkite.com/our-world-in-data/etl-build-public-datasets-branch/builds/1019#018582d0-eb32-440e-8577-83a54305a850). It shows that some of the new tables have `index` columns which probably shouldn't be there)
> 
> image

Ah thanks @Marigold! I'm not sure if I have been dealing with this error in the right way: `ValueError: feather does not support serializing a non-default index for the index; you can .reset_index() to make the index into column(s)`. 

I was using `reset.index()` but then this introduces an index column, which I now then drop. Is there a better way to handle this?

Marigold:
> I was using reset.index() but then this introduces an index column, which I now then drop. Is there a better way to handle this?

That's correct. You could use `.reset_index(drop=True)` to make it slightly shorter.

Marigold (approved):
Looks good! I'm wondering about how much boilerplate code is needed for this and whether it's fine with walkthrough or if we should come up with more concise way of doing these things. Let me know if this bothers you.

PR #742, opened 2023-01-05
feature(step): add new static:// step to add static files to catalog

Marigold:
We've [recently discussed](https://owid.slack.com/archives/C0193RW5E2J/p1672916983670529) what's the easiest way how to get CSV file into MySQL through ETL. Here's a [python script](https://github.com/owid/notebooks/tree/main/EdouardMathieu/gravitational_waves) that generates the data

There are a few ways how to currently do it:

1. **CSV import** in admin - that's gonna get the data into MySQL, but not to ETL
2. **Fast-track** - copy your data, fill metadata and then import it through fast-track (sample [spreadsheet](https://docs.google.com/spreadsheets/d/1NKoZMe6bkXMS29mORNw3o25cENx3MnpYFJk2y3cWXOc/edit?usp=sharing))
3. **Walkthrough** - go through all walkthrough steps and create snapshot & meadow & garden & grapher
4. **Code** - if you know ETL, you should be able to create Snapshot and grapher step

I've tried all of them. Honestly, I understand why people are still using CSV import. Other than that, **fast-track** was the most straightforward to use, although creating a google sheet just for the sake of this small dataset and having my script push data there (or copy&paste manually every time I make a change) is a bit annoying.

Using walkthrough is an overkill for such a dataset and putting together code from the top of my head was hard, even for me! (it took me ~30 mins with debugging typos and such) I guess that only power-users are able to use `etl` at the moment.

[This commit](https://github.com/owid/etl/pull/742/commits/f564ff26688c47e77cda7aaa457384f849691ac7) shows how complex is it to do it at the moment just with code.

I've also experimented with a new approach - let **users generate csv however they like, add it to DVC (= keep it locally, but don't commit to git) and add metadata in YAML**.

In more detail - users would generate csv file using whatever tools they like (python, R, stata) and **add it to S3** with DVC (so it's kind of a snapshot). Then we add a special `static://` step that loads the snapshot, **add metadata** from YAML file and save it as a dataset (that will live in a catalog and could be sent to grapher DB). Here are the steps I made (all in [this commit](https://github.com/owid/etl/pull/742/commits/a134ece23b6f82af6fba055b55f19a3da4e7b2c9)):

1. Create folder `etl/steps/data/grapher/technology/2023-01-06`
2. Copy `script.py` there that generates file `gravitational_waves.csv` (note the script doesn't use ETL at all and it could be R script as well)
3. Run `dvc add etl/steps/data/grapher/technology/2023-01-06/gravitational_waves.csv` to automatically move the file to S3 and only leave the placeholder `gravitational_waves.csv.dvc` in git
4. Add file `gravitational_waves.meta.yml` with metadata
5. Add line `static://grapher/technology/2023-01-06/gravitational_waves.csv:` to DAG
6. Run `etl grapher/technology/2023-01-06/gravitational_waves.csv` to build it (or with `--grapher` flag to upload it to grapher DB)

### How is this `static://` from running my script as part of snapshot?

It's pretty similar. The `script.py` could be run as part of snapshot and we'd have just a very simple `grapher` step that would copy snapshot and add metadata. We would need to fill in snapshot metadata and it would make using non-python scripts more challenging.

### DIFF
--- a/.dvcignore
+++ b/.dvcignore
@@ -8,4 +8,11 @@ tests/
 vendor/
 .cachedir/
 .venv/
+.github
+__pycache__/
+
 !data/snapshots/
+
+# temporary for static step
+!etl/steps/data/**/*.csv
+!etl/steps/data/**/*.csv.dvc
--- a/dag/main.yml
+++ b/dag/main.yml
@@ -292,6 +292,10 @@ steps:
   data://grapher/fasttrack/2023-01-03/long_term_homicide_rates_in_europe:
     - snapshot://fasttrack/2023-01-03/long_term_homicide_rates_in_europe.csv
 
+  # Gravitational waves
+  data://grapher/technology/2023-01-05/gravitational_waves:
+    - snapshot://technology/2023-01-05/gravitational_waves.csv
+
 # Sub-dags to import
 include:
   - dag/open_numbers.yml
--- a/dag/main.yml
+++ b/dag/main.yml
@@ -296,6 +296,9 @@ steps:
   data://grapher/technology/2023-01-05/gravitational_waves:
     - snapshot://technology/2023-01-05/gravitational_waves.csv
 
+  # Gravitational waves with the new static step
+  static://grapher/technology/2023-01-06/gravitational_waves.csv:
+
 # Sub-dags to import
 include:
   - dag/open_numbers.yml
--- a/etl/steps/__init__.py
+++ b/etl/steps/__init__.py
@@ -25,6 +25,7 @@ import structlog
 import yaml
 from dvc.dvcfile import Dvcfile
 from dvc.repo import Repo
+from owid.catalog import Dataset, Table
 
 # smother deprecation warnings by papermill
 with warnings.catch_warnings():
@@ -229,6 +230,9 @@ def parse_step(step_name: str, dag: Dict[str, Any]) -> "Step":
 
     elif step_type == "backport":
         step = BackportStep(path, dependencies)
+
+    elif step_type == "static":
+        step = StaticStep(path)
 
     elif step_type == "data-private":
         step = DataStepPrivate(path, dependencies)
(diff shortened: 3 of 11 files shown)


danyx23:
This looks like a nice addition for cases like the one @paarriagadap recently described, right? I.e. you are already comfortable with working in the ETL and just want a faster way of dealing with a single csv file but inside the current ETL system.

This would not be so nice for the case that Ed had I think - it might seem like a small thing but I do think that this still adds quite a bit of friction vs other solutions like for the authors. For that group I think a very similar step that just pulls things from github would be idea - i.e. the authors would add the scripts and the metadata and data (!) to github in the notebooks repo and we would have a new kind of step that pulls that in and puts it into the catalog and grapher. I'm not saying this PR has to become that, I'm just saying it's still valuable to do that thing too :). What do you think?

Marigold:
Oh, thanks for feedback @danyx23! I wasn't expecting any yet, I'm still thinking this through.

> This would not be so nice for the case that Ed had I think - it might seem like a small thing but I do think that this still adds quite a bit of friction vs other solutions like for the authors. For that group I think a very similar step that just pulls things from github would be idea - i.e. the authors would add the scripts and the metadata and data (!) to github in the notebooks repo and we would have a new kind of step that pulls that in and puts it into the catalog and grapher. I'm not saying this PR has to become that, I'm just saying it's still valuable to do that thing too :). What do you think?

The `static://` step is pretty much the same thing, but it lives in ETL and not a different repository. Check out [this commit](https://github.com/owid/etl/pull/742/commits/a134ece23b6f82af6fba055b55f19a3da4e7b2c9):
* there's [a script](https://github.com/owid/etl/pull/742/commits/a134ece23b6f82af6fba055b55f19a3da4e7b2c9#diff-483cee3d31ae30bfc02475f4b4651d3ca47bff1f2a3751bddaac04e1e33613b5) that generates CSV file (it would be the same in `notebooks`)
* there's [metadata](https://github.com/owid/etl/pull/742/commits/a134ece23b6f82af6fba055b55f19a3da4e7b2c9#diff-8470b34e2a9145083ee3e147e3dd73712ac004a3ac12581ebbf2965507cfdfa9) (again, same would be in `notebooks`)

The question is whether you want to commit your data or not. If you don't, you can run `dvc add ...` and that's it. That could be done from both `etl` or `notebooks` repo. If you want to commit it, you can do it here too. So the only extra thing that has to be done in ETL is adding line `static://grapher/technology/2023-01-06/gravitational_waves.csv:` to DAG. You **don't need working ETL** to run all this (you just need DVC installed) and you don't need any knowledge about ETL. Once we have automated deploys you'd just push your changes and it'd materialise in catalog and grapher.

(Of course, having all this in a different repository isn't technically challenging, it's just that this sounds simpler)

Marigold:
This is again relevant, keeping it open.

Marigold:
Closing this, we already have ETL and fast-track that work quite well. This is somewhere in between, but there's no need to learn another way of creating datasets.

PR #746, opened 2023-01-06
enhance(fast-track): add commits to fast-track and improve UX

Marigold:
Updates to fast-track:
* commit and push new datasets
* include `dag/fasttrack.yml` in the main.yml DAG (so that we can use fast-track datasets as dependencies)
* automatic `black` for generates files
* report errors to bugsnag
* smarter default options based on selected sheet


fasttrack/cli.py
```diff
@@ -327,6 +363,16 @@ def _load_data_and_meta(
...
+        except urllib.error.HTTPError:
+            _bail(
+                [
+                    sheets.ValidationError(
+                        "Sheet not found, have you copied the template? Creating new Google Sheets document or new "
+                        "sheets with the same name in the existing document does not work."
+                    )
+                ]
+            )
+            continue
```
lucasrodes, line 375: It would be nice if we could add a URL-validator. So if the user inputs a URL that does not work in the first input field, an error message box is shown to them with sth like "The URL appears to be invalid. Please check!".

lucasrodes (approved):
LGTM! Just a minor comment, no need to address it now.

fasttrack/cli.py
Marigold: That's a good idea, putting it on my list of improvements.

PR #747, opened 2023-01-08
chore(fuzzywuzzy): use thefuzz, fix levenshtein

lucasrodes:
- Removes dependency on `fuzzywuzzy`:
  - This library is no longer maintained and has been replaced by `thefuzz`. See https://github.com/seatgeek/fuzzywuzzy for more details.
  - We were already using `thefuzz` in some instances.
-  Pins `python-Levenstein` to a specific version. Otherwise, `thefuzz` usage will fail, because of a known bug: See https://github.com/seatgeek/thefuzz/issues/43, https://github.com/maxbachmann/python-Levenshtein/issues/1, https://github.com/seatgeek/thefuzz/pull/10, 

I noticed this error when trying to run `walkthrough charts`. It would not work, because it relies on `thefuzz`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ Unidecode = ">=1.3.2"
 python-dotenv = ">=0.19.0"
 frictionless = {extras = ["pandas"], version = ">=4.40.8"}
 thefuzz = ">=0.19.0"
+# pinned because of https://github.com/seatgeek/thefuzz/issues/43
 python-Levenshtein = "^0.12.2"
 regex = ">=2022.1.18"
 wikipedia = ">=1.4.0"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -40,7 +40,7 @@ Unidecode = ">=1.3.2"
 python-dotenv = ">=0.19.0"
 frictionless = {extras = ["pandas"], version = ">=4.40.8"}
 thefuzz = ">=0.19.0"
-python-Levenshtein = ">=0.12.2"
+python-Levenshtein = "^0.12.2"
```
Marigold, line 43: Could be useful to add a note explaining why is it pinned and link the issue. This will make it easier to unpin in the future.

Marigold (approved):
Looks good, `datadiff` doesn't show any changes either.

pyproject.toml
lucasrodes: Good point!

PR #754, opened 2023-01-09
data(snapshot): add conflict datasets from spreadsheets

lucasrodes:
This PR adds several conflict datasets to Snapshot. To this end, I have created a new namespace: `conflicts`.

These datasets were previously manipulated and saved in [Google Drive](https://docs.google.com/spreadsheets/d/1lh4Ra3TaLt10ZRL4-C16--X1Zu4atofQO5iLwSBuOEY/edit#gid=1175422923). Here, we download these files from Google Drive and port them to Snapshots.

I have added [technical notes](https://docs.google.com/document/d/1aoBrW5UXkFNCZVM18teD9nYWcQH82_y5-uqeEH9c30E/edit#) to the corresponding dataset metadata.

### DIFF
--- a/snapshots/conflicts/2023-01-09/bouthoul_carrere_1978.csv.dvc
+++ b/snapshots/conflicts/2023-01-09/bouthoul_carrere_1978.csv.dvc
@@ -1,0 +1,70 @@
+meta:
+  namespace: conflicts
+  short_name: bouthoul_carrere
+  name: A list of the 366 major armed conflicts of the period 1740 - 1974
+  source_name: Bouthoul and Carrere (1978)
+  publication_year: 1978
+  publication_date: 1978-07-01
+  url: https://www.jstor.org/stable/23609587
+  source_data_url: https://docs.google.com/spreadsheets/d/1JXkz05kb_gkGXuvdnI21B6lhHIVZJD5BaiEOZi8mqRk/edit?usp=sharing
+  file_extension: csv
+  license_url: https://about.jstor.org/terms/
+  license_name: JSTOR
+  date_accessed: 2023-01-09
+  is_public: true
+  description: >-
+    What years and countries are covered? 1740–1974, worldwide
+
+      - "List of all major armed conflicts – domestic and international – for the
+    period 1740 - 1974" (Bouthoul et al. 1978: 83)
+
+
+    Which conflicts are covered?
+
+      - Major armed conflicts, domestic and international.
+
+      - "Only those conflicts were included which satisfied at least one of the following
+    six quantitative criteria and could 
+      therefore be considered as 'major armed conflicts', wars and revolutions being
+    the most representative ones." (Bouthoul et al. 1978: 83).
+
+      - 'Major': more than one state, affected a large area, lasted longer than one
+    year, killed more than 1,000, or had important results,
+      such as regime change or independence; "Those six criteria were: a. undertaken
+    by more than one state, b. Affected a space larger 
+      than one province or a capital, c. lasted longer than one year, d. Was of great
+    intensity (more than 1,000 killed)[...], e. Had important 
+      internal results (secession or regime change), f. Had important international
+    results (annexation or independence, emergence or 
+      disappearance of a state" (Bouthoul et al. 1978: 83).
+
+      - "The list thus includes the following types of conflicts: foreign and civil
+    wars, occupations by force (e.g. Austria, 1938), 
+      military penetrations, revolutions, uprisings and insurrections, massacres and
+    genocide, and other violent troubles with important 
+      consequences." (Bouthoul et al. 1978: 83-84).
+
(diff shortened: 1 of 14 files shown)


pabloarosado (approved):
Hi @lucasrodes , I have a few comments/questions:
1. I'd suggest taking the `SNAPSHOT_VERSION` from `Path(__file__).parent.name`, to avoid hardcoding versions.
2. Couldn't you just use one script, called, e.g. "conflicts", which has a list of all `short_names` to process (assuming they all share the same version)?
3. I don't understand what file you are manually uploading each time (using `path-to-file`). I thought all data was in google documents.

Apart from that, it looks good.

lucasrodes:
@pabloarosado, thanks for your feedback.

> 1. I'd suggest taking the `SNAPSHOT_VERSION` from `Path(__file__).parent.name`, to avoid hardcoding versions.
The `SNAPSHOT_VERSION` value was written by cookiecutter based on the parameters inputted during the `walkthrough` step. I think this is fine. I don't see a clear win here.


> 2. Couldn't you just use one script, called, e.g. "conflicts", which has a list of all `short_names` to process (assuming they all share the same version)?
Yes, that's an option. I preferred keeping them separate in case (i) versions for specific datasets change or (ii) we want to update specific datasets.

> 3. I don't understand what file you are manually uploading each time (using `path-to-file`). I thought all data was in google documents.

For each script, you will a URL to the file to be downloaded in the docstring. For instance, `sutton_1971.py` needs file at https://docs.google.com/spreadsheets/d/1Xo5e8aiFJnmv_aasFXU8MtzizSU0LxiysoUuErEJx3Q/edit?usp=sharing.

pabloarosado:
> @pabloarosado, thanks for your feedback.
> 
> > 1. I'd suggest taking the `SNAPSHOT_VERSION` from `Path(__file__).parent.name`, to avoid hardcoding versions.
> 
> The `SNAPSHOT_VERSION` value was written by cookiecutter based on the parameters inputted during the `walkthrough` step. I think this is fine. I don't see a clear win here.

I think we should be moving in the direction of not having hardcoded versions in data steps. But as I mentioned in [the other PR](https://github.com/owid/etl/pull/756), we can look into this at another time.

> 
> > 2. Couldn't you just use one script, called, e.g. "conflicts", which has a list of all `short_names` to process (assuming they all share the same version)?
> 
> Yes, that's an option. I preferred keeping them separate in case (i) versions for specific datasets change or (ii) we want to update specific datasets.
> 
OK!

> > 3. I don't understand what file you are manually uploading each time (using `path-to-file`). I thought all data was in google documents.
> 
> For each script, you will a URL to the file to be downloaded in the docstring. For instance, `sutton_1971.py` needs file at https://docs.google.com/spreadsheets/d/1Xo5e8aiFJnmv_aasFXU8MtzizSU0LxiysoUuErEJx3Q/edit?usp=sharing.

I'm not sure I understood. So you have to manually download a file from google drive, and then input it via `--path-to-file`, is that right? If so, I think these instructions are not very clear in the docstring. And also, wouldn't it be possible to fetch the file directly from google drive? If you did it this way just for simplicity, maybe mention it in the docstring.

On a separate note, are the google drive files expected to be temporary, or to be maintained? I just checked the [sutton_1971](https://docs.google.com/spreadsheets/d/1Xo5e8aiFJnmv_aasFXU8MtzizSU0LxiysoUuErEJx3Q/edit?usp=sharing) file and its permissions are set to "Anybody with the link can edit". I supposed those permissions should be changed to read only.

lucasrodes:
> I'm not sure I understood. So you have to manually download a file from google drive, and then input it via --path-to-file, is that right? If so, I think these instructions are not very clear in the docstring. And also, wouldn't it be possible to fetch the file directly from google drive? If you did it this way just for simplicity, maybe mention it in the docstring.

I tried making it more obvious in the docstrings, please let me know what you think. 

We could automatically download the files from Google Drive. I tried using `https://docs.google.com/spreadsheets/d/{key}/gviz/tq?tqx=out:csv` style link to download the file, but doesn't seem to work (some cells disappear). 

Instead, we could use `owid-datautils` in `Snapshot. download_from_source `, so that it could download google drive files, too.

https://github.com/owid/etl/blob/428066476d44fe0012334578816be38bf04e8482/etl/snapshot.py#L64-L68

> On a separate note, are the google drive files expected to be temporary, or to be maintained? I just checked the [sutton_1971](https://docs.google.com/spreadsheets/d/1Xo5e8aiFJnmv_aasFXU8MtzizSU0LxiysoUuErEJx3Q/edit?usp=sharing) file and its permissions are set to "Anybody with the link can edit". I supposed those permissions should be changed to read only.

Thanks for the catch. I changed the permissions!

pabloarosado:
OK, if you think we will have to update these datasets often (more than once per year?), then it would be good if they could fetch data directly from google. Otherwise, a manual download is fine. Up to you!

PR #756, opened 2023-01-09
enhance(walkthrough,snapshot): add option for manual dataset import

lucasrodes:
Sometimes we have to import a dataset into Snapshot manually. This is when the raw dataset file can't be automatically downloaded but needs to be obtained manually. Then we would run something like:

```
python snapshots/<namespace>/<version>/<short_name>.py ーpath-to-file <path-to-local-file>
```

This PR adds a checkbox option in `walkthrough` for this specific scenario, as the base script changes a bit depending on whether the dataset is manually imported.


pabloarosado:
Hi @lucasrodes, I was thinking of changing the default snapshot script so that by default it ingests a list of files, instead of a single file. The same would apply to files manually ingested (in which case, instead of `path_to_file`, the argument would be a list of files). Do you think it's a good idea, and would you like to implement it on this PR? Otherwise I'll do a separate one after you are done with this.

pabloarosado (approved):
I left two comments, feel free to ignore them, apart from that it looks good.

lucasrodes:
@pabloarosado  thanks for your feedabck.



> 1. Instead of hardcoding the snapshot version in the generated script, I'd define at the beginning `SNAPSHOT_VERSION = Path(__file__).parent.name`.

I don't see why this would be preferred over using `cookiecutter`'s parameters.

> 2. I was thinking of changing the default snapshot script so that by default it ingests a list of files, instead of a single file. The same would apply to files manually ingested (in which case, instead of `path_to_file`, the argument would be a list of files).

That looks cool. I would leave this for another PR, so we can merge this one and keep improving. I may not have the bandwidth to work on this specific enhancement. I think we would need to re-think how the `walkthrough snapshot` step works, too (e.g. we should allow the user to input several file info).

pabloarosado:
> @pabloarosado thanks for your feedabck.
> 
> > 1. Instead of hardcoding the snapshot version in the generated script, I'd define at the beginning `SNAPSHOT_VERSION = Path(__file__).parent.name`.
> 
> I don't see why this would be preferred over using `cookiecutter`'s parameters.

If the version is hardcoded and you want to create a new step from an existing one (without using walkthrough) you would need to copy/paste the existing step, and then manually change the version. But don't worry, we can do this in a different PR.

> 
> > 2. I was thinking of changing the default snapshot script so that by default it ingests a list of files, instead of a single file. The same would apply to files manually ingested (in which case, instead of `path_to_file`, the argument would be a list of files).
> 
> That looks cool. I would leave this for another PR, so we can merge this and keep improving. I may not have the capacity to work on this specific enhancement. For the improvements that you propose, besides changing the template, we would need to re-think how the `walkthrough snapshot` step works, too, as we should allow the user to input several file info (at the moment we only allow one file).

OK!

Marigold (approved):
Nice one. I've been thinking about snapshots lately too. There's a lot of boilerplate code in most of ingest scripts while everything we do is adding file or downloading file from URL. That could be almost one-liner (with docs). On the other hand, being explicit is good and code is generated by walkthrough anyway. Does it bother you in any way @pabloarosado or @lucasrodes?

lucasrodes:
@Marigold thanks for your comment. It does not bother me at all. I do think though that we could reduce the number of lines in the future, so we have something like:

```
add_snapshots(snapshots=snapshot, upload=csv, local_file=local_file) 
```

However, I would first wait until we improve how we support multiple snapshot files (as @pabloarosado mentioned)

PR #760, opened 2023-01-10
fix(backport): add missing number of workers to backport

Marigold:
There's an option to set workers, but it wasn't applied in `ThreadPool` and backporting was using maximum number of workers.

### DIFF
--- a/backport/backport.py
+++ b/backport/backport.py
@@ -1,5 +1,5 @@
 import datetime as dt
-from typing import List
+from typing import List, Optional
 
 import click
 import pandas as pd
@@ -19,8 +19,6 @@ from . import utils
 config.enable_bugsnag()
 
 log = structlog.get_logger()
-
-engine = get_engine()
 
 
 @click.command()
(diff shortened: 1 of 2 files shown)


pabloarosado (approved):

larsyencken (approved):
Nice.

I wasn't sure off-the-bat if SQLAlchemy engines were thread safe, but [apparently they are](https://davidcaron.dev/sqlalchemy-multiple-threads-and-processes/#:~:text=Every%20pool%20implementation%20in%20SQLAlchemy,will%20also%20be%20thread%2Dsafe.). I may even become comfortable using thread pools myself eventually 😅

Marigold:
> I wasn't sure off-the-bat if SQLAlchemy engines were thread safe, but [apparently they are](https://davidcaron.dev/sqlalchemy-multiple-threads-and-processes/#:~:text=Every%20pool%20implementation%20in%20SQLAlchemy,will%20also%20be%20thread%2Dsafe.). I may even become comfortable using thread pools myself eventually 😅

Yeah, they're, but I got burned by creating separate engines in multiple threads. Better to not be as smart about dynamically creating them and instead passing them as arguments.

PR #761, opened 2023-01-10
refactor: Use geo from etl.data_helpers instead of owid.datautils

pabloarosado:
This PR removes (almost) all uses of the `geo` module of `owid.datautils`, to use the one from `etl.data_helpers` instead. I checked that several important datasets did not change. I haven't checked all datasets, but I expect buildkite will (thanks @Marigold !).

All changes in this PR are trivial (importing `geo` from `etl.data_helpers` instead of `owid.datautils`) except in one file, namely `etl.data_helpers.geo` itself. I had to make a couple of amendments here to make it work.

There are still 2 places where we are using the old `owid.datautils.geo`, namely:
* `tests/data_helpers/test_geo`: When using the one from `data_helpers`, tests fail. I will look into this in a different PR.
* `key_indicators` dataset: This is a bit tricky. To build table `land_area` we are using `owid.datautils.geo.add_region_aggregates`, which depends on dataset `key_indicators`. The solution would be to create a new simple function for `add_regions` in `table_land_area`. Additionally, to construct `population` dataset in `demography`, we are also using `add_region_aggregates`. But this function, again, uses `key_indicators`. This dependency is not explicitly stated in the dag (when doing so, ETL complains because there is a circular dependency, which indeed is the case). Given how important these datasets are, let's solve these issues in a different PR.


Marigold (approved):
Nice! Getting closer to removing population from datautils...

lucasrodes:
On a separate note,
I think you can easily fix the problem you mention in:

> tests/data_helpers/test_geo: When using the one from data_helpers, tests fail. I will look into this in a different PR.

It seems to come from the mock definition

https://github.com/owid/etl/blob/e27763bd07e58585dde96d75cb2625133a99d16f/tests/data_helpers/test_geo.py#L47

which is no longer valid.

I tried changing it to

```python
@patch.object(catalog.catalogs.CatalogMixin, "__getitem__", mock_population_load)
```

but did not work either.

lucasrodes (approved):

pabloarosado:
Thanks for the comments @lucasrodes:
* A circular dependency in ETL would break things. All recipes should work from scratch, and a circular dependency makes that impossible. The reason why it works now is because either a dependency is loaded (by datautils) from the remote catalog (where the dataset already exists) or because the dependency already exists on your local catalog.
* In the long term, we should remove `key_indicators`, since it's not an informative name, and we also have many other key indicators elsewhere. We should instead use demography/population directly. But that would need to be done carefully, given that it's probably the most referenced dataset.
* For the time being, the easiest solution would be to simply not use `add_region_aggregates` anywhere in `key_indicators`, `land_use`, `population` (is there any other use?) and instead use a custom function created for that particular step.
* I also tried solving the issue with tests as you said, but the assertions were failing, so it may need a bit more thought (hence a different PR).

PR #764, opened 2023-01-12
fix(requirements): pin frictionless to 4.x.x version

Marigold:
Frictionless got recently upgraded to 5.x.x version which broke nightly build. Pinning it to 4.x.x for now (it's not clear whether it's a bug in 5.x.x or if data format of open numbers supports only 4.x.x)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1827,22 +1827,20 @@ files = [
 
 [[package]]
 name = "frictionless"
-version = "5.0.4"
+version = "4.40.8"
 description = "Data management framework for Python that provides functionality to describe, extract, validate, and transform tabular data"
 category = "main"
 optional = false
 python-versions = "*"
 files = [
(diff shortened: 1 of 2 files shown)


larsyencken (approved):
I see it, confirmed the error `Field is not valid: 'type' is a required property` and that this PR fixes it.

This is a good example that would have been caught in CI if the checksum of `poetry.lock` had been used in the `DataStep` checksum.

Since `frictionless` is a key library for open numbers, shall we add it to the input checksum?

It would mean something like this in `DataStep.checksum_input()`:

```
if 'open_numbers' in self.path:
    checksums['__frictionless__'] = frictionless.__version__
```

Marigold:
@larsyencken up to you, I'm a bit scared of rebuilding our entire catalog on minor version updates (well, we could extract only major versions...). We found the error in the nightly build, got alerted and solved this relatively fast. It could have been faster, but our monitoring works well enough I think.

larsyencken:
No worries. Let's leave things as they are for now then.

PR #765, opened 2023-01-12
Kyoto cherry blossom dataset - annual data and multi-year averages

spoonerf:
The data is a little patchy in the first few hundred years so the multi-year averages look a bit clunky. Is there a better (and simple) way of having a smoother line through the data? 

Draft chart on staging [here](https://staging.owid.cloud/admin/charts/6298/edit).

Maybe a GAM or spline or would work - but also want something that can be easily explained. Any ideas or comments are appreciated!


spoonerf:
I've decided to go with a 20 year average but with the requirement that at least five of the years in the 20 year window have to have data. This omits most of the problematic years at the beginning.

dag/walkthrough.yml
```diff
@@ -25,4 +25,4 @@ steps:
   data://garden/ihme_gbd/2020-12-19/child_mortality:
   - data://meadow/ihme_gbd/2020-12-19/child_mortality
   data://grapher/ihme_gbd/2020-12-19/child_mortality:
-  - data://garden/ihme_gbd/2020-12-19/child_mortality
+  - data://garden/ihme_gbd/2020-12-19/child_mortality
```
pabloarosado, line 28: Maybe not for this PR, but should we create a separate dag file for `health` (or some other appropriate name) and move all these steps from `dag_walkthrough`?

etl/steps/data/garden/biodiversity/2023-01-11/cherry_blossom.py
```diff
@@ -0,0 +1,55 @@
...
+from owid.catalog import Dataset, Table
+from structlog import get_logger
+
+from etl.helpers import Names
+from etl.paths import DATA_DIR
+
+log = get_logger()
+
+# naming conventions
+N = Names(__file__)
```
pabloarosado, line 11: Instead of `Names` you could use `PathFinder`, but you can just leave it for the next update if you prefer.

etl/steps/data/meadow/biodiversity/2023-01-11/cherry_blossom.py
```diff
@@ -0,0 +1,71 @@
...
+    return df.rename(columns={"Entity": "country", "Year": "year"}).drop(
+        columns=["Source code", "Data type code", "Reference Name", "Unnamed: 7"]
+    )
+
+
+def convert_date(df: pd.DataFrame) -> pd.DataFrame:
+    """
+    The full flowering date is formated like MDD, we should change this to day of the year for better biological meaning
+     - Zero pad the year so it is 4 digits long
+     - Zero pad the full-flowering date so it is 4 digits long
```
pabloarosado, line 60: I don't understand this without looking at the data, maybe it would be good to give an example.

snapshots/biodiversity/2023-01-11/cherry_blossom.py
```diff
@@ -0,0 +1,30 @@
+"""This script has been autogenerated by `walkthrough snapshot`."""
```
pabloarosado, line 1: Could you explain here in the docstring the instructions to obtain the data file?

pabloarosado (approved):
It looks good (I just made minor comments). Without looking at the data, I'm not sure what smoothing procedure would be best, but a moving average is a meaningful choice.

pabloarosado:
Oh, I just saw [the chart on staging](https://staging.owid.cloud/admin/charts/6298/edit), I'm wondering if we could have dates on the y axis, somehow. Otherwise, it looks nice.

dag/walkthrough.yml
spoonerf: Good idea - I can do a separate PR for this

PR #766, opened 2023-01-12
Add published by field to snapshot

pabloarosado:
This PR adds `published_by` field to snapshot, and refactors accordingly. It also adapted dummy data files to the recent changes in walkthrough.

I added a dummy data file to use in walkthrough `--dummy-data`, to avoid downloading a real external data file every time.
NOTE: Running the snapshot step will not work until merged (since I used a url for `dummy_data.csv` in the master branch).


Marigold (approved):
Looks good.

PR #768, opened 2023-01-13
feature(linter): add yaml schemas and lint metadata files in VSCode

Marigold:
The first commit contains [jsonschema](https://json-schema.org/) files with schemas for snapshot and dataset metadata in YAML format. If you use VSCode, you can install [this YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) and add the following lines to your settings.json (hit `Command + Shift + P` and select `Preferences: Open Folder Settings (JSON)`)

```
    "yaml.schemas": {
        "schemas/snapshot-schema.json": "**/*.dvc",
        "schemas/dataset-schema.json": "**/*.meta.yml"
    }
```

This will enable **autocompletion and linting** on your YAML files and give you nice warnings if something's not right.

image

I went through all our YAML files looking for linting errors and fixing them. The most common mistake was including  
```
dataset:
  publication_year: ...
  publication_date: ...
```
class `DatasetMeta` doesn't have these fields and they get thrown away. I've also found a couple of minor issues with labeling and fixed them. It helps discover typos and autocomplete is also quite helpful.


etl/steps/data/garden/bp/2022-07-11/statistical_review.meta.yml
```diff
@@ -3,7 +3,6 @@ dataset:
   version: 2022-07-11
   title: Statistical Review of World Energy - BP (2021)
   short_name: statistical_review
-  publication_year: 2021
   sources:
     -
       name: Our World in Data based on BP Statistical Review of World Energy (2021)
```
pabloarosado, line 8: Would it be possible to impose that dates must always be in the format "YYYY-MM-DD", to avoid things like "8th July 2021" here?

etl/steps/data/garden/health/2022-12-28/deaths_karlinsky.meta.yml
```diff
@@ -1,5 +1,4 @@
-dataset:
-  sources:
+dataset: {}
```
pabloarosado, line 1: Hmm, is this the official way to say: "Nothing new to add in dataset metadata"?

etl/steps/data/garden/hmd/2022-11-04/life_tables.meta.yml
```diff
@@ -1,5 +1,5 @@
 dataset:
-  sources:
+  sources: []
```
pabloarosado, line 2: Is this the official way to say "Nothing new in metadata". I find it a bit ugly that for datasets it's {}, while for sources it's []. I suppose this is YAML complexity in action.

pabloarosado (approved):
Thanks a lot for all the cleaning. This is going to help a lot having more homogeneous metadata files!

etl/steps/data/garden/hmd/2022-11-04/life_tables.meta.yml
Marigold: It's {} for `dataset` because it contains keys and [] for sources because it's a list. I'm not sure why `sources` was there, I removed them and kept only `dataset: {}`. We could make `dataset:` optional and remove it too (and fix function for loading metadata). I don't mind either.

etl/steps/data/garden/health/2022-12-28/deaths_karlinsky.meta.yml
Marigold: Same as https://github.com/owid/etl/pull/768#discussion_r1070967733, perhaps it'd be better to have `dataset` optional so that we don't have to write empty fields in yaml?

PR #769, opened 2023-01-13
fix(geo): fix circular dependency of population dataset

Marigold:
Merging [geo refactoring](https://github.com/owid/etl/pull/761#issuecomment-1378946768) introduced circular dependency of population dataset on itself. Step `data://garden/demography/2022-12-08/population` calls function `add_region_aggregates` calls  `list_countries_in_region_that_must_have_data` which tries to load population table from `data://garden/owid/latest/key_indicators` (that doesn't exist yet). Since we know population in `demography/2022-12-08/population`, the easiest fix is to pass it explicitly as a parameter for region calculation.

There might be a better solution, this is just a hotfix. @pabloarosado feel free to suggest a better one.

In addition, I've also added `data://garden/owid/latest/key_indicators` as dependencies to steps that use `geo.add_population_to_dataframe` function.

(To replicate the bug you can run:)
1. `rm -Rf data/garden/owid/latest/key_indicators` 
2. `etl data://garden/health/2022-12-28/deaths_karlinsky`

### DIFF
--- a/dag/demography.yml
+++ b/dag/demography.yml
@@ -4,6 +4,7 @@ steps:
     - data://garden/papers/2022-11-04/riley_2005
     - data://garden/un/2022-07-11/un_wpp
     - data://garden/hmd/2022-11-04/life_tables
+    - data://garden/owid/latest/key_indicators
 
   data://grapher/demography/2022-11-30/life_expectancy:
     - data://garden/demography/2022-11-30/life_expectancy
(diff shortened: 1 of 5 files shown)


pabloarosado (approved):
Thanks. As a temporary solution, if this avoids breaks, it's good to go. But soon we should simply rethink and simplify these steps.

lucasrodes:
Nice. I think could change the values of `DATASET_KEY_INDICATORS` in `etl.data_helpers.geo`.


```
DATASET_KEY_INDICATORS = DATA_DIR / "garden" / "demography" / "2022-12-08" / "population"
TNAME_KEY_INDICATORS = "population"
```

This, however would mean that we should revisit the DAG and change the dependencies wherever we have `key_indicators` with `population`. This might be something to discuss in the future and a separate PR, as it is unclear at the moment what the future of `key_indicators` is.

PR #772, opened 2023-01-16
enhance(fast-track): minor updates

Marigold:
* remove `deepdiff` from datadiff and replace it with `difflib`
* better error for missing sources in fast-track


lucasrodes (approved):
LGTM!

PR #775, opened 2023-01-16
refactor(dag): moving all health related steps into dag/health.yml

spoonerf:
There were a bunch of health related steps in walkthrough.yml so this was mostly to move them somewhere sensible but then it seemed like a good idea to have all health related steps in the same place.


lucasrodes (approved):
LGTM!

Perhaps we can move this too:

```
data://garden/owid/latest/covid:
    - etag://raw.githubusercontent.com/owid/covid-19-data/master/public/data/owid-covid-data.csv
```

PR #777, opened 2023-01-17
884 data update near surface temperature met office hadley centre

pabloarosado:
I updated this dataset (which didn't really need an update) just to adapt it to the new walkthrough workflow and to see if everything works well. Nothing has changed with respect to the previous steps, so there's no need to review the steps.

Everything went well, except for one small issue: Initially, in the garden metadata, I only had:
```yaml
dataset:
  title: Near-surface temperature anomaly (Met Office Hadley Centre, 2022b)
```
Given that all metadata was well defined since the snapshot step. However, on grapher admin I saw that the datset description (which shows in the SOURCES tab) was duplicated. That's why I had to include `description: ""`. I think this happens because, when transferring the ETL grapher step to grapher, the sources' description are appended to the dataset description.

I think in general it's good to concatenate dataset description and sources' description, except in this kind of scenarios. The only downside is that it creates an extra line at the beginning of the description. @Marigold do you think there's a better solution? Thanks.

### DIFF
--- a/etl/steps/data/garden/met_office_hadley_centre/2023-01-17/near_surface_temperature.py
+++ b/etl/steps/data/garden/met_office_hadley_centre/2023-01-17/near_surface_temperature.py
@@ -1,0 +1,52 @@
+"""Load a meadow dataset and create a garden dataset."""
+
+import pandas as pd
+from owid.catalog import Dataset, Table
+from structlog import get_logger
+
+from etl.data_helpers import geo
+from etl.helpers import PathFinder
+
+log = get_logger()
+
+# Get paths and naming conventions for current step.
+paths = PathFinder(__file__)
+
+
+def run(dest_dir: str) -> None:
+    log.info("near_surface_temperature.start")
+
+    #
+    # Load inputs.
+    #
+    # Load meadow dataset.
+    ds_meadow: Dataset = paths.load_dependency("near_surface_temperature")
+
+    # Read table from meadow dataset.
+    tb_meadow = ds_meadow["near_surface_temperature"]
+
+    # Create a dataframe with data from the table.
+    df = pd.DataFrame(tb_meadow)
+
+    #
+    # Process data.
+    #
+    # Create a new table with the processed data.
+    tb_garden = Table(df, like=tb_meadow)
+
+    #
+    # Save outputs.
+    #
+    # Create a new garden dataset with the same metadata as the meadow dataset.
+    ds_garden = Dataset.create_empty(dest_dir, metadata=ds_meadow.metadata)
+
+    # Add table of processed data to the new dataset.
+    ds_garden.add(tb_garden)
+
+    # Update dataset and table metadata using the adjacent yaml file.
+    ds_garden.update_metadata(paths.metadata_path)
+
+    # Save changes in the new garden dataset.
+    ds_garden.save()
+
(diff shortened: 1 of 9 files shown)


etl/steps/data/garden/met_office_hadley_centre/2023-01-17/near_surface_temperature.py
```diff
@@ -0,0 +1,51 @@
+"""Load a meadow dataset and create a garden dataset."""
```
Marigold, line 1: I'm getting different values in both versions, here's an example

```
ds = Dataset(DATA_DIR / "garden/met_office_hadley_centre/2023-01-17/near_surface_temperature")
print(ds['near_surface_temperature'].loc['Southern hemisphere'].loc[2022])

ds = Dataset(DATA_DIR / "garden/met_office_hadley_centre/2023-01-02/near_surface_temperature")
print(ds['near_surface_temperature'].loc['Southern hemisphere'].loc[2022])
```

etl/steps/data/meadow/met_office_hadley_centre/2023-01-17/near_surface_temperature.py
```diff
@@ -0,0 +1,77 @@
...
+    df = df.set_index(["region", "year"], verify_integrity=True).sort_index()
+
+    # Create a new table and ensure all columns are snake-case.
+    tb = Table(df, short_name=paths.short_name, underscore=True)
+
+    #
+    # Save outputs.
+    #
+    # Create a new meadow dataset with the same metadata as the snapshot.
+    ds_meadow = Dataset.create_empty(dest_dir, metadata=convert_snapshot_metadata(snapshots["Global"].metadata))
```
Marigold, line 66: This inherits short name `near_surface_temperature**_global**` from the snapshot. The step itself is called `near_surface_temperature` which causes confusion. I wonder if we should make it mandatory that `namespace/version/short_name` matches path of a dataset.

Marigold (approved):
Datadiff showed that new version has different values. Would you mind checking it first before merging?

Marigold:
> I think in general it's good to concatenate dataset description and sources' description, except in this kind of scenarios. The only downside is that it creates an extra line at the beginning of the description. @Marigold do you think there's a better solution? Thanks.

For this simple case we can only add the description if it is different from source's description. I'll add that in a different PR. (using `description: ""` is totally fine for this PR)

etl/steps/data/garden/met_office_hadley_centre/2023-01-17/near_surface_temperature.py
pabloarosado: Thanks for pointing this out, the data for northern and southern hemisphere has changed significantly, I don't know why... I'm checking.

pabloarosado: Ouch, it was my mistake. It happened when I manually copied the snapshot metadata of the northern and southern files... I also accidentally copied the URLs. Thanks a lot for pointing this out!

pabloarosado: The new data now is identical, except for the latest data point, which has changed a bit. But this is expected.

Issue #759, opened 2023-01-10
Project: Bake Grapher style data.json files for most columns in ETL


danyx23:
In 2022 we tried a few ways of serving data from ETL in a way the grapher admin can use it (python api, node api as part of the grapher admin). None of these were really satisfactory. 

What we currently do is that we have Grapher steps at the end of etl that upsert etl columns to the grapher database. In this process, rows are created in the `datasets`, `variables` and `sources` tables so that the metadata of these entities is registered in the DB. At the same time, the data values are upserted into the data_values table. This table is by far the largest part of the grapher MySQL DB, blowing a DB that would be a few hundred MB to over 70 GB. In light of the upcoming work on provisioning staging servers it would be very desirable if we could kill data_values for good and make it much faster to get a full copy of our database running in another instance.

To kill data_values, the data for all backported or fully ETL datasets (at the grapher level of ETL processing steps) has to be transformed into data.json files. The point where we know best what needs to be regenerated is the ETL which track this information on a dataset level granularity. 

This project is then about creating a new export step that could be part of the ETL but it could also be a separate downstream process, that makes the data of a column available as data.json files and stores them in S3. Every variable (=ETL column) in a dataset in a grapher step would have it's own file.

For the admin, the metadata fetching should still happen as before via the grapher admin api that queries Mysql directly. For baking, the metadata would similarly be consumed from mysql directly.

Open questions:

- [ ] Should the S3 bucket be exposed to the public and cached via cloudflare? If no then the baking step would involve copying the data.json files for the used variables to netlify. If yes then we would serve data files from S3 via cloudflare.
- [ ] Can we bake Garden level variables as well? And if so do we want to? This is more of a question of what to use for ids in this case (variable ids are only generated in the grapher step) and if we can assemble enough metadata for these variables from the ETL alone and what to do if there is a data model mismatch (i.e. if the ETL Garden variable has more dimensions than year + entity)
- [ ] Should separate summary statistics be generated as well? These are currently part of the metadata of a variable but for creating them at baking time would mean that the whole data would have to be fetched again which we'd probably like to avoid. At ETL time we will probably have the data loaded into a pandas dataframe already and can just fetch e.g. the set of entities and the time range.

Marigold:
Let me first review what we've done until now

### Hosting static files on S3 instead of Netlify to reduce its bandwidth (and save on Netlify plan)

```mermaid
graph LR

baking --> |sync| S3 -.-> website
baking --> bakedSite/ -.-> website
```


[...]

PR #778, opened 2023-01-17
feature(grapher_import): replicate `data_values` as [variable_id].json files on S3

Marigold:
Add CLI `backport-datasync` for replicating `data_values` table from MySQL to `s3://owid-catalog/baked-variables/[db_name]/data/[variable_id].json` files ([an example](https://catalog.ourworldindata.org/baked-variables/staging_grapher/data/525715.json)) with structure that can be consumed by the grapher, i.e.

```
{"years": [...], "entities": [...], "values": [...]}
```

Once variable is replicated, we point its `dataPath` column in MySQL to the file. In theory we wouldn't need extra column and we could get `dataPath` dynamically from env variables, but keeping it as variable attribute comes in handy when you start working with staging environments.

We also create metadata.json files ([an example](https://catalog.ourworldindata.org/baked-variables/staging_grapher/metadata/525715.json)), but these are not used by grapher yet.

Only datasets with at least one variable having a chart are replicated at the moment (the plan is to sync everything).

This is intended to run as [continuous service](https://github.com/owid/ops/pull/46), but in the long run we should be creating those JSON files when running `etl --grapher`.

More details in [this Concept](https://github.com/owid/etl/issues/759) or [PR with motivation for using `dataPath`](https://github.com/owid/owid-grapher/pull/1867).


Marigold:
@pabloarosado I should have said how should this be reviewed. You could take a look at staging and production databases at new columns `dataPath` and `metadataPath` and check if those files exist and are in a reasonable format (I've already checked parity with what we have in production). Then you could check if the logic for updating these fields is ok (if you can think of any race conditions for instance). You don't have to run the service locally (although you could run it against staging for certain dataset if you want, i.e. `ENV=.env.staging backport-datasync -d 2405 --dry-run --force`)

backport/datasync/data_metadata.py
```diff
@@ -0,0 +1,190 @@
...
+    try:
+        return values.map(int)
+    except ValueError:
+        pass
+    # perhaps they're all floats
+    try:
+        return values.map(float)
+    except ValueError:
+        pass
+    # otherwise return them as strings
```
pabloarosado, line 185: This function doesn't look very robust, but if you have tested it, I hope it works in practice.

backport/datasync/datasync.py
```diff
@@ -0,0 +1,335 @@
...
+from etl.db import get_engine
+from etl.publish import connect_s3
+
+log = structlog.get_logger()
+
+config.enable_bugsnag()
+
+
+S3_BUCKET = "owid-catalog"
+S3_PREFIX = f"baked-variables/{config.DB_NAME}"
```
pabloarosado, line 40: Minor suggestion: Shouldn't we call it simply `variables`? The name `baked-variables` sounds a bit like they have been manipulated.

backport/datasync/datasync.py
```diff
@@ -0,0 +1,335 @@
...
+        where = """
+        (
+            -- must be used in at least one chart
+            id in (
+                select distinct v.datasetId from chart_dimensions as cd
+                join variables as v on cd.variableId = v.id
+            )
+        )
+        -- we would like to exclude archived datasets, but we cannot do it
+        -- because some charts still use them (e.g. dataset 88)
```
pabloarosado, line 302: I don't understand why archived datasets should be included (and I don't see any charts using [dataset 88](https://owid.cloud/admin/datasets/88)).

pabloarosado (approved):
Hi Mojmir, this looks good! But I'd need more time to follow the logic in detail. In particular, I'd like to understand why entity names and codes are stored in metadata and not in data. If you want we can have a short call about this.

backport/datasync/data_metadata.py
```diff
@@ -0,0 +1,190 @@
...
+    time_format = "%Y-%m-%dT%H:%M:%S.000Z"
+    for col in ("createdAt", "updatedAt"):
+        variableMetadata[col] = variableMetadata[col].strftime(time_format)  # type: ignore
+
+    return variableMetadata
+
+
+def _infer_variable_type(values: pd.Series) -> str:
+    try:
+        values = pd.to_numeric(values)
```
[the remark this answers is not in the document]
Marigold: Sorry, forgot to add a note (and assert) that data_values can't be null (they won't exist in data_values). I'll fix it. (You'd be right if there were nulls).

backport/datasync/data_metadata.py
Marigold: Yeah, it sucks. I copied it from grapher codebase. I'll add some unit tests and make it clear what it does.

backport/datasync/datasync.py
Marigold: I took this naming from grapher where we use `bakedSite/grapher/variables`. You're right though, it's a bit weird. I already have all the data there and won't be changing it as part of this PR, but I'll reopen this discussion when we start switching URLs in grapher.

backport/datasync/datasync.py
Marigold: Uh, maybe it was a different dataset. Anyway, I'll remove this part of code. We need to be syncing all datasets, even the archived ones (because even archived datasets can have charts).

pabloarosado: Fair enough, although I don't understand why a dataset would be archived if it still contains active charts.

Marigold: > Fair enough, although I don't understand why a dataset would be archived if it still contains active charts.

Neither do I. I'll see if I can find it and create an issue for it.

PR #780, opened 2023-01-18
enhance(walkthrough-garden): accept different meadow version

lucasrodes:
Sometimes, a garden dataset might rely on a meadow dataset with a different version, e.g.

```yaml
data://garden/namespace/2022-11-25/short_name:
    - data://meadow/namespace/2022-11-30/short_name
```

Therefore, I have added an extra field if we want to specify the Meadow version separately.

### DIFF
--- a/walkthrough/garden.py
+++ b/walkthrough/garden.py
@@ -32,6 +32,7 @@ class GardenForm(BaseModel):
     short_name: str
     namespace: str
     version: str
+    meadow_version: str
     add_to_dag: bool
     include_metadata_yaml: bool
     generate_notebook: bool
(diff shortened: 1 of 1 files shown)


pabloarosado (approved):
I'm not sure how often this change is convenient, given that it introduces an additional field to be filled every time in the garden step, which in the most common case is unnecessary (because meadow and garden versions are the same). But it looks good to me, if you think this can be helpful.

lucasrodes:
@pabloarosado I think this is common. At least in my case.

> given that it introduces an additional field to be filled every time in the garden step

A default value is set in the walkthrough form, the same as the default one for the Garden dataset. So in cases where they are the same, you don't need to set it.

> which in the most common case is unnecessary (because meadow and garden versions are the same)
I find this often, and it would be nice if I didn't have to manually edit the DAG and files after running `walkthrough`. Also, it prompts me with the message "Meadow dataset not found..." which is inaccurate cause there is a Meadow dataset, just not with that version.

A typical scenario for me:

When I port several datasets to Grapher, I go channel by channel. First, I add all datasets to Snapshot, Meadow, Garden, etc. We can go into the details of why this makes sense in another discussion if you want. The point is, once I start working on Garden, there might have passed some days since I worked on Meadow, so I will set a different version (I use the current date as the version).

Also, it follows the logic of the `walkthrough meadow` step, where Meadow and Snapshot versions have different fields.

PR #784, opened 2023-01-19
Create Luxembourg Income Study dataset

paarriagadap:
Create Luxembourg Income Study from local files obtained from the LISSY platform.


etl/steps/data/garden/lis/2023-01-18/shared.py
```diff
@@ -0,0 +1,332 @@
...
+    for var, slug in var_slugs.items():
+        for wel in inc_cons_vars:
+            for e in equivalence_scales:
+
+                # For variables that use income variable and equivalence scale
+                col_name = f"{var}_{wel}_{e}"
+
+                if col_name in cols:
+
+                    tb_garden[
```
Marigold, line 218: Perhaps having a function returning Metadata would be more readable (and possibly testable), e.g.

```python
def var_metadata_income_and_equivalence_scale(slug, inc_cons, equivalence_scales) -> VariableMeta:
                      meta = VariableMeta(
                          title=f"{slug['title']} ({inc_cons['name']}, {equivalence_scales['name']})",
                          description=f"{slug['description']}{new_line}{inc_cons_dict['description']}{new_line}{equivalence_scales['description']}",
                          unit = slug["unit"],
                          short_unit = slug["short_unit"],
                      )
                      meta.display = {
                          "name": meta.title,
                          "numDecimalPlaces": slug["numDecimalPlaces"],
                      }
                      return meta

...

tb_garden[col_name].metadata = var_metadata_income_and_equivalence_scale(slug, inc_cons_dict[wel], equivalence_scales_dict[e])
```

This might be used to DRY similar code below.

etl/steps/data/meadow/lis/2023-01-18/lis_distribution.py
```diff
@@ -0,0 +1,85 @@
+"""
```
Marigold, line 1: This module is almost identical to `lis_abs_poverty`. You could wrap most of its functionality into a single function with two parameters (dependency name and target column) and put it to `shared.py` module.

Marigold: You could also have single dataset **with two tables**. That would let you have it in a single module. Do you think it would be worth it?

Marigold (approved):
Only checked that it runs fine, code quality and a couple of charts in grapher. There are some small opportunities for refactoring, but otherwise looks good!

etl/steps/data/meadow/lis/2023-01-18/lis_distribution.py
paarriagadap: Ah, the latter could work. I didn't think about it. In this case, it could be with three tables: `keyvars`, `abs_poverty` and `distribution`. I will try to do that.

etl/steps/data/garden/lis/2023-01-18/shared.py
paarriagadap: Thanks, I will think about this. It is convoluted because there are several variables that "behave" differently from each other, but this could work.

etl/steps/data/meadow/lis/2023-01-18/luxembourg_income_study.py
```diff
@@ -0,0 +1,95 @@
...
+        # Process data.
+        #
+        # Create a new table and ensure all columns are snake-case.
+        tb = Table(df, short_name=ds_name, underscore=True)
+
+        #
+        # Save outputs.
+        #
+        # Create a new meadow dataset with the same metadata as the snapshot.
+        ds_meadow = Dataset.create_empty(dest_dir, metadata=convert_snapshot_metadata(snap.metadata))
```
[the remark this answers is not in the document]
paarriagadap: Please also take a look if the rest of the things make sense. I am just trying to upload the three snapshots into three tables in one dataset.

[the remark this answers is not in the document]
Marigold: Hey @paarriagadap, I pushed a [commit with a fix](https://github.com/owid/etl/pull/784/commits/717f8601394c01b107605b642f513053333153fd). You were almost there, but you had to create dataset only once and then call `ds_meadow.add(tb)` for all three tables. Also `paths.load_dependency("step_name")` always points to short name of a step, i.e.`paths.load_dependency("luxembourg_income_study")`.

etl/steps/data/garden/lis/2023-01-18/luxembourg_income_study.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def run(dest_dir: str) -> None:
+    log.info("luxembourg_income_study.start")
+
+    #
+    # Load inputs.
+
+    ######################################################
+    # Key variables
```
Marigold, line 45: This function is really long. It would help readability if you split it to several functions, e.g. `load_keyvars_table`, `load_abs_poverty`, ... or something like that.

etl/steps/data/meadow/lis/2023-01-18/luxembourg_income_study.py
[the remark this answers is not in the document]
paarriagadap: Thank you very much, @Marigold! So I guess it's OK to call the snapshot `lis_keyvars` twice (before the loop and inside it) to make the `Dataset.create_empty` work

etl/steps/data/garden/lis/2023-01-18/luxembourg_income_study.py
paarriagadap: Do you mean creating three functions in `shared` for each table or much more functions?

etl/steps/data/meadow/lis/2023-01-18/luxembourg_income_study.py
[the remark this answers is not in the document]
Marigold: Up to you, I thought it would be better to call it again to make it transparent that we use metadata from `lis_keyvars` snapshot. (Calling it twice doesn't impact performance)

etl/steps/data/garden/lis/2023-01-18/luxembourg_income_study.py
Marigold: It can be in the same module `luxembourg_income_study.py`. It's just that any function longer than 100 lines is not very readable and prone to errors. Splitting it into multiple functions should improve that and maybe even DRY the code. (I'm leaving how granular should it be up to you. You don't need to create tens of functions, but logically related code should be ideally wrapped in a function)

paarriagadap: Ah, OK, I asked because I didn't quite know what "very long" meant for a code.

PR #785, opened 2023-01-20
fix(data): remove wrong version of lighting efficiency uk

Marigold:
Wrong date format was used as a version in fast-track. Fix it and raise validation error if version is not in the right format.

Correct version was [committed by fasttrack](https://github.com/owid/etl/commit/886bd2a282c67d2f189397231990e8b1043dd8bb)


pabloarosado (approved):
Looks good. I'm just wondering if the steps on lighting efficiency should be committed (with the appropriate version format).

Marigold:
> Looks good. I'm just wondering if the steps on lighting efficiency should be committed (with the appropriate version format).

That was a big question whether to commit every change made by fast-track. We ultimately settled on "why not?". Committing everything is technically much less complex and we can always just delete committed files and prune git history.

PR #787, opened 2023-01-20
Improve harmonize countries

pabloarosado:
* Generalize `etl.data_helpers.harmonize_countries` so that it can ingest a path to a file of excluded country names.
* Change tests to load `geo` from `etl.data_helpers` instead of `owid.datautils`.
* Add new tests for `harmonize_countries`.
* Simplify walkthrough garden to use the new `harmonize_countries`.
* Reformat garden dummy step.

### DIFF
--- a/etl/steps/data/garden/dummy/2020-01-01/dummy.py
+++ b/etl/steps/data/garden/dummy/2020-01-01/dummy.py
@@ -48,6 +48,9 @@ def run(dest_dir: str) -> None:
     # Add table of processed data to the new dataset.
     ds_garden.add(tb_garden)
 
+    # Update dataset and table metadata using the adjacent yaml file.
+    ds_garden.update_metadata(paths.metadata_path)
+
     # Save changes in the new garden dataset.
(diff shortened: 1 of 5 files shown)


etl/steps/data/garden/dummy/2020-01-01/dummy.py
```diff
@@ -52,38 +48,7 @@ def run(dest_dir: str) -> None:
     # Add table of processed data to the new dataset.
     ds_garden.add(tb_garden)
 
-    # Update dataset and table metadata using the adjacent yaml file.
```
Marigold, line 55: Why it got removed? Metadata yaml file is still there.

Marigold (approved):
Nice!

etl/steps/data/garden/dummy/2020-01-01/dummy.py
pabloarosado: Thanks, that was a mistake, I just put those lines back.

PR #789, opened 2023-01-20
clean(edstats): remove backport of datasets 88 and 2796

lucasrodes:
I am planning on archiving the following datasets in Grapher:

- [Adult literacy proficiency - World Bank EdStats and STEP Skills Measurement Program](https://owid.cloud/admin/datasets/2796)
- [World Bank Economy & Growth - World Bank EdStats (2015)](https://owid.cloud/admin/datasets/88)

I have seen that there are snapshot steps backporting these datasets. My guess is that these are safe to delete, as long as there are no downstream dependencies (which is the case), right?


Marigold:
> My guess is that these are safe to delete, as long as there are no downstream dependencies (which is the case), right?

Exactly, checking whether they're used in DAG is enough. You don't even have to remove those snapshots as these would live forever in S3, so we could either leave them in ETL or prune them automatically at some point.

PR #790, opened 2023-01-23
enhance(datadiff): improve performance

Marigold:
* improve performance of `datadiff` so that running `etl-datadiff REMOTE data/ --include "garden" --verbose` runs in reasonable time (it could still take an hour or so if your bandwidth is slow)
* avoid recomputing dependencies checksums by caching them, this speeds up `etl garden` by 3x or so
* save country profiles as other formats to make them available in the catalog

### DIFF
--- a/etl/datadiff.py
+++ b/etl/datadiff.py
@@ -68,8 +68,16 @@ class DatasetDiff:
             for col in ds_b[table_name].columns:
                 self.p(f"\t\t[green]+ Column [b]{col}[/b]")
         else:
-            table_a = _sort_index(ds_a[table_name]).reset_index()
-            table_b = _sort_index(ds_b[table_name]).reset_index()
+            table_a = ds_a[table_name]
+            table_b = ds_b[table_name]
+
+            # only sort index if different to avoid unnecessary sorting for huge datasets such as ghe
+            if not _index_equals(table_a, table_b):
+                table_a = _sort_index(table_a)
+                table_b = _sort_index(table_b)
+
+            table_a = table_a.reset_index()
+            table_b = table_b.reset_index()
 
             # compare table metadata
             diff = _dict_diff(_table_metadata_dict(table_a), _table_metadata_dict(table_b), tabs=3)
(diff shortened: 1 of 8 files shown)


etl/steps/__init__.py
```diff
@@ -372,8 +372,13 @@ def _output_dataset(self) -> catalog.Dataset:
         return catalog.Dataset(self._dest_dir.as_posix())
 
     def checksum_output(self) -> str:
-        # This cast from str to str is IMHO unnecessary but MyPy complains about this without it...
-        return cast(str, self._output_dataset.checksum())
+        # cache the output checksum to avoid unnecessarily re-computing checksums
+        # for the same datasets
+        key = str(self)
```
pabloarosado, line 377: I've never seen this before, but I guess it makes sense :-).

etl/datadiff.py
```diff
@@ -261,6 +269,14 @@ def _append_and_print(x):
     exit(1 if any_diff else 0)
 
 
+def _index_equals(table_a: pd.DataFrame, table_b: pd.DataFrame, sample: int = 1000) -> bool:
+    """Check if two tables have the same index. Sample both tables to speed up the check."""
+    return (
+        table_a.sample(min(len(table_a), sample), random_state=0).index
+        == table_b.sample(min(len(table_b), sample), random_state=0).index
+    ).all()  # type: ignore
```
pabloarosado, line 277: Interesting idea. I'm a bit surprised that this would be quicker than just comparing `(table_a.index == table_b.index).all()` or `table_a.index.equals(table_b.index)`. But I suppose you have checked and for big tables sampling is better.
Maybe you could include
```
if (len(table_a) < sample):
    return table_a.index.equals(table_b.index)
else:
    return (...).all()
```

pabloarosado (approved):
Looks good, although I don't fully understand the logic of cache and locks.

etl/datadiff.py
Marigold: This optimisation was done for `ghe` where it speeds it up from almost 2 minutes to 4 seconds. It has huge multi-level index which is probably why it takes so long.

Thanks, used your approach.

etl/steps/__init__.py
Marigold: All our steps implement `__str__` which returns their URI from dag (e.g. `data://garden/...`). Calling `str(self)` then calls implicitly `__str__`.

Marigold:
Improved caching to cache `is_dirty` method instead. This speeds up `etl garden` from ~180s to ~20s. The downside is that the dynamic caching looks like magic (though it was the easiest way how to make it work and only cache in `select_dirty_steps` method)

PR #792, opened 2023-01-23
enhance(wt-charts): use new backend version by default

lucasrodes:
By default, `walkthrough charts` and `etl-chart-suggester` will now use the new backend version. Previously, new version was optional (see #776).

To use old (legacy) version, use:

```
etl-chart-suggester MAPPING.json -o
```

or in walkthrough check "Use old version":

image

---
Good enough review would be to test `walkthrough charts` for some datasets locally and see that it makes sense.

### DIFF
--- a/etl/chart_revision/cli.py
+++ b/etl/chart_revision/cli.py
@@ -12,6 +12,8 @@ import click
 from rich_click.rich_command import RichCommand
 from structlog import get_logger
 
+# TBD
+from etl.chart_revision.deprecated import ChartRevisionSuggester
 from etl.chart_revision.revision import (
     get_charts_to_update,
     submit_revisions_to_grapher,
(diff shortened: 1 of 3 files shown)


etl/chart_revision/cli.py
```diff
@@ -12,6 +12,8 @@
 from rich_click.rich_command import RichCommand
 from structlog import get_logger
 
+# TBD
```
pabloarosado, line 15: TBD means "To be determined", I guess you could specify "To be deleted after testing period".

pabloarosado (approved):
It looks good. I just tested it on staging for [this small dataset](https://staging.owid.cloud/admin/datasets/5857) of one chart and worked well. I'll try for a bigger dataset on local grapher, and let you know if anything goes wrong.

pabloarosado:
I checked on local grapher with a big dataset (the Energy mix) and while going through the chart revisions, I got the following error:
Screen Shot 2023-01-24 at 11 59 06

Then I tried again **selecting "Use old version"** and I didn't get that error (but I did get the usual issue that some charts don't respect the choice of timeline selection).

But then I tried one more time **without selecting "Use old version"**. This time I did not get any error, and the timeline selections were always respected :-) . So I guess the error I got was some other small local grapher issue.

So I think this is a good improvement and seems to work well (at least with local grapher and staging).

Issue #793, opened 2023-01-23
Mismatch between Datasets path and metadata


Marigold:
There are tens of datasets with metadata different from their path. Dataset's URI is computed from metadata as `[channel]/[namespace]/[version]/[short_name]`. Path could theoretically determine all these fields in metadata, but we don't enforce it and if users don't set it explicitly there's gonna be a mismatch. This is annoying because lots of our tools depend on dataset URIs (`datadiff` or `compare`) and we use either metadata or path assuming they match.

Here are some problematic datasets I found

```
URI `garden/hmd/2022-12-07/hmd` does not match dataset path `garden/hmd/2022-12-07/life_tables`
URI `garden/homicide/None/homicide_long_run_omm` does not match dataset path `garden/homicide/2023-01-03/homicide_long_run_omm`
URI `garden/irena/2022-10-20/renewable_electricity_capacity_and_generation` does not match dataset path `garden/irena/2022-10-20/renewable_electricity_capacity`
URI `garden/irena/2023-01-04/renewable_electricity_capacity_and_generation` does not match dataset path `garden/irena/2023-01-04/renewable_electricity_capacity`
URI `garden/malnutrion/2022-10-18/malnutrition` does not match dataset path `garden/malnutrition/2022-10-18/malnutrition`
URI `garden/met_office_hadley_centre/2023-01-02/near_surface_temperature_global` does not match dataset path `garden/met_office_hadley_centre/2023-01-02/near_surface_temperature`
URI `garden/owid/2023-01-23/covid19` does not match dataset path `garden/owid/latest/covid`
URI `garden/rff/2022-09-14/emissions_weighted_carbon_price__economy` does not match dataset path `garden/rff/2022-09-14/emissions_weighted_carbon_price`
URI `garden/rff/2022-10-11/emissions_weighted_carbon_price__economy` does not match dataset path `garden/rff/2022-10-11/emissions_weighted_carbon_price`
URI `garden/wpp/2019/standard_projections` does not match dataset path `garden/un/2019/un_wpp`
URI `meadow/andrew/2019-12-03/co2_mitigation_curves_2celsius` does not match dataset path `meadow/andrew/2019-12-03/co2_mitigation_curves`
URI `meadow/aviation_safety_network/2022-10-12/aviation_statistics_by_period` does not match dataset path `meadow/aviation_safety_network/2022-10-12/aviation_statistics`
[...]

PR #794, opened 2023-01-24
fix(data): align metadata with dataset paths

Marigold:
Manual fix for https://github.com/owid/etl/issues/793. It's likely we'll solve this in a different more automatic way.

* covid dataset changed short name from `covid19` to `covid` and version from today's date to `latest` to match dataset path (we still have `data_accessed` from source if anyone wants to check the date of latest update)
* grapher datasets that changed `short_name` to match their path will have to be updated in production DB


etl/steps/data/grapher/cait/2022-08-10/carbon_dioxide_emissions_by_sector.py
```diff
@@ -0,0 +1,38 @@
+from owid import catalog
+
```
pabloarosado, line 2: Isn't this dataset identical to `cait/2022-08-10/carbon_dioxide_emissions_by_sector`? I guess your plan was to remove `co2_emissions`, but you forgot?

pabloarosado: Or, alternatively, for consistency with `ch4_emissions` and `n2o_emissions`, why not removing `carbon_dioxide_emissions_by_sector`? Maybe I'm misunderstanding the problem.

pabloarosado (approved):
Thanks for fixing all this. It looks good (except that, as I commented below, I think file `carbon_dioxide_emissions_by_sector` should not exist, but I'm not sure if I'm missing something). But I'm not sure if there could be unintended changes with this big refactor. Can you think of any?

etl/steps/data/grapher/cait/2022-08-10/carbon_dioxide_emissions_by_sector.py
Marigold: My bad, I originally thought I'd rename those files to keep `short_name` same as in our DB, but then I changed my mind and forgot to delete it.

Marigold:
[Datadiff](https://buildkite.com/our-world-in-data/etl-build-public-datasets-branch/builds/1214#0185e349-e21f-462b-b724-8a6d55a22e15) looks ok. It's gonna change metadata in our data catalog, but I don't think anyone would notice. We definitely have to change grapher short names in both production and staging. I can't think of anything other than that.

PR #799, opened 2023-01-26
Check for outdated datasets

pabloarosado:


etl/helpers.py
```diff
@@ -263,3 +267,405 @@ def load_dependency(
             dataset = catalog.Dataset(dataset_path)
 
         return dataset
+
+
+# For backwards compatibility.
+Names = PathFinder
```
Marigold, line 273: We've stopped using `Names`. There shouldn't be any more `Names` in ETL (and if there are, please replace them with `PathFinder`)

etl/helpers.py
```diff
@@ -263,3 +267,405 @@ def load_dependency(
...
+
+
+def extract_step_attributes(step: str) -> Tuple[str, str, str, str, str, str, str]:
```
Marigold, line 272: This is gonna be useful for other parts of code. Would it be possible to return dictionary instead of 7-tuple? Working with tuples would become cumbersome if I needed only e.g. namespace.

pabloarosado: Thanks. Yes, I was thinking the same thing. I'll do that.

Marigold: Also would you mind moving it under `parse_step` [here](https://github.com/owid/etl/blob/master/etl/steps/__init__.py#L246)? It's more logical to have it together.

etl/helpers.py
```diff
@@ -263,3 +267,405 @@ def load_dependency(
...
+        # Extract attributes from root of the step.
+        channel, namespace, version, name = root.split("/")
+
+        # Define an identifier for this step, that is identical for all versions.
+        identifier = f"{channel}/{namespace}/{name}"
+
+    return step, kind, channel, namespace, version, name, identifier
+
+
+def list_all_steps_in_dag(dag: Dict[str, Any]) -> List[str]:
```
Marigold, line 348: These all look like they could be methods of some DAG object. Right now we're wasting its potential with using just `DAG = Dict[str, Any]`. No need to do anything right now, just noting that we might refactor it in the future.

etl/helpers.py
pabloarosado: Done.

Marigold:
Would it get simpler if we had two DAGs `main.yml` and `archive.yml` instead of having additional `archive:` in each of the files?

`archive.yml` would look like this

```yaml
...
include:
  # archived steps are moved to their own file
  - dag/energy_archive.yml
  # we also include main because it can contain dependencies for archived steps
  - dag/main.yml
```

then run it with `etl --dag-path dag/archive.yml`. If our current checks work correctly and you archive step (move it to `energy_archive.yml`) that is necessary for some other un-archived step and run `etl`, it should throw an error.

(Thinking about it more, I should have done the same for private datasets which would give me consistency checks for free)

pabloarosado:
> Would it get simpler if we had two DAGs `main.yml` and `archive.yml` instead of having additional `archive:` in each of the files?
> 
> `archive.yml` would look like this
> 
> ```yaml
> ...
> include:
>   # archived steps are moved to their own file
>   - dag/energy_archive.yml
>   # we also include main because it can contain dependencies for archived steps
>   - dag/main.yml
> ```
> 
> then run it with `etl --dag-path dag/archive.yml`. If our current checks work correctly and you archive step (move it to `energy_archive.yml`) that is necessary for some other un-archived step and run `etl`, it should throw an error.
> 
> (Thinking about it more, I should have done the same for private datasets which would give me consistency checks for free)

You may be right, I wasn't sure of what was better, either to have `steps` and `archive` sections in each yaml file, or to have separate files. I'm thinking that separate files might also make it easier to move steps automatically from active to archive (e.g. once a step has been updated and should be archived). I don't have a strong opinion. Let me know if you think we should use separate files.

Marigold:
> I don't have a strong opinion. Let me know if you think we should use separate files.

Hard to tell honestly. I don't have a strong opinion either. With two files we _might_ end up with less code. If you are unsure, you can give it a quick try and see what you prefer. I'm doing that quite often and it helps me to quickly realise what approach feels most natural.

pabloarosado:
Hi @Marigold I've just refactored so that we have dag files and dag/archive files, and by default ETL will load the dag archive. So, for now, this PR shouldn't change anything in production.

I haven't created a command-line tool to check for outdated datasets, but I think that would not be very helpful. As we discussed the other day, we need something a bit more elaborate (like my prototype dashboard) to be able to visualize what datasets need to be updated, and the implications of that update. Also, we'd need to have some info from grapher DB to be able to decide what to update and what to archive. For example, to know whether an ETL grapher step can be archived, we need to know if the grapher grapher dataset has active charts.

So, if you agree, we can do that in a separate PR.

etl/command.py
```diff
@@ -22,6 +22,11 @@
 # if the number of open files allowed is less than this, increase it
 LIMIT_NOFILE = 4096
 
+# DAG file to use by default.
+# Use paths.DAG_ARCHIVE_FILE to load the complete dag, with active and archive steps.
+# Otherwise use paths.DAG_FILE to load only active steps, ignoring archive ones.
+DEFAULT_DAG_FILE = paths.DAG_ARCHIVE_FILE
```
Marigold, line 28: I'd be eager to do the leap of faith and make un-archived DAG the default one (unless you want to test it first in production with archived+ DAG and then do a second PR for un-archived DAG).

etl/helpers.py
```diff
@@ -263,3 +267,327 @@ def load_dependency(
...
+    * To archive those active steps.
+    * To un-archive those archive steps.
+    """
+
+
+class LatestVersionOfStepShouldBeActive(ExceptionFromDocstring):
+    """The latest version of each data step should be in the dag as a main step (maybe it was accidentally removed)."""
+
+
+class VersionTracker:
```
Marigold, line 422: Is it used anywhere when we use special archive DAG file? If not, I'd be more inclined to merge only the archival-related part and move this to a different PR and wait for a moment when we actually need this. These methods are useful, but perhaps refactoring them and having them as [methods on DAG/Step classes](https://github.com/owid/etl/issues/801) might be better?

Marigold (approved):
Two suggestions, but I don't feel strongly about them, so merge at your will.

etl/command.py
pabloarosado: Yes, I'm happy to do the leap of faith too :-D. But for now I'll stick to what we decided with Lars last time we talked about it. It's easy enough to change it in a second PR.

etl/helpers.py
pabloarosado: I agree that some of the components of VersionTracker should be moved to a `DAG` class. I like the idea. But I think we still need a class that keeps those functions from DAG, plus additional ones (for example, it should also interact with grapher, to get the number of charts used by each dataset, or even user analytics about each dataset). The goal in the end is to have a tool that makes it easy to:
* Have an overview of our datasets, to detect which ones are outdated, and how to prioritize their updates
* Create/update/archive datasets.

So I'd merge it for now (it's not used anywhere else in the code anyway) and we can rethink its components in a separate PR. I will add a message in the docstring mentioning that it should not be used for now.

PR #802, opened 2023-01-27
Data: Updating UN SDG to latest data and converting from walden to snapshot

spoonerf:


etl/steps/data/grapher/un/2023-01-24/un_sdg.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    Load the existing json which maps the raw sources to a cleaner version of the sources.
+    """
+    with open(paths.directory / "un_sdg.sources.json", "r") as f:
+        sources = json.load(f)
+        return cast(Dict[str, str], sources)
+
+
+@cache
+def get_metadata_link(indicator: str) -> str:
```
Marigold, line 206: Just a warning here - steps ideally shouldn't use data from other sources than snapshots. If those URLs suddenly stop working, ETL would start failing (or if they change it, we might end up with different metadata).

We've been using it with old SDGs and there weren't any problems, so no need to change it now, just be aware of it.

dag/main.yml
```diff
@@ -117,12 +113,26 @@ steps:
...
+    - walden://un_sdg/2022-07-07/un_sdg
+  data://garden/un_sdg/2022-07-07/un_sdg:
+    - data://meadow/un_sdg/2022-07-07/un_sdg
   data://grapher/un_sdg/2022-07-07/un_sdg:
     - data://garden/un_sdg/2022-07-07/un_sdg
   data://grapher/un/2022-07-11/un_wpp:
     - data://garden/un/2022-07-11/un_wpp
 
+ # UN SDG - current
+  data://meadow/un/2023-01-24/un_sdg:
```
Marigold, line 126: I've noticed we're changing namespace from `un_sdg` to `un`. Could you also change namespace of the old dataset from `un_sdg` to `un`? This would let us archive it. If we're lucky we won't have to make any changes to production DB or anything.

Marigold (approved):
Looks good! Left a couple of comments (the most important being changing the namespace of the old dataset). Right on time ;)

etl/steps/data/grapher/un/2023-01-24/un_sdg.py
spoonerf: That's a very good point, I don't think they are likely to change it, but you're right it's a bit dangerous. I will make a note to grab the metadata and store it in a snapshot for the next update.

PR #809, opened 2023-01-30
feat: Add version tracker checks to make test

pabloarosado:
Add version tracker checks to make test.

This new check will:
* Raise an error if an archive dataset is a dependency of an active dataset.
* Raise a warning if an active dataset can safely be archived.

@larsyencken: @Marigold and I were discussing that the `VersionTracker` still needs some work, so we may need to refactor it a bit. But I still agree that it's good to add it to `make test` in its current version. If we refactor it later, we can also change the way we call the checks.


Makefile
```diff
@@ -115,3 +115,7 @@ dependencies.pdf: .venv dag.yml etl/to_graphviz.py
 deploy:
 	@echo '==> Rebuilding the production ETL from origin/master'
 	ssh -t owid@analytics.owid.io /home/owid/analytics/ops/scripts/etl-prod.sh
+
+version-tracker:
```
larsyencken, line 119:
```suggestion
version-tracker: .venv
```

larsyencken (approved):
Looks good :shipit: 

Just one suggestion to add a `.venv` dependency to your rule for cleanliness.

Marigold (approved):
Nice! Agree that everything can be refactored in the future.

PR #811, opened 2023-01-30
fix(etl): use DEFAULT_DAG_FILE everywhere

Marigold:
Some steps were not building because `load_dag` was using `DAG_FILE` and not `DEFAULT_DAG_FILE`. This is a bit confusing, it would be better do the [leap of faith](https://github.com/owid/etl/pull/799#discussion_r1090485357), avoid using `DEFAULT_DAG_FILE` and use just `DAG_FILE` & `DAG_ARCHIVE_FILE`

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -22,11 +22,6 @@ config.enable_bugsnag()
 # if the number of open files allowed is less than this, increase it
 LIMIT_NOFILE = 4096
 
-# DAG file to use by default.
-# Use paths.DAG_ARCHIVE_FILE to load the complete dag, with active and archive steps.
-# Otherwise use paths.DAG_FILE to load only active steps, ignoring archive ones.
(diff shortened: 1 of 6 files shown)


pabloarosado (approved):
It looks good. Thanks for catching those additional paths to the dag file that I didn't see.
I think it's fine if we decide to use the active dag by default, and only load the archive dag when running the etl on archive datasets. What do you think @larsyencken?

larsyencken:
@pabloarosado Hmm... so, so far it looks like we don't have any `--archive` flag to enable/disable the behaviour.

We could totally use that flag to just load different DAGs, and default to the non-archival DAG.

The alternative would be to load everything but tag some steps as being archived, then do filtering logic using tags. Maybe it's not worth doing that unless we had a good reason to.

PR #815, opened 2023-01-31
refactor(dag): moving old unodc homicide data to archive dag

spoonerf:
Just a quick PR to check I am archiving correctly

### DIFF
--- a/dag/archive/main.yml
+++ b/dag/archive/main.yml
@@ -3,7 +3,11 @@ steps:
     - walden://wpp/2019/standard_projections
   data://garden/un/2019/un_wpp:
     - data://meadow/un/2019/un_wpp
-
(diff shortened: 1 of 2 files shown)


pabloarosado:
Hi @spoonerf I've run `make test` (which applies some checks to ensure that there are no missing dependencies) and it all tests passed. However, from what I can see, you didn't actually archive the datasets, you simply removed them from the dag. Shouldn't you move the steps from `dag/main.yml` to `dag/archive/main.yml`? Or is there a reason why you want to remove them instead of archiving them?

spoonerf:
@pabloarosado Ah sorry that was a silly mistake and a careless commit, they are meant to be in the archive and are back there now!

pabloarosado (approved):
Looks good!

PR #816, opened 2023-02-01
enhance(helpers): add helper function create_dataset

Marigold:
Implementation of `create_dataset` function suggested by Pablo as a fix https://github.com/owid/etl/issues/793. With this function we can omit `namespace`, `short_name` and `version` from YAML metadata and infer it from path instead. Walkthrough was updated to reflect this.

This should prevent mismatching metadata with paths (e.g. [this one](https://github.com/owid/etl/blob/master/etl/steps/data/garden/homicide/2023-01-27/unodc.meta.yml#L40) that got recently merged)


pabloarosado:
Hi @Marigold I haven't looked into it in detail yet, but a quick question: Why not including `ds.save()` inside the `create_dataset()` function? That would save a lot of redundant code (and I think it's something that can easily be forgotten). And, if someone wants to make further modifications to the dataset, they could simply `ds.save()` again after the modifications. Or am I missing something?

Marigold:
@pabloarosado initially I had it in `create_dataset`, but then I found several steps where we modify dataset metadata programatically after creating it. It's also nicer if `create_dataset` doesn't have side effect and just returns an object. You can always call `create_dataset(...).save()` (I avoided that and made it explicit)

> and I think it's something that can easily be forgotten

There's a warning displayed if we forget to call `.save()`

pabloarosado (approved):
Looks good, and very nice to get rid of so much redundant code!

Marigold:
> Looks good, and very nice to get rid of so much redundant code!

Yeah, creating that function was a really nice idea.

PR #817, opened 2023-02-01
enhance(wt-charts): use new backend version

lucasrodes:
For some reason I forgot to change the function called in the `etl-chart-suggester` command tool. The web app client is already using the new backend version


pabloarosado (approved):
Looks good!

PR #762, opened 2023-01-10
data(conflicts): add steps for meadow

lucasrodes:
This PR adds meadow steps for all conflict datasets (part of the project https://github.com/owid/owid-issues/issues/443).

- Module `shared` with standard tools across steps.
- Most datasets have the same column names, except three of them, where a re-definition of the column renaming mapping dictionary is needed.
- Fix a small typo in the short name of a snapshot dataset.

PR #818, opened 2023-02-03
data(war): new steps

lucasrodes:
This PR adds the datasets needed for the History of War project. These consist of 12 datasets, which we label with the new namespace `war`.

- 7 datasets are ported from transcriptions made by Joe in a Google spreadsheet. We use this process as the data was manually extracted from old resources (such as papers, books, etc.). These are Bouthoul and Carrere, Clodfelter, Dunnigan and Martel, ORAE, Eckhardt, Sorokin and Sutton.
- The remaining five datasets (Brecke, Prio v1.0, Prio v3.1, CoW and UCDP) are available online; thus, we follow the usual procedure.

This follows previous PRs:
- https://github.com/owid/etl/pull/754: Work on snapshots manually porting data files from Google spreadsheets.
- https://github.com/owid/etl/pull/762: Work on formatting snapshot data files into Meadow-friendly datasets.

This PR focusses on implementing the remaining steps so that all the datasets are available in Garden.


larsyencken (commented):
Thanks for all this work Lucas! 🙌 

Looking over it, the main thing I was looking at was what we mean by things being in `garden`, which is meant to be where we have harmonised the data. 

It looks like `garden` means you've selected common column names for the different metrics we have on conflicts, and done some basic standardisation of participant names.

To properly harmonise participants in future, we might need a table like `countries_regions` for them, which gives each one a unique ID and possibly multiple aliases. That's just to say that I think there'll be a lot of work in future in all that harmonisation, you might need some kind of semi-interactive tool like `harmonize` to help with that.

Anyhow, that's all for the future. Feel free to :shipit: now!

lucasrodes:
Thanks, @larsyencken! Indeed, further harmonisation is needed. I will link your comment in the topic issue.

Issue #779, opened 2023-01-18
Redefine countries-regions dataset


pabloarosado:
We need to redefine our current countries-regions dataset.

# Requirements
* We definitely need the list of current countries (which is a static list). We don't need to have all countries that have ever existed.
  * Each current country is defined by their current geographical border.
* We need a list of aliases of current countries, e.g. "United States" -> "USA", "US", ...
* We need to know the current continent of each current country. This is also a static list.
  * If we consider only continents, each country would be allocated to just one continent. However, if we include other regions, like EU(27) or OECD, and others, then a country can belong to multiple regions.
* We need the population of the current country, and the population of that geographical region for each year in the past and projected into the future. We want this population regardless of whether the country existed a specific year. For example, we want the population of Russia (the geographical region of current Russia) even when it did not exist.
  * We also need the population of historical regions. A priori, we could apply the same criterion as with current regions, and have the population (in the past and in the future) of the geographical region that it occupied. However, in practice, we may only need population for the years when the region existed.
* We need to map the transitions between geographical regions. For example, we need to know that East Germany and
  West Germany became Germany in 1990, and that USSR became Armenia, Azerbaijan, ...
* We may need to have a special way to identify former regions, like suffix "(former)" or prefix "Former".
  For example, we would have "USSR (former)" or "Former USSR". In the case of Sudan, we would need "Sudan", and
  "Sudan (former)" or "Former Sudan".
* We need alternative definitions of regions, given by the data providers. For example, "North America (BP)".
  * We may need the mapping of these regions to current regions, but it may not be always possible (they may not provide such explicit definitions).
  * We may also need alternative definitions of countries. For example, some providers may include Zanzibar in Tanzania and others would not, or define China as "China (mainland)" and "China" (including Taiwan and Macau). In that case, it would be good to have, e.g. "Tanzania mainland (Institution)" and "Zanzibar (Institution)".
* We may need different definitions of country codes.
* We need income groups (although this already is a separate World Bank dataset).

# Issues
* How do we deal with transition years? For example, when does the USSR end and when does Russia start?
* How do we deal with regions that change borders? For example, USSR is different in 1989 and 1990. We could define, e.g. "USSR (pre-1990)" and "USSR (1990)", but it may be an overkill.

# Other considerations
* How do we edit this data, manually or through tools (like harmonize)?
* Should we consider sub-country regions and their dependencies, e.g. Bavaria < Germany?
[...]

Issue #796, opened 2023-01-24
Regions and aliases that must be corrected


pabloarosado:
## Changes to make

Regions whose name needs to be changed:
* Turkey -> Türkiye. Should we apply this change? I think so far all our country names are free of accents and umlauts. So this change feels like an outlier.
* Timor -> Timor-Leste. Since Timor refers to the entire island (where West Timor belongs to Indonesia), but we usually refer to the country East Timor within the island.

Aliases to be corrected:
* Yemen Arab Republic should be an alias of former region North Yemen, instead of Yemen. I wrote this down some time ago, but I think it has already been fixed. @larsyencken was it you?

Additional changes:
* We need a short name for "Democratic Republic of Congo", for example for Marimekko charts. What about "DR Congo"?

## Todo

- [ ] Countries regions dataset change in ETL
- [ ] Update grapher so that new regions actually show on map
- [ ] Fix every existing country mapping file in ETL

lucasrodes:
> Turkey -> Türkiye. Should we apply this change? I think so far all our country names are free of accents and umlauts. So this change feels like an outlier.

We can use "Turkiye" instead, but I think we should consider this rename [[ref](https://www.aljazeera.com/news/2022/6/2/un-registers-turkiye-as-new-country-name-for-turkey)].

> Timor -> Timor-Leste. Since Timor refers to the entire island (where West Timor belongs to Indonesia), but we usually refer to the country East Timor within the island.

An alternative name could be "East Timor". 

> 
Regions whose name needs to be changed:

Turkey -> Türkiye. Should we apply this change? I think so far all our country names are free of accents and umlauts. So this change feels like an outlier.
Timor -> Timor-Leste. Since Timor refers to the entire island (where West Timor belongs to Indonesia), but we usually refer to the country East Timor within the island.
Aliases to be corrected:

Yemen Arab Republic should be an alias of former region North Yemen, instead of Yemen. I wrote this down some time ago, but I think it has already been fixed. @larsyencken was it you?
Additional changes:

> We need a short name for "Democratic Republic of Congo", for example for Marimekko charts. What about "DR Congo"?

I like it!

spoonerf:
This sounds great and I think we should definitely do it. However, I don't think it is currently possible to use accents in grapher? I made a quick experiment [here](https://neurath.owid.cloud/admin/charts/5864/edit) and the 'ü' doesn't show up.

edomt:
Thanks Pablo!

[...]

Issue #156, opened 2022-04-05
Use "Former" prefix for historic countries


Marigold:
@pabloarosado had an idea to use "Former " prefix for historic countries, e.g. `USSR -> Former USSR`, `Yugoslavia -> Former Yugoslavia`. Hannah and others approved it. 

## Scope
* [ ] Add a column to the `countries_regions` table to indicate if this country is historic
* [ ] Rename those countries in `countries_regions` and make sure regions grouping doesn't change
* [ ] Make sure `population` dataset works correctly after this change as it is used outside this repository
* [ ] Rename them in the grapher DB tables (entities themselves and the harmonization tables)
* [ ] Check for countries that need renaming in harmonized ETL steps

In case this turns out to be trickier than expected, we might do it as part of larger countries_regions refactoring.

larsyencken:
Since this would touch a lot of ETL datasets, it would really make it easier to have a simple sanity checker, at least something that compared the range of values in each column in the primary key.

Marigold:
Removing `wontfix`, I think this is still important. We're still using custom solutions [such as this](https://github.com/owid/etl/blob/master/etl/steps/data/garden/bp/2022-07-14/energy_mix.py#L85), but having a unified approach would be nice.

larsyencken:
@pabloarosado Hey, marking this one as "prio 3 - nice to have" for now, since we have some other correctness issues we'd like to fix first. But if at some stage you wanted to champion this, we'd be happy to help out.

pabloarosado:
Hi @Marigold now that we have the new `regions` dataset, where we have information about historical regions, I think we don't need to impose this suffix to historical regions. So, if you agree, feel free to close this issue.

PR #824, opened 2023-02-08
Redefine countries regions

pabloarosado:
This PR creates a new dataset for countries-regions.

## Theoretical goals
1. We need to ensure that, across different datasets, regions mean the same thing. For example, if in one dataset China includes `Hong Kong` and `Macao` while another dataset doesn't, then they should not be called the same thing.
2. When creating aggregates (e.g. Europe) we need to ensure that we are adding data from all relevant regions and that we are not double-counting any region.

## Practical goal
Finding a solution for the previous goals may require a complicated data architecture and many hours of work. Or it may be impossible (given that data providers are usually unclear about region definitions). So we accept that there will always be some small geographical inaccuracies that are impossible to fix given our limitations. Given this, the goals become:
1. We need to ensure that inaccuracies are not very significant (e.g. we should not double-count Russia's contribution when aggregating data for Europe).
2. We need to save time when harmonizing country names and creating region aggregates.

## For reviewers
For major comments (e.g. "I think we should remove this table") it would be better to keep the conversation on [this issue](https://github.com/owid/etl/issues/779). For small comments, please use this PR.
* [x] @spoonerf regarding [this issue](https://github.com/owid/owid-issues/issues/13) currently the only aggregates we include are continents, EU 27 and the World (plus some other small geographical regions that happened to be in the old countries-regions dataset). Do you think there's any other region aggregates to include in this dataset (e.g. land-locked countries)? Otherwise, if those regions are created by specific institutions (e.g. income groups are defined by the World Bank and change periodically) I'd say they should be in different datasets.
* [x] @lucasrodes do you mind reviewing the `regions.py` garden step, `regions.meta.yml` metadata, and the `migration.py` script? The latter needs to be run only once, so it doesn't require a thorough review.
* [x] The `regions.codes.csv` doesn't need to be reviewed. It contains the legacy codes that were in the old dataset.
* [x] @bastianherre do you mind reviewing the content of `regions.yml`? Or if it's easier, you can review the definitions at the beginning of the `migrations.py` script. There you have some explanations (and you have country names instead of codes). If you need any clarifications, just let me know.
* [x] @Marigold I'd like to know if you agree with my comments below.

## Changes with respect to the old `countries_regions`
* New columns: `region_type`, `short_name`, `is_historical`, `successor`, `end_year`, `related`, and `defined_by`.
* New structure: Now we have one dataset, called `regions`, which contains the following tables:
  * `definitions`: Region definitions (name, country type, etc.).
  * `aliases`: Region aliases (i.e. variants of the region name).
  * `members`: Region members (roughly, sub-regions that would need to be added up when aggregating data for the region).
  * `related`: Other possible region members to be aware of (e.g. regions with an unclear official status that may be members of the region according to some data providers).
  * `transitions`: Historical transitions between regions.
  * `legacy_codes`: Legacy codes (e.g. wikidata code).
* The name of TLS has been corrected to "East Timor" (instead of "Timor").
* German and Italian subregions that were in the old dataset have been removed (I don't think we need country sub-regions for now). All other regions should be the same as in the old dataset.
* Some regions have been removed from the list of members of continents.
* Added a short name for the Democratic Republic of Congo (DR Congo).

## Minor comments
* I wasn't sure if the namespace should be `demography` or `geography`, so I simply called it `regions`, but let me know if anyone disagrees.
* I added the version `2023-01-01` just for simplicity, since it's unlikely that we will change it often.
* The idea here is that, if we want to include aliases or new regions, we can manually edit the `regions.yml` file directly, without creating a new dataset version. We should create a new dataset version only after a major change.
* We don't have to start replacing the `reference` dataset by this one for now. But we can start using this one when creating new datasets (after we agree on its content).
* This dataset will start making more sense once we create new helper functions for harmonization and to create region aggregates. For now, maybe we should not use this dataset.
* Currently the dataset only has one garden step that loads two adjacent files. This is an easy solution, but if we think it clashes too much with the ETL philosophy, I can move the two input files (`regions.yml` and `regions.codes.csv`) to snaphot, and the manual data gets ingested there.


etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -0,0 +1,1054 @@
+# Region definitions (see more details in the description of the 'regions' garden step).
+# Each region must contain the following fields:
+# code: Region code (unique for each region).
+# name: Region name.
+#
+# Additionally, each region can contain the following fields:
```
Marigold, line 6: (We could have JSON schema validation to make sure we're not making mistakes when manually editing the file, but that could be added in the future)

Marigold (approved):
Can't really imagine cleaner way how to define regions. I hope it's gonna work well once we use it for harmonisation.

Marigold:
Agree with everything, especially that we don't need snapshot for this and that we should be modifying existing version on minor updates.

> We don't have to start replacing the reference dataset by this one for now. But we can start using this one when creating new datasets (after we agree on its content).

It could be useful to create another PR that would use this dataset instead of reference (and compare its output with datadiff). Just to make sure that it works in practice and we didn't miss anything. (That doesn't block this PR though, it could be done in parallel or even after we merge this)

After we start using this file we'll have to modify [harmonize](https://github.com/owid/etl/blob/master/etl/harmonize.py#L265) to start saving aliases in the regions YAML file (`ruamel.yaml` is useful for modifying YAML files in-place). We can do it together with new helpers.

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+}
+
+# Manually define "related member" territories, which includes all dependent and overseas territories.
+# Here, the definition of member is vague.
+# The status of some of these regions are unclear or contested, and by defining these dependencies we are not
+# making any political statement.
+# This list will only be used to raise warnings on potential overlaps in the data.
+# For example, if there is data for both China and Hong Kong, it would be good to raise the warning that we may
+# be double-counting Hong Kong's data (as some data providers include Hong-Kong as part of China).
+related_territories = {
```
lucasrodes, line 571: nice idea

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+
+        # Add related members only if there is any.
+        if len(region["related"]) > 0:
+            text += f"  related: {region['related']}\n"
+        text += "\n"
+
+    return text
+
+
+def main():
```
lucasrodes, line 734: I would place this function before `_create_yaml_content_from_df`

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+
+    # Add a short name for DRC.
+    df.loc[df["code"] == "COD", "short_name"] = "DR Congo"
+
+    # Assign TLS the right name and short name, which is East Timor.
+    df.loc[df["code"] == "TLS", "name"] = "East Timor"
+    df.loc[df["code"] == "TLS", "short_name"] = "East Timor"
+
+    # Add a short name for DRC.
+    df.loc[df["code"] == "COD", "short_name"] = "DR Congo"
```
lucasrodes, line 776: Duplicate row (see row 768-769)

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+
+    # Assign TLS the right name and short name, which is East Timor.
+    df.loc[df["code"] == "TLS", "name"] = "East Timor"
+    df.loc[df["code"] == "TLS", "short_name"] = "East Timor"
```
lucasrodes, line 780: Duplicate rows (see rows 771-773)

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+
+    # Remove unnecessary regions.
+    df = df[~df["name"].isin(regions_to_remove)].reset_index(drop=True)
+
+    # Edit the rows corresponding to historical regions.
+    for historical_region in historical_regions:
+        end_year = historical_regions[historical_region]["end_year"]
+        successors = historical_regions[historical_region]["successors"]
+        successor_codes = sorted([df[df["name"] == successor]["code"].item() for successor in successors])
+        region_sel = df["name"] == historical_region
```
lucasrodes, line 802: We could maybe assert that `sum(region_se) == 1`.

etl/steps/data/garden/regions/2023-01-01/regions.py
```diff
@@ -0,0 +1,149 @@
...
+# Path to input file of region definitions.
+REGION_DEFINITIONS_FILE = paths.directory / "regions.yml"
+# Path to input file of region codes.
+REGION_CODES_FILE = paths.directory / "regions.codes.csv"
+
+# Expected region types.
+REGION_TYPES = ["country", "continent", "aggregate", "other"]
+
+
+def parse_raw_definitions(df: pd.DataFrame) -> pd.DataFrame:
```
lucasrodes, line 32: would place it after method `run`

etl/steps/data/garden/regions/2023-01-01/regions.py
```diff
@@ -0,0 +1,149 @@
...
+        df["defined_by"] = pd.Series(dtype=str)
+    df["defined_by"] = df["defined_by"].fillna("owid")
+
+    # If "is_historical" is not given, fill with False.
+    df["is_historical"] = df["is_historical"].fillna(False)
+
+    return df
+
+
+def run_sanity_checks(df: pd.DataFrame) -> None:
```
lucasrodes, line 50: would place it after method `parse_raw_definitions`, after `run`.

lucasrodes (commented):
LGTM! Thanks for working on this. Added minor comments.

lucasrodes (approved):

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+        "Chad",
+        "Comoros",
+        "Congo",
+        "Cote d'Ivoire",
+        "Democratic Republic of Congo",
+        "Djibouti",
+        "Egypt",
+        "Equatorial Guinea",
+        "Eritrea",
+        "Eritrea and Ethiopia",
```
bastianherre, line 56: Does this refer to Ethiopia before Eritrea split from it? I thought it was more commonly known as simply 'Ethiopia'.

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+        "Sri Lanka",
+        "Syria",
+        "Taiwan",
+        "Tajikistan",
+        "Thailand",
+        # 'Timor',
+        "Turkey",
+        "Turkmenistan",
+        "United Arab Emirates",
+        "United Korea",
```
bastianherre, line 158: Is this the official name for the country before it split into North and South? I thought it was more commonly known as just 'Korea'.

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,884 @@
...
+    "Somaliland",
+    "South Ossetia",
+    "Svalbard and Jan Mayen",
+    "Transnistria",
+    "United States Minor Outlying Islands",
+]
+
+# Manually list regions to remove.
+regions_to_remove = [
+    # Is there any reason to keep these historical regions?
```
bastianherre, line 424: Some datasets have data on these historical regions/states, before they became Germany and Italy — V-Dem is one prominent example. Would that be a reason to keep them?

bastianherre (commented):
This is very nice work!

I left a few comments. Which sources did you use for all this information? Is this all based on UN data/information — or did you use other sources as well?

Please also let me know if I went about this review the wrong way — it is my first one!

spoonerf:
Hey @pabloarosado - awesome that you have tackled this! 

Would it be worth adding in regions we are confident are unlikely to change in their composition? I can see that we might want to avoid the World Bank income groups here as these change periodically.

However, I think it could be valuable to add the World Bank regions and other organisational regions (WHO, UN) defined on [this page](https://ourworldindata.org/world-region-map-definitions) as I think these are unlikely to change. 

I think it is possible that the other groupings listed on [this issue](https://github.com/owid/owid-issues/issues/13) might change and therefore we wouldn't want to include them here.

spoonerf (approved):

etl/steps/data/garden/regions/2023-01-01/migration.py
danyx23: Probably not - the two main uses of this file for us are a) to compute aggregations in a more automated way and b) to join data across datasets (e.g. when doing a scatterplot). I doubt that we have enough datasets that contain comparable definitions for these entities that either case is compelling. We can always _use_ countries within one dataset as we like - removing them here just communicates that we don't view these as useful enough that for either aggregation or joining. Does that make sense?

etl/steps/data/garden/regions/2023-01-01/regions.py
```diff
@@ -0,0 +1,149 @@
+"""Region definitions, aliases, members, historical transitions and codes.
+
+This step will load the adjacent files of region definitions and codes, and create a garden dataset with tables:
+* "definitions": Region definitions (name, country type, etc.).
+* "aliases": Region aliases (i.e. variants of the region name).
+* "members": Region members (roughly, sub-regions that would need to be added up when aggregating data for the region).
+* "related": Other possible region members to be aware of (e.g. regions with an unclear official status that may be
+  members of the region according to some data providers).
+* "transitions": Historical transitions between regions.
+* "legacy_codes": Legacy codes (e.g. wikidata code).
```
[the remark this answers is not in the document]
danyx23: I'd keep them as it might be useful in the future for updating old datasets or redoing datasets that come from literature. Wikidata codes might be useful at some point to fetch data from there (e.g. capitals for countries)

danyx23:
Great work, @pabloarosado! This all looks very good. I have two points to add:

* I think we should consider if we really want the aliases to be stored in this file. The reason for this about ETL re-computation. In my mind, by far the most frequent change in the regions.yml file will be alias additions/changes. The way the ETL is currently set up this would then invalidated all downstream datasets (which would eventually be all, right?). There is a good chance that the added convenience and improved resilience is worth it but it's something I'd like to chat about in a data architecture meeting or similar.
* I think we should add a README.md file to this folder that explains how to make changes to the files (i.e. what is the threshold at which new entities should be added here, what are the fields we track and how to use them).

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,506 @@
...
+        "Niue",
+        "Norfolk Island",
+        "Pitcairn",
+        "Samoa",
+        "Tokelau",
+        "Tonga",
+        "Tuvalu",
+        "Wallis and Futuna",
+    ],
+}
```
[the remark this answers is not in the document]
pabloarosado: That's what the `migration.py` script does already. Or do you mean something else?

etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -0,0 +1,2148 @@
...
+  defined_by: "owid"
+  is_historical: False
+
+- code: "ALA"
+  name: "Aland Islands"
+  short_name: "Aland Islands"
+  region_type: "country"
+  defined_by: "owid"
+  is_historical: False
+  aliases: ['Ã\x81EEland Islands', 'Åland Islands']
```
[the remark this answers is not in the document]
pabloarosado: Good idea, I'm changing it now.

etl/steps/data/garden/regions/2023-01-01/migration.py
```diff
@@ -0,0 +1,506 @@
...
+contested_regions = [
+    "Nagorno-Karabakh",
+    "Northern Cyprus",
+    "Abkhazia",
+    "South Ossetia",
+    "Transnistria",
+    "Western Sahara",
+    "Somaliland",
+    "Kosovo",
+]
```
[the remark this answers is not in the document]
pabloarosado: Just reading from wikipedia and other sources, nothing extremely consistent, it was hard to find a consistent criterion...

etl/steps/data/garden/regions/2023-01-01/migration.py
pabloarosado: I just added a similar check above, ensuring that there are no repeated names.

etl/steps/data/garden/regions/2023-01-01/regions.py
pabloarosado: I usually define `run` at the very end, and all additional functions are defined before `run`, and tend to appear in order of appearance in `run`. What's your criterion?

etl/steps/data/garden/regions/2023-01-01/migration.py
pabloarosado: Yes, I believe that's what it refers to. I didn't create this entity, but I suppose it was called `Eritrea and Ethiopia` to avoid confusion with `Ethiopia` current and former. But we could redefine `Eritrea and Ethiopia` -> `Ethiopia (former)`. Let me know if you think that's a better choice.

etl/steps/data/garden/regions/2023-01-01/migration.py
pabloarosado: Yes, again, I don't know who introduced this definition in our dataset, but `United Korea` refers to the country before the split. I suppose it was called `United Korea` to avoid confusion. Feel free to propose a better name, e.g. `Korea (former)`.

etl/steps/data/garden/regions/2023-01-01/migration.py
pabloarosado: Thanks Daniel, that's also how I see it. In fact I have been very permissive letting regions stay in the dataset, maybe I should have removed other complicated regions. But for this specific case, if we are not expecting these historical regions to appear in any dataset apart from V-Dem, then there's no strong reason to keep them.
In any case, if we change our mind, we can add them in the future.

pabloarosado:
> This is very nice work!
> 
> I left a few comments. Which sources did you use for all this information? Is this all based on UN data/information, because some of this is controversial — or did you use other sources as well?
> 
> Please also let me know if I went about this review the wrong way — it is my first one!

Thank you @bastianherre! Regarding sources, I couldn't find one specific source where all the required definitions were clearly stated. So I used various sources, mostly Wikipedia. If you have found something that is particularly controversial, or that we should change, please let me know.

And yes, you did well with the review, adding comments by line is very useful, thanks!

pabloarosado:
> Hey @pabloarosado - awesome that you have tackled this!
> 
> Would it be worth adding in regions we are confident are unlikely to change in their composition? I can see that we might want to avoid the World Bank income groups here as these change periodically.
> 
> However, I think it could be valuable to add the World Bank regions and other organisational regions (WHO, UN) defined on [this page](https://ourworldindata.org/world-region-map-definitions) as I think these are unlikely to change.
> 
> I think it is possible that the other groupings listed on [this issue](https://github.com/owid/owid-issues/issues/13) might change and therefore we wouldn't want to include them here.

Hi @spoonerf thanks for the clarification and the suggestion. To avoid blocking this one, I've created [a separate issue](https://github.com/owid/owid-issues/issues/945) to add those regions.

pabloarosado:
> Great work, @pabloarosado! This all looks very good. I have two points to add:
> 
> * I think we should consider if we really want the aliases to be stored in this file. The reason for this about ETL re-computation. In my mind, by far the most frequent change in the regions.yml file will be alias additions/changes. The way the ETL is currently set up this would then invalidated all downstream datasets (which would eventually be all, right?). There is a good chance that the added convenience and improved resilience is worth it but it's something I'd like to chat about in a data architecture meeting or similar.
> * I think we should add a README.md file to this folder that explains how to make changes to the files (i.e. what is the threshold at which new entities should be added here, what are the fields we track and how to use them).

Hi @danyx23 thanks!
* Regarding aliases, I'm not sure what the best solution is. If you move the aliases outside of this dataset to another dataset, then that new dataset would have the same problem, wouldn't it? We could simply omit that dependency, but that would not be an ideal solution. I think aliases don't change that often anymore, so I don't think it's a big deal. Locally, we usually have to run only the steps that we need to work on, and in production, all datasets are executed regularly anyway. What do you think? We can talk about it tomorrow during the data architecture chat.
* Regarding the README, no problem. I can add one to `etl/docs/data`. It's quite a hidden place (and there's only one more dataset README there) but I think it's probably the right place. Let me know if you have a better idea.

etl/steps/data/garden/regions/2023-01-01/migration.py
bastianherre: Yes, that makes sense. I expect us to add more datasets with these historical states, and we may want to create some aggregates there — but the number of datasets probably will likely stay low.

etl/steps/data/garden/regions/2023-01-01/migration.py
bastianherre: Yes, if we want a different name for when Eritrea was part of Ethiopia, I would prefer Ethiopia (former). That would make clear that it was indeed one country.

etl/steps/data/garden/regions/2023-01-01/migration.py
pabloarosado: I don't think it's a problem to include those regions, but we'd simply need to add the necessary information (e.g. their end year). But we can do this as a separate issue, to avoid delaying this one. It would be better to add those regions once we have better tools in place (i.e. `harmonize` and `add_region_aggregates` are updated).

etl/steps/data/garden/regions/2023-01-01/migration.py
bastianherre: In this case, I would not use '(former)', as some may understand that as there being a 'Korea' in the present. I would simply use 'Korea', which is the name usually used.

pabloarosado:
Hi @bastianherre, so you are suggesting two changes:
* `Eritrea and Ethiopia` -> `Ethiopia (former)`.
* `United Korea` -> `Korea`.

We have [another issue](https://github.com/owid/etl/issues/156) where we proposed to use `(former)` for all historical regions. But lately I've been thinking that it may not be a good idea. For example, `USSR (former)` may also sound like there is another `USSR` now.

I created [a separate issue](https://github.com/owid/etl/issues/796) for renaming countries, and people were generally quite reticent to do so. But these two changes you suggested are used probably in very few datasets, so it shouldn't be a big problem. If you want, feel free to propose the changes as a new comment in that issue, or in Slack, and (if others agree) I can implement those changes in a separate pull request (for the moment we are not using the `regions` dataset anywhere, so it should be fine to update it again soon). Thanks.

pabloarosado:
Final changes:
* I've implemented all suggestions above.
* I've realised we still had `Caribbean Netherlands`, which, as we mentioned in the past, should not be included as a separate country, since it's the same region as the BES Islands. Therefore, I added `Caribbean Netherlands` in the list of aliases of `Bonaire Sint Eustatius and Saba`.
* I added a README in `docs/data/regions.md`, @danyx23 feel free to take a look.

Issue #825, opened 2023-02-09
Helper function `create_dataset` will not work for explorer steps


pabloarosado:
I think `create_dataset` will not work for explorer steps, given that you have to specify `ds_explorer.add(tb_garden, formats=["csv"])`. I suppose one solution would be to include `formats` as an argument of `create_dataset`.
Additionally (not totally sure if it's a good idea), `create_dataset` could add the `formats=["csv"]` by default if the step is an explorer step.

Marigold:
Passing `formats` works well, doing it implicitly could be tricky. Since this is a one-liner, I'm gonna wait until someone does it as part of their explorer work.

PR #826, opened 2023-02-10
feat(helpers): Add formats to create_dataset, so that it can be used …

pabloarosado:
Add `formats` to `create_dataset` function from `etl.helpers`, so that it can be used in explorer steps.

Given that Pablo A is currently working on an explorer step, I thought it would be good to fix this (since explorer steps are not that commonly used).

This PR should fix: https://github.com/owid/etl/issues/825


Marigold (approved):
Looks good!

PR #833, opened 2023-02-13
Create explorer step for Luxembourg Income Study

paarriagadap:
Create explorer step for Luxembourg Income Study. It follows the same structure as [WID](https://github.com/owid/etl/blob/master/etl/steps/data/explorers/wid/2023-01-27/world_inequality_database.py).


Marigold (approved):
Looks solid. (I've only run the step and checked code)

PR #844, opened 2023-02-15
fix(life-exp): complete metadata

lucasrodes:
Enhanced how sources are defined in the metadata. Also added some missing fields.

Closes https://github.com/owid/owid-issues/issues/565


edomt (changes requested):
See https://github.com/owid/owid-issues/issues/565#issuecomment-1433219287

edomt (approved):
All good, thank you!

PR #846, opened 2023-02-16
feature(walkthrough): add explorer step

Marigold:
Add `walkthrough explorers`. It's pretty basic, might need some iterations on real examples.


pabloarosado (approved):
Looks good!

PR #847, opened 2023-02-16
feature(super-fasttrack): add option to load local CSV instead of Google sheet

Marigold:
Many authors are still using "Import CSV" functionality in admin to explore their datasets. It is understandable that they prefer it over Google sheets that require some effort (~5min to set up). Since we need to deprecate csv import soon, we have to make this possible in fast-track too and make it even faster - super fast!

This PR adds new option for uploading your CSV file. Metadata is then inferred from CSV in a similar way as from Google sheet. Datasets created this way have `DRAFT` prefix and their metadata are not editable (which shouldn't matter since metadata is edited at chart level).


pabloarosado (approved):
Nice! I just tested it and only noticed a minor issue, which is that the
```
[Dataset in admin](http://127.0.0.1:8082/admin/datasets/5881)
```
Doesn't work, because the port is wrong (it's using fasttrack's port, not my local grapher's port).

The rest looks good!

Marigold:
> Doesn't work, because the port is wrong (it's using fasttrack's port, not my local grapher's port).

True, you'd have to set env variable `ADMIN_HOST=http://localhost:3030` to make it work.

PR #853, opened 2023-02-20
fix(walkthrough): add some fields to metaata to make dummy dataset from walkthrough work

Marigold:
Before this `walkthrough grapher --dummy-data` generated grapher dataset that was raising error about missing properties like title. This PR adds missing metadata when we use flag `--dummy-data`.

(I could have fixed just the dummy dataset, but someone would have overwritten it soon anyway, so it's better to fix it when being generated)


lucasrodes (approved):
LGTM! I tested it and works for me.

PR #855, opened 2023-02-20
fix(data): fix failing grapher upsert steps

Marigold:
Couple of minor fixes needed to run `etl grapher --grapher` successfully on staging / live. This is a prerequisite to automated deploys.

It also fixes bug with excessive number of open connections (by using shared engine).

In addition to changes in this PR, I'll have to run the following SQLs against live DB (should be harmless)

```
update datasets set shortName = 'global_carbon_budget_additional'
where name = 'Global Carbon Budget (Global Carbon Project, v2021b)' and shortName = 'global_carbon_budget_global';

update datasets set name = 'United Nations Office on Drugs and Crime - Intentional Homicides'
where name = 'United Nations Office on Drugs and Crime - Intentional Homicides (2022)' and version = '2023-01-04';
```

### DIFF
--- a/etl/steps/data/grapher/energy/2022-07-29/primary_energy_consumption.py
+++ b/etl/steps/data/grapher/energy/2022-07-29/primary_energy_consumption.py
@@ -13,6 +13,9 @@ def run(dest_dir: str) -> None:
     garden_dataset = catalog.Dataset(DATASET_PATH)
     dataset = catalog.Dataset.create_empty(dest_dir, garden_dataset.metadata)
 
+    # Backward compatibility
+    dataset.metadata.title = "Primary energy consumption (BP & EIA, 2022a)"
+
     # There is only one table in the dataset, with the same name as the dataset.
     table = garden_dataset[garden_dataset.table_names[0]].reset_index().drop(columns=["gdp", "population", "source"])
     dataset.add(table)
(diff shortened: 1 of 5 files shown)


etl/steps/data/grapher/energy/2022-07-29/primary_energy_consumption.py
```diff
@@ -13,6 +13,9 @@ def run(dest_dir: str) -> None:
     garden_dataset = catalog.Dataset(DATASET_PATH)
     dataset = catalog.Dataset.create_empty(dest_dir, garden_dataset.metadata)
 
+    # Backward compatibility
+    dataset.metadata.title = "Primary energy consumption (BP & EIA, 2022a)"
```
pabloarosado, line 17: Maybe better call it "Primary energy consumption (BP & EIA, archive)", to avoid confusion (although this step should not be used to create a grapher dataset ever again).

pabloarosado (approved):
Looks good!

etl/steps/data/grapher/energy/2022-07-29/primary_energy_consumption.py
Marigold: Good idea, thanks.

PR #856, opened 2023-02-22
enhance(grapher_import): upsert to S3 in addition to data_values

Marigold:
When running `etl --grapher`, push to S3 in addition to upserting to `data_values`. Stop syncing datasets from ETL in `datasync` service. After we make sure loading data from S3 works well, we can remove upserting to data_values and prune all ETL datasets from there. That should make `data_values` significantly smaller and give us better visibility into what's writing there and still needs to be migrated.

Running `ENV=.env.prod etl ... --grapher` now does the following:

1. Upsert to `data_values`
2. Upload data to https://catalog.ourworldindata.org/baked-variables/live_grapher/data/[variable_id].json (and metadata)
3. Set `dataPath` and `metadataPath` of all upserted variables
4. (Skip `datasync`)

Grapher admin will fetch data and metadata from S3 instead of data_values.

Note that we're **pushing to S3** even when running locally (to bucket `s3://owid-test`). That could slow things down a bit. If this becomes too annoying I can add option to have local variables catalog on disk (it'd make sense to do it as a separate PR anyway as it needs also changes to grapher codebase).

Shortly after this gets merged and tried out, there'll be another PR that removes upserting to `data_values`.

### DIFF
--- a/backport/datasync/data_metadata.py
+++ b/backport/datasync/data_metadata.py
@@ -46,6 +46,10 @@ def variable_data_df_from_s3(engine: Engine, data_path: str) -> pd.DataFrame:
     if df.empty:
         return empty_df
 
+    return add_entity_code_and_name(engine, df)
+
+
+def add_entity_code_and_name(engine: Engine, df: pd.DataFrame) -> pd.DataFrame:
     # add entities from DB
     q = """
     SELECT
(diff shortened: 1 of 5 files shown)


Marigold:
> Looks good. Although I'm a bit concerned about writing to the test bucket when working on local grapher.

@pabloarosado that's an extremely valid concern! Writing to S3 when doing local development is indeed really weird, but it simplifies it a lot. (I think we do the same for GDocs publishing if I'm not mistaken). Anyway, if we find it's problematic (e.g. because of performance) we can switch it for local dir.

> Couldn't we have clashes of variable ids if different people are creating different datasets in their local grapher?

By default it adds your [laptop's name to the S3 path](https://github.com/owid/etl/pull/856/files#diff-65c3044b26e69b8a41f84a5c6dd8fe22ceb6f6c91b7b1815186968c2177dc5c5R51), so I'm hoping that could take care of clashes.

PR #857, opened 2023-02-22
Update faostat snapshots

pabloarosado:
Main changes of this PR:
* We stop using the old walden ingest script to check for updates in FAOSTAT data, and instead create snapshots.
* In each snapshot folder now there are only metadata files, and no scripts.
* The script to check for updates and to create new snapshots is in `etl/scripts/faostat/create_new_snapshots.py`, next to the old `etl/scripts/faostat/create_new_steps.py` (which needs to be refactored after these changes, but in a different PR).
* Added all snapshot metadata files of FAOSTAT domain datasets.

**Note to reviewers**: Only two files needs to be reviewed, namely the `faostat.md` README, and the `create_new_snapshots.py` script. All other files are just auto-generated snapshot metadata files.


snapshots/faostat/2023-02-22/faostat_ef.zip.dvc
```diff
@@ -0,0 +1,31 @@
...
+  namespace: faostat
+  short_name: faostat_ef
+  name: 'Land, Inputs and Sustainability: Fertilizers indicators - FAO (2022)'
+  description: 'The FAOSTAT domain Fertilizers Indicators provides information on
+    three rations: a) the ratio between the totals by nutrient of agricultural use
+    of chemical or mineral fertilizers, reported in the FAOSTAT domain �Inputs/Fertilizers
+    by Nutrient� for nitrogen (N), phosphorus (expressed as P2O5) and potassium (expressed
+    as K2O) and the area of cropland reported in the FAOSTAT domain �Inputs/Land Use�;
+    b) The ratio of fertilizers use and the annual population reported in the FAOSTAT
+    domain �Population and Employment/Population�; and c) The ratio of fertilizers
```
Marigold, line 11: Are these unknown characters supposed to be there?

etl/scripts/faostat/create_new_snapshots.py
```diff
@@ -0,0 +1,377 @@
...
+    def metadata(self):
+        """
+        Snapshot-compatible view of this dataset's metadata.
+
+        """
+        if self._dataset_metadata["DatasetDescription"] is None:
+            # Description is sometimes missing (e.g. in faostat_esb), but a description is required in index.
+            self._dataset_metadata["DatasetDescription"] = ""
+            print(f"WARNING: Description for dataset {self.short_name} is missing. Type one manually.")
+        return {
```
Marigold, line 191: Nitpick - Is there a reason why return dict and not `SnapshotMeta`? If we could use it we could leverage its `.uri` method.

etl/scripts/faostat/create_new_snapshots.py
```diff
@@ -0,0 +1,377 @@
...
+    def __init__(self):
+        # Assign current date to additional metadata.
+        self.publication_date = dt.datetime.today().date()
+        self.publication_year = self.publication_date.year
+
+        # Initialise the combined metadata to be downloaded from FAOSTAT.
+        self.faostat_metadata = None
+
+    @property
+    def metadata(self):
```
Marigold, line 272: (Methods are missing type hints, but we don't enforce them anymore so up to you.)

Marigold (approved):
Some nits, in general looks good.

snapshots/faostat/2023-02-22/faostat_ef.zip.dvc
pabloarosado: Hmmm probably not, I'll check what happens in the upcoming steps and how these descriptions compare to the old descriptions. There's already another issue for improving description formats.

etl/scripts/faostat/create_new_snapshots.py
pabloarosado: I basically adapted the code we used to have in walden ingest, so I didn't put much additional effort. But I'll try to add type hints now.

etl/scripts/faostat/create_new_snapshots.py
pabloarosado: Good idea, I've just done that (to avoid making many changes, I created an additional property).

PR #858, opened 2023-02-23
Update script to create faostat steps

pabloarosado:
* Update script that creates new FAOSTAT steps, to use Snapshot instead of Walden.
* TODO (in separate PRs):
  * Make script work for explorers step.
  * Remove lines related to Walden (which are still needed until the data steps are updated).
  * Apply script to create new FAOSTAT steps.

Side note: The logic of the scripts `create_new_snapshots` and `create_new_steps` should be at some point generalized to work for any steps, not just for FAOSTAT.


etl/scripts/faostat/create_new_steps.py
```diff
@@ -81,12 +52,16 @@ def get_channel_from_dag_line(dag_line: str) -> str:
...
+    elif dag_line.startswith("snapshot://"):
+        channel = "snapshot"
     elif dag_line.startswith("walden://"):
         channel = "walden"
-    elif dag_line.startswith("grapher://"):
+    elif dag_line.startswith("data://grapher"):
         channel = "grapher"
     else:
-        raise ValueError("dag line not understood")
+        channel = "other"
```
lucasrodes, line 64: Perhaps we could have something like

```python
prefix_to_channel = {
  "data://garden": "garden",
  "data://meadow": "meadow".
...
}

channel = "other"
for prefix, channel_ in prefix_to_channel.items():
   if dag_line.startswith(prefix):
        channel = channel_
        break

**Review by lucasrodes — APPROVED** [review]
LGTM!

As a side note, maybe at some point, we can consider having an [Enum](https://docs.python.org/3/library/enum.html) class called `Channel` or something like that. It would look like this:

```python
class Channel(Enum):
    SNAPSHOT = "snapshot"
    MEADOW = "meadow"
    GARDEN = "garden"
    EXPLORERS = "explorers"
    GRAPHER = "grapher"
    WALDEN = "walden"
```

This way, we would not need to write the channel names every time (which could be prone to human typos)

etl/scripts/faostat/create_new_steps.py
pabloarosado: Thanks Lucas, yes, I think we could recycle some of the stuff in this script, improve it, and use it to create the tool that automatically (or semi-automatically) updates data steps.

PR #859, opened 2023-02-23
Create new faostat meadow steps

pabloarosado:
Create new faostat meadow steps.

Only two files need review (`shared.py` and `faostat_metadata.py`). These files are based on the previous version of the dataset, after adapting them to the recent ETL changes (and to read from snapshot instead of walden) plus some minor changes to accommodate for a few new issues found in the data. So these files don't require a detailed review.


spoonerf (approved):
LGTM and works for me locally!

PR #865, opened 2023-02-28
Update ember electricity data and energy dataset

pabloarosado:
This PR:
* Adds snapshots of the latest yearly electricity data from Ember.
  * In the end, the European Electricity Review 2023 is added to garden, but not used in grapher or elsewhere, since it lacks data between 1990 and 1999, so we still use the 2022 release.
* Creates new steps of energy datasets and adapts them to the latest changes in ETL.
* Makes all dataset dependencies explicit.
* Creates new variables combining solar and wind.
* Uses new population dataset and reduces redundant code that used to create population from historical regions.
* Cleans up other unused code.

For reviewers:
Most of the code is duplicated from the previous version, and adapted for the various changes mentioned above. So there's no need to do a thorough code review. I'd suggest checking:
* [x] @lucasrodes Changes in the active and archive dag.
* [ ] @lucasrodes Simple fields in metadata (e.g. `date_accessed`, `title`, etc.), but no need to check dataset and variables descriptions.
* [ ] @Marigold Quick look through `run` function of data steps, without going through all functions in the step or shared modules.


etl/steps/data/garden/bp/2023-02-20/shared.py
```diff
@@ -0,0 +1,106 @@
...
+
+from etl.data_helpers import geo
+
+log = get_logger()
+
+CURRENT_DIR = Path(__file__).parent
+VERSION = CURRENT_DIR.name
+
+
+def add_population(
```
Marigold, line 15: Could this be somewhere in `etl.data_helpers`? Looks pretty useful for other datasets too.

Marigold (approved):
Massive... run functions look good, it runs fine locally.

dag/energy.yml
```diff
@@ -219,4 +229,56 @@ steps:
...
+    - data://garden/smil/2017-01-01/global_primary_energy
+    - data://garden/bp/2022-12-28/statistical_review
+  data://grapher/energy/2022-12-28/global_primary_energy:
+    - data://garden/energy/2022-12-28/global_primary_energy
+  # BP - Energy mix 2022.
+  data://garden/bp/2022-12-28/energy_mix:
+    - data://garden/bp/2022-12-28/statistical_review
+    - data://garden/owid/latest/key_indicators
+    - data://garden/wb/2021-07-01/wb_income
+  data://grapher/bp/2022-12-28/energy_mix:
```
lucasrodes, line 282: I might be missing something out, but I see that several `grapher` recipes are not in `archive/energy.yml`, including:

- `data://grapher/bp/2022-12-28/energy_mix`
- `data://grapher/energy/2022-12-28/global_primary_energy`
- `data://grapher/energy/2022-12-28/uk_historical_electricity`
- `data://grapher/energy/2022-12-28/primary_energy_consumption`
- `data://grapher/energy/2022-12-28/fossil_fuel_production`
- `data://grapher/energy/2022-12-28/electricity_mix`

Is this expected or should we put these in the archived dag too?

etl/steps/data/garden/bp/2023-02-20/energy_mix.meta.yml
```diff
@@ -0,0 +1,61 @@
+dataset:
+  namespace: bp
+  title: Energy mix (BP, 2023)
```
lucasrodes, line 3: I have used `Dataset Name - Source (Year)` formatting in the past, but I see that you have been using `Dataset Name (Source, Year)` in this PR. Is this the preferred format at the moment? I don't have any strong preference, just double checking!

etl/steps/data/garden/bp/2023-02-20/energy_mix.meta.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    Additional metrics have been calculated by Our World in Data:
+    - Annual change in energy consumption by source: this is calculated as the difference from the previous year.
+    - % of total primary energy: calculated as each source's share of primary energy (direct energy and primary energy using the substitution method) from all sources.
+    - Per capita energy by source: calculated as primary energy consumption by source, divided by population.
+
+    Per capita figures have been calculated using a population dataset that is built and maintained by Our World in Data, based on <a href="https://ourworldindata.org/population-sources">different sources</a>.
+
```
lucasrodes, line 58: nice!

etl/steps/data/garden/bp/2023-02-20/energy_mix.meta.yml
lucasrodes: Just checked the metadata guidelines on Notion, resolved!

lucasrodes (approved):
LGTM

dag/energy.yml
pabloarosado: Hi @lucasrodes you are totally right, however those datasets still exist in grapher db. I think the safest order of events would be to:
1. Merge this PR.
2. Archive the unnecessary grapher db datasets.
3. Archive the grapher etl dataset (and then VersionTracker will warn about other unnecessary dependencies to be archived).

Otherwise, we could find an issue during the PR review and need to fall back to the old grapher db dataset (although this is unlikely).
In the future, VersionTracker should have the option to check grapher db as well, but for now this has to be done manually. I'll do all the necessary archiving after merging!

etl/steps/data/garden/bp/2023-02-20/energy_mix.meta.yml
pabloarosado: Hey @lucasrodes you are also right about being confused here. We have contradictory information in [the Notion metadata guidelines](https://www.notion.so/owid/Metadata-guidelines-29ca6e19b6f1409ea6826a88dbb18bcc?pvs=4) (“Human-readable dataset title (Source name, Year)”.) and on the short description in the grapher web admin ("Short name for this dataset, followed by the source and year. Example: Government Revenue Data – ICTD (2016)").
Given that the dataset title is shown only internally, we can for now allow for both options. But feel free to raise the issue somewhere so we decide about which option is preferrable, thanks.

etl/steps/data/garden/bp/2023-02-20/shared.py
pabloarosado: Thanks. I think it would be good to give this function a bit of extra thought before sharing it in data helpers, to avoid what happened in the past with datautils.geo functions, so I'll leave it for a different PR.

dag/energy.yml
lucasrodes: makes sense, thanks for clarifying!

PR #866, opened 2023-02-28
feat(grapher_helpers): Ensure short and long sources citation ends in a period

pabloarosado:
So far we have not been consistent about whether sources short/long citations end in a period.
This PR ensures that the ETL grapher step adds a '.' at the end of the source name and `published_by` fields.

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -335,7 +335,7 @@ def combine_metadata_sources(sources: List[catalog.Source]) -> catalog.Source:
 
         # Ensure combined source names (short and long versions) end in a period.
         if ((attribute == "name") or (attribute == "published_by")) and not combined_value.endswith("."):
-            combined_value += '.'
+            combined_value += "."
 
         # Instead of leaving an empty string, make any empty field None.
         if combined_value == "":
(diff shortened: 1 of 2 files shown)


edomt:
Thanks, Pablo.

I tend not to perceive this kind of metadata as actual sentences, so I hadn't thought about adding periods to these fields.

I could see this resulting in awkward outputs sometimes, e.g., when the source's name is relatively short (`World Bank.`)

Do you have examples of cases where this has been a problem so far?

pabloarosado:
Hi @edomt, I was just looking at different random charts, and noticed that sometimes we add a period and sometimes we don't, but I don't think our choice is based on any criterion, but rather random. It's not at all a problem anywhere, but it's just a matter of consistency.
With this PR, by default, new charts will have a period at the end of the source, which I guess in most cases is the right thing.
We could alternatively decide to **not** have a period at the end of sources. If so, I could also change this PR so that a final period is automatically removed.

pabloarosado:
Following our discussion in [this Slack thread](https://owid.slack.com/archives/C0193RW5E2J/p1677570202469339), we have decided that we should not have periods at the end of short or long source citations. So I'll close this PR without merging.

PR #868, opened 2023-02-28
data(excess-mortality): update

lucasrodes:
Update Excess mortality data:

- New snapshot version: `2023-02-28`
- New countries available in Kobak & Karlinsky excess data: `Turkey` and `United Arab Emirates`
- Updated `version`, `date_accessed` and `publication_date` fields in metadata.


Marigold (approved):
Looks good. You could avoid copying versions if you used `latest` for snapshots too. Then you would just run snapshot scripts which would download latest data and replace `*.csv.dvc` files. I don't think we have to store recipes for old versions in the repo (they'll be available in git history).

lucasrodes:
will prune that in the next update, thanks @Marigold !

PR #869, opened 2023-03-02
fix(data): fix wrong dependency version for country_profile

Marigold:
Our nightly full build was failing because of outdated dependency. We've updated life expectancy dataset, but forgot to update it country profiles code. Switching to `PathFinder` should fix it.


spoonerf (approved):
LGTM! If this `country_profile` namespace is continually annoying we could think about archiving it for now. It was a bit of an experiment and we might use it in the future, but if it's wasting your time feel free to archive it!

PR #874, opened 2023-03-03
data(vdem): Add updated data for V-dem index

Marigold:
Migrate our existing V-dem dataset from grapher to ETL and copy all metadata there. We'd normally use fast-track to upload it, but the file has 35mb and won't fit into Google Sheets. It would also make future updates easier.

(I've noticed that some variables like [this one](https://owid.cloud/admin/variables/447875) and [this one](https://owid.cloud/admin/variables/419298) have identical display names. Is this intentional @bastianherre?)


bastianherre (changes requested):
This almost all looks great, thank you very much!

I only found two issues:

On the one hand, there was an error in my code which affects the population-based metrics. I have sent you a revised table in Slack which should correct this issue.

On the other hand, in [this](https://staging.owid.cloud/admin/charts/5518/edit) and [this](https://staging.owid.cloud/admin/charts/5517/edit) chart, the numeric values seem now to be treated as strings. How could we fix that?

PS: The sometimes identical display names are intentional. Thanks for checking in!

Marigold:
@bastianherre I've uploaded your new CSV and deployed to staging. If you're happy with the results, please approve this and I'll deploy it to production.

bastianherre (approved):

bastianherre:
@Marigold! The charts affected by the two issues all look good now, you can deploy to production. I will keep you posted if I catch any additional issues as I update our related writing. Thank you very much!

PR #885, opened 2023-03-06
fix(data): use vdem ETL dataset instead of backport

Marigold:
We've recently migrated V-dem dataset to ETL. Country profiles were still using backports which caused weird circular dependency. Switch to ETL version in this PR.


spoonerf (approved):
LGTM! Will bear this in mind for future updates

Marigold:
> LGTM! Will bear this in mind for future updates

Yeah, me too as this was totally my fault :).

PR #889, opened 2023-03-08
data(vdem): add variable metadata

Marigold:
Adding metadata for vdem dataset.


bastianherre (approved):
Thank you very much for recovering these descriptions @Marigold! They all look good!

PR #891, opened 2023-03-08
experiment(red_list): changing name of fasttrack dataset

spoonerf:


spoonerf:
Hey @Marigold is it possible to change fasttracked variable names via the etl or elsewhere after they have been uploaded?

Marigold:
> Hey @Marigold is it possible to change fasttracked variable names via the etl or elsewhere after they have been uploaded?

(I'm assuming you're using "super"-fast-track and uploaded it via csv). In general, changes to **titles** are fine, but you'd have to be very careful when changing short names (I'd avoid that at all... they're not shown anywhere anyway). Changing short name could raise an error if the old variable is using any charts.

As for this PR, changing `DRAFT Table 5  Threatened species in each major group by country - show all (1)` to `Threatened species by country (IUCN, 2022)` in YAML (both in snapshot and grapher step) works fine, but as I said, I'd keep the original short names. Same for variables.

snapshots/fasttrack/latest/table_5__threatened_species_in_each_major_group_by_country__show_all__1.csv.dvc
```diff
@@ -1,10 +1,9 @@
 meta:
   namespace: fasttrack
-  short_name: table_5__threatened_species_in_each_major_group_by_country__show_all__1
```
Marigold, line 3: Even though it looks ugly, changing `short_name` would create a new dataset (which is problematic because the existing one has charts). So either you keep it as it is or you'll have to match old variables to the new ones.

spoonerf:
> 

Ah I see - I haven't used fasttrack before and I didn't realise I was using super fasttrack. Is it possible to delete this dataset so I can upload it via regular fast track?

Marigold:
Deleting existing datasets is not possible yet, so just create a new one for now and I'll delete this one when I implement it.

spoonerf:
> Deleting existing datasets is not possible yet, so just create a new one for now and I'll delete this one when I implement it.

Okay that's good to know, thanks!

PR #894, opened 2023-03-08
data: Add microprocessor trends

edomt:
Another tiny dataset to add, sourced from https://github.com/karlrupp/microprocessor-trend-data/.

Hopefully, everything looks good this time! 🤞🏻


Marigold (approved):
Yup, this time everything's perfect. It might be worth having a discussion about using ETL vs using fast-track. We could theoretically go all in ETL, but with less boilerplate code than is needed now.

PR #895, opened 2023-03-09
Create dataset from trust questions in the World Values Survey

paarriagadap:
Data issue here https://github.com/owid/owid-issues/issues/987


paarriagadap:
I think I got vendor/owid-catalog-py when I rebase

paarriagadap:
I also wanted to squash, but many changes beyond this were included, so I left it as it is now

Marigold (approved):
Looks good! Code works locally (haven't checked data itself, only metadata). Could you update owid-catalog with
```
cd vendor/owid-catalog-py
git pull
```
and then add a commit **in ETL** (as usual). That should get rid of those owid-catalog-py commits.

PR #897, opened 2023-03-09
fix(PathFinder): support for private steps

Marigold:
Add support for private steps in `PathFinder`. I could have done it more explicit by e.g. using `load_dependency("short_name", private=True)` which would be cleaner, but as long as we have tests even this implicit approach should be fine.


pabloarosado:
Hey, I've created [an alternative PR](https://github.com/owid/etl/pull/904), feel free to check it out and choose the solution you find best :-).

pabloarosado:
Hey, I've created [an alternative PR](https://github.com/owid/etl/pull/904), feel free to check it out and choose the solution you find best :-).

pabloarosado:
Hey, I've created [an alternative PR](https://github.com/owid/etl/pull/904), feel free to check it out and choose the solution you find best :-).

PR #898, opened 2023-03-09
Flunet update - testing snapshot update

spoonerf:


spoonerf:
Hey @Marigold - I'm just experimenting with updating the snapshot here for these two datasets as it's something we will hopefully be able to automate soon. 

I ran `python snapshots/who/latest/fluid.py --skip-upload` and `python snapshots/who/latest/flunet.py --skip-upload`. Do I need to run it without the `--skip-upload` to get the CI to pass?

Marigold:
@spoonerf you have to run it **without** `--skip-upload` flag to upload it to S3 (where CI can then download it from). To be honest, that flag is pretty useless and just confusing.

That CI failure with "Warning: Checkout failed! exit status 128 (Attempt 1/3 Retrying in -1.334129189s)" is a bit weird. It could either be that github is down or that your commits in this PR are a bit messy (there are 34 commits for just 2 lines of code). Could you try applying your changes to the current master branch? (for instance by running `git checkout master` then `git reset --hard origin/master` then `git checkout -b flunet-update` and then creating PR)

spoonerf:
Ah I see - that's a good GH tip, thanks! 

That would be great if we could set up a CRON job - I think a daily update would be ideal, if that's not too much?

Marigold:
Sounds good, I've created an issue https://github.com/owid/etl/issues/900 and will work on it next week.

PR #899, opened 2023-03-09
feat(flu): experimenting with updating the snapshot

spoonerf:


Marigold (approved):
Datadiff looks good. (It only shows "changed shape" for all columns, we could add more interesting statistics if you'd find them helpful, like % of added / changed values)

spoonerf:
> 

That's good to hear! It would be great to see % of changed values. I have a feeling the most recent datapoints are often adjusted with every update, it would be good to know to what extent that happens.

PR #902, opened 2023-03-09
data(gho): Add age-std suicide rate metrics

lucasrodes:
Importing the complete GHO dataset is complex; we are waiting for them to update their platform and publish a new release.

Currently, we do not have any GHO dataset in ETL, which prevents us from rapidly editing it and its metadata. In the meantime, I have added a subset of their current dataset, with the metrics relevant to the Suicides topic page redesign (which I am currently working on with @saloni-nd).


spoonerf (commented):
Code looks good to me but when I run it locally the values are showing up as categories rather than numbers? Is this happening for you also?

Screenshot 2023-03-09 at 16 58 30

spoonerf:
Ah weird, mine is still like it but if it works for you then that's okay. Otherwise LGTM!

spoonerf (approved):

PR #903, opened 2023-03-09
enhance(ghe): dimensions for cause=self-harm

lucasrodes:
For our projection on the Suicide topic page, we need more granularity for the suicide rates. That is, we need the metric for different age groups.

This PR includes all age groups for metrics with `cause="Self-harm"`.

The Grapher step takes a while to run for this one. I am currently running it, and the dataset should be available soonish on [staging](https://staging.owid.cloud/admin/datasets/5790).


spoonerf (approved):
LGTM - thanks for improving the code!!

PR #904, opened 2023-03-10
Let path finder work with private steps

pabloarosado:
This is an alternative solution to https://github.com/owid/etl/pull/897/files, to let PathFinder handle private steps.

@Marigold I created it with the intention of having a better solution, where `is_private` is an optional argument. Then I realised it wasn't as trivial as initially thought. So feel free to discard this PR and use yours (which is definitely simpler).

Additionally, in either case, please add a `$` at the end of the end of the regular expression of `step_name` when `channel` is not defined in `create_step_name` function (line 267 in this PR). This avoids issues when having a dependency short name that includes another dependency, like `faostat_fbs` and `faostat_fbsh`.


etl/helpers.py
```diff
@@ -180,6 +180,10 @@ def __init__(self, __file__: str):
         if len(self.f.relative_to(paths.STEP_DIR).parts) == 6:
             self.f = self.f.parent
 
+        # Start by assuming the current step is public.
+        # Then, if the step is not found in the dag, but it's found as private, is_private will be set to True.
+        self.is_private = False
```
Marigold, line 185: I like the idea of using `is_private` as an attribute. How about adding it as an optional argument to `__init__` with default `None` for auto-detection? (you could just call `self.step` to set automatically if `is_private is None`)

Marigold (approved):
Happy to go with your approach!

etl/helpers.py
pabloarosado: Thanks! I've just commited a change, please let me know if that's what you meant.

PR #906, opened 2023-03-10
Metadata cleanup and other minor improvements

Marigold:
Minor improvements to ETL that accumulated this week.


lucasrodes (approved):
LGTM! Just wondering, how come we are removing the field `short_name` for tables?

Marigold:
> LGTM! Just wondering, how come we are removing the field `short_name` for tables?

`short_name` is taken from key in the YAML file, we don't need it repeated. For instance

```
tables:
  statistical_review:  # this is the short_name
    short_name: "statistical_review"  # this is unnecessary
```

PR #918, opened 2023-03-13
Data/suicides ghe ages

lucasrodes:
- Metadata improvements
- Define broad age groups. The selection is inspired by [this chart](https://ourworldindata.org/grapher/suicide-rates-by-age-detailed-who). To do this, we have to "un-normalize" rate metrics with population values, aggregate and then normalize again.


dag/health.yml
```diff
@@ -30,6 +30,7 @@ steps:
     - data://garden/reference
   data://garden/who/2022-09-30/ghe:
     - data://meadow/who/2022-09-30/ghe
+    - data://garden/un/2022-07-11/un_wpp
```
Marigold, line 33: Do we need this? (it doesn't have any effect on the memory problem, just asking)

lucasrodes: Yeah, it is kind of hidden, but the function `add_population` uses this dataset.

spoonerf (approved):
This looks great - thanks for improving the metadata as well.

PR #919, opened 2023-03-14
feature(scripts): script for comparing data parity

Marigold:
Small script for comparing data files in S3 and baked files on live. They should be identical to safely merge [this grapher PR](https://github.com/owid/owid-grapher/pull/1997). You don't have to try it on your local, just check the code for any obvious errors, please.

They're identical except of variable 468301 used in [this chart of road traffic deaths](https://ourworldindata.org/grapher/road-traffic-deaths-to-alcohol). If you check the data with `select * from data_values where variableId = 468301;` you'll see that it contains non-numeric values (e.g. `13 (GB), 23 (NI)`)

```
"value","year","entityId","variableId"
"13 (GB), 23 (NI)",2017,1,468301
"38.5",2017,2,468301
"29",2017,3,468301
"11 - 24",2017,5,468301
"7",2017,6,468301
"13",2017,7,468301
"20 - 25",2017,8,468301
"17 Males, 6.7
Females",2017,9,468301
"24",2017,10,468301
"12.7",2017,11,468301
...
```

even though it seems to work in chart, grapher truncates `13 (GB), 23 (NI)` to `13`, `11 - 24` to 11, etc. It's just a couple of values, so I think it's reasonable to fix this in MySQL directly and replace them with sensible values. What do you think @lucasrodes ?

### DIFF
--- a/scripts/data_parity.py
+++ b/scripts/data_parity.py
@@ -1,0 +1,93 @@
+import gzip
+import json
+from pathlib import Path
+from typing import List
+
+import click
+import pandas as pd
+import structlog
+
+from etl.db import get_engine
+
+log = structlog.get_logger()
+
+engine = get_engine()
+
+
+@click.command()
+@click.argument("s3_path")
+@click.argument("baked_path")
+@click.option("--variables", "-v", type=int, multiple=True)
+def data_parity_cli(s3_path: Path, baked_path: Path, variables: List[int]) -> None:
+    """Compare two sets of generated data files, typically from S3 and baked files from live
+    server. Find all differences between the two sets.
+
+    First sync data from S3:
+
+        aws s3 sync s3://owid-catalog/baked-variables/live_grapher/ playground/s3-live/
+
+    and sync baked data from live
+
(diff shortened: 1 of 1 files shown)


lucasrodes (commented):
Just to confirm that I understand more or less what's happening here:

- If I understand correctly, if no value for `variables` is given it will compare all variables used in charts, right? And these should be available both in `s3_path` and `baked_path` otherwise it will fail.
- Will this fail because of the example you mention (road traffic deaths)? Like, the values in `s3_path` will have the correct values (as exported from ETL), but the ones in `baked_path` will have the truncated ones (as exported from the DB)?


Other than that, it LGTM!

Marigold:
> If I understand correctly, if no value for variables is given it will compare all variables used in charts, right? And these should be available both in s3_path and baked_path otherwise it will fail.

Exactly. Sorry for the lack of documentation! It's a one off thing and I didn't spend that much time on it.

> Will this fail because of the example you mention (road traffic deaths)? Like, the values in s3_path will have the correct values (as exported from ETL), but the ones in baked_path will have the truncated ones (as exported from the DB)?

It's not gonna fail, but log problematic variables.

lucasrodes:
Nice one!

PR #921, opened 2023-03-14
data: Add registered vehicles per 1,000 people (WHO, UNWPP)

edomt:


edomt:
@Marigold Do you know what the `etl-unit-tests` issue is about?

lucasrodes:
@Marigold I think this is because object `Snapshot` (which is a potential return type of `load_dependency`) is not a list-like object (missing `__getitem__` method), and thus the `["population"]` would fail.


@edomt You could replace your code with something like:

```python
# Load population.
population_dataset: Dataset = paths.load_dependency("key_indicators")
population_table: Table = population_dataset["population"]
pop = population_table["population"] / 1000  # Divide by 1000 to get vehicles per 1000 people

# Load registered vehicles.
short_name = "dataset_5676_global_health_observatory__world_health_organization__2022_08"
who_gh_dataset: Dataset = paths.load_dependency(short_name)
who_gh = who_gh_dataset[short_name]
```

That is, forcing the return type of `load_dependency` to be `Dataset`.

Marigold:
Lucas is correct. Type inference doesn't work well with `load_dependency`. You can also just add `# type: ignore` to the problematic line, e.g.

```python
population_table: Table = paths.load_dependency("key_indicators")["population"]  # type: ignore
```

edomt:
Thank you guys for the prompt response!

PR #923, opened 2023-03-14
Add Poverty and Inequality Platform to explorer (+ LIS update)

paarriagadap:
I have introduced a temporal update for the `poverty_inequality` `explorers` step: As the World Bank Poverty and Inequality Platform (PIP) dataset is currently outside the ETL we decided to merge the current version of the exported dataset (in the notebooks repo) to both LIS and WID explorers datasets to create a [comparison data explorer](https://github.com/owid/owid-issues/issues/983).

This is a temporary solution because I plan to move PIP to the ETL [in May](https://github.com/owid/data-research-roadmap/#pablo-a-gantt-chart)

There is also a [2023 LIS update](https://www.lisdatacenter.org/news-and-events/2023-spring-data-splash/) that I include

### DIFF
--- a/etl/steps/data/explorers/poverty_inequality/2023-03-01/poverty_inequality.py
+++ b/etl/steps/data/explorers/poverty_inequality/2023-03-01/poverty_inequality.py
@@ -1,7 +1,8 @@
-"""World Inequality Database explorer data step.
+"""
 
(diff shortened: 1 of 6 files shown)


Marigold (approved):
Looks good! I like your temporary solution and plan to move it to ETL (which is explicitly mentioned near code, nice!).

Issue #925, opened 2023-03-15
Flunet updates sometimes contains invalid rows


Marigold:
We're now updating our flue net automatically, however the snapshots we download occasionally contain invalid rows and we get error

```
# ParserError: Error tokenizing data. C error: Expected 49 fields in line 50053, saw 56
```

If you print those invalid lines with

```python
def foo(line):
    print(line)

pd.read_csv("data/snapshots/who/latest/flunet.csv", on_bad_lines=foo, engine='python')
```

you'd get two invalid lines with wrong number of rows

```
['AFR', 'YR', 'SH', 'FLU_MID_AFR', 'COD', 'Democratic Republic of the Congo', '2021-01-18', '2021', '3', '2021-01-17', '2021', '3', 'NOTDEFINED', '25', '25', '2', '', '0', '', 'NOTDEFINED', '1753', '865', '0', '0', '8', '0', '', '0', '', '0', '', '8', '', '', '0', '', '0', '4', '4', '12', '', '3', '', '', '', '', '', '', '0', '', '', '', '', 'RU', '201337', '201337']
['EUR', 'NH', 'NH', 'FLU_SW_EUR', 'MLT', 'Malta', '2018-02-19', '2018', '8', '2018-02-18', '2018', '8', 'SENTINEL', '0', '', '0', '0', '0', '', '', '0', '', '', '', '0', '', '', '0', '', '0', '0', '0', '0', '', '', '', '', '', '', '', '', '', '', '', '3', '', '', '', '', '', '', '', '', 'ParaInflu 1', '', 'It seems that Influenza A[/H1](https://file+.vscode-resource.vscode-cdn.net/H1) started slowly in South of the country (2 A[/H1](https://file+.vscode-resource.vscode-cdn.net/H1) in Gabes, Sfax and Djerba) , in total 6 A[/H1](https://file+.vscode-resource.vscode-cdn.net/H1) viruses were isolated and we wait for seology to confirm the subtype.', 'TN', '200546', '200546']
```

@spoonerf any idea what should we do? We could ignore them (assuming there are only a few) or abort automatic update if we see that the file is unreadable.

Marigold:
It's also strange that some successful downloads of fluid are way smaller than they should be (~130mb). This [snapshot has for instance](https://github.com/owid/etl/blob/master/snapshots/who/latest/fluid.csv.dvc#LL24C9-L24C17) just ~25mb. I checked it and it only has 165099 lines compared to expected 865988.

[...]

PR #926, opened 2023-03-15
fix(flu): raise an error when creating snapshot if CSV cannot be parsed

Marigold:
Make sure we're not updating snapshots that are not working. This should hotfix https://github.com/owid/etl/issues/925 until we either decide what to do or until provider fixes their data. It also rolls back fluid to the latest working file.

It has also checks on CSV size since sometimes we're getting valid files which are [however much smaller](https://github.com/owid/etl/issues/925#issuecomment-1470365608).


spoonerf (approved):
LGTM!

PR #933, opened 2023-03-20
feat(happiness): adding latest happiness report data

spoonerf:


etl/steps/data/meadow/happiness/2023-03-20/happiness.meta.yml
```diff
@@ -0,0 +1,37 @@
+dataset:
+  title: World Happiness Report
```
lucasrodes, line 2: I would probably use the format `Name (year, source)`. However, I am unsure about the source name, seems like dataset name = source name. Perhaps we can go with "World Happiness Report (2023)" then.

etl/steps/data/meadow/happiness/2023-03-20/happiness.meta.yml
```diff
@@ -0,0 +1,37 @@
...
+  - url: https://worldhappiness.report
+  - date_accessed: 2023-03-20
+tables:
+  happiness:
+    title: World Happiness Report
+    variables:
+      country:
+        title: Country name
+        short_unit: ''
+        unit: ''
```
lucasrodes, line 20: I haven't checked, but if we set `country` (along with `year`) as indices in `garden`, does this disappear? Like, there will not be a "country" variable.

lucasrodes (changes requested):
Minor changes suggested

PR #936, opened 2023-03-22
enhance(gho): suicide death rate

lucasrodes:
Add the suicide death rate ratio between males and females, using WHO GHO data.

This dataset has data available since 2000 and 2019.


Marigold:
@lucasrodes it looks like you've manually deployed this dataset which already exists. The problem is that there are automatic deploys that overwrite any manual updates of **existing datasets**. I've merged this to make ETL CI pass again, please open a new PR if you need to make any changes (@spoonerf please make a review even though it's been already closed)

spoonerf (commented):
LGTM!

PR #941, opened 2023-03-23
fix(ozone): metadata

lucasrodes:
This PR:

- Improves metadata for datasets:
  - [Ozone hole area and concentration (NASA, 2023)](https://owid.cloud/admin/datasets/5912)
  - [Consumption of controlled substances (UNEP, 2023)](https://owid.cloud/admin/datasets/5925)
- Adds a new variable, `consumption_rel_1986`, which normalizes the variable `consumption` value so that it equals 100 for the year 1986. This should allow us to update this chart: https://ourworldindata.org/grapher/ozone-depleting-substances-index

### DIFF
--- a/etl/steps/data/garden/nasa/2023-03-06/ozone_hole_area.meta.yml
+++ b/etl/steps/data/garden/nasa/2023-03-06/ozone_hole_area.meta.yml
@@ -50,10 +50,10 @@ tables:
         short_unit: "km²"
         unit: "square kilometres"
         description: |
-          Ozone hole area mean (07 September -- 13 October) for each year.
+          Ozone hole area mean (07 September - 13 October) for each year.
       mean_hole_concentration:
         title: Mean daily concentration
         short_unit: "DU"
         unit: "Dobson Units (DU)"
         description: |
-          Minimum ozone (21 September -- 16 October) for each year.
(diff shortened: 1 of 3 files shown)


spoonerf (commented):
Looks good code-wise but I wonder if it the index or world totals should be calculated for 2022 as there are so few countries with data for this year?

It is perhaps a little misleading that the index goes down to 0 for 2022 - when it's only really due to missing data, if I understand it correctly.

spoonerf:
> Looks good code-wise but I wonder if it the index or world totals should be calculated for 2022 as there are so few countries with data for this year?
> 
> It is perhaps a little misleading that the index goes down to 0 for 2022 - when it's only really due to missing data, if I understand it correctly.

Seems like this was also an issue in [the existing chart] (https://ourworldindata.org/grapher/ozone-depleting-substances-index)- the value for 2018 is shown as 0.3 here but now they data for 2018 shows 1.6. Perhaps it is better to not include the latest year when calculating the index, or world total?

lucasrodes:
Thanks, Fiona ー That's a very reasonable point! I have removed 2022 data for regions.

spoonerf (approved):
Looks good!

PR #952, opened 2023-03-28
enhance(chart_revision): error message, simplified API

lucasrodes:
- Improves a slightly confusing error message.
- Simplifies the submission of chart revisions with the new function `create_and_submit_charts_revisions`. This aims to improve the experience of users that want to submit chart revisions programmatically. In such cases, users have a dictionary mapping old to new variable IDs and want to use this to create and submit the necessary chart revisions.


pabloarosado (approved):
Looks good! Thanks for this new function, I'll use it straight away!

PR #953, opened 2023-03-28
fix(food_explorer): load correct version of faostat_food_explorer

Marigold:
Our [nightly build](https://buildkite.com/our-world-in-data/etl-full-private-rebuild-nightly-master/builds/220#01872596-e464-467e-b707-05c235ecc0c0) failed because food_explorer tried to load wrong version of garden dataset.

Our DAG defines it as
```
data://explorers/owid/latest/food_explorer:
    - data://garden/faostat/2022-05-17/faostat_food_explorer
```
but code in `owid/latest/food_explorer` tries to load **the latest** version which has a table with different name. This PR switches to PathFinder which loads the correct version found in DAG.


pabloarosado (approved):
Ah thanks a lot, I was planing to work on the explorer step after creating the grapher steps, but forgot that the old explorer step would cause issues in production.

PR #959, opened 2023-03-29
Adding deaths from drug use disorders from Global Burden of Disease

spoonerf:
Max wants it as a chart so just adding this as a temporary measure until we have the full GBD in the etl.


spoonerf:
@Marigold - Just single variable that I need to add for a chart that Max wants. I just had a feeling that the other `ihme_gbd` steps were currently not included in the automatic update of live as they are 'complicated'. 

I'm not sure if this particular dataset would still automatically be integrated into live once merged?

Marigold (approved):
Looks good!

> I'm not sure if this particular dataset would still automatically be integrated into live once merged?

It should be.

PR #966, opened 2023-03-30
Correct known faostat anomalies

pabloarosado:
I have checked all previously known anomalies in FAOSTAT data. Some have been removed, some remain, and new anomalies appeared (the new ones will be handled in a separate PR).

Previously we had a dictionary with the item code, element code, etc. of the rows affected by anomalies. However:
* For certain anomalies the json format was very limiting.
* There was no easy way to check if, after an update, the anomalies are still there.

In this PR I implemented an alternative way to handle anomalies. The main file to review is `etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py`. All others are just adaptations to use this new method.


lucasrodes:
@pabloarosado unsure why but it seems that there are modifications to files in this PR that should not be there. E.g. work on history of war.

pabloarosado:
Hey @lucasrodes I've fixed the conflicts, and the PR is ready for review.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
+"""Detected anomalies.
+
+This module contains a class for each data anomaly detected. The class has the following methods and properties:
```
lucasrodes, line 3: I'd slightly rephrase this as:

> This module contains a class for each type of data anomaly detected.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
...
+* description: A human-readable text that describes the anomaly.
+  NOTE: The description will be added to the dataset metadata description, and hence will be shown in grapher.
+* checks: A method that ensures the anomaly exists in the data.
+  This is useful to detect if an anomaly has been corrected after a data update.
+* fix: A method that removes the anomaly.
+* inspect: An optional method that plots (in the browser) a visualization that shows the anomaly.
+  It can be used before and after removing the anomalies.
+* handle_anomalies: A helper method that uses all the previous methods in the usual order.
```
lucasrodes, line 11: I'd probably have this text in the methods from the abstract class `DataAnomaly`, too, as one-line docstrings, for instance.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
...
+    """Abstract class for data anomaly."""
+
+    def __init__(self) -> None:
+        pass
+
+    @property
+    @abc.abstractmethod
+    def description(self) -> str:
+        # Description of the anomaly, which will be added to the dataset description.
+        pass
```
lucasrodes, line 40: Instead of using `pass`, I'd raise a `NotImplementedError` exception to ensure this method is never called from the abstract class.

From the python [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError):

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
...
+
+    @abc.abstractmethod
+    def check(self, df: pd.DataFrame) -> None:
+        pass
```
lucasrodes, line 44: Instead of using `pass`, I'd raise a `NotImplementedError` exception to ensure this method is never called from the abstract class.

From the python [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError):

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
...
+
+    @abc.abstractmethod
+    def fix(self, df: pd.DataFrame) -> pd.DataFrame:
+        pass
```
lucasrodes, line 48: Instead of using `pass`, I'd raise a `NotImplementedError` exception to ensure this method is never called from the abstract class.

From the python [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError):

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def inspect(self, df: pd.DataFrame) -> None:
+        pass
```
lucasrodes, line 51: Instead of using `pass`, I'd raise a `NotImplementedError` exception to ensure this method is never called from the abstract class.

From the python [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError):

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

etl/steps/data/garden/faostat/2023-02-22/detected_anomalies.py
```diff
@@ -0,0 +1,358 @@
...
+        df_fixed = self.fix(df=df)
+
+        if inspect_anomalies:
+            log.info("Inspect anomaly after fixing.")
+            self.inspect(df=df_fixed)
+
+        return df_fixed
+
+
+def _split_long_title(text: str) -> str:
```
lucasrodes, line 72: Maybe add a short summary of this function? Something like "Formats a given text to have at most 100 characters per line. Line breaks are implemented with the HTML tag `<br>`.

lucasrodes (approved):
Superb work, Pablo! Very interesting approach to detecting and sanity-checking anomalies. This can serve as inspiration for future work on anomaly detection in other datasets.

Just left minor comments. Other than that, LGTM!

pabloarosado:
Thanks a lot, @lucasrodes! I've implemented all your suggestions.

PR #968, opened 2023-03-31
enhance(walkthrough): fix bug, add field in Grapher step

lucasrodes:
Only affects Grapher step:

- Minor bug in Grapher template: `ds_garden` was used instead of `ds_grapher` when running sanity checks.
- It was assumed that Garden dataset version was the same as Grapher's. I have added an extra field to allow users to define different versions for each step. We are currently doing this in the Garden step for Meadow/Garden.


Marigold (approved):
🎉

PR #975, opened 2023-03-31
data(gapminder): population v7

lucasrodes:
Gapminder released version 7 of their Population dataset. This dataset will be used to update our [population OMM](https://owid.cloud/admin/datasets/5959).

I have checked any changes between v6 and v7; they are all as expected. Changes are due to new estimates from UN WPP 2022, as v6 used UN WPP 2019.

## UN WPP 2022
This version uses UN WPP 2022 estimates (instead of 2019) for data starting in 1950.

## Minor adjustments before 1950
For data before 1950, Gapminder relies on CLIO Infra and US Census Bureau.

We have noticed that Gapminder has made some minor adjustments to data before 1950 (when comparing with v6). 

For 22 countries, values have slightly changed in some years. These changes go from 5% to 22%. Find below the list of countries based on their maximum relative change in population: 
  - **5% - 7.5%**: India, North Korea, Tonga, Mauritius, Malta, Fiji, United States, Samoa, United Arab Emirates
  - **7.5% - 10%**: Sudan, Azerbaijan, Vanuatu, Saint Vincent and the Grenadines, Central African Republic, Jordan, Suriname
  - **10% - 15%**: Serbia, Uganda, Kiribati
  - **\>15%**: Solomon Islands, Lesotho, Papua New Guinea

These adjustments are likely caused due to the changes introduced by the 2022 revision by the UN. Gapminder adjusts the values before 1950 to align with UN's. Therefore, changes in UN estimates will influence these adjustments.

An example is Papua New Guinea, for which the UN estimated a population of 1,927,846 in 2019 and 1,542,854 in 2022. Therefore, Gapminder has adjusted their estimates for 1949 (roughly a 22%).

In [their metadata](https://docs.google.com/spreadsheets/d/1hkLbEilJbl630IG68q-aQJlUjuTFm9b_12nQMVd1sZM/edit#gid=0) for the period 1800 - 1949, Gapminder states:

> Mainly Maddison via ClIo infra with some adjustments made by Gapminder to have estimates meet the UN data in 1950.

---
More details on the original dataset: https://docs.google.com/spreadsheets/d/1c1luQNdpH90tNbMIeU7jD__59wQ0bdIGRFpbMm8ZBTk/edit#gid=501532268

PR #976, opened 2023-03-31
data(population): update OMM using Gapminder v7

lucasrodes:
This PR should be merged after merging https://github.com/owid/etl/pull/975.

Brief summary on changes:
- [ ] `etl/steps/data/garden/demography/2023-03-31/population/`: Update our population OMM recipe with Gapminder population v7 data. The previous version used Gapminder v6. Most of the code hasn't changed from the previous version. Most noticeable: some minor adjustments to the metadata to reflect the change in Gapminder's version.
- [ ] `etl/steps/data/garden/demography/2022-12-08/population/meta.yaml`: Change the URL for Gapminder v6 in metadata for the previous version of our population OMM.
- [ ] `etl/steps/data/grapher/demography/2023-03-31/population.py`: Mostly copied from `etl/steps/data/grapher/demography/2022-12-08/population.py`. In this step we create "only historical" and "only projection" versions of the variables.
- [ ] `dag/demography.yml`: Added new steps.

In general, creating the new version of this OMM basically involved copy-pasting the code from `etl/steps/data/garden/demography/2022-12-08/population/` to `etl/steps/data/garden/demography/2023-03-31/population/`


pabloarosado:
Hi @lucasrodes shouldn't this PR be to branch `data/gapminder-population` instead of `master`? Otherwise I think I may be reviewing double.
Also, if some of the code has been duplicated from previous versions, could you please specify which files you need to review? Thanks.

lucasrodes:
@pabloarosado thanks for your feedback, changed the PR accordingly.

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -0,0 +1,289 @@
...
+    # add sources for world
+    df.loc[df["country"] == "World", "source"] = "; ".join(sorted(SOURCES_NAMES.values()))
+    return df
+
+
+def add_historical_regions(df: pd.DataFrame) -> pd.DataFrame:
+    """Add historical regions.
+
+    Systema Globalis from Gapminder contains historical regions. We add them to the data. These include
+    Yugoslavia, USSR, etc.
```
pabloarosado, line 228: Just to confirm: A country like Russia has data that is extended backwards even in USSR times, right? Otherwise historical regions would need to be added prior to creating region aggregates.

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -0,0 +1,289 @@
...
+    log.info("population: loading data (Gapminder Systema Globalis)")
+    gapminder_sg = load_gapminder_sys_glob_former()
+    # map source name
+    gapminder_sg["source"] = SOURCES_NAMES["gapminder_sg"]
+
+    df = pd.DataFrame(pd.concat([df, gapminder_sg], ignore_index=True))
+    return df
+
+
+def filter_rows(df: pd.DataFrame) -> pd.DataFrame:
```
pabloarosado, line 241: Rename it to something more meaningful, like `fix_anomalies` or something like that.
By the way, let me see if I understand correctly (following my previous comment):
+ Current regions have data for many years in the past (even before they existed).
+ But historical regions have data only until they stopped existing.

Is this the expected behaviour?

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -0,0 +1,289 @@
...
+    df["world_pop_share"] = (100 * df["population"].div(df.world_pop)).round(2)
+    df = df.drop(columns="world_pop")
+    return df
+
+
+def df_to_table(df: pd.DataFrame) -> Table:
+    """Create table from dataframe."""
+    log.info("population: converting df to table...")
+    # fine tune df (sort rows, columns, set index)
+    df = df.set_index(["country", "year"]).sort_index()[["population", "world_pop_share", "source"]]
```
pabloarosado, line 280: For safety, add `verify_integrity=True` to `set_index()`.

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -0,0 +1,289 @@
...
+
+# naming conventions
+N = PathFinder(__file__)
+METADATA_PATH = os.path.join(N.directory, "meta.yml")
+
+# sources names
+# this dictionary maps source short names to complete source names
+SOURCES_NAMES = {
+    "unwpp": "United Nations, World Population Prospects 2022 (https://population.un.org/wpp/Download/Standard/Population/)",
+    "gapminder_v6": "Gapminder v6 (https://www.gapminder.org/data/documentation/gd003/)",
```
pabloarosado, line 46: For consistency, you could rename this key to simply `gapminder`, and then change all instances of `gapminder_v6` in the file.

etl/steps/data/garden/demography/2023-03-31/population/meta.yml
```diff
@@ -0,0 +1,80 @@
...
+    For a more detailed description, please refer to the field "source" in table "population".
+
+  version: "2022-12-08"
+  sources: &population-sources
+    - name: Gapminder (2022)
+      published_by: Gapminder (v7)
+      url: https://www.gapminder.org/data/documentation/gd003/
+      date_accessed: 2023-03-31
+    - name: UN (2022)
+      published_by: UN, World Population Prospects (2022)
```
pabloarosado, line 23: Again, I'd say `UN` -> `United Nations`.

pabloarosado (approved):
It looks fine, although I'm a bit worried about the use of `geo.add_region_aggregates`. If you share my worry, then you could probably do a simple `groupby(...).agg()` and avoid possible issues. Feel free to disregard the rest of my comments (I haven't had time to look much deeper into it, so maybe some of my comments are irrelevant).

etl/steps/data/garden/demography/2023-03-31/population/meta.yml
lucasrodes: Will keep `UN (2022`. Otherwise, the full citation (`UN (2022), Gapminder (2022); HYDE (2017)`) is too long (even more if combined with other sources).

lucasrodes:
Thanks for your feedback, @pabloarosado.

As a general note, I did not re-write most of the code, hence the usage of functions that we are no longer using. I will look into migrating some of this to the new methods, but might leave this for another PR as this is currently blocking progress on the topic page re-design.

etl/steps/data/garden/demography/2023-03-31/population/meta.yml
pabloarosado: The full citation can be as long as we want, this doesn't show up in charts, but only in the SOURCES tab. But up to you.

etl/steps/data/garden/demography/2023-03-31/population/gapminder_sg.py
```diff
@@ -0,0 +1,155 @@
...
+
+In there, open sheet "Data" and check the Source per country and year to find out how they
+obtained that data.
+
+"""
+from owid.catalog import Dataset
+
+from etl.paths import DATA_DIR
+
+DATASET_GAPMINDER_SYSTEMA_GLOBALIS = (
```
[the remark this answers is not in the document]
lucasrodes: No. `load_dependency` currently does not support `open_numbers` channel.

etl/steps/data/garden/demography/2023-03-31/population/gapminder.py
```diff
@@ -0,0 +1,16 @@
+from owid.catalog import Dataset
+
+from etl.helpers import PathFinder
+
+# Get paths and naming conventions for current step.
+paths = PathFinder(__file__)
```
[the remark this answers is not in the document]
lucasrodes: This is loading the garden Gapminder dataset. You can find it at `data://garden/gapminder/2023-03-31/population`.

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
lucasrodes: Yes, countries like Russia ("current states") extend back in time. Former countries do not replace any of these.

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -0,0 +1,289 @@
...
+    df = df.loc[~df["country"].isin(regions)]
+
+    # keep sources per countries, remove from df
+    # remove from df: otherwsie geo.add_region_aggregates will add this column too
+    sources = df[["country", "year", "source"]].copy()
+    df = df.drop(columns=["source"])
+
+    # re-estimate region aggregates
+    for region in regions:
+        df = geo.add_region_aggregates(df=df, region=region, population=df)
```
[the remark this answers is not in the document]
lucasrodes: I don't remember the details, we checked the completeness of the dataset, and it looked fine.

Also, having population values for regions is a priority.

etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -0,0 +1,289 @@
...
+
+def run(dest_dir: str) -> None:
+    log.info("population.start")
+
+    # create table
+    tb = make_table()
+
+    # create dataset
+    log.info("population: create dataset")
+    ds = Dataset.create_empty(dest_dir)
```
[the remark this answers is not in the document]
lucasrodes: I think we can use `create_dataset` leaving the default metadata field empty. I later call `update_metadata` to fill it in.

PR #986, opened 2023-04-04
Fix further faostat anomalies

pabloarosado:
Main changes:
* Fixed issue on `detect_anomalies.py`, whereby the correction was removing more data for aggregate regions than necessary. Additionally, create an anomaly class related to tea production (if you want more details, see [this issue](https://github.com/owid/owid-issues/issues/632)).
* In `faostat_qcl`, when the number of slaughtered poultry is not given, fill it with the number of slaughtered chicken, if possible. Then, if poultry is not informed, remove the total number of slaughtered animals (given that chicken is the most commonly slaughtered animal, by far). We do this to avoid spurious jumps in the number of slaughtered animals, caused by missing chicken data.
* Add major fishing areas to country mapping (since they are used in one dataset).
* Add per capita land use and region aggregates to `faostat_rl` dataset.
* Rename some items to fix issues after some FAO item name changes (now all items used in the food explorer have custom names, so this should not happen again).

@spoonerf  there's no need to review the changes in the `custom_*` files.


spoonerf (approved):
LGTM!

PR #997, opened 2023-04-10
Archive old faostat datasets as well as other unnecessary steps

pabloarosado:
In this PR:

* I created a script that should ease the task of archiving unnecessary FAOSTAT DB datasets, as well as unnecessary ETL steps. In the future, this script could easily be generalised to be part of `VersionTracker`.
* Thanks to this script (although it has some issues, as explained in the code and documentation) I archived unnecessary FAOSTAT DB datasets and ETL steps.
* However, the `sdg` dataset was backporting two FAOSTAT datasets, so I updated the `sdg` scripts and dependencies (I hope I did it properly).
* I then realised that there were several other ETL steps that could be archived, so I did that.

@lucasrodes could you please review these files?
* [x] `docs/data/faostat.md`
* [x] `etl/helpers.py`
* [x] `etl/scripts/faostat/archive_old_datasets.py`

@spoonerf could you please review these files?
* [ ] `dag/*`
* [ ] `etl/steps/data/garden/sdg/latest/sdg_sources.csv`
* Also, do you know why the `sdg` dataset is using a backported FAOSTAT dataset, instead of its corresponding garden step?

There's no need to review all the `*.dvc` files, since they were generated automatically.

Thanks!


etl/helpers.py
```diff
@@ -729,6 +726,21 @@ def check_that_all_active_steps_are_necessary(self) -> None:
...
+        """Get list of ids of DB datasets that are used as backported datasets in active steps of ETL.
+
+        Returns
+        -------
+        backported_dataset_ids : List[int]
+            Grapher DB dataset ids that are used in ETL backported datasets.
+        """
+        backported_dataset_names = [step for step in self.all_active_dependencies if step.startswith("backport://")]
+        backported_dataset_ids = sorted(
+            set([int(step.split("dataset_")[1].split("_")[0]) for step in backported_dataset_names])
```
lucasrodes, line 739: I'd lean towards having a dedicated function can recognize the different fields from a backported dataset URI. For instance

```python
def extract_dataset_id_from_backported_dataset(dataset_uri: str) -> int:
    rex = r"backport://backport/(?P<namespace>[a-z]+)/(?P<version>[\d+\-]|latest)/dataset_(?P<dataset_id>\d+)_(?P<dataset_slug>[a-z_\d]+)"
    match = re.match(rex, dataset_uri)
    if match:
        parameters = match.groupdict()
        # return id
        try:
            dataset_id = int(parameters["dataset_id"])
        except ValueError:
            raise ValueError(f"Invalid dataset id `{parameters['dataset_id']}`!")
        return dataset_id
    raise SyntaxError(f"Unrecognized backported dataset URI `{dataset_uri}`.")
```

And then call it from here:

```python
backported_dataset_ids = sorted(
            set(extract_dataset_id_from_backported_dataset(step) for step in backported_dataset_names)
```

Also, note that you do not need the brackets inside `set()`.

etl/scripts/faostat/archive_old_datasets.py
```diff
@@ -0,0 +1,267 @@
...
+
+def list_active_db_datasets(db_conn: Connection) -> pd.DataFrame:
+    query = """
+    SELECT d.id, d.name, d.namespace, d.version, d.shortName
+    FROM datasets d
+    JOIN variables v
+    ON v.datasetId = d.id
+    JOIN chart_dimensions cd
+    ON v.id = cd.variableId
+    WHERE d.isArchived IS FALSE
```
lucasrodes, line 32: maybe not relevant but: do we want to check if the dataset is private/public here?

etl/scripts/faostat/archive_old_datasets.py
```diff
@@ -0,0 +1,267 @@
...
+    ;
+    """
+    with db_conn.cursor() as cursor:
+        cursor.execute(query)
+        result = cursor.fetchall()
+    df = pd.DataFrame(result, columns=["dataset_id", "dataset_title", "namespace", "version", "short_name"])
```
lucasrodes, line 38: Maybe you have already considered this, but you can also use

```python
df = pd.read_sql(query, db.get_engine())
```

and then rename columns

```python
df = df.rename(columns={...})
```

etl/scripts/faostat/archive_old_datasets.py
```diff
@@ -0,0 +1,267 @@
...
+        JOIN variables v
+        ON v.id = cd.variableId
+        )
+    ;
+    """
+    with db_conn.cursor() as cursor:
+        cursor.execute(query)
+        result = cursor.fetchall()
+
+    df = pd.DataFrame(result, columns=["dataset_id", "dataset_title", "namespace", "version", "short_name"])
```
lucasrodes, line 73: (idem as before)
Maybe you have already considered this, but you can also use

```python
df = pd.read_sql(query, db.get_engine())
```

and then rename columns

```python
df = df.rename(columns={...})
```

lucasrodes (approved):
Nice work!

I left minor suggestions, other than that LGTM.

etl/scripts/faostat/archive_old_datasets.py
pabloarosado: Thanks, I was looking for something like this, but didn't see the `get_engine` function. Anyway, not worth changing for this script. If we migrate to `VersionTracker` then I'll improve this.

etl/scripts/faostat/archive_old_datasets.py
pabloarosado: I think I prefer to be open, just in case a private dataset is using any of the relevant variables.

etl/helpers.py
pabloarosado: Thanks for the suggestion! For this script it's not worth the effort of changing it (and I already have a subbranch, so I'd prefer to skip unnecessary changes). We can improve these functions if we move them to `VersionTracker`.

pabloarosado:
Hi @spoonerf I've just checked that (at least for the World) the old and new data for the `sdg` dataset is identical, so in principle I guess it's safe to merge this PR. But please check it out when you are back and have a minute and let me know if you see something odd, thanks.

Issue #791, opened 2023-01-23
Allow placeholders in metadata yaml files


pabloarosado:
Currently, every time we update an existing dataset, we need to do some manual changes to the metadata file, for example, in years and dates. This is prone to error, and, if we ever want to be able to automatically update datasets, we should avoid it.

I think one solution (the only one?) would be to accept placeholders in yaml files, e.g.
```yaml
dataset:
  title: Some dataset (Institution, %%year%%)
```
Or something like that. Then, the function `dataset.update_metadata` would fill the placeholders with the appropriate information from the dataset, or it would ingest them as arguments.

Marigold:
Nice idea, this could be really useful. With this we could also add parameters at the top of the YAML file and use them later inline, e.g.

```yml
parameters:
  a: 1
  b: 2
  
...
  description: %%a%% and %%b%%
```

that could be used to clean up meta file [like this](https://github.com/owid/etl/blob/master/etl/steps/data/garden/ophi/2022-12-13/multidimensional_poverty_index.meta.yml#L75).

larsyencken:
Typically in Python you use Jinja2 for templating and curly braces, it might be a good angle.

larsyencken:
Marking it as "nice to have" until we get around to automating updates.

The idea is that the YAML file on-disk is always read using something that fills basic template slots.

Marigold:
Tried to use Jinja2 as a templating engine, but it's not as convenient as I was hoping for. Here is an example of multi-line strings from this [real example](https://github.com/owid/etl/pull/1002)

```yaml
{% set description_ethnic_group %}
The Ethnic Power Relations dataset defines ethnicity as any subjectively experienced sense of commonality based on the belief in common ancestry and shared culture. Given this definition, an ethnic group (i.e. a group of individuals sharing a common ethnicity) is included in the EPR Core dataset if it is politically relevant at least once in the sample period. An ethnic group is classified as politically relevant if at least one political organization claims to represent it in national politics or if its members are subjected to state-led political discrimination.
{% endset %}

{% set description_egip %}
An ethnic group in power (EGIP) is a politically relevant ethnic group which has access to power, in one of these options:
  <ul>
    <li>The group rules alone:</li>
      <li>Monopoly: Elite members hold power in the executive to the exclusion of members of all other ethnic groups.</li>
      <li>Dominance: Elite members hold dominant power in the executive, but there is some limited inclusion of "token" members of other groups who however do not have real influence on decision making.</li>
[...]

PR #1002, opened 2023-04-13
Create Ethnic Power Relations Dataset + aggregations

paarriagadap:
Create the Ethnic Power Relations Dataset and  include regional aggregations (continents, income regions). It uses the `add_region_aggregates` from `geo` to sum up several variables. Some of them have further processing to get weighted percentages.

I am mostly interested that the aggregations are calculated correctly (as far as I can see it is the case). This is done in the `add_regional_aggregations` function in `garden`


etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,247 @@
...
+
+    df[share_vars] *= 100
+
+    return df
+
+
+def load_population() -> pd.DataFrame:
+    # Load population table from key_indicators dataset.
+    ds_indicators = paths.load_dependency("population")
+    tb_population = ds_indicators["population"]
```
paarriagadap, line 90: There's an error with this when running `make test` that I don't know how to solve

Marigold: That's a typing thing, `load_dependency` can return both Dataset and Snapshot so you might tell it what is it that you want (in this case `Dataset`)

This should fix it
```
ds_indicators: Dataset = paths.load_dependency("population")
tb_population = ds_indicators["population"]
```

paarriagadap: Thank you @Marigold! I will mention this to add the right function in https://github.com/owid/etl/pull/1004

dag/main.yml
```diff
@@ -366,6 +366,16 @@ steps:
   - backport://backport/owid/latest/dataset_235_world_happiness_report__2022
   data://grapher/happiness/2023-03-20/happiness:
   - data://garden/happiness/2023-03-20/happiness
+  # Ethnic Power Relations Dataset
+  data://meadow/eth/2023-03-15/growup:
+  - snapshot://eth/2023-03-15/growup.csv
+  data://garden/eth/2023-03-15/ethnic_power_relations:
+  - data://meadow/eth/2023-03-15/growup
+  - data://garden/demography/2023-03-31/population
+  - data://garden/reference
```
pabloarosado, line 375: Since you are using `geo.add_region_aggregates`, you also need to include the dependence on income groups:
`- data://garden/wb/2021-07-01/wb_income`

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,247 @@
...
+        "maxexclpop_headcount": "sum",
+        "regautpop_headcount": "sum",
+        "regautexclpop_headcount": "sum",
+        "regautegippop_headcount": "sum",
+        "population": "sum",
+    }
+
+    # Add regional aggregates, by summing up the variables in `aggregations`
+    for region in regions:
+        df = geo.add_region_aggregates(df, region=region, aggregations=aggregations, population=df)
```
pabloarosado, line 153: Just so you know, this function is a bit "too smart" and does a few things you may not be aware of (which is why we need to create a better one). If the most populated countries in a region are not informed, the aggregate will be nan. If you don't want this behaviour, add the argument `countries_that_must_have_data=[]`.

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,247 @@
...
+    return df
+
+
+# EPR classifies some countries as both their historical and current name. This is not ideal for some cases where their borders changed completely
+def separate_historical_countries(df: pd.DataFrame) -> pd.DataFrame:
+
+    df["country"] = df["country"].astype(str)
+    # RUSSIA / SOVIET UNION
+    # Change the country name for the Soviet Union to Russia for years less or equal to 1991
+    df.loc[(df["country"] == "Russia (Soviet Union)") & (df["year"] <= 1991), "country"] = "Soviet Union"
```
pabloarosado, line 187: "Soviet Union" is not a harmonized country, if anything it should be "USSR". But doesn't that region already exist in the data?

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,247 @@
...
+    df.loc[(df["country"] == "Russia (Soviet Union)") & (df["year"] > 1991), "country"] = "Russia"
```
pabloarosado, line 188: Wouldn't it suffice to simply add the mapping "Russia (Soviet Union)": "Russia" in the countries file?

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # VIETNAM
+    # Drop values for "Vietnam, Democratic Republic of" between years 1954 and 1975
+    df = df[
+        ~((df["country"] == "Vietnam, Democratic Republic of") & (df["year"] >= 1954) & (df["year"] <= 1975))
```
pabloarosado, line 193: In our regions dataset we have "Vietnam" and "Republic of Vietnam". Maybe we should rename "Republic of Vietnam" as "South Vietnam (former)" and add "North Vietnam (former)". Then you could add "Vietnam, Democratic Republic of": "North Vietnam" (at least that's how I understand it, I'm not sure about this). Anyway, if that's too many changes, you can simply drop these values for now.

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
pabloarosado: Just to clarify: It's ok to have data for Russia prior to 1991. The meaning of it is "the data for the region that is geographically known today as Russia".

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,247 @@
...
+    ].reset_index(drop=True)
+
+    # YEMEN
+    # Rename "Yemen (Arab Republic of Yemen)" to "Yemen" for years greater than 1990
+    df.loc[(df["country"] == "Yemen (Arab Republic of Yemen)") & (df["year"] > 1990), "country"] = "Yemen"
+
```
pabloarosado, line 199: I haven't looked into the data, but wouldn't you need to also distinguish West Germany from Germany?

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.meta.yml
```diff
@@ -0,0 +1,413 @@
...
+        unit: "%"
+        short_unit: "%"
+        display:
+          name: Share of population in EGIP groups
+          numDecimalPlaces: 1
+        sources:
+          - *source-epr
+      legippop:
+        title: Share of ethnically relevant population in ethnic groups in power (EGIP)
+        description: EGIP population as a fraction of ethnically relevant population in this country (egippop / lpop).
```
pabloarosado, line 118: Refer to variables by their title, not their name.

pabloarosado (changes requested):
Good work! I see how this dataset is quite hard to harmonize. The one change you definitely have to do is to include the dependency of income groups dataset in the dag. Apart from that, I made some suggestions regarding harmonization, but maybe they are not relevant (I haven't looked into the data). So feel free to disregard them.

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.countries.json
```diff
@@ -0,0 +1,185 @@
...
+  "Zanzibar": "Zanzibar",
+  "Belarus (Byelorussia)": "Belarus",
+  "Bosnia-Herzegovina": "Bosnia and Herzegovina",
+  "Burkina Faso (Upper Volta)": "Burkina Faso",
+  "Cambodia (Kampuchea)": "Cambodia",
+  "Congo, Democratic Republic of (Zaire)": "Democratic Republic of Congo",
+  "German Democratic Republic": "East Germany",
+  "German Federal Republic": "Germany",
+  "Iran (Persia)": "Iran",
+  "Italy/Sardinia": "Italy",
```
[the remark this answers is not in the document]
paarriagadap: No, it's Italy (+ Sardinia)

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.countries.json
```diff
@@ -0,0 +1,185 @@
...
+  "Korea, People's Republic of": "North Korea",
```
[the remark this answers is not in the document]
paarriagadap: I know, but this is the way they refer to NK from 1946 to 2020

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
paarriagadap: I saw that in the documentation and it doesn't generate much problem because there is data for almost every country in the world.

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
paarriagadap: > "Soviet Union" is not a harmonized country, if anything it should be "USSR". But doesn't that region already exist in the data?

Yes, I leave it like that and the harmonization tool converts it to USSR, so we have it anyway. With @bastianherre preferred to split it, because it shows fairly different data between 1991 and 1992: there's a big drop which we prefer to separate. They can have it like that because they use a geographical tool that [changes the borders while changing years](https://growup.ethz.ch/pfe/Russia%20(Soviet%20Union)).

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.countries.json
```diff
@@ -0,0 +1,185 @@
...
+  "Yugoslavia": "Yugoslavia",
+  "Zambia": "Zambia",
+  "Zanzibar": "Zanzibar",
+  "Belarus (Byelorussia)": "Belarus",
+  "Bosnia-Herzegovina": "Bosnia and Herzegovina",
+  "Burkina Faso (Upper Volta)": "Burkina Faso",
+  "Cambodia (Kampuchea)": "Cambodia",
+  "Congo, Democratic Republic of (Zaire)": "Democratic Republic of Congo",
+  "German Democratic Republic": "East Germany",
+  "German Federal Republic": "Germany",
```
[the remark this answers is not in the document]
paarriagadap: Yeah, I have modified it for West Germany between 1946-1990, though there's no major impact compared to the other cases. Thanks!

snapshots/eth/2023-03-15/growup.csv.dvc
```diff
@@ -0,0 +1,25 @@
...
+  publication_date: 2021-06-08
+  source_name: Geographical Research On War, Unified Platform - GROW<sup>up</sup>
+    (2021)
+  source_published_by: Luc Girardin, Philipp Hunziker, Lars-Erik Cederman, Nils-Christian
+    Bormann, Seraina Rüegger, and Manuel Vogt. 2015. GROW<sup>up</sup> - Geographical
+    Research On War, Unified Platform. ETH Zurich. http://growup.ethz.ch/
+  url: https://growup.ethz.ch/
+  source_data_url:
+  file_extension: csv
+  license_url:
```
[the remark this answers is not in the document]
paarriagadap: No, but I just asked

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
paarriagadap: Now I did it, thanks. In this case nothing much changes, because there is only one ethnic groups considered in the dataset for any Germany ("Germans")

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.meta.yml
paarriagadap: That's right, thanks!

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
```diff
@@ -0,0 +1,258 @@
...
+    tb_garden = Table(df)
+    tb_garden.metadata.short_name = "ethnic_power_relations"
+
+    #
+    # Save outputs.
+    #
+    # Create a new garden dataset with the same metadata as the meadow dataset.
+    ds_garden = create_dataset(dest_dir, tables=[tb_garden])
+
+    # For now the variable descriptions are stored as a list of strings, this transforms them into a single string
```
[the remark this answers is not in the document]
paarriagadap: I am using YAML anchors in the garden metadata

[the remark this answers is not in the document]
Marigold: Oh yeah, that's tricky. The reason is doesn't work is that even though calling `create_dataset` updates metadata of dataset and table, it does it on a copy of the table, so the input `tb_garden` won't be modified. You have to load it from dataset like this

```python
# Create a new garden dataset with the same metadata as the meadow dataset.
    ds_garden = create_dataset(dest_dir, tables=[tb_garden])

    # For now the variable descriptions are stored as a list of strings, this transforms them into a single string
    tb_garden = ds_garden["ethnic_power_relations"]
    for col in tb_garden.columns:
        if isinstance(tb_garden[col].metadata.description, list):
            tb_garden[col].metadata.description = "\n\n".join(tb_garden[col].metadata.description)
    ds_garden.add(tb_garden)
```

Even better would be to implement [placeholders in YAML files](https://github.com/owid/etl/issues/791) to avoid this hack with lists, but that ball is on my side.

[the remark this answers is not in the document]
paarriagadap: Oh, thank you! I actually tried something very similar (just replacing `tb_garden` to  `ds_garden["ethnic_power_relations"]` in the loop) and it failed, because it said that some other variable unit was empty. (I think I forgot to also change the `ds_garden.add(tb_garden)` to `ds_garden.add(ds_garden["ethnic_power_relations"])`!)

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.py
paarriagadap: And regions are currently fixed in time, right? For instance, if for the 80s I don't have Russia or the other former Soviet republics will it not calculate the aggregation?

etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.meta.yml
```diff
@@ -0,0 +1,594 @@
...
+        sources:
+          - *source-acd2epr
+      onset_ko_gov_eth_flag:
+        title: Governmental ethnic conflict onset (KO)
+        description:
+          - Binary flag indicating governmental ethnic conflict onset / ko option
+          - *description_ethnic_group
+          - *description_acd_conflict
+          - *description_conflict_onset
+        unit: ""
```
paarriagadap, line 396: Do you know the reason why the jump spaces behave differently in cases like [this variable](https://owid.cloud/admin/variables/576990), @Marigold? When there's a text instead of the anchor (like in the first item of the list), there is no space between this and the second item. And when I use an HTML list (as `*description_acd_conflict`) there seems to be a double jump with `*description_conflict_onset`.

No space:
image

Double space:
image

Marigold: If you mean the space before "Each conflict..." then that's not actually a double space, but a space and bottom margin for `<ul>` element. If you want you can eliminate the space by writing it as
```
- acd_conflict:
      description: &description_acd_conflict |
        An ACD (Armed Conflict Database) conflict for this dataset includes:
            <ul>
              <li>Intrastate conflict: A conflict between a government and a non-governmental party, with no interference from other countries.</li>
              <li>Intrastate conflict internationalized: An armed conflict between a government and a non-government party where the government side, the opposing side, or both sides, receive troop support from other governments that actively participate in the conflict.</li>
            </ul>Each conflict is assigned a unique UCDP (Uppsala Conflict Data Program) ID. ACD conflicts are merged with the Ethnic Power Relations country-level data by assigning each country-year observation one or several UCDP IDs if the ACD dataset reports an ongoing conflict in the respective country and year.
```

paarriagadap: Ah, you're right, thanks! And how about the first case?

Marigold: @paarriagadap sorry, can't replicate it now. I've just uploaded the [dataset to staging](https://staging.owid.cloud/admin/datasets/5921). Could you tell me what variable is problematic?

paarriagadap: Any variable, for instance [this one](https://staging.owid.cloud/admin/variables/576837). There is no jump between the first and the second paragraph. In this case, I have formatted the first paragraph not as an anchor but a text.

Marigold: I see. Try adding `<br>` at the end of all texts, that should help, i.e.

```
description:
- Count variable indicating the number of EGIP groups in this country.<br>
- *description_ethnic_group
- *description_egip
```

I should really get those placeholders working...

paarriagadap: Thanks! I did that last time with the multidimensional poverty index, but I thought I could do something else

paarriagadap: I will let you know, @Marigold. Thank you very much!

Issue #1003, opened 2023-04-14
walkthrough: walkthrough meadow generates old dependencies for regions and population


paarriagadap:
### One-liner
When running `walkthrough meadow` the options **_Load countries regions in the script_** and **_Load population in the script need_** to be updated

### Environment

- `python -V`: Python 3.11.1
- Operating system: macOS Ventura 13.3.1
- Working with local, staging or live: None (or local)

### Context & details

When selecting these two options at the bottom of `walkthrough meadow` I add countries regions and population as dependencies, but

- We can't call `data://garden/reference` via `paths.load_dependency("reference")`, because we get this error:
```
etl.helpers.NoMatchingStepsAmongDependencies: No steps found among dependencies of current ETL step, that match the given specifications.
```
- Population needs to be updated to `data://garden/demography/2023-03-31/population` instead of `data://garden/owid/latest/key_indicators`

When fixed, it should also be nice to have these options in `walkthrough garden`

pabloarosado:
A few comments on this issue:
* I think adding regions and population is something that should happen in the garden step, not meadow. Meadow is supposed to be for the minimum amount of preparation of the original data, so we don't need countries-regions or population for that.
* The error when loading `reference` is a small bug in `PathFinder`, simply because the `reference` dataset has an unusual URI. But given that we want to move away from `countries_regions` and start using `regions` soon, I don't think it's worth fixing.

paarriagadap:
Thanks @pabloarosado. I agree with both points: in fact, I wanted to include this data on a garden step (currently unavailable). And regarding countries_regions, probably we should do nothing yet, but it's important to highlight it's broken for the most common use case.

lucasrodes:
Issued a PR for this https://github.com/owid/etl/pull/1004

PR #1004, opened 2023-04-14
fix(walkthrough): load population and reference datasets in garden step

lucasrodes:
Previously, `walkthrough` allowed users to add our population and reference datasets as dependencies to our DAG _only_ in the meadow step. However, these datasets should not be listed as dependencies in any meadow step. Instead, they should be used in Garden.

This PR:

- Allows users to add population and reference datasets as dependencies in a Garden step from `walkthrough` interface. It removes these options from the Meadow step, as these should not be used there.
- Updates the population dataset used by default to the latest available one: `data://garden/demography/2023-03-31/population`

---
related to https://github.com/owid/etl/issues/1003
tagged you both, @pabloarosado @paarriagadap ー but one review should suffice I think.

### DIFF
--- a/walkthrough/utils.py
+++ b/walkthrough/utils.py
@@ -17,6 +17,8 @@ from etl.steps import DAG
 
 DAG_WALKTHROUGH_PATH = DAG_DIR / "walkthrough.yml"
 WALDEN_INGEST_DIR = Path(walden.__file__).parent.parent.parent / "ingests"
+DATASET_POPULATION_URI = "data://garden/demography/2023-03-31/population"
+DATASET_REFERENCE_URI = "data://garden/reference"
 
 DUMMY_DATA = {
     "namespace": "dummy",
--- a/walkthrough/utils.py
+++ b/walkthrough/utils.py
@@ -18,6 +18,7 @@ from etl.steps import DAG
 DAG_WALKTHROUGH_PATH = DAG_DIR / "walkthrough.yml"
 WALDEN_INGEST_DIR = Path(walden.__file__).parent.parent.parent / "ingests"
 
+# Load latest population version
 POPULATION_LATEST_VERSION = (
     sorted((STEP_DIR / "data/garden/demography").glob("*/population/"))[-1].as_posix().split("/")[-2]
 )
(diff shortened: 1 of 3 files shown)


walkthrough/utils.py
```diff
@@ -17,6 +17,8 @@
 
 DAG_WALKTHROUGH_PATH = DAG_DIR / "walkthrough.yml"
 WALDEN_INGEST_DIR = Path(walden.__file__).parent.parent.parent / "ingests"
+DATASET_POPULATION_URI = "data://garden/demography/2023-03-31/population"
```
pabloarosado, line 20: Instead of hardcoding this specific version of the population dataset, it would be good to fetch the latest version. The optimal solution would be to look in the dag, but just for simplicity you could do something like:
```
from etl.paths import STEP_DIR
sorted((STEP_DIR / "data/garden/demography").glob("*/population/"))[-1].as_posix().split("/")[-2]
```
To get the latest version.

pabloarosado (approved):
Thanks for fixing this. I just suggested a minor enhancement.

walkthrough/utils.py
lucasrodes: Oh, ー sweet!

I was thinking about the risks of automatically getting the population URI this way.

Like, we are assuming that:

- we want to always use the latest population OMM (I think this is correct)
- the garden recipe for the latest available population dataset works (that should be the case)

We should continue to be extra careful when adding a new version of our population OMM (this occurs every 1-2 years, approx).

paarriagadap:
Hi, @lucasrodes! Thanks! I tried this, and it seems it only adds the dependencies in the dag, but not the function to load the population, as it existed before. I also mention this because when loading the dataset we need to add `: Dataset` to [solve ambiguities between snapshots and datasets](https://github.com/owid/etl/pull/1002/files#r1168402467) and avoid a buildkite error.

So something like
```
def load_population() -> pd.DataFrame:
    # Load population table from key_indicators dataset.
    ds_indicators: Dataset = paths.load_dependency("population")
    tb_population = ds_indicators["population"]

    return tb_population

```

can work

PR #1006, opened 2023-04-17
enhance(fasttrack): add instructions for importing large CSVs

Marigold:
Add instructions for importing large CSVs to fast-track and make it work with Google Drive URLs. We are already using this feature, but the documentation was not available.


pabloarosado (approved):
Looks good!

PR #1014, opened 2023-04-18
Update aviation statistics

pabloarosado:
* Fix bug in walkthrough meadow related to file extensions.
* Add new snapshots, meadow, garden and grapher steps for aviation statistics dataset.

### DIFF
--- a/etl/steps/data/garden/aviation_safety_network/2023-04-18/aviation_statistics.meta.yml
+++ b/etl/steps/data/garden/aviation_safety_network/2023-04-18/aviation_statistics.meta.yml
@@ -1,0 +1,68 @@
+dataset:
+  title: Fatal aviation accidents and fatalities (ASN, 2023-04-18)
+  description: |
+    Data on fatal accidents, hijacking incidents and numbers of fatalities are based on airliners of 14+ passengers, and do not include corporate jet and military transport accidents.
+
+    Air traffic figures are based on the number of airliner flight departures (not the number of passengers) per year.
+  sources:
+    -
+      name: Aviation Safety Network
+      published_by: The Flight Safety Foundation's Aviation Safety Network (ASN)
+      publisher_source: |
+        Information contained in the Aviation Safety Network database is based on information from official sources (authorities, safety boards).
+      publication_year: 2023
+      date_accessed: 2023-04-18
+      url: "https://aviation-safety.net/statistics/"
+    -
+      name: World Bank's World Development Indicators
+      published_by: World Bank's World Development Indicators
+      publication_year: 2022
+      date_accessed: 2022-05-26
+
+tables:
+  aviation_statistics:
+    variables:
+      accidents_per_million_flights:
+        title: Fatal accidents per million commercial flights
+        short_unit:
+        unit: accidents
+      accidents_with_passenger_and_cargo_flights_including_hijacking_etc:
+        title: Fatal accidents from commercial airliners
+        short_unit:
+        unit: accidents
+      departures_worldwide:
+        title: Number of flight departures
+        short_unit:
+        unit: departures
+      fatalities_per_million_flights:
+        title: Fatalities per million flights
+        short_unit:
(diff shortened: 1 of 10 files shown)


snapshots/aviation_safety_network/2023-04-18/aviation_statistics.py
```diff
@@ -0,0 +1,95 @@
+"""Script to create a snapshot of dataset 'Aviation statistics'.
+
+In total, 3 snapshots will be created:
+* Data from their shared Google spreadsheet (from one specific sheet).
+* Data extracted directly from their page on aviation statistics by period.
+* Data extracted directly from their page on aviation statistics by nature.
```
lucasrodes, line 6: You could add a note explaining that the data for these two is sourced dynamically (and not from snapshot dvc metadata) from source HTML site. As a result, `source_data_url` in the DVC files is empty.

etl/steps/data/meadow/aviation_safety_network/2023-04-18/aviation_statistics.py
```diff
@@ -0,0 +1,86 @@
...
+    #
+    # Process data.
+    #
+    # Select necessary columns and rename them appropriately.
+    df = df[list(COLUMNS)].rename(columns=COLUMNS)
+    df_by_period = df_by_period.rename(columns={"Year": "year"})
+    df_by_nature = df_by_nature.rename(columns={"Year": "year"})
+
+    # Drop last column (which should be the only one without a year), which gives a grand total.
+    df = df.dropna(subset="year").reset_index(drop=True).astype({"year": int})
```
lucasrodes, line 63: I don't understand the comment here. Which last column?

It looks as if you are dropping rows that have `NaNs` in field `year`.

etl/steps/data/garden/aviation_safety_network/2023-04-18/aviation_statistics.py
```diff
@@ -0,0 +1,136 @@
...
+# Select columns from WDI table, and how to rename them.
+WDI_COLUMNS = {
+    "country": "country",
+    "year": "year",
+    "is_air_dprt": "departures_worldwide",
+    "is_air_psgr": "passengers_carried",
+}
+
+
+def run(dest_dir: str) -> None:
```
lucasrodes, line 55: In general, I lean towards defining functions for the different code sections. E.g. `process_wdi_data`.

Just a preference. Most important are comments, and that's here!

etl/steps/data/garden/aviation_safety_network/2023-04-18/aviation_statistics.meta.yml
```diff
@@ -0,0 +1,68 @@
+dataset:
+  title: Fatal aviation accidents and fatalities (ASN, 2023-04-18)
+  description: |
+    Data on fatal accidents, hijacking incidents and numbers of fatalities are based on airliners of 14+ passengers, and do not include corporate jet and military transport accidents.
+
+    Air traffic figures are based on the number of airliner flight departures (not the number of passengers) per year.
+  sources:
+    -
+      name: Aviation Safety Network
```
lucasrodes, line 9: This is the name that is surfaced in charts, right?

I tend to have the publication year, too, as that is our standard on charts. So I'd have `Aviation Safety Network (2023)` instead.

In the future, I think charts should bake the source name based on fields `name` and `publication_year` to produce something like `[name] ([publication_year])`. So in that case, the year wouldn't be required in the field `name`.

lucasrodes (approved):
LGTM ー added minor suggestions

etl/steps/data/meadow/aviation_safety_network/2023-04-18/aviation_statistics.py
pabloarosado: Oh yes, the explanation is wrong, it should be "Drop last row". I'll change that.

PR #1027, opened 2023-04-21
data(regions): add grapher/regions dataset to replace hardcoded regions in grapher codebase

Marigold:
Assemble dataset with countries and regions that could be used by grapher instead of having duplicated hardcoded definitions such as [WorldRegionsToProjections.ts](https://github.com/owid/owid-grapher/blob/master/packages/%40ourworldindata/grapher/src/mapCharts/WorldRegionsToProjection.ts)

This PR uses our garden regions dataset and adds a few extra columns needed by grapher like mappable countries or no country page. These are currently hardcoded in this script, however since they are grapher specific, it might make more sense to **move them from ETL to grapher** (if we decide to do that, we can just replace those columns there and later remove them from ETL).

This dataset is published automatically on S3 as CSV under https://catalog-staging.ourworldindata.org/grapher/regions/2023-01-01/regions/regions.csv (**this is staging, prod link would be without `-staging`**).

There are some differences from ETL regions and current Grapher regions, namely Grapher regions contain extra

```
entities['OWID_AFR']['members'].extend(['OWID_SML','OWID_ZAN' ])
entities['OWID_ASI']['members'].extend(['OWID_ABK','OWID_AKD','OWID_NAG','OWID_CYN','OWID_SOS' ])
entities['OWID_EUR']['members'].extend(['OWID_CIS','SJM','OWID_TRS' ])
```
I assume that ETL regions are more recent and therefore omit these in the script.

One quirk is that even though this dataset lives in `grapher` channel, it's not meant to be imported to MySQL (or is it?) and was excluded from automatic deploy.

[Slack discussion](https://owid.slack.com/archives/CQQUA2C2U/p1681768079599289)


samizdatco:
> I assume that ETL regions are more recent and therefore omit these in the script.

These entities are still present in the [`migration.py`](https://github.com/owid/etl/blob/master/etl/steps/data/garden/regions/2023-01-01/migration.py)’s `continent_members` dictionary, but have been commented out so it wasn't quite clear whether the entities truly don't 'belong' in the continent as a geographical judgment or if it was more about calculating aggregate statistics. Maybe @pabloarosado can offer some guidance here.
 
> One quirk is that even though this dataset lives in `grapher` channel, it's not meant to be imported to MySQL (or is it?) and was excluded from automatic deploy.

It looks like the MySQL tables that could make use of this data include `entities` & `country_name_tool_countrydata` (in the case of country name changes) but that seems like it might be better handled through a migration kicked off by the update script being added to grapher. When I compared the inlined grapher entity data to the ETL version I spotted a few places where the names have changed:
|code|old name|new name|
|-----|---------|----------|
| ALA | Åland Islands | Aland Islands |
| BLM | Saint Barthélemy | Saint Barthelemy |
| TLS | Timor | East Timor |

It's not yet clear to me whether changing these in the grapher code without also updating the database will cause problems but maybe @marcelgerber has an idea.

samizdatco:
One other question I have is about the URL: would it be possible to remove the date component? My understanding was that one advantage of reading from S3 rather than poking around in the ETL repo's source files was that we could have a stable URL across updates to the underlying data.

samizdatco:
It also looks like the country `aliases` lists aren't being included (presumably because their main function is for harmonizing entity names). I've come across a few instances of alias names in older datasets though, so it would be handy to be able to correct for those on the client side.

Marigold:
> One other question I have is about the URL: would it be possible to remove the date component? My understanding was that one advantage of reading from S3 rather than poking around in the ETL repo's source files was that we could have a stable URL across updates to the underlying data.

You're totally right, fixed. It's available on https://catalog-staging.ourworldindata.org/grapher/regions/latest/regions/regions.csv now.

> historical countries (and the is_historical field)

Added!

(I can't answer the other questions, Pablo and Marcel will do better job than me.)

samizdatco:
The script on the other side of this is starting to settle down and there are now only a couple more tweaks needed to get things in sync:

1. Could you change the name of the `omit_country_page` column to `is_unlisted`
2. It's probably not a bad idea to generate slugs for all entities, not just countries (or at the least to include them for all the rows with a `region_type` of `other`)

etl/steps/data/grapher/regions/latest/regions.py
```diff
@@ -0,0 +1,305 @@
...
+    "VNM",
+    "VUT",
+    "WSM",
+    "YEM",
+    "ZAF",
+    "ZMB",
+    "ZWE",
+]
+
+# Countries with no country page in Grapher.
```
pabloarosado, line 231: I think this list is incomplete. For example, "Abkhazia" doesn't have a country page either. Maybe it's better to define:
```
# All mappable countries should have a country page, and there are some that cannot be mapped but should also have a page.
HAS_COUNTRY_PAGE = MAPPABLE_COUNTRIES + [...]
```
And if there's any mappable country that shouldn't have a page, we could also remove it from the list here.
(This is my intuition, we would need to check)

etl/steps/data/grapher/regions/latest/regions.py
```diff
@@ -0,0 +1,305 @@
...
+    # Drop unneeded columns
+    regions: Table = regions.drop(columns=["defined_by"])  # type: ignore
+
+    # Create slugs for all countries and keep track of legacy slugs.
+    regions["slug"] = regions["name"].astype(str).map(slugify)
+
+    # Countries that are mappable in Grapher.
+    regions["is_mappable"] = regions.index.isin(MAPPABLE_COUNTRIES)
+
+    # Countries that have a country page in Grapher.
```
pabloarosado, line 285: Update the comment.

etl/steps/data/grapher/regions/latest/regions.py
```diff
@@ -0,0 +1,305 @@
...
+paths = PathFinder(__file__)
+
+
+# Legacy slugs for countries that have changed their name.
+LEGACY_SLUGS = {
+    "CZE": "czech-republic",
+    "MKD": "macedonia",
+    "SWZ": "swaziland",
+    "TLS": "timor",
+}
```
pabloarosado, line 23: So, are you using `LEGACY_SLUGS`? If not, you could remove it. And if it's needed, maybe it could be added as some new field, like `legacy_redirects` or something like that.

pabloarosado (approved):
Looks good! There are a few open questions, but we can discuss them somewhere else.

etl/steps/data/grapher/regions/latest/regions.py
Marigold: @samizdatco do you know if there's any relationship between mappable and listed countries? (like that `All mappable countries should have a country page`). If yes then I agree with Pablo and can explicitly whitelist them instead of blacklisting.

etl/steps/data/grapher/regions/latest/regions.py
Marigold: Ouch, you're right. I removed them by accident. If redirects are doable then it would certainly be cleaner (though if we don't know or would be too complicated, it doesn't need to block this PR). Maybe we could just move legacy slugs to grapher so that it's closer to redirects once we decide to refactor it.

samizdatco: Agreed. I'll make the slug substitutions on my end until it's time to set up redirects on wordpress.

etl/steps/data/grapher/regions/latest/regions.py
```diff
@@ -0,0 +1,307 @@
...
+# Get paths and naming conventions for current step.
+paths = PathFinder(__file__)
+
+
+# Legacy slugs for countries that have changed their name.
+LEGACY_SLUGS = {
+    "CZE": "czech-republic",
+    "MKD": "macedonia",
+    "SWZ": "swaziland",
+    "TLS": "timor",
```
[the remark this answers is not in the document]
samizdatco: Right, we want to update these but it's going to have to happen in tandem with setting up the redirects from the old slugs to the new.

etl/steps/data/grapher/regions/latest/regions.py
Marigold: All right! Removed legacy slugs, ball is on your side Christian.

etl/steps/data/grapher/regions/latest/regions.py
Marigold: (Implemented your suggestions @samizdatco)

This is crosstab of mappability and whether it is unlisted.
```
is_unlisted  False  True
is_mappable
False           62     21
True           200      2
```

Unfortunately "All mappable countries should have a country page" doesn't hold because Western Sahara (ESH) and French Southern Territories (ATF) are mappable, but doesn't have country profiles.

I guess I could do
```
# All mappable countries should have a country page, and there are some that cannot be mapped but should also have a page.
HAS_COUNTRY_PAGE = MAPPABLE_COUNTRIES + NON_MAPPABLE_WITH_COUNTRY_PAGE - MAPPABLE_WITHOUT_COUNTRY_PAGE
```
but that's not much simpler as the original blacklisting approach. I'd probably keep it as it is now.

samizdatco: Thanks @Marigold. For some reason I'm not seeing the OWID_ZAN change reflected in the csv yet. Also, if we're keeping the blocklist approach, can you add PS_GZA and SXM to the non-country-page group?

Marigold: My bad, forgot to push this to staging. It should be there in a couple minutes.

danyx23 (approved):
Looks good - I wonder a bit if Antarctica should become a continent but that doesn't have to be part of this PR

etl/steps/data/grapher/regions/latest/regions.py
```diff
@@ -0,0 +1,299 @@
...
+    "ZAF",
+    "ZMB",
+    "ZWE",
+]
+
+# Countries with no country page in Grapher.
+NO_COUNTRY_PAGE = [
+    "ALA",
+    "ANT",
+    "ATA",
```
samizdatco, line 227: > Looks good - I wonder a bit if Antarctica should become a continent but that doesn't have to be part of this PR

The one thing that would affect this PR is that if Antarctica is going to be set to `region_type = "other"` we could remove the `ATA` item from the `NO_COUNTRY_PAGE` list.
> Looks good - I wonder a bit if Antarctica should become a continent but that doesn't have to be part of this PR

The one thing that would affect this PR is that if Antarctica is going to be set to `region_type = "other"` we could remove the `ATA` item from the `NO_COUNTRY_PAGE` list.

Otherwise, this is all looking good to go

samizdatco (approved):

etl/steps/data/grapher/regions/latest/regions.py
```diff
@@ -0,0 +1,299 @@
+"""Regions dataset adjusted for Grapher. It is published as CSV so that it can
+be easily loaded in Grapher codebase.
```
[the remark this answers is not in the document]
samizdatco: Using JSON might have simplified things initially, but the csv-specific part of the update script in `devTools/regionsImporter` is pretty minimal compared to the rest of the merging & validation that it's doing.

samizdatco:
The other thing that could be more easily added to a JSON version of the region data is the collection of the geojson features we use for each of the countries on the map. I've [posted a script](https://gist.github.com/samizdatco/80a0bdca4976ba8d697b871a5a0eac82) that builds up a dictionary of country-specific feature data that would ideally be added to the corresponding entities (and could effectively make the `is_mappable` flag unnecessary). The one difference from the current set of mappable countries is that an outline for Antarctica is included, though that could easily be filtered out by a rule on the grapher side when consuming the ETL data. 

The upside of this is that it would allow us to have a repeatable process for recreating/updating the map data we use and would keep the relationship between `isMappable` and the presence of an actual outline in sync.

etl/steps/data/grapher/regions/latest/regions.py
Marigold: Removed, good catch.

Marigold:
> but generating JSON from the ETL would still be preferable since it paves the way for further (structured) additions to the entity records.

ETL unfortunately only supports tabular data at the moment. Adding JSON would be possible, but could take some time. I've changed members to use `;` separator so that you don't need to do eval magic in grapher. We can change it in the future if we add JSON support, but it shouldn't be blocking this one.

PR #1029, opened 2023-04-24
enhance(harmonize): save aliases to regions.yml file

Marigold:
In addition to `countries_regions.csv`, save aliases also to `regions.yml` file that contains metadata about our new regions dataset.

It uses regexes to find `aliases` and replace them, which is not great, but works... I've tried using ruamel.yaml, but keeping formatting of `regions.yml` unchanged was surprisingly hard. We'd have convert it to a more standardised format first to be able to effectively use ruamel.yaml.

### DIFF
--- a/etl/harmonize.py
+++ b/etl/harmonize.py
@@ -5,6 +5,7 @@
 
 import cmd
 import json
+import re
 from collections import defaultdict
 from pathlib import Path
 from typing import DefaultDict, Dict, List, Optional, Set, cast
@@ -14,7 +15,10 @@ import pandas as pd
 from owid.catalog import Dataset
 from rapidfuzz import process
 
-from etl.paths import REFERENCE_DATASET
+from etl.paths import REFERENCE_DATASET, STEP_DIR
+
+# Path to latest regions definitions file. We update aliases from harmonize
+REGION_DEFINITIONS_FILE = STEP_DIR / "data/garden/regions/2023-01-01/regions.yml"
 
 
 @click.command()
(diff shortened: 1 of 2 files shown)


etl/harmonize.py
```diff
@@ -14,7 +15,10 @@
 from owid.catalog import Dataset
 from rapidfuzz import process
 
-from etl.paths import REFERENCE_DATASET
+from etl.paths import REFERENCE_DATASET, STEP_DIR
+
+# Path to latest regions definitions file. We update aliases from harmonize
+REGION_DEFINITIONS_FILE = STEP_DIR / "data/garden/regions/2023-01-01/regions.yml"
```
pabloarosado, line 21: Instead of needing to update this every time we create a new version, you can ensure you always load the latest version:
```

REGION_DEFINITIONS_FILE = sorted((STEP_DIR / "data/garden/regions/").glob("*/regions.yml"))[-1]
```

pabloarosado (approved):
How different would the `regions.yml` file look if you created it with `ruamel`? If you think it's helpful, we could adapt its format now. If, however, the resulting format is inconvenient, then `ruamel` is simply not very helpful.

etl/harmonize.py
Marigold: Sweet!

pabloarosado:
> Here's a [comparison](https://www.diffchecker.com/Qgr0CTZt/). We could make it more similar, but formatting with ruamel is exactly as annoying as using regex for editing YAML files... I think it's ok to use regex for now and if we find that we need to do more editing, then switch to ruamel.

OK, up to you. If we have to eventually move to ruamel format, it doesn't look too bad (and maybe there's a way to customize the output a bit).

PR #1032, opened 2023-04-24
Flu meta

spoonerf:


spoonerf:
Hi @Marigold!
I was hoping you might know why the build tests is failing, I can't figure it out :( 

I think the main thing that has changed if that I'm setting more rows to NA and that might be clashing with the categorical dtype somehow?

spoonerf:
@Marigold Ah I see - yeh that would be expected now I am pretty sure, I've raised the thresholds for dropping values for most columns. 

Let me check whether it is okay to drop them entirely or if we should keep them in with the anticipation they might not be all NA in the future.

Marigold:
@spoonerf pushed new version of owid-repack to your PR which fixes the error.

spoonerf:
Amazing, thanks! Will merge it now

PR #1041, opened 2023-04-25
enhance(walkthrough): use quotation marks in snapshot metadata fields

lucasrodes:
Sometimes, snapshot metadata field values may have special characters (such as `:`). This will raise an error due to incorrect formatting of the DVC file.

Let's force all fields with potential special characters to use quotation marks.


Marigold (approved):
Looks good. We could also get this step further and automatically format all YAML files with prettier (just like we format python files with black).

lucasrodes:
Agree; that should be the next improvement here.

PR #1043, opened 2023-04-26
enhance(walkthrough docs): adding info about local development set-up to grapher step

spoonerf:


Marigold (approved):
If setting up a local grapher is too hard, we can also spin up grapher automatically. Let me know if this is something you'd find this useful.

spoonerf:
@Marigold Ah that's good to know - it's something I had trouble with before but am mostly okay with now. I think just adding this will be okay for now. I realised it might not be super clear to new users where there local grapher is.

PR #1047, opened 2023-04-27
Add agriculture land use evolution

pabloarosado:
Add a table to the dataset of additional FAOSTAT variable, with the change in agricultural land (and croplands and pastures) with respect to one decade back.


paarriagadap (approved):
Looks great! I don't see any problem with this

PR #1052, opened 2023-04-28
Add ukraine agriculture data

pabloarosado:
New tables added to the FAOSTAT dataset of additional variables:
* Fertilizer exports in the context of the Ukraine war.
* Maize and wheat exports and uses in the context of the Ukraine war.


lucasrodes (approved):
LGTM!

Just wondering, why don't you define the variable metadata in the YML file instead of dynamically in the code?

pabloarosado:
> LGTM!
> 
> Just wondering, why don't you define the variable metadata in the YML file instead of dynamically in the code?

Thanks. I suppose I could. In this case, it's because the garden variables are not directly used in grapher, so the metadata would need to be adjusted again at the grapher step. But the original dataset didn't have any metadata either, so for now I'll leave it like this and I'll look into it again if we come back to this topic in the future.

PR #1054, opened 2023-04-30
data(wb_income): add WB classifications, including historical

lucasrodes:
Adds complete historical WB classifications to ETL.
  - [x] Snapshot
  - [x] Meadow
  - [x] Garden
  - [x] Grapher 

The dataset powering [this chart](https://ourworldindata.org/grapher/world-banks-income-groups) was uploaded manually.

Having this dataset in ETL allows us to:

- Bring WB income groups to regions
- Therefore, standardise the naming convention for these. (more details https://github.com/owid/owid-issues/issues/11)


etl/steps/data/garden/wb/2023-04-30/wb_income.py
```diff
@@ -0,0 +1,84 @@
...
+    }, f"Unknown income group label! Check {labels}"
+    # Check unusual LM* label
+    msk = df["classification"] == "LM*"
+    lm_special = set(df[msk]["country_code"].astype(str) + df[msk]["year"].astype(str))
+    assert lm_special == {"YEM1987", "YEM1988"}, f"Unexpected entries with classification 'LM*': {df[msk]}"
+    # Rename labels
+    MAPPING_CLASSIFICATION = {
+        "..": np.nan,
+        "L": "Low-income countries",
+        "H": "High-income countrires",
```
edomt, line 76: There's a small typo here: countrires -> countries

lucasrodes: good catch, fixed! thanks

dag/main.yml
```diff
@@ -337,6 +337,14 @@ steps:
   data://grapher/happiness/2023-03-20/happiness:
   - data://garden/happiness/2023-03-20/happiness
 
+  # WB income groups
+  data://meadow/wb/2023-04-30/wb_income:
```
pabloarosado, line 341: Now that you are here... I have never liked the dataset short name "wb_income". It would make way more sense to call this dataset something "income_groups", wouldn't it? Up to you, if it's too much hassle leave it.

etl/steps/data/meadow/wb/2023-04-30/wb_income.py
```diff
@@ -0,0 +1,90 @@
...
+    ds_meadow.save()
+
+    log.info("wb_income: end")
+
+
+def extract_data_from_excel(df: pd.DataFrame) -> pd.DataFrame:
+    # Sanity shape check
+    log.info("wb_income: sanity check on table dimensions")
+    assert (shape := df.shape) == (
+        shape_expected := (239, 37)
```
pabloarosado, line 62: Is this a good thing to assert? On the next update it will definitely change. Maybe it's better to assert that the shape is = 239 >= 37.

etl/steps/data/garden/wb/2023-04-30/wb_income.py
```diff
@@ -0,0 +1,84 @@
...
+        "L",
+        "LM",
+        "LM*",
+        "UM",
+        np.nan,
+    }, f"Unknown income group label! Check {labels}"
+    # Check unusual LM* label
+    msk = df["classification"] == "LM*"
+    lm_special = set(df[msk]["country_code"].astype(str) + df[msk]["year"].astype(str))
+    assert lm_special == {"YEM1987", "YEM1988"}, f"Unexpected entries with classification 'LM*': {df[msk]}"
```
pabloarosado, line 71: Venezuela used to have a nan income group. Wouldn't it be useful to warn if there are unexpected new nans?

etl/steps/data/garden/wb/2023-04-30/wb_income.py
```diff
@@ -0,0 +1,84 @@
...
+    df = df.drop(columns=["country_code"]).set_index(["country", "year"])
+    # Create a new table with the processed data.
+    tb_garden = Table(df, short_name=paths.short_name)
+
+    #
+    # Save outputs.
+    #
+    # Create a new garden dataset with the same metadata as the meadow dataset.
+    ds_garden = create_dataset(dest_dir, tables=[tb_garden], default_metadata=ds_meadow.metadata)
+    ds_garden.update_metadata(paths.metadata_path)
```
pabloarosado, line 49: The `update_metadata` happens already inside `create_dataset`, so you can delete the previous line.

etl/steps/data/garden/wb/2023-04-30/wb_income.py
```diff
@@ -0,0 +1,84 @@
...
+
+    #
+    # Process data.
+    #
+    log.info("wb_income.harmonize_countries")
+    df = geo.harmonize_countries(df=df, countries_file=paths.country_mapping_path)
+    # Harmonize income group labels
+    df = harmonize_income_group_labels(df)
+    # Set index, drop code column
+    df = df.drop(columns=["country_code"]).set_index(["country", "year"])
```
pabloarosado, line 40: Verify integrity and sort?

pabloarosado (approved):
Hey, thanks for doing this! A few comments:
* Would you mind renaming the dataset to `income_groups`?
* I understand that now the resulting dataset will have country and year, which didn't use to be the case in the previous version. I think this is good, useful information, but we may forget (or when updating other datasets it may happen without us knowing), and then income groups would be changing over time. Given that we want to define income groups as they are today, maybe it would make sense to have a table for income groups (which is the frozen classification of the latest year) and income groups historical (with all years).

etl/steps/data/garden/wb/2023-04-30/income_groups.py
```diff
@@ -0,0 +1,123 @@
...
+    #
+    # Get classifications for each country for the latest year available.
+    df_latest = df.reset_index().drop_duplicates(subset=["country"], keep="last")
+    # Sanity checks of missing countries for latest release
+    missing_countries = set(df_latest.loc[df_latest["year"] != df_latest["year"].max(), "country"])
+    assert (
+        missing_countries == EXPECTED_MISSING_COUNTRIES_IN_LATEST_RELEASE
+    ), f"Unexpected missing countries in latest release. All missing countries: {missing_countries}"
+    # Get data only for latest release (and remove column year)
+    df_latest = df_latest[df_latest["year"] == df_latest["year"].max()].drop(columns=["year"])
```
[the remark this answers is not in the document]
lucasrodes: Indeed. I sanity check that missing countries are as expected.

The way I thought about this table table is that it should only contain countries as per the latest release (2022). So, the classification for Venezuela is none, as expected. 

But ofc, if you think it makes more sense to have data for whatever the latest year available, that's easy to change

etl/steps/data/garden/wb/2023-04-30/income_groups.py
```diff
@@ -0,0 +1,123 @@
...
+    # Set index, drop code column
+    df_latest = df_latest.set_index(["country"], verify_integrity=True).sort_index()
+
+    #
+    # Create tables from dataframes
+    #
+    tb_garden = Table(df, short_name=paths.short_name)
+    tb_garden_latest = Table(df_latest, short_name=f"{paths.short_name}_latest")
```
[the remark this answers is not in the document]
lucasrodes: I think the default should be the one with years, as most of our datasets, which have `country, year`.

But, again, happy to change. I don't have a very strong opinion on this.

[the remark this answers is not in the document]
pabloarosado: I'm not sure if this has important consequences. On the one hand, I think all countries should have their "latest" income group classification (including, say, USSR). On the other hand, if WB decided to not assign any income group to a certain country (e.g. Venezuela) because of recent changes and lack of a meaningful classification, then it should be nan.

etl/steps/data/garden/wb/2023-04-30/income_groups.py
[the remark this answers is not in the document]
pabloarosado: I'm not sure if this has important consequences. On the one hand, I think all countries should have their "latest" income group classification (including, say, USSR). On the other hand, if WB decided to not assign any income group to a certain country (e.g. Venezuela) because of recent changes and lack of a meaningful classification, then it should be nan.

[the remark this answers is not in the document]
pabloarosado: OK, given that the only affected countries are those in `EXPECTED_MISSING_COUNTRIES_IN_LATEST_RELEASE`, maybe it's a good criterion to only take the latest year (I think that would be what we currently have, since we don't have income group for, say, USSR). If more countries are affected in a future release, we can rethink it.

PR #1056, opened 2023-05-01
data(oecd): Pharmaceutical market

lucasrodes:
The [OECD](https://stats.oecd.org/) has several datasets. This PR imports only the "Pharmaceutical market" dataset from the [Health database](https://www.oecd.org/health/health-data.htm).

The metadata for variables is generated dynamically in the code using fields from the data frame.

AFAIK there is only one variable relevant in the context of mental health. A description for this variable is written manually in the metadata YAML file. All other variables are kept in the dataset in case they turn out to be relevant too.

Bonus: alias for South Korea added.


dag/health.yml
```diff
@@ -150,3 +150,11 @@ steps:
     - data://garden/wb/2021-07-01/wb_income
   data://grapher/health/2023-04-26/wgm_2018:
     - data://garden/health/2023-04-25/wgm_2018
+
+  # OECD: Pharmaceutical Market
+  data://meadow/oecd/2023-05-01/oecd_pharma_market:
```
pabloarosado, line 155: Consider whether having "oecd_" in the dataset short name is necessary.

etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -1350,6 +1350,7 @@
     - "Korea, Rep."
     - "Korea, Republic Of"
     - "Republic of Korea"
+    - "Korea"
```
pabloarosado, line 1353: I'm not sure this is a good alias. "Korea" does not mean "South Korea", even if for this particular dataset it may mean that.

data/garden/reference/countries_regions.csv
```diff
@@ -226,7 +226,7 @@ SLB,Solomon Islands,SB,SLB,813.0,SOL,940.0,SOI,BP,,SLB,SLB,http://www.wikidata.o
 SOM,Somalia,SO,SOM,726.0,SOM,520.0,SOM,SO,SMLA,SOM,SOM,http://www.wikidata.org/entity/Q1045,Somalia,82.0,747.0,,
 ZAF,South Africa,ZA,ZAF,199.0,SAF,560.0,SAF,SA,SAFR,ZAF,ZAF,http://www.wikidata.org/entity/Q258,South Africa,81.0,788.0,,
 SGS,South Georgia and the South Sandwich Islands,GS,SGS,,,,,,,,,http://www.wikidata.org/entity/Q35086,South Georgia and the South Sandwich Islands,352.0,807.0,,
-KOR,South Korea,KR,KOR,542.0,ROK,732.0,ROK,KO,KORS,KOR,KOR,http://www.wikidata.org/entity/Q884,South Korea,127.0,673.0,,"[""Korea, Rep."", ""Korea, Republic Of"", ""Republic of Korea""]"
+KOR,South Korea,KR,KOR,542.0,ROK,732.0,ROK,KO,KORS,KOR,KOR,http://www.wikidata.org/entity/Q884,South Korea,127.0,673.0,,"[""Korea"", ""Korea, Rep."", ""Korea, Republic Of"", ""Republic of Korea""]"
```
pabloarosado, line 229: I'm not sure this is a good alias. "Korea" does not mean "South Korea", even if for this particular dataset it may mean that.

pabloarosado (approved):
Looks good, although I'm not sure the mapping of Korea is a good idea.

data/garden/reference/countries_regions.csv
lucasrodes: good point

PR #1058, opened 2023-05-02
Improve gcp steps

pabloarosado:
Since I need to update the CO2 dataset, I decided to update the main source of data (the Global Carbon Budget dataset). The original data has not changed, but I have improved the handling of the steps, to use the latest ETL conventions and also the latest auxiliary datasets (e.g. population and regions).

Main changes to be reviewed:
* When running the `statistical_review` step, ETL was failing when reading the metadata (not sure why that did not happen on the live server). It complained because the `table` section of the metadata yaml file didn't have a `variables` field. I fixed it, but I'm not sure if there's a better solution.
* I added the function `data_helpers.geo.list_members_of_region`, which ingests the regions and income groups datasets and returns the list of members of a region. This should eventually replace the old `list_countries_in_region` which implicitly loads data.

Although I've refactored the GCP steps, there's no need to thoroughly review them, since I checked that the data has not changed.


snapshots/gcp/2023-04-28/global_carbon_budget_fossil_co2_emissions.csv.dvc
```diff
@@ -0,0 +1,41 @@
...
+    \ N. M., Munro, D. R., Nakaoka, S.-I., Niwa, Y., O'Brien, K., Ono, T., Palmer,\
+    \ P. I., Pan, N., Pierrot, D., Pocock, K., Poulter, B., Resplandy, L., Robertson,\
+    \ E., Rödenbeck, C., Rodriguez, C., Rosan, T. M., Schwinger, J., Séférian, R.,\
+    \ Shutler, J. D., Skjelvan, I., Steinhoff, T., Sun, Q., Sutton, A. J., Sweeney,\
+    \ C., Takao, S., Tanhua, T., Tans, P. P., Tian, X., Tian, H., Tilbrook, B., Tsujino,\
+    \ H., Tubiello, F., van der Werf, G. R., Walker, A. P., Wanninkhof, R., Whitehead,\
+    \ C., Willstrand Wranne, A., Wright, R., Yuan, W., Yue, C., Yue, X., Zaehle, S.,\
+    \ Zeng, J., and Zheng, B.: Global Carbon Budget 2022, Earth Syst. Sci. Data, 14,\
+    \ 4811-4900, https://doi.org/10.5194/essd-14-4811-2022, 2022."
+  url: https://globalcarbonbudget.org/
```
lucasrodes, line 28: alternatively, you could use `>` or `|` for this long text

etl/steps/data/meadow/gcp/2023-04-28/global_carbon_budget.py
```diff
@@ -0,0 +1,241 @@
...
+    # Set an appropriate index and sort conveniently.
+    df_fossil_co2 = df_fossil_co2.set_index(["Country", "Year"], verify_integrity=True).sort_index().sort_index(axis=1)
+
+    # Create a new table and ensure all columns are snake-case.
+    tb_fossil_co2 = Table(df_fossil_co2, short_name="global_carbon_budget_fossil_co2_emissions", underscore=True)
+
+    return tb_fossil_co2
+
+
+def prepare_historical_budget(df_historical_budget: pd.DataFrame) -> Table:
```
lucasrodes, line 37: In this case, I'd lean towards function names like `make_table_historical_budget` (also in the following functions).

etl/steps/data/garden/bp/2022-12-28/statistical_review.meta.yml
```diff
@@ -41,5 +41,4 @@ dataset:
     Where the individual countries in each region are defined <a href="https://ourworldindata.org/world-region-map-definitions">in this map</a>. Additional BP regions are ignored, since they belong to other regions already included (e.g. the data for "Other Western Africa" is included in "Other Africa"). Finally, income groups are constructed following the definitions <a href="https://ourworldindata.org/grapher/world-banks-income-groups">in this map</a>.
 
 tables:
-  statistical_review:
-    title: "Statistical Review of World Energy"
+  {}
```
lucasrodes, line 44: I think you can delete this, right?

lucasrodes (approved):
LGTM!

I see that several of the processing happens in `meadow` (e.g. renaming columns). I usually do such things in `garden`. Is this a particular case?

snapshots/gcp/2023-04-28/global_carbon_budget_fossil_co2_emissions.csv.dvc
pabloarosado: This was created automatically by walkthrough.

etl/steps/data/garden/bp/2022-12-28/statistical_review.meta.yml
pabloarosado: If I delete it, ETL complains:
````
for table_name in metadata["tables"].keys():
AttributeError: 'NoneType' object has no attribute 'keys'
````

pabloarosado:
> LGTM!
> 
> I see that several of the processing happens in `meadow` (e.g. renaming columns). I usually do such things in `garden`. Is this a particular case?

Thanks Lucas! I'm not sure if all processing was totally necessary, but I thought it would be convenient to create all tables with a similar format, rather than having a dataset with very dissimilar tables. Maybe on the next version I'll move some of the processing to garden.

etl/steps/data/meadow/gcp/2023-04-28/global_carbon_budget.py
lucasrodes: The function's output will always be a `Table`, right? If it was not, I guess it would be a different function. I think it is slightly better than having generic names. But no big deal.

PR #1065, opened 2023-05-03
Add national contributions to climate change

pabloarosado:
Add snapthots, meadow, garden and grapher steps for Jones et al. (2023) dataset on national contributions to climate change.


etl/steps/data/garden/emissions/2023-05-02/national_contributions.excluded_countries.json
```diff
@@ -0,0 +1,9 @@
+[
+    "ANNEXI",
+    "ANNEXII",
+    "BASIC",
+    "EIT",
+    "LMDC",
+    "NONANNEX",
+    "Pacific Islands (Palau)"
```
spoonerf, line 8: Is this not equivalent to the country Palau?

etl/steps/data/garden/emissions/2023-05-02/national_contributions.countries.json
```diff
@@ -0,0 +1,231 @@
...
+  "Uruguay": "Uruguay",
+  "Uzbekistan": "Uzbekistan",
+  "Vanuatu": "Vanuatu",
+  "Venezuela": "Venezuela",
+  "Viet Nam": "Vietnam",
+  "Wallis and Futuna Islands": "Wallis and Futuna",
+  "Yemen": "Yemen",
+  "Zambia": "Zambia",
+  "Zimbabwe": "Zimbabwe",
+  "LDC": "Least developed countries (Jones et al. 2023)",
```
spoonerf, line 222: Having (Jones et al. 2023) doesn't feel like the perfect solution here, but I can't think of an alternative.

spoonerf (approved):
LGTM, couple of minor thoughts on entity names but no strong feelings on either.

etl/steps/data/garden/emissions/2023-05-02/national_contributions.excluded_countries.json
pabloarosado: They have data for both "Pacific Islands (Palau)" and "Palau". In GCP dataset this also happened, but it turned out that the former had data only until a certain year, and from then on, the latter had data. But now we had data for both countries overlapping. I could have looked deeper into this to figure out a better solution, but decided to simply exclude one of them...

etl/steps/data/garden/emissions/2023-05-02/national_contributions.countries.json
pabloarosado: I know, it's not a great solution :-/.

PR #1072, opened 2023-05-04
Update co2 dataset

pabloarosado:
* Improve metadata of the Jones et al. (2023) dataset on national contributions to climate change.
* Add the data from Jones et al. (2023) to the `owid_co2` dataset.

There's no need to thoroughly review `etl/steps/data/garden/emissions/2023-05-03/owid_co2.py` since it's duplicated from its previous version, with some adjustments. The main adjustment is to use the latest regions and populations datasets, as well as the Jones et al. (2023) dataset.


pabloarosado:
> LGTM but not sure why it's failing CI?

Thanks! I think it was some buildkite randomness. I've re-run it again and now it works.

PR #1075, opened 2023-05-04
fix(population): version in metadata

lucasrodes:
Version in metadata was wrong.


Marigold:
@lucasrodes none of the fields `namespace`, `short_name` and `version` should be there. They're all taken from step path.

lucasrodes:
@Marigold fix in https://github.com/owid/etl/pull/1078

PR #1077, opened 2023-05-04
refactor(data): rename countries (Timor -> East Timor, etc.)

Marigold:
Rename countries with
```
const renameMap = {
    Timor: "East Timor",
    "Saint Barthélemy": "Saint Barthelemy",
    "Åland Islands": "Aland Islands",
    "Faeroe Islands": "Faroe Islands",
    "Eritrea and Ethiopia": "Ethiopia (former)",
    "United Korea": "Korea (former)",
}
```

There's also an audit script for finding all occurrences of old names that will be run after ETL rebuilds all datasets (the script has the ability to rename them as well, but we rely on ETL to do that).

I've also added a couple of fixes / improvements to datadiff to handle renaming countries better.

## TODO
* [x] make sure datadiff only shows relevant changes


etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -474,9 +474,9 @@
     - "DEU"
 
 - code: "TLS"
-  name: "Timor"
+  name: "East Timor"
   aliases:
-    - "East Timor"
+    - "Timor"
```
pabloarosado, line 479: I think we shouldn't have this alias. Timor is an island that includes East Timor and a part of Indonesia.

etl/steps/data/garden/un/2022-07-11/un_wpp/un_wpp.countries.json
```diff
@@ -0,0 +1,267 @@
...
+  "Cambodia": "Cambodia",
+  "Indonesia": "Indonesia",
+  "Lao People's Democratic Republic": "Laos",
+  "Malaysia": "Malaysia",
+  "Myanmar": "Myanmar",
+  "Philippines": "Philippines",
+  "Singapore": "Singapore",
+  "Thailand": "Thailand",
+  "Timor-Leste": "East Timor",
+  "Timor": "East Timor",
```
pabloarosado, line 111: I don't think this mapping is correct. You could rename "Timor": "Timor (UN)"  (in principle you could rename it simply "Timor", but given all the confusion we've had, it's probably better to avoid that name).

pabloarosado: (Unless for some reason in this dataset "Timor" actually meant "East Timor")

pabloarosado (approved):
Massive! It looks good, except for two minor suggestions. Thanks a lot for this!

PR #1080, opened 2023-05-05
data(regions): rename East Timor back to Timor to stay consistent with ETL & grapher

Marigold:
`regions.yml` with East Timor is inconsistent with `countries_regions.csv` with Timor. That prevents us from region [centralisation in grapher](https://github.com/owid/owid-grapher/pull/2135). We have a migration prepared that would rename Timor -> East Timor, but that might take time and we want to unblock the grapher PR first.


pabloarosado (approved):
All right.

PR #1084, opened 2023-05-05
Archive unused steps and update country profile dependency

pabloarosado:
Update `country_profile` dependency on global carbon budget (to use the latest version), and archive unused steps.


spoonerf (approved):
LGTM, sorry for the delay!

PR #1098, opened 2023-05-11
enhance(regions): add related

lucasrodes:
- Add Greenland as "related" to Denmark
- Add Niue and Cook Island as "related" to New Zealand


pabloarosado (approved):
Looks good.

PR #1099, opened 2023-05-12
enhance(walkthrough): improve walkthrough navigation

Marigold:
* Add "Go to the next step" at the end of every step to avoid opening new pywebio app
* Save form values into shared state and prefill next step
* Remove walden step

`walkthrough all` shows index of all steps with links and `walkthrough meadow` opens meadow step directly.


walkthrough/meadow.py
```diff
@@ -88,14 +87,14 @@ def app(run_checks: bool, dummy_data: bool) -> None:
...
-                value=dummies.get("version", str(dt.date.today())),
+                value=state.get("version", str(dt.date.today())),
                 help_text="Snapshot version (usually the same as the meadow version).",
             ),
             pi.input(
                 "Snapshot file extension",
                 name="snapshot_file_extension",
                 placeholder=DEFAULT_EXTENSION,
-                value=dummies.get("snapshot_file_extension"),
+                value=state.get("snapshot_file_extension"),
```
pabloarosado, line 97: I think this should be:
`value=state.get("file_extension"),`

pabloarosado (approved):
Thanks for this, it's going to be very useful! Everything looks good, except for a small bug.

PR #1103, opened 2023-05-12
Create Women, Business and the Law additional dataset

paarriagadap:
It creates the WBL additional dataset available [here](https://wbl.worldbank.org/en/wbl-data). This dataset contains some variables not covered by new versions of the World Bank Gender dataset. Discussion in [this thread](https://github.com/owid/owid-issues/issues/537).

It seems I can't manually push the dataset to staging or live anymore, so maybe some further details are needed to edit after seeing the output in the grapher admin.


spoonerf (approved):
LGTM!

PR #1106, opened 2023-05-15
:sparkles: chart revision v2

lucasrodes:
- New back-end of our chart-revision baker (the ETL side of the chart-revision)
  - It now uses `etl.grapher_model`
  - Also attempts to simplify the logic and generalise the concept of "updating" a chart.
- New web app `walkthrough charts`, which is now based on `streamlit`, instead of pywebio
- I had to bump the minor required python version to `3.10`. I also tried pin pointing the version of `pyright`.

On this last point, for some reason, I am getting several errors when running `make test`, and also, lots of files are being refactored after running `make format`. Any ideas on this? @Marigold 

Related: https://github.com/owid/owid-grapher/pull/2217 (new Grapher-side chart approval version)


etl/chart_revision/v2/updaters/variable_update.py
```diff
@@ -0,0 +1,471 @@
...
+        # Update dimensions
+        config_new = update_config_dimensions(config_new, self.variable_mapping)
+        # Update sort
+        config_new = update_config_sort(config_new, self.variable_mapping)
+        # Validate the  configuration of the chart and remove default values (if any)
+        config_new = validate_chart_config_and_remove_defaults(config_new, self.schema)
+        return config_new
+
+
+def find_charts_from_variable_ids(variable_ids: Set[int]) -> List[gm.Chart]:
```
Marigold, line 124: This could go into grapher_helpers as a classmethod of `Variable`, might be useful elsewhere.

etl/steps/data/grapher/demography/2023-03-30/Untitled.ipynb
```diff
@@ -0,0 +1,602 @@
+{
```
Marigold, line 1: Does this file belong here?

etl/steps/data/grapher/demography/2023-03-31/Untitled.ipynb
```diff
@@ -0,0 +1,602 @@
+{
```
Marigold, line 1: Does this file belong here?

walkthrough/charts_v2/__main__.py
```diff
@@ -0,0 +1,358 @@
...
+    * **DB_USER**: `{config.DB_USER}`
+    * **DB_NAME**: `{config.DB_NAME}`
+    * **DB_HOST**: `{config.DB_HOST}`
+    """
+    )
+
+
+@st.cache_resource
+def _check_env_and_environment():
+    print(1111)
```
Marigold, line 69: Does this file belong here?

Marigold (approved):
It's huuuge, but code looks good (it's not critical anyway, so I'd encourage you to go ahead and merge, then we can open new PRs that will be more manageable). How would I test this? Do you have a good example?

etl/chart_revision/v2/updaters/variable_update.py
lucasrodes: Interesting! 

Did you mean `grapher_model`? I can't find `Variable` class in `grapher_helpers`.

If we were to put it under `Variable` in `grapher_model`, it would only be able to retrieve charts used by a single variable with one DB call, right?

As it is now, with one DB call we query for multiple variables.

Maybe we can just place this function in `grapher_helpers` as it is now. ? @Marigold

walkthrough/charts_v2/__main__.py
lucasrodes: forgot to remove the print!

lucasrodes:
@Marigold, sorry for the long PR and lack of an example.

I'd suggest the best way to test this is to play with the tool. 

You could run `walkthrough charts` and see if works or raises any error. As test datasets, I have been using:

- Old dataset: `Global Health Observatory: Age-standardized suicide rates (WHO, 2022)`
- New dataset: `Global Health Observatory: Age-standardized suicide rates (WHO, Invented)`

If it works as expected, you should be able to create and push the revisions to Grapher (I've tested it with my local Grapher). Then you can review from there in the admin page. Also, if you want to test the new UI for the grapher side, you can check out to branch `enhance/chart-rev` in `owid-grapher` (https://github.com/owid/owid-grapher/pull/2217)

lucasrodes:
It looks like there is [an error](https://buildkite.com/our-world-in-data/etl-unit-tests/builds/5016#01882e6a-d390-450d-b23f-00654132e94d) coming from one of our open-numbers pipelines. I'll re-run the build later and see if it builds.

etl/chart_revision/v2/updaters/variable_update.py
Marigold: Ugh, sorry, I meant `grapher_model.py`. There's an example of this method
```method
@classmethod
    def load_variable(cls, session: Session, variable_id: int) -> "Variable":
        return session.exec(select(cls).where(cls.id == variable_id)).one()
```
which is useful syntactic sugar for getting variable by ID. We could have similar method for `find_charts_from_variable_ids`. Having these methods on sqlmodel objects makes them more discoverable and reusable.

PR #1110, opened 2023-05-15
:sparkles: improve walkthrough experience

Marigold:
* Use `Table` instance everywhere
* Replace `: Dataset` with `cast(Dataset, ...` to fix some linting errors
* Rerun walkthrough for dummy dataset


pabloarosado (approved):
Looks good!

PR #1115, opened 2023-05-16
:hammer: gbd: adding share of population variable and removing erron…

spoonerf:
…eous drop duplicates(). This makes the size of the data significantly bigger.  @Marigold do you think there would be any issues with the size of these files? 

I only really need the `gbd_prevalence` dataset at the moment, so we could always experiment by only including the grapher step for this datasets?


Marigold:
I don't think there should be any major issues. Give it a shot, merge it and if it doesn't work we'll see the pipeline fail.

spoonerf:
@Marigold it also runs for me locally so let's give it a go and see what happens 🙏🏼

PR #1116, opened 2023-05-16
:bug: un: remove sint maarten inaccurate estimates

lucasrodes:
We were trying to filter Sint Maarten before but using the ISO code. Instead, we should be using the complete country name.

TODO: review "natural_disasters" pipeline, seems to sanity check the countries in un_wpp. Now that Sint Maarten is missing, it fails.


spoonerf (approved):
Faster than I could write an issue! LGTM.

PR #1118, opened 2023-05-16
:hammer: gbd: trying dask

spoonerf:


spoonerf:
@Marigold - perhaps this could work in combination with your memory increases?

Marigold:
Interesting! I haven't used dask much, could be worth a try since we have 32 cores on foundation. I'd focus on other steps though, `drop_duplicates` is not as slow after given enough memory. This one is taking a looong time
```
12. data://garden/ihme_gbd/2019/gbd_cause...
--
  | 2023-05-17 08:39.31 [info     ] gbd_cause.start
  | 2023-05-17 08:39.36 [info     ] exclude_countries
  | 2023-05-17 08:39.53 [info     ] harmonize_countries
  | 2023-05-17 08:40.35 [info     ] prepare_garden
  | 2023-05-17 08:53.42 [info     ] gbd_cause.end
  | OK (876s)
```
That's because of those `_pivot_number`, ... operations. I'm not sure if we can speed that up though.

spoonerf:
@Marigold Oh wow that is mad - I think you can pivot with dask too, let me give that a shot and see if it helps

spoonerf:
@Marigold 😢 Dask can't pivot with multi-index

spoonerf:
I tried to use [Modin](https://github.com/modin-project/modin) too but installing that was a real pain. Not sure if we'll find a solution here - might just close this PR?

Marigold:
I guess so. It takes long time, but it won't be run often anyway. If we find it annoying we can keep trying!

PR #1122, opened 2023-05-17
:sparkles: Improve data-diff

Marigold:
* improve data-diff by printing info about differing columns
* fix iso2 for Namibia - `NA` was loaded as NaN by pandas

### DIFF
--- a/etl/steps/data/garden/regions/2023-01-01/regions.py
+++ b/etl/steps/data/garden/regions/2023-01-01/regions.py
@@ -98,7 +98,14 @@ def run(dest_dir: str) -> None:
         df = pd.DataFrame.from_dict(yaml.safe_load(_file))
 
     # Load file of region codes.
-    df_codes = pd.read_csv(REGION_CODES_FILE)
+    # NOTE: Namibia has iso_code "NA" which would be interpreted as NaN without extra arguments.
+    df_codes = pd.read_csv(
+        REGION_CODES_FILE,
+        keep_default_na=False,
+        na_values=[
+            "",
+        ],
+    )
 
     #
     # Process data.
--- a/tests/test_tempcompare.py
+++ b/tests/test_tempcompare.py
@@ -7,8 +7,8 @@ from etl.tempcompare import df_equals
 
 def test_df_equals():
     # Test with two identical dataframes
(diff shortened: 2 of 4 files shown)


etl/steps/data/garden/regions/2023-01-01/regions.py
```diff
@@ -98,7 +98,14 @@ def run(dest_dir: str) -> None:
...
     # Load file of region codes.
-    df_codes = pd.read_csv(REGION_CODES_FILE)
+    # NOTE: Namibia has iso_code "NA" which would be interpreted as NaN without extra arguments.
+    df_codes = pd.read_csv(
+        REGION_CODES_FILE,
+        keep_default_na=False,
+        na_values=[
+            "",
+        ],
+    )
```
pabloarosado, line 108: Nice catch! Luckily the code that we used to index Namibia is "NAM" which is not interpreted as NaN.

tests/test_tempcompare.py
```diff
@@ -0,0 +1,68 @@
...
+    assert not df_equals(df1, df2, absolute_tolerance=1e-05).all().all()
+
+    # Test with categorical columns that are not identical
+    df1 = pd.DataFrame({"col1": pd.Categorical(["a", "b", "c"])})
+    df2 = pd.DataFrame({"col1": pd.Categorical(["a", "b", "d"])})
+    assert not df_equals(df1, df2).all().all()
+
+    # Test with one column having NaN values
+    df1 = pd.DataFrame({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
+    df2 = pd.DataFrame({"col1": [1, 2, np.nan], "col2": ["a", "b", "c"]})
```
pabloarosado, line 43: What about testing whether two dataframes having nan on the same position are considered equal?
```
df1 = pd.DataFrame({"col1": [1, 2, np.nan], "col2": ["a", "b", "c"]})
df2 = pd.DataFrame({"col1": [1, 2, np.nan], "col2": ["a", "b", "c"]})
assert df_equals(df1, df2).all().all()
```

pabloarosado (approved):
I haven't tried it in action, but the code looks good, thanks for catching that bug in the iso2 codes.

tests/test_tempcompare.py
Marigold: Added that. Shame on you chatgpt...

PR #1124, opened 2023-05-18
:sparkles: improve dependency graph

Marigold:
Wanted to show dependency graph so I fixed a few things to make it work and look nicer. Still could use some love.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,8 +107,8 @@ grapher: .venv
 
 dot: dependencies.pdf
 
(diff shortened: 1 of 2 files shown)


lucasrodes (approved):
LGTM ー out of curiosity, do you use the PDF anytime?

Marigold:
> LGTM ー out of curiosity, do you use the PDF anytime?

Nope. I thought it'd be useful for data pages, but the output is huge and not very useful... but I fixed a few bugs along the way, so I thought I'd merge them.

PR #1127, opened 2023-05-18
:sparkles: grapher_model: new methods for Class

lucasrodes:
Following dicussion from https://github.com/owid/etl/pull/1106#discussion_r1197624562, I'm adding some methods to class `Chart` in our `etl.grapher_model` ORM.

What do you think, @Marigold? I was unsure if we wanted a `load_chart` or a `load_charts`. Or if one could use the logic of the other without making unnecessary calls to the table.


Marigold (approved):
Really nice! It doesn't matter whether we use `load_chart` or `load_charts`, we can change it on the fly if we find a need.

PR #1133, opened 2023-05-19
Add population vars to Ethnic Power Relations dataset

paarriagadap:
Minor add to the Ethnic Power Relations dataset


Marigold (approved):
Looks good, adding variables is harmless.

PR #1134, opened 2023-05-19
✨etl.population: allowing the population age-group to just be one year in add_population()

spoonerf:


lucasrodes (commented):
Hi @spoonerf, thanks for looking into this.

What would be the motivation for having this if-else clause? The current code should work if you use `age_min` and `age_max = age_min + 1`, right?

spoonerf:
Ah yes I see, sorry I understand now. 

I wonder if having `age_max` would be more intuitive, so: 

`pop_g = pop[(pop["age"] >= age_min) & (pop["age"] <= age_max)].copy()` , as it would be more similar to how many age-groups are coded in the datasets I'm aware of. For example, the age-group 0-4 years (inclusive) would currently have to be coded with the `age_max` as 5, right? 

Maybe it would be a pain to change now though and it's not a big deal, so happy to close this if you think it is best to keep it as is.

lucasrodes:
@spoonerf oh, right. That makes total sense! I'll look into that later today and will ping you.

Issue #1131, opened 2023-05-19
walkthrough: Introduce variable mapping quality indicator in `walkthrough charts`


paarriagadap:
### One-liner
In the `walkthrough charts`' _Map variables_ section it could be nice to have an indicator of the precision of the matching, to guide through good or bad matches.

### Environment

- `python -V`: Python 3.11.1
- Operating system: macOS Ventura 13.3.1 (a)
- Working with local, staging or live: staging

### Context & details
I was testing a fix for https://github.com/owid/etl/issues/819 using the new `walkthrough charts` on staging with two completely different datasets. The platform automatically matched 10 variables, but none of them are actually even similar, so this extreme case might show the need for reinstating a mapping accuracy indicator (as it was available on the previous version of the platform)
![image](https://github.com/owid/etl/assets/63430031/9edf0152-4268-4d27-9229-51477dfaa852).

When there are multiple variables to assign the chances of missing a bad match increase, so adding the accuracy somewhere (as a number or with colors) might reduce this risk.

lucasrodes:
Thanks for raising this. Potential solution for this: #1136

PR #1136, opened 2023-05-19
✨ chart-revision: add similarity score for closes variable

lucasrodes:
The previous `walkthrough charts` version displayed the similarity score between old variables and their suggested new variables. This score appeared in the drop-down of new variables.

In this version, I chose not to have it in the dropdown as it affects search. For instance, if a user searched for variable id "93", variables with a similarity score of "93" will be returned.

Consequently, I have added a small text area next to the dropdown that displays the similarity score between an old variable and the "closest" variable in the new set. Not ideal, but should be helpful in the context of https://github.com/owid/etl/issues/1131.

image

### DIFF
--- a/walkthrough/charts_v2/__main__.py
+++ b/walkthrough/charts_v2/__main__.py
@@ -124,7 +124,6 @@ if "charts_obtained" not in st.session_state:
 old_var_selectbox = []
 new_var_selectbox = []
 ignore_selectbox = []
-variable_display_to_id = {}
 charts = []
 updaters = []
 num_charts = 0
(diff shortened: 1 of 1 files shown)


paarriagadap (approved):
It's a very good solution! The only thing I would add for now (if possible) is ordering these matches by score. That way I know there are more chances for the bottom matches to be wrong (when the coincidence is not exact, of course)
image

paarriagadap (changes requested):
Ah, I see the scores are not dynamic when I change the variable from the new dataset. That can lead to confusion.

paarriagadap (changes requested):
Ah, I see the scores are not dynamic when I change the variable from the new dataset. That can lead to confusion.

lucasrodes:
@paarriagadap Good suggestion! I'll re-order the list.

Regarding it being dynamic, I am afraid that the score cannot be adjusted based on the metric selected from the dropdown.

This comes from how [forms](https://docs.streamlit.io/library/api-reference/control-flow/st.form) behave in streamlit.

Also, that would likely affect the sorting of how variables are presented in the list, which adds complexity to the code.

I'll leave this for future improvements. Maybe something new in streamlit comes out that makes this easy.

lucasrodes:
@paarriagadap Added your suggestion on sorting + some colouring.

paarriagadap (approved):
Thank you very much! Looks nicer, I think 
image

PR #1138, opened 2023-05-20
🎉 chart-rev: add preview of variables

lucasrodes:
- Add preview of variables
- Improve code organisation for walkthrough app (now in different modules)
- Added option to skip revision of the time slider.
- Add tool `etl-chartgpt` to generate chatGPT-based chart revision suggestions.

For reviewers: It is fine just to play with the tool itself and report any errors  (`walkthrough charts`)


paarriagadap:
I will add the comments here, as I just am playing with the tool.

- [ ] The link here doesn't take you to the tool: **Chart revisions submitted successfully! Now review these at the [approval tool](http://localhost:8082/None/suggested-chart-revisions/review)!**
- [ ] Using both the recommended option and Yes it doesn't expand the slider of [this old chart](https://owid.cloud/admin/charts/385/edit):
image

- [ ] The mapping worked worse that expected, because I wanted to get Gini data:
image
I needed to change it manually:
image

- [ ] For now I don't see the difference with the experimental feature, as I see the same results. But it took like 10 seconds to load.

The old dataset is 318 and the new one is 5873:
image

paarriagadap:
Ah, now I see the experimental feature is activated through the checkbox, it wasn't clear for me at the beginning. I will add comments for that then

paarriagadap:
I like the Explore feature, and it's good that values are sorted with the (absolute?) percentual difference. I think it's good to have that already, but if filters can be added to those tables it can be even better. Also if there is color coding for that relative difference (like a traffic light or with bars)

paarriagadap:
Maybe the relative difference is better expressed in percentage terms? Here I had the case of using Gini and the numbers looked similar. I am aware if dealing with percentage variables I would face a similar problem, though

paarriagadap (changes requested):
Looks awesome! I added the comments above. There seems to be something to be done with the slider-expansion feature, which didn't work for me on this example

paarriagadap (changes requested):
Looks awesome! I added the comments above. There seems to be something to be done with the slider-expansion feature, which didn't work for me on this example

lucasrodes:
@paarriagadap, thanks for your help! Very much appreciated. Regarding your points:

> The link here doesn't take you to the tool: Chart revisions submitted successfully! Now review these at the [approval tool](http://localhost:8082/None/suggested-chart-revisions/review)!

I think this is fixed now

> Using both the recommended option and Yes it doesn't expand the slider of [this old chart](https://owid.cloud/admin/charts/385/edit):

This is the expected behaviour. Hardcoded years will pass along. I've improved the help text for that option to make this more obvious.

>  The mapping worked worse that expected, because I wanted to get Gini data:

There is not much we can do at the moment other than changing the similarity function used. In the future we can work on improving these similarity functions.

>  For now I don't see the difference with the experimental feature, as I see the same results. But it took like 10 seconds to load.

has been clarified

> I like the Explore feature, and it's good that values are sorted with the (absolute?) percentual difference. I think it's good to have that already, but if filters can be added to those tables it can be even better. Also if there is color coding for that relative difference (like a traffic light or with bars)

I'll change that using percentual difference makes it more readable, thanks for the suggestion!

I have added some filters for the countries.

I have added some colouring to the cells.

> Looks awesome! I added the comments above. There seems to be something to be done with the slider-expansion feature, which didn't work for me on this example

What feature do you mean here?

PR #1139, opened 2023-05-22
✨ Add age-groups to WHO mortality database homicide dataset

spoonerf:


lucasrodes (approved):

spoonerf:
Thank you so much for updating the age-groups function @lucasrodes and for creating it in the first place, it's incredibly useful! I'm just checking in with Bastian about what age-groups we might want to use and then I will merge this.

PR #1140, opened 2023-05-22
:construction: Migrate excess mortality economist

Marigold:
Migration of Excess Mortality from the economist as per https://github.com/owid/owid-issues/issues/921

Migrated with `ENV=.env.prod backport-migrate --dataset-id 5429 --namespace excess_mortality --version latest --short-name excess_mortality_economist` and then by manually changing steps to private ones.


lucasrodes:
Thanks for doing this, @Marigold! This is very valuable!

Two things from the top of my head, both related to the tasks done in the `meadow` step:

- I see that external files are accessed from the `meadow` step. Why not from `snapshot`? We could have a `.dvc` file per each and import them there, right? Or am I missing something here?
- Most processing occurs in `meadow` (and not in `garden`). Was this simpler to do there, as you were already merging and combining different files?

Marigold:
> I see that external files are accessed from the meadow step. Why not from snapshot? We could have a .dvc file per each and import them there, right? Or am I missing something here?

`etag://...` step works similarly to snapshot. If a file on github gets updated, its etag will change and we will rerun its downstream steps. With snapshots we'd have to do more work like running the ingest script ourselves.

> Most processing occurs in meadow (and not in garden). Was this simpler to do there, as you were already merging and combining different files?

I didn't think of it much. Does it matter if I do the merging in garden vs meadow? My idea was to get all the data together in meadow and harmonize in garden.

lucasrodes:
thanks for the summary and for doing this work!

I think it looks good as it is

etl/steps/data/garden/excess_mortality/latest/excess_mortality_economist/excess_mortality_economist.meta.yml
```diff
@@ -0,0 +1,133 @@
+dataset:
+  title: Excess Mortality Estimates – The Economist (2023)
```
lucasrodes, line 2: I would change the name to "Excess Mortality (The Economist, 2023)", to be more consistent with the other dataset ["Excess Mortality (various sources)"](https://owid.cloud/admin/datasets)

lucasrodes (approved):
LGTM! Just minor suggestion on the dataset title.

PR #1141, opened 2023-05-23
:bug: switch key_indicators from datautils.geo to data_helpers.geo

Marigold:
Key indicators were still using `from owid.datautils import geo` instead of `from etl.data_helpers import geo` which caused problems for nightly build. Fix this dependency by passing population directly.

@pabloarosado is this the right time to remove `geo` module from `datautils` package? Or do we want to keep it there for some reason?

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -35,6 +35,7 @@ NUM_ALLOWED_NANS_PER_YEAR = None
 DATASET_KEY_INDICATORS = DATA_DIR / "garden" / "owid" / "latest" / "key_indicators"
 TNAME_KEY_INDICATORS = "population"
 # Path to Key Indicators dataset
+# TODO: we should update it to latest garden/wb/*/income_groups dataset
 DATASET_WB_INCOME = DATA_DIR / "garden" / "wb" / "2021-07-01" / "wb_income"
 TNAME_WB_INCOME = "wb_income_group"
 
@@ -157,12 +158,15 @@ def list_countries_in_region_that_must_have_data(
         Countries that are expected to have the largest contribution.
 
     """
+    # TODO: we should be passing countries_regions explicitly and get rid of `_load_countries_regions`
     if countries_regions is None:
         countries_regions = _load_countries_regions()
 
+    # TODO: we should be passing population explicitly and get rid of `_load_population`
     if population is None:
         population = _load_population()
 
+    # TODO: we should be passing income groups explicitly and get rid of `_load_income_groups`
     if income_groups is None:
         income_groups = _load_income_groups().reset_index()
 
--- a/etl/paths.py
+++ b/etl/paths.py
@@ -10,11 +10,13 @@ SNAPSHOTS_DIR = BASE_DIR / "snapshots"
 STEP_DIR = BASE_DIR / "etl" / "steps"
 
 # Regions paths
-LATEST_REGIONS_REGIONS_YML = sorted((STEP_DIR / "data/garden/regions/").glob("*/regions.yml"))[-1]
-LATEST_REGIONS_DATASET_PATH = BASE_DIR / LATEST_REGIONS_REGIONS_YML.relative_to(STEP_DIR).with_suffix("")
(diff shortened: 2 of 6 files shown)


etl/paths.py
```diff
@@ -10,11 +10,13 @@
...
 # Regions paths
-LATEST_REGIONS_REGIONS_YML = sorted((STEP_DIR / "data/garden/regions/").glob("*/regions.yml"))[-1]
-LATEST_REGIONS_DATASET_PATH = BASE_DIR / LATEST_REGIONS_REGIONS_YML.relative_to(STEP_DIR).with_suffix("")
+LATEST_REGIONS_VERSION = sorted((STEP_DIR / "data/garden/regions/").glob("*/regions.yml"))[-1].parts[-2]
+LATEST_REGIONS_YML = STEP_DIR / "data/garden/regions" / LATEST_REGIONS_VERSION / "regions.yml"
+LATEST_REGIONS_DATASET_PATH = BASE_DIR / "data/garden/regions" / LATEST_REGIONS_VERSION / "regions"
 
 # WB Income
-LATEST_INCOME_DATASET_PATH = BASE_DIR / LATEST_REGIONS_REGIONS_YML.relative_to(STEP_DIR).with_suffix("")
+LATEST_INCOME_VERSION = sorted((STEP_DIR / "data/garden/wb/").glob("*/wb_income.ipynb"))[-1].parts[-2]
```
pabloarosado, line 18: Now that we are here, we could try to use the latest income groups dataset, which is called `income_groups`, and stop using `wb_income`. However, within `key_indicators` we should keep using the old `wb_income` dataset, to avoid abrupt changes.

etl/data_helpers/geo.py
```diff
@@ -54,7 +51,7 @@ def _load_countries_regions() -> pd.DataFrame:
 
 @functools.lru_cache
 def _load_income_groups() -> pd.DataFrame:
-    income_groups = Dataset(DATASET_WB_INCOME)[TNAME_WB_INCOME]
+    income_groups = Dataset(LATEST_INCOME_DATASET_PATH)["wb_income_group"]
```
pabloarosado, line 54: We should at some point stop using these kinds of functions. If you think about it, they will, under the hood, upload the latest versions of e.g. the income groups dataset. But in the dag, the version of the income groups dataset (if it's even explicitly given) may be a different one. So we should move to use functions that always ingest the data they need.

etl/steps/data/garden/owid/latest/key_indicators/utils.py
```diff
@@ -20,5 +21,5 @@ def add_regions(df: pd.DataFrame) -> pd.DataFrame:
     ]
     df = df.loc[-df.country.isin(regions)]
     for region in regions:
```
pabloarosado, line 23: Here we could use 
```
countries_in_region = geo.list_members_of_region(region=region, ds_regions=ds_regions, ds_income_groups=ds_income_groups)
df = geo.add_region_aggregates(df=df, region=region, countries_in_region=countries_in_region, population=population)
```
Where `ds_income_groups` should be the old hardcoded version of the `wb_income` dataset, and not the latest.

pabloarosado (approved):
If you want you can merge the PR as it is now. But I pointed out that there's a newer version of the income groups dataset. So, in principle, in `etl.paths` we should write the path to the newest version. However, I don't think it's a good idea to modify income groups in the `key_indicators` dataset (and many other datasets). If we did that, we would need to modify the dependencies of `wb_income` in many places in the dag.
So, for a different (big) PR, what we need to do is find all uses of `add_region_aggregates` and ensure we don't load population, regions or income groups under the hood. Then we can finally get rid of the problematic `_load_population()`, `_load_countries_regions()` and `_load_income_groups()` functions.

pabloarosado:
Yes, if we are not using `geo` anywhere, we could finally get rid of that module in datautils. Lucas suggested creating a future warning, and on next version getting rid of it. But I suppose (given that it's an unofficial version) you can simply bump to a new version without geo module in it.
Thanks a lot for doing this!

PR #1146, opened 2023-05-23
:honeybee: set default VSCode settings

Marigold:
Add more defaults that could be useful for new hires
* format on save
* isort on save
* enable linting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ ign.*
 .idea
 examples/data/*
 
+playground*
 playground.ipynb
 !walkthrough/**/playground.ipynb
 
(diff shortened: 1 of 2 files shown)


Marigold:
@veronikasamborska1994 could you try whether this works for you? When you update a file (e.g. add 5 new lines between functions) and save it, it should be formatted automatically.

veronikasamborska1994:
hmm just after saving? doesn't seem to work for me but might be doing something wrong

Marigold:
Great! Can you then approve this PR, please?

veronikasamborska1994 (approved):

PR #1153, opened 2023-05-25
📊 (UNODC): Adding further tables and data disaggregations

spoonerf:
@Marigold - I've added a bunch of disaggregations to this existing dataset and as a result the variable names have changed but the underlying data has not. Do I need to change the dataset version to not anger the etl gods? I think the answer is yes but just wanted to check with you first.


etl/steps/data/garden/homicide/2023-01-27/unodc.py
```diff
@@ -29,13 +30,14 @@ def run(dest_dir: str) -> None:
...
 
     # create new dataset with the same metadata as meadow
-    ds_garden = create_dataset(
-        dest_dir, tables=[Table(df, short_name=tb_meadow.metadata.short_name)], default_metadata=ds_meadow.metadata
-    )
-    ds_garden.save()
+    ds_garden = Dataset.create_empty(dest_dir, metadata=ds_meadow.metadata)
+    for tb in tb_garden_list:
+        ds_garden.add(tb)
+        ds_garden.save()
```
Marigold, line 40: You should be able to use
```
ds_garden = create_dataset(dest_dir, tables= tb_garden_list, default_metadata=ds_meadow.metadata)
```

Marigold (approved):
Looks good!

PR #1157, opened 2023-05-26
Add data from Wuepper et al. (2020) paper

pabloarosado:
Add snapshots, meadow, garden and grapher steps for Wuepper et al. (2020) data, and update dag.


pabloarosado:
Hi @Marigold I've passed this PR to you, since @paarriagadap is off today. It has two subbranches waiting, so I'd appreciate if someone could take a quick look (it's quite simple). Thanks!

Marigold (approved):
Looks good! (I think it'd be fine to merge even without review given that it blocks you and it only adds new dataset)

PR #1159, opened 2023-05-26
Add ray et al. data on the impact of climate change on crop yields

pabloarosado:
Add ray et al. data on the impact of climate change on crop yields.


etl/steps/data/garden/papers/2023-05-26/ray_et_al_2019.meta.yml
```diff
@@ -0,0 +1,94 @@
...
+      maize_change_in_yield:
+        title: "Maize yield change"
+        unit: "tonnes per hectare"
+        short_unit: "t/ha"
+      maize_change_with_respect_to_current:
+        title: "Maize yield change with respect to current yield"
+        unit: "%"
+        short_unit: "%"
+      oilpalm_change_in_yield:
+        title: "Oilpalm yield change"
```
spoonerf, line 32: The tiniest quibble, I think it is two words as Oil palm, that is how they refer to it in the main text of the paper.

snapshots/papers/2023-05-26/ray_et_al_2019.xlsx.dvc
```diff
@@ -0,0 +1,31 @@
...
+    ONE 14(5): e0217148. https://doi.org/10.1371/journal.pone.0217148'
+  url: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0217148
+  source_data_url: https://doi.org/10.1371/journal.pone.0217148.s017
+  file_extension: xlsx
+  license_url: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0217148
+  license_name: Creative Commons Attribution License
+  date_accessed: 2023-05-26
+  is_public: true
+  description: |
+    Ray et al. (2019) modelled the impact of observed climate change on yields of the top ten global crops. These ten crops account for ~83% of global kilocalorie production.
```
spoonerf, line 22: I don't think it's that clear what period the paper and the data values are covering? Is the start 1974 and the values shown here are the average of 2009-13? I think it would be useful to clarify what exactly is shown.

spoonerf (approved):
LGTM, just a couple of points on the metadata front

etl/steps/data/garden/papers/2023-05-26/ray_et_al_2019.meta.yml
pabloarosado: Thanks!

snapshots/papers/2023-05-26/ray_et_al_2019.xlsx.dvc
pabloarosado: I agree, it's not clear. I added a note in the code, but for now I'll simply keep the dataset description that we have [in grapher](https://owid.cloud/admin/datasets/5344).

PR #1164, opened 2023-05-29
added World and tourism adjusted emissions

veronikasamborska1994:


pabloarosado (approved):
The metadata changes look good. I'm a bit curious about the choice of using columns for regions (instead of having a "country" column, as we usually do). But feel free to merge if this is blocking you and we can discuss it some other time.

PR #1167, opened 2023-05-31
Add long run yields

pabloarosado:
Create a new dataset that has all data, including long-run series, on crop yields.


pabloarosado:
After some discussions on [this issue](https://github.com/owid/owid-issues/issues/139) and on [this slack thread](https://owid.slack.com/archives/C03NA1B2P1U/p1686037107635999), we have decided to use the long-term crop yields data by default in all our charts and the explorer.
So I have made two changes:
* I created a new version of the `attainable_yields` dataset, that ingests `long_term_crop_yields` instead of `faostat_qcl`.
* I created a new version of the crop-yields explorers step, that ingests the latest `attainable_yields` dataset.
I'll merge this PR, check the new explorer and charts, and possibly come back to this on a different PR.

PR #1169, opened 2023-05-31
Add new detected anomalies to FAOSTAT QCL

pabloarosado:
Add new detected anomalies to FAOSTAT QCL.

Fixes https://github.com/owid/owid-issues/issues/1071


Marigold (approved):
Your anomaly framework is really cool. We should consider using it for other datasets too.

PR #1177, opened 2023-06-02
🐛 chart-rev: update sortBy.sortColumnSlug

lucasrodes:


pabloarosado (approved):
Thanks! I've tried it on staging and it seems to fix the issue.

PR #1184, opened 2023-06-05
Fix missing sources in energy-related datasets

pabloarosado:
Some variables in some energy-related were missing sources. This PR should ensure that all variables have the right sources.

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -150,6 +150,44 @@ def create_dataset(
     return ds
 
 
+def create_dataset_with_combined_metadata(
+    dest_dir: Union[str, Path], datasets: List[catalog.Dataset], tables: List[catalog.Table], *args, **kwargs
+) -> catalog.Dataset:
+    # Gather unique sources from the original datasets.
+    sources = []
+    licenses = []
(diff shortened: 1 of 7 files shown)


pabloarosado:
Hey @Marigold a bit of further context. I'm doing this PR because, for example, [this variable](https://owid.cloud/admin/variables/700435) does not have all sources mentioned in the "Data published by" field in the SOURCES tab (it shows only BP, but it should show BP and Shift). So this PR attempts to ensure that all variables have all sources.
However, I'm running this code on staging, and it doesn't seem to fix the issue: [That variable](https://staging.owid.cloud/admin/variables/712169) still shows only BP as a source.
But if I load the ETL grapher dataset, I see that the variable actually has two sources. So do you know what's going on? Thanks!

pabloarosado:
Hi @Marigold I'll merge this for now (since currently the sources' tab is showing incomplete sources). But feel free to take a look whenever you have some time, and let me know if you agree on this solution, thanks!

Marigold:
Good call on merging this without review. The new helper function is nice.

PR #1186, opened 2023-06-05
Update long-run UK electricity dataset and archive old datasets

pabloarosado:
Created new garden and grapher steps for the long-run UK electricity dataset (with only minor changes with respect to the previous version), and archived unnecessary steps.
NOTE: I changed one of the dependencies of the country profiles dataset, but the data should not be affected.


spoonerf (approved):
LGTM!

PR #1188, opened 2023-06-06
:bar_chart: migrate wheat prices from grapher to ETL

Marigold:
Wheat prices dataset that was generated by running

```python
ENV=.env.prod backport-migrate --dataset-id 3222 --namespace papers --short-name wheat_prices --version 2023-06-06
```

The other changes in this PR are improvements to backport-migrate (mainly moving sources to snapshot).

Dataset was [uploaded to staging](http://staging-site-mojmir/admin/datasets/6021).

Some questions / ideas for @pabloarosado:
1. I wasn't able to find license for the data from the [github package](https://github.com/robjhyndman/fma). Is it ok if we omit it or should we go with some default one?
2. [SnapshotMeta](https://github.com/owid/etl/blob/e95466156f792955db6c79c51457ae88087cc37e/etl/snapshot.py#L93) has fields like this:
```
meta:
  ...
  description: ...
  source_name: ...
  url: ...
  source_published_by: ...
  license_url: ...
  license_name: ...
```
those fields are identical to fields in `Source` and `License` object that we use later for datasets. Would it be useful to start using them in Snapshot?
```
meta:
  ...
  source:
    name: ...
    description: ...
    url: ...
    published_by: ...
  license:
    url: ...
    name: ...
```
I think it would be much cleaner and we could avoid a lot of boilerplate code when propagating metadata from snapshot to meadow. (I actually started working on this already, but then realised that there's a lot to update in walkthrough etc. and wasn't sure whether it's worth the time. What do you think? I find it more and more annoying.)

3. This is identical copy of dataset we have in grapher. Is it ok to merge it like this or should we improve / update anything?


pabloarosado:
Hi @Marigold after [this discussion](https://github.com/owid/owid-issues/issues/694) I think the best would be to add a docstring to `snapshots/papers/2023-06-06/wheat_prices.py`, explaining that the data can be accessed from R, by running the following commands:
```
install.packages("fma")
library(fma)
plot(wheat)
```
I suppose the real sources of the data would be stated in the Makridakis (1990) book, but I can't access it. So I suppose it's fine to cite Makridakis et al. 1997.
So feel free to merge if you agree.

PR #1191, opened 2023-06-06
Co2 source edit

veronikasamborska1994:


etl/steps/data/garden/oecd/2023-05-19/co2_air_transport.meta.yml
```diff
@@ -1,12 +1,15 @@
...
+  description: This database includes annual, quarterly, and monthly information on
+    CO2 emissions related to commercial passenger, freight, and general aviation flights,
+    on both a territory and a residence basis, for 186 countries. These CO2 emissions
+    are estimated by the OECD, based on a consistent methodology across countries.
   licenses:
   - name: Creative Commons BY 4.0
     url: https://www.oecd.org/termsandconditions/#:~:text=You%20must%20give%20appropriate%20credit,(accessed%20on%20(date)).
   sources:
-  - name: OECD (2023), Air Transport CO2 Emissions
+    name: Clarke et al. (2022)
```
spoonerf, line 11: I think you need the '-' back in here, I believe that's why the CI tests are failing atm

etl/steps/data/garden/oecd/2023-05-19/co2_air_transport.meta.yml
```diff
@@ -78,8 +85,15 @@ tables:
         short_unit: lb
```
spoonerf, line 85: I'm curious why we are using lb here - I would have thought kg would have been the more natural unit?

spoonerf (commented):
Looks good to me! I really like the monthly stacked bar chart. 

I think the missing '-' in the metadata should fix the buildkite issue and just had a query about why you used pounds as the unit, rather than kg?

PR #1193, opened 2023-06-07
Add data on food expenditure in the US

pabloarosado:
Add snapshots/meadow/garden/grapher steps on food expenditure in the US from the USDA ERS.

The resulting grapher dataset should replace [this one](https://owid.cloud/admin/datasets/2659).
I couldn't find data for "Restaurant prices relative to retail prices", which is used in [one chart](https://owid.cloud/admin/charts/2402/edit). However, this chart is [safe to be deleted](https://docs.google.com/spreadsheets/d/1RFKqagbs8eSdfUOAZTKSOXO7RdopOelxiGGkSA6KPLc/edit#gid=616176227), so we may not need that indicator anymore.


Marigold (approved):
Looks good, ran it with local grapher and checked some indicators.

PR #1194, opened 2023-06-07
co2 emissions source change + changed small to medium in oil spills

veronikasamborska1994:


spoonerf (approved):
Looks good! 

Slight nit-pick, maybe you could add descriptions to the Month and Region variables in the co2 metadata? 

Something like: 

```
january:
     title: January
     description: Total CO2 emissions from aviation (is it domestic and international combined?) 
```
Something like that would be helpful for understanding what is shown in the grapher admin.

PR #1198, opened 2023-06-08
:bar_chart: Commodity price index

Marigold:
Replacement & update of [Real Commodity Price Index since 1850 - Jacks (2016)](https://owid.cloud/admin/datasets/699) dataset.

### DIFF
--- a/etl/steps/data/garden/papers/2023-06-07/commodity_prices.py
+++ b/etl/steps/data/garden/papers/2023-06-07/commodity_prices.py
@@ -1,0 +1,37 @@
+"""Load snapshot and create a garden dataset."""
+
+import pandas as pd
+from owid.catalog import Table
+from structlog import get_logger
+
+from etl.helpers import PathFinder, create_dataset
+
+log = get_logger()
+
+# Get paths and naming conventions for current step.
+paths = PathFinder(__file__)
+
+
+def run(dest_dir: str) -> None:
+    #
+    # Load data from snapshot.
+    #
+    snap = paths.load_snapshot_dependency()
+    df = pd.read_excel(snap.path, skiprows=1).rename(columns={"(1900=100)": "year"})
+    df["country"] = "World"
+    df = df.set_index(["country", "year"])
+
+    #
+    # Process data.
+    #
+    # Create a new table and ensure all columns are snake-case.
+    tb = Table(df, short_name=paths.short_name, underscore=True)
+
+    #
+    # Save outputs.
+    #
+    # Create a new garden dataset with the same metadata as the snapshot.
+    ds_garden = create_dataset(dest_dir, tables=[tb], default_metadata=snap.metadata)
+
+    # Save changes in the new garden dataset.
+    ds_garden.save()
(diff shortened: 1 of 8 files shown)


snapshots/papers/2023-06-07/commodity_prices.xlsx.dvc
```diff
@@ -0,0 +1,24 @@
...
+  source_name: 'Jacks, D.S. (2019), "From Boom to Bust: A Typology of Real Commodity
+    Prices in the Long Run." Cliometrica 13(2), 202-220.'
+  source_published_by: 'Jacks, D.S. (2019), "From Boom to Bust: A Typology of Real
+    Commodity Prices in the Long Run." Cliometrica 13(2), 202-220.'
+  url: http://www.sfu.ca/~djacks/data/boombust/index.html
+  source_data_url: 
+    http://www.sfu.ca/~djacks/data/boombust/Real%20commodity%20prices,%201850-2020.xlsx
+  file_extension: xlsx
+  date_accessed: 2023-06-07
+  is_public: true
```
pabloarosado, line 16: I'd add as license either "Public domain" or "CC BY" (given that the author says "These data are freely provided to all comers, and questions and/or comments are welcome.
At the same time, I ask that if you use any of these files that you cite the following: ..."). You can use that link as the license_url:
https://www.sfu.ca/~djacks/data/boombust/index.html

pabloarosado (approved):
Looks good.

etl/steps/data/garden/papers/2023-06-07/commodity_prices.py
```diff
@@ -0,0 +1,35 @@
...
+    # Load data from snapshot.
+    #
+    snap = paths.load_snapshot_dependency()
+    df = pd.read_excel(snap.path)
+
+    #
+    # Process data.
+    #
+    # Create a new table and ensure all columns are snake-case.
+    tb = Table(df, short_name=paths.short_name, underscore=True)
```
pabloarosado, line 26: Maybe you could add a country column (with "World" in all rows). Actually you may need it for grapher.

snapshots/papers/2023-06-07/commodity_prices.xlsx.dvc
Marigold: Thanks! I'm new to licensing.

etl/steps/data/garden/papers/2023-06-07/commodity_prices.py
Marigold: Done!

Marigold:
@pabloarosado I realised we don't use "Data publisher's source" in ETL and can't propagate it all the way down to grapher (we are not [converting it from Snapshot](https://github.com/owid/etl/blob/master/etl/steps/data/converters.py#L49)). Have we ever discussed this? Should "Data publisher's source" be a thing?

image

pabloarosado:
> @pabloarosado I realised we don't use "Data publisher's source" in ETL and can't propagate it all the way down to grapher (we are not [converting it from Snapshot](https://github.com/owid/etl/blob/master/etl/steps/data/converters.py#L49)). Have we ever discussed this? Should "Data publisher's source" be a thing?
> 
> image

We have had some discussions about whether we should use it. I have never used, but I know that others have (but in a quite inconsistent way). If you think about it, having the option to list the source's sources is possibly a good idea. But then we have to decide how to structure this field properly (and what to do with it). Yes, this is something we should discuss.

PR #1206, opened 2023-06-09
fix(space): converting column name entity -> country

spoonerf:


spoonerf:
Hey @edomt - I think the etl was failing because grapher needs the entity column to be called country, so just changing that here for the exoplanet dataset.

spoonerf:
Oops saw you have one open already!

edomt:
Yes! Thank you though ☺️

PR #1211, opened 2023-06-12
fixed descriptions

veronikasamborska1994:
sorry, need to fix this as I didn't realised it would show up in sources tab


spoonerf (approved):
Good point! Might also be worth editing the descriptions for the region variables.

PR #1214, opened 2023-06-13
:tada: upload data to R2 instead of S3

Marigold:
When running --grapher step, upload data to R2 rather than to S3. Note that catalog is still published to S3 (R2 migration is likely planned at some point,  there's [even PR](https://github.com/owid/etl/pull/583) that supports publishing to R2).

After this change, everyone will have to setup extra .env variables
```
R2_ACCESS_KEY=xxx
R2_SECRET_KEY=xxx
R2_ENDPOINT_URL=https://078fcdfed9955087315dd86792e71a7e.r2.cloudflarestorage.com
```
It'd keep working if people still push to S3 because there's a fallback in Cloudflare worker, but the sooner we start pushing to R2 the better.

## Checklist for Data API
* [x] merge [MySQL paths migration](https://github.com/owid/owid-grapher/pull/2315)
* [x] set R2 credentials on `etl-prod-1` and `etl-staging-1` servers
* [ ] inform others to set R2 credentials
* [x] merge this PR
* [ ] reset all checksums in live DB to trigger update with `update datasets set sourceChecksum = 'refresh' where sourceChecksum is not null;` (do it overnight)
* [ ] (manually upsert [datasets](https://github.com/owid/ops/blob/main/templates/etl-prod/etl-prod.sh#L89) that are excluded from automatic deploy)
* [ ] enable backup of R2 buckets

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,5 +7,9 @@ DB_PASS=grapher
 DB_PORT=3307
 DB_HOST=127.0.0.1
 
+R2_ACCESS_KEY=xxx
+R2_SECRET_KEY=xxx
+R2_ENDPOINT_URL=https://078fcdfed9955087315dd86792e71a7e.r2.cloudflarestorage.com
+
 # optional
 # BUGSNAG_API_KEY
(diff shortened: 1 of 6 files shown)


pabloarosado (approved):
Looks good. But what's going to happen with private data?

Marigold:
> Looks good. But what's going to happen with private data?

All variables (even private ones) must be in public bucket so that grapher can load them in the browser. We do it in the same way even now.

PR #1215, opened 2023-06-13
Add coah dataset to faostat

pabloarosado:
I added the FAOSTAT CoAHD dataset and, now that I was in the midst of it, I also updated all other datasets. It was surprisingly easy (except for the expected changes in FAOSTAT data).

The bad news is that, now that we have the data, I have realised that it is not enough to replace [Food affordability (World Bank; FAO, 2022)](https://owid.cloud/admin/datasets/5651). Even if the underlying data is the same, they add some new variables, so it may be easier to get the data from WB.

So, in a way, this PR was unnecessary... Although it was a good exercise to see how a new FAOSTAT update would be, and I fixed a few things.

TODO: Once this PR is merged, branch `add-coah-dataset-to-faostat-base-branch` will be merged to `master`.


Marigold (approved):
Looking good, I skimmed through the code and found no issues and was able to run it locally.

PR #1225, opened 2023-06-14
:sparkles: don't remove datasets from remote catalog

Marigold:
Deleting old dataset from remote S3 catalog could cause problems when incrementing version of a dataset (and archiving an old one). With this PR and turning off [pruning](https://github.com/owid/etl/blob/master/etl/prune.py) all previous versions will stay in the catalog and in S3.


pabloarosado (approved):
Thanks, looks good!

PR #1226, opened 2023-06-14
📊 ihme: custom dataset for a chart (using causes as entities)

lucasrodes:
Adding a new custom dataset so we can create a specific chart where we show DALYs and prevalence for different diseases for the World.

more context here: https://owid.slack.com/archives/C04VBQ00HQR/p1686749360343129


spoonerf (approved):
Looks cool!

PR #1232, opened 2023-06-15
✨ underscore: format camelCase fields as snake_case

lucasrodes:
It is often the case that the source codes their fields using `camelCase`. In such instances, we want to preserve the token separation with a `_` instead of a "lower- to upper-case" change.

An example:

- Currently: `numberOfParticipants` → `numberofparticipants`
- New: `numberOfParticipants` → `number_of_participants`

To this end, I've slightly modified our function `underscore` to perform camelCase to snake_case formatting. This is only triggered if the argument `camel_to_snake` is True. By default, this argument is set to False so we don't break any of the pipelines (i.e. some metadata files might expect the older formatting).

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -86,6 +86,7 @@ def create_dataset(
     tables: Iterable[catalog.Table],
     default_metadata: Optional[Union[SnapshotMeta, catalog.DatasetMeta]] = None,
     underscore_table: bool = True,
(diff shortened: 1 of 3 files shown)


Marigold (approved):
Looks good, could you add a [couple of tests](https://github.com/owid/etl/blob/master/lib/catalog/tests/test_utils.py) please? Just to make sure no one breaks it in the future.

PR #1234, opened 2023-06-15
:honeybee: Reduce import overhead of etl.helpers

larsyencken:
Reduce by about 40% the import time for `etl.helpers`, which by convention we're importing on all new steps, by lazy loading modules that might not be used.

Changes:

- :honeybee: only lazy dvc and papermill on demand
- :honeybee: lazy load boto3
- :honeybee: lazy load requests and grapher helpers

Benchmark with `time etl data://grapher/fasttrack/ --force --only`:

- 10.2s before the change
- 7.8s with lazy loading (~25% faster)


Marigold (approved):
Nice speedup, let's give it a try!

PR #1235, opened 2023-06-15
Update LGBT rights dataset

paarriagadap:
Update Velasco's dataset for the [LGBT+ rights topic page](https://github.com/owid/owid-issues/issues/1045)


paarriagadap:
Hi @Marigold! It is a minor update that I was about to merge, but these submodules (walden, owid-catalog-py) were included in a different way to the past. Is there a way to get rid of them?

Marigold:
@paarriagadap these submodules were recently moved to `/lib`. Just remove the entire `vendor` folder, run `poetry install` and commit it.

PR #1236, opened 2023-06-16
:bar_chart: move Google mobility trends from covid repo to ETL

Marigold:
Proof of concept for migrating COVID datasets from [covid-19-data](https://github.com/owid/covid-19-data) to ETL.

The dataset was first migrated to get all metadata using
```
ENV=.env.prod backport-migrate --dataset-id 5085 --namespace covid --short-name google_mobility_trends --version latest
```
and then modified to use `etag://...` of the CSV file in covid repo.

ETL is still missing some features implemented in [GrapherBaseUpdater](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/grapher/db/base.py#L10)
1. Slack notifications
2. Incrementing [charts versions](https://github.com/owid/covid-19-data/blob/master/scripts/src/cowidev/grapher/db/utils/db_imports.py#L243) to trigger rebake
3. `enqueue_deploy` that enqueues deploy

I'm not sure how charts work at the moment. I thought that all of them get rebaked irrespective of version, but I might be wrong. As for deploy, we could have a single CRON job instead that would be triggering deploys daily or every few hours.

If this dataset works, I can start migrating other datasets.


lucasrodes:
@Marigold, thanks for doing this.

The code looks good! Some comments/questions:

1. Currently, the data flow looks like `covid-19-data → Grapher → backport → ETL → Grapher`? I guess I am missing something. Is the idea to move the scraping/getting data part from `covid-19-data` to `snapshot`?
2. Side comment, Google stopped updating this dataset during (I believe) October 2022.
3. Slack notifications are especially useful for vaccinations/hospitalizations pipelines. For the remaining ones these are good to have, but it is fine.
4. On the chart version increment, I honestly don't know what the impact of this is. In general, I am not sure if we even make any use of the chart versions. AFAIK the chart will display the updated version irrespective of the version being bumped up or not.
5. `enqueue_deploy`: Are you talking about running something like `etl covid` once a day? I think that'd be a nice idea.

Marigold:
> 1. Currently, the data flow looks like `covid-19-data → Grapher → backport → ETL → Grapher`?

It's actually `covid-19-data -> .csv -> ETL -> Grapher`. The `backport-migrate` is not backporting periodically, I just used it to generate ETL steps (with metadata) that I updated.

> 2. Side comment, Google stopped updating this dataset during (I believe) October 2022.

Yeah, I know. It was just a proof of concept.

> 3. Slack notifications are especially useful for vaccinations/hospitalizations pipelines. For the remaining ones these are good to have, but it is fine.

Would it be enough to send notifications after `.csv` is generated?

> 4. and 5.

Good!

If you think it's viable approach I'll remove the `.csv -> Grapher` part from covid repo and run it from ETL. If it works I can start with other covid datasets.

lucasrodes:
Right, all makes sense!

Re: removing `.csv → Grapher`, it is currently not being executed on the covid repo:

https://github.com/owid/covid-19-data/blob/c5c053b41c65e3661592f5d50821968430ad3ff9/scripts/scripts/grapherupdate.sh#L61

But for the other processes, I think we could start with those listed in:

https://github.com/owid/covid-19-data/blob/c5c053b41c65e3661592f5d50821968430ad3ff9/scripts/scripts/grapherupdate.sh#L56-L65

- cases/deaths
- decoupling
- hospitalisations
- oxcgrt
- uk-nations
- sweden

PR #1245, opened 2023-06-20
:bug: set latest date_accessed for excess mortality

Marigold:
Having fixed `date_accessed: "2023-04-24"` in YAML metadata was preventing updates of excess mortality in the covid repo. This is a hotfix that sets `date_accessed` of all sources to the latest one.

Proper fix would involve propagating sources from snapshots through channels instead of defining them in YAML metadata. It'd be better to wait for metadata updates we're working on though and @pabloarosado way of propagating metadata (this would be a good proof of concept).

### DIFF
--- a/etl/steps/data/garden/excess_mortality/latest/excess_mortality/__init__.py
+++ b/etl/steps/data/garden/excess_mortality/latest/excess_mortality/__init__.py
@@ -54,6 +54,22 @@ def run(dest_dir: str) -> None:
     # Update dataset and table metadata using the adjacent yaml file.
     ds_garden.update_metadata(paths.metadata_path)
 
+    # NOTE: ideally we wouldn't define sources in YAML, but inherit them from snapshots
+    # setting date_accessed is a workaround
+    # Get latest date_accessed from all sources and use it for variables and dataset.
+    max_date_accessed = max(
+        ds_hmd.metadata.sources[0].date_accessed,  # type: ignore
+        ds_wmd.metadata.sources[0].date_accessed,  # type: ignore
+        ds_kobak.metadata.sources[0].date_accessed,  # type: ignore
+    )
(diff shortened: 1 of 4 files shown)


lucasrodes:
I am a bit confused with my code here.

I see that the fields in snapshot and garden metadata differ a bit. For instance, the `publication_date` for the _Human Mortality Database (2023)_:

- `etl/steps/data/garden/excess_mortality/latest/excess_mortality/excess_mortality.meta.yml`: value is `2023-04-21`
- `snapshots/excess_mortality/latest/hmd_stfm.csv.yml`: value is `2023-06-19`

Marigold:
@lucasrodes I was a bit confused with your code too :). I wanted to implement the propagation of sources from snapshots to steps, but **snapshot source names weren't consistent with garden source names** (in snapshot it's called `Karlinsky and Kobak (2021)`, in garden there are suddenly two `World Mortality Dataset (2023)` source names).

Ideally you'd define sources only once in snapshots and they would be propagated. I did it by only defining source names instead of anchors, e.g.
```
sources:
          - name: Human Mortality Database (2023)
          - name: World Mortality Dataset (2023)
```
and then adding detail information about the source from dataset sources. This way it would get propagated correctly.

**Could you please pull this branch and update source names in snapshots to reflect source names in garden YML file?** It doesn't work at the moment, but should after sources are fixed. (I'm a bit hesitant to do it myself)

(We could do this trick automatically in `create_dataset(...)` if we find it useful)

lucasrodes:
Right, sorry, I got confused with the diff! You actually got rid of all that hard coding (lol). Thanks for doing that.

I just changed the source_name fields in snapshot accordingly.

We have one file for `Human Mortality Database (2023)`, another for `World Mortality Dataset (2023)` and two for `Karlinsky and Kobak (2021)`

PR #1247, opened 2023-06-21
📊 Cantril ladder - adding archive of reports to etl 

spoonerf:
Previously we were using a backport of the Cantril ladder data, this changes to source the data directly from the reports and also corrects a data source error along the way. 

* Sourcing the data for 2012-2019 directly from the reports rather than a table from the 2019 report, the values are slightly different and it is more correct to source directly from the reports. 
* Change the year we refer to in the charts. Each report presents data as a combination from the previous 3 years, i.e. the 2023 report uses survey data from 2020-2022 and we were showing this as the mid-point, 2021. Based on feedback from Este we'll now show it as the latest year of survey data, so 2022 in the example above.


snapshots/happiness/2019-03-20/happiness.xls.dvc
```diff
@@ -0,0 +1,25 @@
+meta:
+  namespace: happiness
+  short_name: happiness
+  name: World Happiness Report (2019)
+  version: '2019-03-20'
+  publication_year: 2019
+
```
lucasrodes, line 7: remove spacing?

snapshots/happiness/2021-03-20/happiness.xls.dvc
```diff
@@ -0,0 +1,26 @@
+meta:
+  namespace: happiness
+  short_name: happiness
+  name: World Happiness Report (2021)
+  version: '2021-03-20'
+  publication_year: 2021
+
```
lucasrodes, line 7: remove spacing?

dag/walkthrough.yml
```diff
@@ -7,4 +7,5 @@ steps:
   data://grapher/dummy/2020-01-01/dummy:
   - data://garden/dummy/2020-01-01/dummy
   data://explorers/dummy/2020-01-01/dummy:
-  - data://garden/dummy/2020-01-01/dummy
\ No newline at end of file
+  - data://garden/dummy/2020-01-01/dummy
+
```
lucasrodes, line 11: if you could remove these so no diff is shown.

dag/main.yml
```diff
@@ -344,8 +344,36 @@ steps:
...
+  - data://meadow/happiness/2022-03-20/happiness
+  - data://meadow/happiness/2021-03-20/happiness
+  - data://meadow/happiness/2020-03-20/happiness
+  - data://meadow/happiness/2019-03-20/happiness
+  - data://meadow/happiness/2018-03-20/happiness
+  - data://meadow/happiness/2017-03-20/happiness
+  - data://meadow/happiness/2016-03-20/happiness
+  - data://meadow/happiness/2015-03-20/happiness
+  - data://meadow/happiness/2012-03-20/happiness
   - backport://backport/owid/latest/dataset_235_world_happiness_report__2022
```
lucasrodes, line 377: is the `backport` dataset a dependency still? I don't see any reference to it in the code.

etl/steps/data/garden/happiness/2023-03-20/happiness.meta.yml
```diff
@@ -5,19 +5,35 @@ dataset:
...
+
+    2018: https://s3.amazonaws.com/happiness-report/2018/WHR2018Chapter2OnlineData.xls
+
+    2017: https://s3.amazonaws.com/happiness-report/2017/online-data-chapter-2-whr-2017.xlsx
+
+    2016: https://s3.amazonaws.com/happiness-report/2016/Online-data-for-chapter-2-whr-2016.xlsx
+
+    2015: https://s3.amazonaws.com/happiness-report/2015/Chapter2OnlineData_Expanded-with-Trust-and-Governance.xlsx
+
+    2012: https://happiness-report.s3.amazonaws.com/2012/2012.xlsx
```
lucasrodes, line 36: do we need to list the links to the individual datasets in here?

We could instead propagate them from the different dataset's metadata and have them in the `sources` field, right?

etl/steps/data/garden/happiness/2023-03-20/happiness.meta.yml
```diff
@@ -5,19 +5,35 @@ dataset:
...
   sources:
   - name: World Happiness Report (2023)
   - published_by: 'Helliwell, J. F., Layard, R., Sachs, J. D., De Neve, J.-E., Aknin, L. B., & Wang, S. (Eds.). (2023). World Happiness Report 2023. New York: Sustainable Development Solutions Network.'
```
lucasrodes, line 39: slightly related to the previous comment ー aren't all happiness report versions sources (and not only the 2023 version)?

etl/steps/data/garden/happiness/2023-03-20/happiness.meta.yml
```diff
@@ -28,20 +44,10 @@ tables:
...
-      upperwhisker:
-        title: 97.5% confidence interval
-        short_unit: ''
-        unit: ''
-      lowerwhisker:
-        title: 2.5% confidence interval
-        short_unit: ''
-        unit: ''
+        display:
+          numDecimalPlaces: 2
```
lucasrodes, line 53: Is there a particular reason to use `numDecimalPlaces=2`? The general rule is to use `numDecimalPlaces=1`, unless more precision is needed (e.g. possible values range from 0 to 1).

lucasrodes (approved):
LGTM, just minor comments on metadata.

etl/steps/data/garden/happiness/2023-03-20/happiness.meta.yml
spoonerf: True - is there a good example of how to do this?

PR #1253, opened 2023-06-22
:honeybee: upgrade notebook to version 7

Marigold:
This might help with some performance issues and kernel crashes.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -171,25 +171,26 @@ frozenlist = ">=1.1.0"
 
 [[package]]
 name = "altair"
-version = "4.2.2"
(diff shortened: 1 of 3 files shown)


pabloarosado:
I'm trying to make jupytext work with the new notebook version and I'm not succeeding, which is what I was fearing. If that doesn't work, I'd prefer to either cancel this PR. Alternatively, given that jupytext doesn't integrate well with either jupyterlab or the latest notebook, maybe it's time for me to find an alternative workflow...

pabloarosado (approved):
It seems to be working well with jupytext and plotly. We can give it a go and, if we find issues, we can revert back to notebook 6.

PR #1256, opened 2023-06-23
🚧 Experiment: add metadata to WDI to create data page

danyx23:
This PR fills the new metadata structure to get a feel for what a fully filled example that renders a very rich data page woudl look like. The goal would be for all of this information to land in the variable metadata.json and the database so that we can preview/bake datapages directly.


Marigold:
Rebased on top of master.

PR #1263, opened 2023-06-26
📊 World Values Survey: terrorism questions

veronikasamborska1994:


etl/steps/data/garden/wvs/2023-06-25/longitudinal_wvs.meta.yml
```diff
@@ -0,0 +1,45 @@
+dataset:
+  title: World Values Survey time-series data (1981-2022)
+  description: &desc |
+    The WVS comprises survey data conducted across 64 countries and societies around the world and more than 80,000 respondents.
+    This dataset includes survey responses comprises integrated survey responses for the 1981-2022 period to the following questions:
+    'Worries: a terrorist attack' corresponding to column 'H006_04'
+    'Effects of immigrants on the development of [your country]: Increase the risks of terrorism' corresponding to column 'G057'
+    'Justifiable: Terrorism as a political, ideological or religious mean' corresponding to column 'F114E'.
+
+  licenses:
```
Marigold, line 10: @veronikasamborska1994 metadata about licenses and sources gets propagated from the snapshot metadata file. It's not necessary to repeat them here.

(It's not clear because of our bad design. Snapshot metadata actually contains `source` and `license`, but it's not clear because their fields are un-nested as `source_name`, `source_published_by`, etc. This should be fixed with the new metadata.)

PR #1268, opened 2023-06-26
:lipstick: Give catalog objects rich html versions

larsyencken:
We work in notebooks often, but today metadata is horrific to view in notebooks.

This PR adds rich HTML display of metadata to make it easier to read, and also adds a title to make it easier to distinguish a `Table` from a `DataFrame` and a `Variable` from a `Series`.

Before and after comparison: https://gist.github.com/larsyencken/5c9aedaaacba4f593d1b18d797cc2c00

It uses a completely generic approach that doesn't know anything about the metadata fields, so it will work just as well when new fields become available.

### DIFF
--- a/lib/catalog/owid/catalog/meta.py
+++ b/lib/catalog/owid/catalog/meta.py
@@ -77,6 +77,9 @@ class License:
     def from_dict(d: Dict[str, Any]) -> "License":
         ...
 
+    def __bool__(self):
+        return bool(self.name or self.url)
+
 
 @pruned_json
 @dataclass_json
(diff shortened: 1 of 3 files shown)


Marigold (approved):
Beautiful! Perhaps [rich](https://github.com/Textualize/rich) could have methods that could be useful too? It'd be also awesome to have nushell-style console output.

larsyencken:
> Beautiful! Perhaps [rich](https://github.com/Textualize/rich) could have methods that could be useful too? It'd be also awesome to have nushell-style console output.

We could possibly do this for ipython and ipydb, if people are using them. IPython has the `_ipython_display_()` hook for pretty display, and you can even customise the tab completions https://ipython.readthedocs.io/en/stable/config/integrating.html

PR #1275, opened 2023-06-27
:hammer: Support strict primary key enforcement

larsyencken:
This change causes a warning to happen any time you save a table without a primary key. It also allows a strict mode in which the warning becomes an exception.

## Motivation

We are trying to move more metadata to the indicator level. The dimensions an indicator has is a really important piece of metadata that is make or break for automatically collecting, indexing and reusing the indicators.

## How it will be used

The Buildkite tasks that build production data and do the nightly full build will not be run in strict mode. However, PRs will be run in strict mode. This is to avoid any downtime for the ETL, whilst strongly encouraging future changes to have a primary key.

### DIFF
--- a/lib/catalog/owid/catalog/datasets.py
+++ b/lib/catalog/owid/catalog/datasets.py
@@ -8,7 +8,7 @@ import shutil
 import warnings
 from dataclasses import dataclass
 from glob import glob
-from os import mkdir
+from os import environ, mkdir
 from os.path import join
 from pathlib import Path
 from typing import Any, Iterator, List, Literal, Optional, Union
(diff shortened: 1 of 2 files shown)


larsyencken:
For more background of where I think we should go with dimensions, have a read here: https://www.notion.so/owid/2023-06-27-Proposal-for-dimensions-in-the-ETL-9e1a26fec3b94ad2a33ca8fab14b090a?pvs=4

pabloarosado (approved):
It's a good idea. The only annoying thing is that, in practice, we are always setting index before saving, and then resetting index after loading. In an ideal world working with multi-index tables would be easy. But overall I think it's a safe approach.

larsyencken:
> It's a good idea. The only annoying thing is that, in practice, we are always setting index before saving, and then resetting index after loading. In an ideal world working with multi-index tables would be easy. But overall I think it's a safe approach.

@pabloarosado Have a read of the proposal I wrote in Notion as well. I think we should move to something simpler, which is just using `dim_` in front of dimension columns. E.g. `dim_country`, `dim_year`. It's a common convention that's also pretty self-explanatory. We can could even support new and old ways of doing this in a backwards-compatible way.

PR #1276, opened 2023-06-27
:honeybee: test for invalid table metadata assignment

larsyencken:
A `Table` can have metadata, and its `TableMeta` class includes a `dataset` field, but that dataset field is only a cached copy that is overwritten at the moment you add to a dataset.

This PR adds a simple code scan to check for any cases where we manually set it in a step. In all of those cases it's invalid, since it will be overwritten at save time.


Marigold:
Force pushed rebase on master because we incremented ETL_EPOCH recently.

Marigold:
Rebased & fixed all cases where we assign to `table.metadata.dataset`. Also deleted a couple of old versions of CO2 datasets.

pabloarosado (approved):
Looks good.

PR #1279, opened 2023-06-28
:hammer: raise strict mode error for non-unique index

larsyencken:
Usually when an index is not unique, you haven't properly specified the dimensions of the data or done roll-ups that you need to.

We would enforce this for new PRs only.

### DIFF
--- a/lib/catalog/owid/catalog/datasets.py
+++ b/lib/catalog/owid/catalog/datasets.py
@@ -110,9 +110,18 @@ class Dataset:
 
         if not table.primary_key:
             if "OWID_STRICT" in environ:
-                raise PrimaryKeyMissing(f"Table `{table.metadata.short_name}` does not have a primary_key set")
+                raise PrimaryKeyMissing(
+                    f"Table `{table.metadata.short_name}` does not have a primary_key set -- please use set_index() to indicate dimensions"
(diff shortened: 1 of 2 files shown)


larsyencken:
Here is an example of changes needed to comply with this strict mode: https://github.com/owid/etl/pull/1282

pabloarosado (commented):
A priori I like this idea. But in practice, I know that there are datasets (e.g. some FAOSTAT domains in meadow) that have non-unique indexes, because of some issues that are later on handled in garden. So I'm wondering if we should impose this uniqueness only at the garden level (and I'm wondering if we should only raise a warning instead of an error if there are non-unique indexes).

Marigold:
There's [already a warning](https://github.com/owid/etl/blob/5af3c72047b6200171af2355717fa9048478a3d8/lib/catalog/owid/catalog/datasets.py#L103) if an index is non-unique. I wanted to make it compulsory too, but then I decided it wasn't worth refactoring all steps (regions in particular).

larsyencken:
> only raise a warning instead of an error

I totally agree that this might be too strict, or not ideal. On this part, Pandas already warns every time, and we have tons of these warnings every day, which is what triggered this.

PR #1282, opened 2023-06-28
:bar_chart: rework regions dataset to have unique indexes

larsyencken:
The regions dataset contains many join tables with one-to-many relationships. In our current strict dimensions model, we have two options:

- Give them an integer index, or
- Put every column in the index

This change does the latter. It's also meant to trigger discussion about what we want in general here.


pabloarosado:
I think that having a table like `aliases`, with both `code` and `alias` as indexes is not very useful.
More generally, I think so far there has been no benefit of having many separate tables. Should we instead consider keeping only the `regions` table (which has a well-defined index)?

pabloarosado:
Hey @larsyencken I've removed all tables except for the `regions` table in the garden step, and adapted the grapher step accordingly.
There are still a few things that need to be refactored:
* [ ] `etl.harmonize`
* [ ] `etl.data_helpers.geo`
* [ ] `owid_co2` garden step
* [ ] `etl.fasttrack`

larsyencken:
Thanks Pablo!

pabloarosado:
Hi @larsyencken I think this PR is ready to be merged. I've removed all tables from the regions dataset except the combined `regions` table (which has a unique index and all data we need about regions), and I've removed all uses of those tables. I'll wait for @Marigold to check it out before merging (hopefully it doesn't clash with any of the other metadata refactors).

larsyencken:
Good work!!!

Marigold (approved):
Thanks, it looks good. However, changing regions triggers a huge ETL rebuild, so please merge it at the end of your workday to avoid disrupting others' work.

PR #1290, opened 2023-06-29
:hammer: remove publisher_source field and move it to description

Marigold:
`Source.publisher_source` field is not used very frequently and it's not clear what should be there. In all cases `Source.published_by` and `Source.description` can do the job.

Getting rid of publisher_source was really easy - either it duplicated some other source field or could be easily moved to the description (sometimes adding "Data comes from ..."). I had to also change the converter from MySQL source to `catalog.Source` by appending it to description (this is only used for backports and their sources are never used as far as I know).


pabloarosado (approved):
Looks good. Just a minor comment is that descriptions sometimes end in a period and sometimes they don't. Ideally they should end in ".".

Marigold:
> Looks good. Just a minor comment is that descriptions sometimes end in a period and sometimes they don't. Ideally they should end in ".".

We could add a linter warning for YAML files in VSCode (i.e. that description should match certain regex).

PR #1291, opened 2023-06-29
:hammer: Replace snapshot fields by source and license

Marigold:
Fields in `SnapshotMeta` like `source_published_by` are just flattened Source and License. Working with Sources and Licenses directly makes further changes to metadata easier.

I've been initially thinking about editing all snapshot YAML files and adding `source:` and `license:` entries there, but that'd be probably too much. To avoid that I'm converting those YAML fields to Source object on load.

### DIFF
--- a/backport/backport.py
+++ b/backport/backport.py
@@ -4,6 +4,7 @@ from typing import List, Optional
 import click
 import pandas as pd
 import structlog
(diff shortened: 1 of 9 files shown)


pabloarosado (approved):
Looks good.

PR #1292, opened 2023-06-29
:hammer: enforce strict mode for new datasets

larsyencken:
Enable strict mode by default for all `garden` and `grapher` datasets, where `version` is `latest` or after `2023-06-25`.

Strict mode for now means the following:

- Every table must have a primary key
- Every primary key must uniquely identify a row

If either of these criteria fail then the step will refuse to build.

When building the ETL manually, you can override the default setting by running `etl --strict` or `etl --no-strict`.

## Todo

- [x] Enable strict mode checks when the criteria matches
- [x] Ensure current ETL passes all these checks

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -272,7 +272,8 @@ def _detect_strictness_level(step: Step, strict: Optional[bool] = None) -> bool:
         return False
 
     # now it depends on the version
-    return step.version == "latest" or step.version >= config.STRICT_AFTER
+    # TODO fix the "latest" cases as well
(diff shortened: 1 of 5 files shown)


Marigold (approved):
Looks good!

larsyencken:
It's breaking for some datasets, so I'm just weighing up whether to fix them or skip ones with version `latest` for now.

larsyencken:
Kicked off a full rebuild, if that passes I'll merge: https://buildkite.com/our-world-in-data/etl-full-private-rebuild-weekly/builds/313

PR #1293, opened 2023-06-29
:hammer: remove logging from walkthrough

Marigold:
I feel like no one's using it (and those wanting to use it know how to import it), so there's no point in having it in walkthrough. We print time taken for every step and when it started, so these logs are kinda useless.


pabloarosado (approved):
Yes, I'd also prefer to avoid redundant log entries.

PR #1300, opened 2023-06-29
fix: decimal places for monetary values on WID

paarriagadap:
Changed the decimal points for monetary values from 1 to 2


Marigold:
@paarriagadap you can push such minor changes straight to master, no need to create PR.

PR #1315, opened 2023-07-05
📊 Update statistical review of world energy dataset

pabloarosado:
# Current situation
The Energy Institute Statistical Review of World Energy (previously owned by BP) has been recently updated. We used to import this dataset via the old `importers` repos. Then, the dataset was backported into ETL.

We have three versions of this dataset in live grapher:
* [And old version of the dataset, uploaded via importers](https://owid.cloud/admin/datasets/5347). 
* [The last BP version of the dataset, uploaded via importers](https://owid.cloud/admin/datasets/5650).
* [An ETL version of the dataset, which combines the previous two](https://owid.cloud/admin/datasets/5840).
The reason for this is that variables on installed capacity were missing in the last BP release. However, now I have realised that this data actually comes from IRENA.

# Situation after merging
Main improvements:
* Importers code can be deleted.
* I have created the new steps to fetch the capacity data directly from its original producer (IRENA). This way we don't need to keep old versions of the statistical review (and instead keep just the latest version).
* I have fixed various issues with the data and created variables with more convenient units. For example, variables are more explicit to show whether they correspond to direct primary energy and to input-equivalents.
* Metadata is now much more rich, although I removed the long description about region definitions (which I think is not very useful).

# For reviewers
@lucasrodes could you please review the changes in the `geo` and `lib/catalog` modules?
* [x] `etl/data_helpers/geo.py`
* [ ] `lib/catalog/owid/catalog/processing.py`
* [ ] `lib/catalog/owid/catalog/tables.py`

@spoonerf could you please review the changes related to the Energy Institute steps?
* [x] `dag/energy.yml`
* [x] `etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.countries.json`
* [x] `etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.excluded_countries.json`
* [x] `etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml`
* [x] `etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.py`
* [x] `etl/steps/data/grapher/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml`
* [x] `etl/steps/data/grapher/energy_institute/2023-06-26/statistical_review_of_world_energy.py`
* [x] `etl/steps/data/meadow/energy_institute/2023-06-26/statistical_review_of_world_energy.py`
* [x] `snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.csv.dvc`
* [x] `snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.py`
* [x] `snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.xlsx.dvc`

@veronikasamborska1994  could you please review the steps related to IRENA steps?
* [x] `etl/steps/data/garden/irena/2023-06-26/renewable_electricity_capacity.countries.json`
* [x] `etl/steps/data/garden/irena/2023-06-26/renewable_electricity_capacity.meta.yml`
* [x] `etl/steps/data/garden/irena/2023-06-26/renewable_electricity_capacity.py`
* [x] `etl/steps/data/grapher/irena/2023-06-26/renewable_electricity_capacity.py`
* [x] `etl/steps/data/grapher/irena/2023-06-26/renewable_electricity_capacity_by_technology.py`
* [x] `etl/steps/data/meadow/irena/2023-06-26/renewable_electricity_capacity.py`
* [x] `snapshots/irena/2023-06-26/renewable_electricity_capacity_and_generation.py`
* [x] `snapshots/irena/2023-06-26/renewable_electricity_capacity_and_generation.xlsm.dvc`

# TO-DO after merging:
* After merging this PR, grapher variables have to be carefully mapped (given that some variables have changed). No conversion factors are needed anymore, and some titles need to be adapted.
* Then, another PR will archive unnecessary ETL steps, and the old live grapher datasets will be archived.
* Another PR needs to fix a minor issue with `lib/catalog`, that raises unnecessary warnings when melting a table.
* I kept the old variable titles to make the variable mapping easier. But after merging, consider improving them (for example they are not well capitalized or very homogeneous).
* Adapt and update the rest of steps of the energy dataset, and update the `energy-data` repos.


etl/data_helpers/geo.py
```diff
@@ -592,12 +605,21 @@ def list_members_of_region(
...
+            # TODO: Remove this block once the old income groups dataset has been archived.
+            # Get the main table from the income groups dataset.
+            df_income = (
+                pd.DataFrame(ds_income_groups["wb_income_group"])
+                .reset_index()
+                .rename(columns={"income_group": "classification"})
+            )
+        else:
+            # Get the table with the current definitions of income groups.
+            df_income = ds_income_groups["income_groups_latest"].reset_index()
```
lucasrodes, line 618: should probably check that `income_groups_latest in ds_income_groups.table_names` and then have an `else` clause raising an error saying something like "table was not found"

etl/data_helpers/geo.py
```diff
@@ -73,6 +82,10 @@ def list_countries_in_region(
 ) -> List[str]:
     """List countries that are members of a region.
 
+    ####################################################################################################################
+    WARNING: This function is deprecated, use list_members_of_region instead.
+    ####################################################################################################################
+
```
lucasrodes, line 88: I see that in the functions above, they have a similar section which is not included in the docstring, but here it is. We should probably include this in the docstring always are never (but not mix).

lucasrodes: Could use `warning.warn` to raise a `DeprecationWarning` message. That could be more useful.

```python
from warning import warn

def some_function():
    ...
    warn('This method is deprecated.', DeprecationWarning, stacklevel=2)
    ...
```

lib/catalog/owid/catalog/tables.py
```diff
@@ -963,21 +962,48 @@ def _add_table_and_variables_metadata_to_table(table: Table, metadata: Optional[
 
 
 def read_csv(
-    filepath_or_buffer: Union[FilePath, ReadCsvBuffer[bytes], ReadCsvBuffer[str]],
+    path: Union[str, Path],
```
lucasrodes, line 965: What is the motivation for this argument name change? I would stick to the old name, which is the one that `pandas` uses.

lucasrodes (approved):
lgtm, some suggestions in my comments.

etl/data_helpers/geo.py
pabloarosado: Hey Lucas, I don't understand your first comment. The warnings in the `_load*` functions are not really necessary, since they are already internal functions, but I added so that we remember to not use them when creating functions in `geo` (and to signal that eventually we should remove them).
Having a deprecation warning implies that we would get many warnings when running old ETL steps.

lib/catalog/owid/catalog/tables.py
pabloarosado: I just thought that `filepath_or_buffer` is a long, hard to understand argument, while `path` is much clearer (and we will pass a path in the vast majority of cases). Function `read_excel` has as first argument `io` which is also ambiguous. That's why I used `path` on both. Also, we are already introducing `metadata` and `underscore`, so the list of arguments is already different to `pandas`. Do you see any harm in changing these argument names?

etl/steps/data/garden/irena/2023-06-26/renewable_electricity_capacity.meta.yml
```diff
@@ -0,0 +1,117 @@
+tables:
+  renewable_electricity_capacity:
+    variables:
+      bagasse:
+        title: Bagasse capacity
```
veronikasamborska1994, line 5: maybe worth adding descriptions for these? 

> Bagasse is a byproduct from the process of sugar cane refining. After the juice has been extracted from the sugar cane, what's left over is a fibrous material called bagasse. It is often used as a biofuel for the production of heat, energy, and electricity, and in the manufacture of pulp and building materials.

veronikasamborska1994:
looks good!

etl/steps/data/garden/irena/2023-06-26/renewable_electricity_capacity.meta.yml
pabloarosado: Thanks, good idea. Where did you get this definition from?

veronikasamborska1994: GPT-4 :) you'd of course need to check it's accurate but it's usually pretty good at definitions. The code looks good otherwise, better than anything I ever write :)

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.csv.dvc
```diff
@@ -0,0 +1,23 @@
...
+  publication_year: 2023
+
+  publication_date: '2023-06-26'
+  source_name: Energy Institute Statistical Review of World Energy 2023
+  source_published_by: Energy Institute Statistical Review of World Energy 2023
+  url: https://www.energyinst.org/statistical-review/
+  source_data_url: 
+    https://www.energyinst.org/__data/assets/file/0007/1055761/Consolidated-Dataset-Panel-format-CSV.csv
+  license_url: 
+    https://www.energyinst.org/__data/assets/pdf_file/0004/1055542/EI_Stat_Review_PDF_single-2.pdf
```
spoonerf, line 12: This link to the license isn't working for me, getting a 404.

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.xlsx.dvc
```diff
@@ -0,0 +1,23 @@
...
+  publication_year: 2023
+
+  publication_date: '2023-06-26'
+  source_name: Energy Institute Statistical Review of World Energy 2023
+  source_published_by: Energy Institute Statistical Review of World Energy 2023
+  url: https://www.energyinst.org/statistical-review/
+  source_data_url: 
+    https://www.energyinst.org/__data/assets/excel_doc/0007/1055545/EI-stats-review-all-data.xlsx
+  license_url: 
+    https://www.energyinst.org/__data/assets/pdf_file/0004/1055542/EI_Stat_Review_PDF_single-2.pdf
```
spoonerf, line 12: Same issue here, the link to the license isn't working for me, getting a 404.

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.csv.dvc
spoonerf: Ah it looks like they may have already changed the link to: https://www.energyinst.org/__data/assets/pdf_file/0004/1055542/EI_Stat_Review_PDF_single_3.pdf

Perhaps a good idea to use a web archive link in case they do this again.

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.csv.dvc
```diff
@@ -0,0 +1,23 @@
+meta:
+  name: Energy Institute Statistical Review of World Energy 2023
+  publication_year: 2023
+
```
spoonerf, line 4: I don't think it matters at all but the line break here seems unusual (I assume it is automatically added?)

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.csv.dvc
```diff
@@ -0,0 +1,23 @@
...
+  publication_date: '2023-06-26'
+  source_name: Energy Institute Statistical Review of World Energy 2023
```
spoonerf, line 6: I think the year should be in brackets to match our usual format for the bottom left source.

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.xlsx.dvc
```diff
@@ -0,0 +1,23 @@
+meta:
+  name: Energy Institute Statistical Review of World Energy 2023
+  publication_year: 2023
+
+  publication_date: '2023-06-26'
+  source_name: Energy Institute Statistical Review of World Energy 2023
```
spoonerf, line 6: Likewise, I think the year should be in brackets here.

etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml
```diff
@@ -0,0 +1,709 @@
...
+        description: |
+          Thermal equivalent efficiency factors used to convert non-fossil electricity in input-equivalent to direct primary energy.
+
+  statistical_review_of_world_energy_fossil_fuel_prices:
+    variables:
+      # Coal prices.
+      coal_price_us_central_appalachian_spot_price_index_current_dollars_per_tonne:
+        title: Coal price - US Central Appalachian coal spot price index
+        unit: 'US dollars per tonne'
+        short_unit: 'US$'
```
spoonerf, line 493: I think the short unit should just be '$'?

etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml
```diff
@@ -0,0 +1,709 @@
+tables:
+  statistical_review_of_world_energy:
```
spoonerf, line 2: I'm curious about how you produce this metadata? Is it copied over from the BP statistical energy metadata?

etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.excluded_countries.json
```diff
@@ -0,0 +1,5 @@
+[
+    "Canadian Oil Sands: Total",
```
spoonerf, line 2: Do the values for Canada and Venezuela not need to be incorporated into regional totals? Just wondering as you keep in "Other *" regions as they are needed to be counted in the regional totals?

etl/steps/data/grapher/energy_institute/2023-06-26/statistical_review_of_world_energy.py
```diff
@@ -0,0 +1,113 @@
...
+        "nuclear_consumption_equivalent_twh",
+        "oil_consumption_twh",
+        "solar_consumption_equivalent_twh",
+        "wind_consumption_equivalent_twh",
+    ]:
+        new_column = f"{column}_zero_filled"
+        tb[new_column] = tb[column].fillna(0)
+        # Note: The following line may not be necessary once "fillna" properly propagates metadata.
+        tb[new_column].metadata = copy.deepcopy(tb[column].metadata)
+        tb[new_column].metadata.title += " (zero filled)"
```
spoonerf, line 74: I think creating zero-filled indicators for stacked area columns is no longer needed after a recent update - https://owid.slack.com/archives/C014SGT6W9X/p1684500482832199?thread_ts=1684500411.887979&cid=C014SGT6W9X ?

spoonerf (approved):
Looks good to me - have added some suggestions and queries, all fairly minor!

snapshots/energy_institute/2023-06-26/statistical_review_of_world_energy.csv.dvc
pabloarosado: Thanks for pointing this out. That's a bit annoying, but the only place I've found where they mention citation and copyrights is the PDF. I've just changed it to the link from the web archive.

etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml
pabloarosado: It's in my to-do list to check the criteria on FASTT for these units, so I'll change it (and possibly variable names and chart titles) later.

etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml
pabloarosado: I took most of it from what we had in importers, I don't know if it was Bobbie who prepared it. Most of the texts seem to come from footnotes in the sheets of the main excel data file (hard to parse).

etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.excluded_countries.json
pabloarosado: Good catch. I already checked and, according to a footnote in the corresponding sheet of the excel data file:

> Reserves and R/P ratio for Canada includes Canadian oil sands. Reserves and R/P ratio for Venezuela includes the Orinoco Belt.  Saudi Arabia's oil reserves include NGLs from 2017.

etl/steps/data/grapher/energy_institute/2023-06-26/statistical_review_of_world_energy.py
pabloarosado: Ah I think recently I tried and it didn't work... I may try again, but I'd rather do it on a separate PR. Keeping it this way will make my life easier for mapping grapher variables, but I'll look into this afterwards, thanks.

etl/data_helpers/geo.py
lucasrodes: > Hey Lucas, I don't understand your first comment

In lines 67-69 (and others, e.g.58-60), the comment `WARNING: ...` is not included in the docstring. However, this one (85-87) is in the docstring.

To be consistent, I suggested having that comment either within or outside the docstring. However, I see that internal functions do not have a docstring; hence it is fine as it is.

> Having a deprecation warning implies that we would get many warnings when running old ETL steps.

I think that it is totally fine and actually positive to remind us that we are using deprecated functions. Also, we'd get such warnings if we use them accidentally now. So overall, it is a good strategy to prevent us from using these deprecated functions.

lib/catalog/owid/catalog/tables.py
lucasrodes: In general, I'd minimise argument names whenever they are precisely equivalent. It is easier to get documentation on them (directly from pandas). Also, maintenance might be easier (as in if pandas were to introduce some API changes, argument descriptions, etc.).

I would not change `filepath_or_buffer` to `path`.

If it were the case that we _only_ accept paths (and not buffers), I'd make that explicit in the docs and adapt the code accordingly to raise an error if a buffer is passed.


Adding new ones, I think, is fine. And users should only use our docs for that.

etl/data_helpers/geo.py
pabloarosado: For now I'd suggest to keep this in mind when reviewing PRs, and not add more warnings to the already long list of ETL unnecessary warnings. There's no rush to deprecate those functions given that we don't have good a good alternative for creating region aggregates (which is the most common goal when using these functions).

lib/catalog/owid/catalog/tables.py
pabloarosado: OK, thanks, I'll keep this in mind on the next refactor when handling metadata (coming soon). The first mandatory argument of these functions is usually not explicitly called, so the name is not very important. Also, keep in mind that:
* We want to eventually create our own functions, instead of inheriting pandas behaviour. So it's likely that we'll end up changing names.
* I found it a bit risky to import obscure pandas types like `ReadCsvBuffer` from pandas hidden modules (see line `from pandas._typing import FilePath, ReadCsvBuffer  # type: ignore`). I think it would be safer to define those types ourselves.

lucasrodes: I still think that, even if we are re-defining names and customising the functions, we should try to avoid distancing ourselves too much from pandas whenever possible, as most people are used to it.

If we are only accepting paths, we can indeed change the name and have a check like `if not isinstance(path, str): raise TypeError("...")`.

etl/data_helpers/geo.py
lucasrodes: Alright, maybe it is personal. I generally don't mind having error messages and actually appreciate them.

Without them, it becomes hard for me to track changes/recommendations like these. So not sure I'll "keep this in mind when reviewing PRs", will do my best to remember, though!

pabloarosado: Feel free to create a separate PR and raise a warning when these functions are used. I don't have a strong opinion on this.

lib/catalog/owid/catalog/tables.py
pabloarosado: I just created https://github.com/owid/etl/pull/1327

PR #1317, opened 2023-07-05
📊 Updating the UNODC homicide dataset based on external feedback of homicide page

spoonerf:
Pretty much just a direct update from the previous version, so code has previously passed review. The only changes are to snapshot dvc.


pabloarosado (approved):
Looks good, I made some minor suggestions, feel free to ignore them.

Next time you do an update of a step without big changes, you could first create a git branch where you simply duplicate the old step (and change its version). Then create a sub-branch from there, and make all changes. Then you create a PR where you aim to merge the sub-branch on the branch. That way the reviewer can see what has changed with respect to the previous version (and once reviewed you can merge the sub-branch to the branch, and the branch to master).

spoonerf:
> Next time you do an update of a step without big changes, you could first create a git branch where you simply duplicate the old step (and change its version).

That's a very good point, I will do that next time!

PR #1324, opened 2023-07-07
Minor improvements on statistical review

pabloarosado:
Improve variable titles and units in the Statistical Review dataset, and remove zero_filled variables.

@paarriagadap would you mind checking the FASTT of these charts on fossil fuel prices?
* https://owid.cloud/admin/charts/851/edit
* https://owid.cloud/admin/charts/1369/edit
* https://owid.cloud/admin/charts/653/edit
* https://owid.cloud/admin/charts/1370/edit
* https://owid.cloud/admin/charts/652/edit
* https://owid.cloud/admin/charts/1368/edit

Thanks!

### DIFF
--- a/etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml
+++ b/etl/steps/data/garden/energy_institute/2023-06-26/statistical_review_of_world_energy.meta.yml
@@ -3,478 +3,474 @@ tables:
     variables:
       # biodiesel_consumption_kboed:
       #   title: 'Biofuels Consumption - Kboed - Biodiesel'
-      #   unit: 'Thousand barrels of oil equivalent per day'
-      #   short_unit: 'KBOED'
+      #   unit: 'thousand barrels of oil equivalent per day'
+      #   short_unit: 'kBOED'
       biodiesel_consumption_pj:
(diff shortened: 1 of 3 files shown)


paarriagadap:
@pabloarosado Looks good; I think that the text is fine, but if you want to follow the FASTT rules as they're written it should be more like the points below. Feel free to disregard it if it breaks the style of many other charts.

* https://owid.cloud/admin/charts/851/edit I changed to specify that only prices are adjusted for inflation
* https://owid.cloud/admin/charts/1369/edit Here I would change it to _This data is expressed in US dollars per tonne. It is not adjusted for inflation._ But I am not quite sure because I don't know how some indexes or market prices work, maybe there are differences between countries involved. By the way, is it Asian marker or market?
* https://owid.cloud/admin/charts/653/edit OK
* https://owid.cloud/admin/charts/1370/edit Similar to 1369
* https://owid.cloud/admin/charts/652/edit _This data is expressed in US dollars per cubic meter. It is not adjusted for inflation._
* https://owid.cloud/admin/charts/1368/edit Similar to 1369

pabloarosado:
Thanks @paarriagadap I think, for consistency with similar charts, the subtitles are good this way (maybe at some point we decide to change them all consistently).
And yes, it is Asian marker price.

PR #1327, opened 2023-07-07
Minor improvements on table functions

pabloarosado:
Following the discussion on https://github.com/owid/etl/pull/1315, I've reverted the arguments of our new `read_csv` and `read_excel` functions to coincide with those in pandas. To avoid importing hidden pandas modules, I had to redefine some types.
I also fixed a test that was failing when finding hidden jupyter notebook files.


lucasrodes (approved):
Thanks! LGTM

PR #1329, opened 2023-07-10
Improve the way we combine units when propagating metadata

pabloarosado:
Improve way we combine units when propagating metadata, to avoid spurious warnings.

This change affects for example the step `data://meadow/energy_institute/2023-06-26/statistical_review_of_world_energy`. Before this PR, it would raise many warnings on a melt operation, and with this PR it doesn't.

 NOTE: I increased ETL_EPOCH to see if datadiff shows any relevant difference with this change.


Marigold (approved):
Looks good! Increasing ETL_EPOCH to get datadiff should work in theory, but I'm not sure how is it going to work in practice. I worry you might run into memory error or get too much output from datadiff. A better way might be running this on your staging server and with datadiff on random datasets (you don't have to do this for this PR, but perhaps for future changes).

PR #1333, opened 2023-07-11
:sparkles: parallel upserts to grapher on table level

Marigold:
Grapher step for `un_sdg` takes almost two hours now. That's because we use thousands of tables and grapher upsert only runs in parallel within table. This PR moves the parallelism up so that multiple tables can be processed in parallel. This makes `un_sdg` run about 10x faster.


pabloarosado (approved):
Nice!

PR #1337, opened 2023-07-13
:hammer: remove (WB) suffix from World Bank regions

Marigold:
Even though adding "(WB)" suffix is technically correct, we have too many charts that already use those regions without the suffix. For now it's easier to drop it. We may reconsider adding it back in the future.

We're not consistent with the suffix. If you search for "(WB)" in the repo, you'll find that we're converting it back and forth.


pabloarosado (approved):
For now, you can go ahead and merge this, but it would be good to check how many charts are using "North America" instead of "North America (WB)", because they actually mean [different things](https://ourworldindata.org/world-region-map-definitions). If you have noticed a few places where we are doing this, please create an issue, so at least we are aware of this inaccuracy, thanks.

Marigold:
Keeping (WB) suffix.

PR #1339, opened 2023-07-13
:sparkles: case sensitive matching of entities

Marigold:
Case from ETL might not be respected since we match on the lower cased name. This PR makes the matching case-sensitive.

Merging this and manually running `etl energy_institute/2023-06-26/statistical_review_of_world_energy --grapher` should fix the [following chart](https://owid.cloud/admin/charts/1369/edit).

### DIFF
--- a/etl/db_utils.py
+++ b/etl/db_utils.py
@@ -15,7 +15,7 @@ UPDATE = 2
 
 
 def normalize_entity_name(entity_name: str) -> str:
-    return unidecode(entity_name.lower().strip())
+    return unidecode(entity_name.strip())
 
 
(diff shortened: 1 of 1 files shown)


pabloarosado (approved):
Thanks for fixing this. Do you think there was any particular reason for the lowering?

Marigold:
> Do you think there was any particular reason for the lowering?

I don't know for sure, but I'd say it was meant to be some kind for "harmonization" if you get your cases wrong. We don't need anything like that anymore.

Marigold:
Verified that the chart is showing correct values now (had to re-add them manually).

PR #1342, opened 2023-07-14
🐛 ucdp: data fixes

lucasrodes:
Fixing bugs and issues in the UCDP dataset raised by Bastian:

- Improve metadata
- Add zeroes (instead of NaNs) where applicable
- Correct a bug that was miss-calculating the number of new conflicts by type.


spoonerf (approved):
LGTM to me - checked a bunch of different aggregations and they all make sense to me.

PR #1343, opened 2023-07-14
Remove walden code

pabloarosado:
If our goal is to eventually get rid of `lib/walden` then maybe it's good to start little by little removing all code that is not needed anymore, until we don't have any more active steps depending on walden.
I checked and I think all these files can be safely removed (**after https://github.com/owid/etl/pull/1341 is merged**).
Eventually we would need to also remove the code of archive ETL steps that rely on walden.

NOTE: I think CI will fail until https://github.com/owid/etl/pull/1341 is merged.


Marigold:
Nice! I wanted to get rid of walden and even [got a PR](https://github.com/owid/etl/pull/685), but then decided not to continue with it as I didn't see having walden longer with us as a problem. Perhaps now is the right time to kill it? It'd be definitely easier now that most of our steps use Snapshots.

Marigold (approved):

pabloarosado:
> Nice! I wanted to get rid of walden and even [got a PR](https://github.com/owid/etl/pull/685), but then decided not to continue with it as I didn't see having walden longer with us as a problem. Perhaps now is the right time to kill it? It'd be definitely easier now that most of our steps use Snapshots.

Thanks. I think it makes sense to keep the old `walden` code as long as we have active ETL steps that depend on it. Otherwise, we lose transparency on how the datasets were created. But we can start removing unnecessary code little by little, and at some point kill `walden`.
I'll merge this after the relevant ETL steps are archived.

pabloarosado:
> Nice! I wanted to get rid of walden and even [got a PR](https://github.com/owid/etl/pull/685), but then decided not to continue with it as I didn't see having walden longer with us as a problem. Perhaps now is the right time to kill it? It'd be definitely easier now that most of our steps use Snapshots.

Thanks. I think it makes sense to keep the old `walden` code as long as we have active ETL steps that depend on it. Otherwise, we lose transparency on how the datasets were created. But we can start removing unnecessary code little by little, and at some point kill `walden`.
I'll merge this after the relevant ETL steps are archived.

PR #1349, opened 2023-07-17
:lipstick: use placeholders for poverty index metadata file

Marigold:
Use placeholders instead of a list of descriptions (that are labelled as invalid by YAML linter).

There are two ways how to use newlines. Either you strip newlines from `definitions` with `|-` and then make newlines in `description` explicitly.

```
description: |
          {definitions.cme}

          Multidimensional Poverty Index.

          {definitions.mpi_first}

          {definitions.core}

          {definitions.intensity}

          {definitions.mpi_last}
```

or you use `definitions` with `|` that keeps newlines and then the description would be more compact, although you have to make a newline after the header which looks a bit odd

```
description: |
          {definitions.cme}
          Multidimensional Poverty Index.

          {definitions.mpi_first}
          {definitions.core}
          {definitions.intensity}
          {definitions.mpi_last}
```

@paarriagadap what do you think, is this better than a list of descriptions?


paarriagadap:
Hi @Marigold ! It looks excellent, better than before because we avoid errors and I prefer the option you chose, the first one: it looks cleaner to me. The only thing that I like how it was before is that the editor uses a different color for the anchors, here they have the same color as the text (orange)
image
image

But that's minor.

I introduced some changes, because there was a text that didn't belong to an indicator's metadata. Additionally, I corrected the line generated after HTML list and I also corrected a link.

paarriagadap (approved):

PR #1351, opened 2023-07-17
📊 mars: add new war dataset

lucasrodes:
This is another dataset providing indicators for the History of War project.

Reported indicators are the same as in UCDP: # ongoing conflicts, # new conflicts and # deaths (high and low estimates).

In this case, only two types of conflicts are considered: civil war and non-civil war (i.e. the rest).

Some complexity in the code comes from the fact that:
- a conflict may take place in several regions.
- a conflict may have different types along its course (i.e. start as a civil war, then become a non-civil war).

These issues are very important to remember to avoid counting conflicts twice. For instance, if the same conflict occurs in Europe and Asia, we should only add one when estimating the number of conflicts in the World.


etl/steps/data/garden/war/2023-07-07/mars.meta.yml
```diff
@@ -0,0 +1,66 @@
...
+  title: History of war (Project Mars, 2020)
+  description: >-
+    This dataset contains information on armed conflicts, both internal and external, in the period of 1800 and 2011.
+
+
+    It has been constructed using the Project Mars dataset by Lyall, Jason.
+
+
+    More details about Project Mars:
+
```
paarriagadap, line 11: Why is this built as a text here + the text from the snapshot? Wouldn't it be better to keep text to modify only in one place?

etl/steps/data/garden/war/2023-07-07/mars.meta.yml
```diff
@@ -0,0 +1,66 @@
...
+          numDecimalPlaces: 0
+      number_deaths_ongoing_conflicts_low:
+        title: Number of soldier deaths in ongoing conflicts (low estimate)
+        unit: deaths
+        description: >-
+          The number of soldiers killed in a year over the course of wars. Excluded from this total are missing soldiers as well as those who
+          died from illness and/or disease. This is the high estimate of those who were killed.
+        display:
+          numDecimalPlaces: 0
+
```
paarriagadap, line 33: Is there no central estimates available?

paarriagadap: I see there aren't. But I was thinking about how to plot this in the future (maps, for example). Maybe separate, or create an average between both estimations.

etl/steps/data/meadow/war/2023-06-09/mars.py
paarriagadap, line 1: Maybe you can set an index to avoid the ETL warning

etl/steps/data/garden/war/2023-07-07/mars.py
```diff
@@ -0,0 +1,298 @@
...
+    - Runs minimal sanity check
+    - Keeps only relevant columns
+    """
+    ## Drop NaNs
+    tb = tb.dropna(how="all")
+
+    ## Check at least one and only one FLAG within each group is always activated
+    assert (
+        tb[COLUMNS_REGIONS].sum(axis=1) == 1
+    ).all(), "Entry found with no region (ore more than one region) assigned!"
```
paarriagadap, line 120: Typo in ore

etl/steps/data/garden/war/2023-07-07/mars.py
```diff
@@ -0,0 +1,298 @@
...
+        |-------|---------|----------|-------|--------|
+        |  1990 |    1    |    1     | 1     |   3    |
+        |  1991 |    1    |    1     | 1     |   3    |
+        |  1992 |    1    |    1     | 1     |   3    |
+        |  1993 |    1    |    1     | 1     |   3    |
+
+    """
+    # Add an entry of the triplet ("warcode", "campcode", "ccode") per year
+    # For that we scale the number of deaths proportional to the duration of the conflict.
+    tb[["kialow", "kiahigh"]] = tb[["kialow", "kiahigh"]].div(tb["yrend"] - tb["yrstart"] + 1, "index")
```
paarriagadap, line 195: What happens with non-integer deaths here?

etl/steps/data/garden/war/2023-07-07/mars.py
```diff
@@ -0,0 +1,298 @@
...
+
+    - Goes from triplets to single warcodes.
+    - Preserving triplets until now helps us have better death estimates per year and regions.
+    """
+    tb = tb.groupby(["year", "warcode", "region"], as_index=False).agg(
+        {
+            "conflict_type": lambda x: "others (non-civil)" if "others (non-civil)" in set(x) else "civil war",
+            "kialow": "sum",
+            "kiahigh": "sum",
+            "yrstart": lambda x: min(x),  # df.groupby("warcode")["yrstart"].transform(min)
```
paarriagadap, line 220: Maybe delete this comment?

paarriagadap:
Just for fun: some hallucination from GitHub Copilot 
image

etl/steps/data/garden/war/2023-07-07/mars.meta.yml
lucasrodes: I am still determining why, but the snapshot metadata from Project Mars is appended to the description (?)

Therefore I added this final line before the text from snapshot.

Is this something happening to you too?

etl/steps/data/garden/war/2023-07-07/mars.py
lucasrodes: In line 79, we assign type `Int64` to these columns. Although this converts floats to integers, it applies always `floor` operation.

Instead I'll add a `round` call, so that actually a rounding operation is performed.

etl/steps/data/garden/war/2023-07-07/mars.meta.yml
paarriagadap: It's weird, uh? It also happens to me in this case. For my newest datasets, I keep the snapshot descriptions whenever possible, because currently things are carried well. And in the past the text was replaced so this hasn't happened to me before

PR #1352, opened 2023-07-17
fix: Add missing index files that are still used in the active dag

pabloarosado:
[ETL is failing to build](https://buildkite.com/our-world-in-data/etl-deploy-master/builds/1813#01896439-46dd-4207-aa0f-e74d6bd8b91e) and I think it's because these walden index files should have not been deleted (in my previous PR, https://github.com/owid/etl/pull/1343).


Marigold (approved):
Removing old code is always tricky, thanks for doing this.

pabloarosado:
I think I've fixed it now. I've also archived the previous owid_energy dataset, to avoid a warning by `version_tracker`.

PR #1357, opened 2023-07-18
Include different sources by variable in Multidimensional Poverty Index

paarriagadap:
@Marigold I am experimenting with two things:

1. Call values to fill three different source definitions that share parameters (in `source_common`)
2. Call the `sources` structure for each indicator (like in `cme_mpi_national` or `cme_mpi_urban`)

Is it possible to make placeholders work for (1) and (2)?

For (1) I thought I was doing right, but I have YAML warnings (strangely not for `hot` (?)):
image

I tried to use placeholders for (2), but I couldn't. For now, I was trying to use anchors, as in [here](https://github.com/owid/etl/blob/master/etl/steps/data/garden/eth/2023-03-15/ethnic_power_relations.meta.yml), but we are trying to avoid that.

### DIFF
--- a/etl/grapher_model.py
+++ b/etl/grapher_model.py
@@ -531,6 +531,9 @@ class Source(SQLModel, table=True):
         if self.description.get("additionalInfo"):
             conds.append(cls.description["additionalInfo"] == self.description["additionalInfo"])  # type: ignore
 
+        if self.description.get("dataPublishedBy"):
+            conds.append(cls.description["dataPublishedBy"] == self.description["dataPublishedBy"])  # type: ignore
+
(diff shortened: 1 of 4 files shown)


paarriagadap:
Now I have fixed the string warnings, but not some date warnings:
image
image

paarriagadap:
Thank you @Marigold! The dates are working now. Is there a way to include different sources with placeholders? I tried anchors, which are not working properly, and we want to get rid of them.

paarriagadap:
Cool! Thank you very much, @Marigold! Yeah, now it's working as expected. This is a very specific use case, where the authors want to see the three papers related to their corresponding indicator, but they are, at the same time, the same data product.

PR #1366, opened 2023-07-20
📊 terrorism dataset

veronikasamborska1994:


etl/steps/data/meadow/terrorism/2023-07-20/global_terrorism_database.py
```diff
@@ -0,0 +1,81 @@
...
+        "region_txt",
+        "attacktype1_txt",
+        "weaptype1_txt",
+        "targtype1_txt",
+        "nkill",
+        "nwound",
+    ]
+
+    # Load data from snapshots.
+    df_2020 = pd.read_excel(snap_2020.path)
```
Marigold, line 60: Loading the excel file took almost 5 minutes on my laptop. It might be better to convert it to `.csv` format on your local and then upload the csv as a snapshot instead of excel file (it might not be crystal "raw", but it looks like there's nothing extra in the excel file).

(For comparison, csv file is 5x smaller and loads in 0.3s. Using `feather` format would compress even better, but it's not as common as csv)

etl/steps/data/garden/terrorism/2023-07-20/global_terrorism_database.py
```diff
@@ -0,0 +1,253 @@
...
+        "total_incident_counts",
+        "total_casualties",
+        "terrorism_wounded_per_capita",
+        "terrorism_deaths_per_capita",
+        "terrorism_casualties_per_capita",
+    ]
+
+    df_pop_deaths = perform_decadal_averaging(df_pop_deaths, cols_for_decadal_av=cols_for_decadal_av)
+
+    # Convert relevant columns to float64 data type (to avoid errors related to this issue -  https://github.com/owid/etl/issues/1334)
```
Marigold, line 138: Oh man, working with both `float64` and `Float64` is a nightmare! What we often do is that we convert everything to `float` right after loading to avoid these problems. You could do

```
tb_terrorism = tb_terrorism.astype(
        {
            "nkill": float,
            "nwound": float,
        }
    )
```
it doesn't really matter that these are really not floats, but ints with NaN values. When saving the dataset we optimize types anyway.

etl/steps/data/garden/terrorism/2023-07-20/global_terrorism_database.py
```diff
@@ -0,0 +1,253 @@
...
+
+    for column in df[cols_for_decadal_av]:
+        # Create a new column name for decadal average
+        decadal_column = f"decadal_{column}"
+
+        # Group the data by decade and compute the mean for each decadal group
+        df[decadal_column] = df.groupby(df["year"] // 10 * 10)[column].transform("mean")
+
+        # Replace non-decadal years with NaN
+        df[decadal_column] = df[decadal_column].mask(df["year"] % 10 != 0, pd.NA)
```
Marigold, line 182: If you decide to work with `float` (see my comment above), then you should use `np.nan` instead of `pd.NA`. I'd stay away from `pd.NA` until we upgrade to pandas 2.0.

Marigold (approved):
Added some performance tips, the dataset in admin looks good!

PR #1371, opened 2023-07-21
📊 war: add prio 3.1 dataset

lucasrodes:
parent issue: https://github.com/owid/owid-issues/issues/446

### 📝 Notes
_(mostly for @bastianherre)_

- Each row in the original dataset is an observation of a conflict (i.e. conflict 1 in 1946). There is a total of ~1950 observations.
- Missing death estimates:
  - Deaths are given as low, best and high estimates
  - While low and high estimates are always present, best estimates are missing for some rows.
  - Missing values here are unlikely to represent zeroes. E.g., for `id=1` and `year=1952`, we have:
    - low death estimate: 450
    - high death estimate: 3000
    - best death estimate: NA
  - From all observations (~1950), there are ~800 with missing best death estimates.
  - Consequently, the value for best estimates for some years cannot be aggregated (some of the values are missing and are not zero).


etl/steps/data/garden/war/2023-07-21/prio_v31.meta.yml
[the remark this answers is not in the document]
paarriagadap: Also, the meadow step could set index to prevent the warning to appear

etl/steps/data/garden/war/2023-07-21/prio_v31.py
```diff
@@ -0,0 +1,294 @@
...
+    ## For columns "number_ongoing_conflicts", "number_new_conflicts"
+    columns = [
+        "number_ongoing_conflicts",
+        "number_new_conflicts",
+        "number_deaths_ongoing_conflicts_battle_high",
+        "number_deaths_ongoing_conflicts_battle_low",
+    ]
+    tb.loc[:, columns] = tb.loc[:, columns].fillna(0)
+
+    # Set number of deaths to zero whenever high and low estimates are zero
```
paarriagadap, line 259: What are the problems with estimating a value for `number_deaths_ongoing_conflicts_battle` when it's missing and high and low are not? Just as the previous dataset, I wonder how you'd solve that on the explorers. Probably showing the three together for all the views

paarriagadap (approved):
Hi @lucasrodes! Looks good! I mostly commented on the snapshot and meadow steps, which are not in this PR, and a question again on how to solve the best estimate indicator when it's missing and the rest is not

etl/steps/data/garden/war/2023-07-21/prio_v31.py
lucasrodes: `number_deaths_ongoing_conflicts_battle` is missing for some rows in the source; hence we can't obtain it.

In these lines, we fill NaNs in `number_deaths_ongoing_conflicts_battle` only when values in high and low estimates are also zero.

paarriagadap: I know, it just came to my mind the question here

PR #1372, opened 2023-07-24
:bug: load aviation statistics from snapshot rather than from walden

Marigold:
Fix the failing nightly ETL build.


pabloarosado (approved):
Thanks!

PR #1376, opened 2023-07-25
:bug: fix Table.to_* methods

Marigold:
The problem is this line
```python
if not isinstance(path, str) or not path.endswith(".csv"):
```
if path has type `Path` then it immediately throws ValueError even if the path is valid and ends with `.csv`. So the only option is to pass string path.

I ran into this when I was calling `df_to_file(table)` - the function converts it to `Path` object which throws the error.

@pabloarosado is this valid bug or am I missing something?


pabloarosado (approved):
The changes regarding paths make sense.

PR #1377, opened 2023-07-25
:hammer: Refactor fast track & add origins

Marigold:
This PR was originally only refactoring, but I ended up adding origins as well. Let me know if this is too hard to review and I can split it in two.

## Refactoring Fast-track

Simplify fast-track. Allow only one source / origin and get rid of source's `short_name` (which has caused just confusion in the past). No one has used multiple sources so far. Having just one source simplified the logic a lot. It's all backward compatible.

## Add origins

Add new sheet `origins_meta`. Here's an example of the [new template](https://docs.google.com/spreadsheets/d/1HqIjmsjrXimiY-dMrwcm0iNQJfZKpUgIz0nLVoeoiKc/edit?usp=sharing) (note that I'll have to update template after merging this to avoid breaking old fast-track spreadsheets). It's very bare-bones at the moment. I'll polish it when we start using it.

## TODO
- [x] test loading from CSV
- [x] test loading from S3 path
- [x] add origins
- [x] make sure as much metadata as possible is saved in Origin YAML
- [x] modify fasttrack template to allow only one source / origin

## Edits to template spreadsheet
- [ ] remove all `sources` fields
- [ ] add `origins` sheet
- [ ] add CHANGELOG sheet with updates


MutugiD:
Are any of the unchecked tasks on Issues open for collaboration? 
Just came across this repo from ourworldindata.org.

Marigold:
Hey @MutugiD. Unfortunately, these tasks need a lot of context, but thanks for your interest!

pabloarosado:
Quick note: I just swapped the values of `producer` and `citation_producer` in the template.

PR #1380, opened 2023-07-25
📊 Improving GBD metadata

spoonerf:


etl/steps/data/garden/ihme_gbd/2019/gbd_tools.py
```diff
@@ -152,12 +153,177 @@ def add_share_of_population(df: pd.DataFrame) -> pd.DataFrame:
...
+    ds_garden = Dataset.create_empty(dest_dir)
+    ds_garden.metadata = ds_meadow.metadata
+
+    df = df.reset_index()
+    df_group = df.groupby(dims)
+    for group_id, group in df_group:
+        dims_id = dict(zip(dims, group_id))
+        cause = dims_id["cause"]
+        sex = dims_id["sex"]
+        age = dims_id["age"]
```
lucasrodes, line 286: Maybe you don't need to define these variables? I see you only use them in line 292, where you could just use the dictionary directly.

Actually, since the keys in the dictionary are named the same as the arguments from `create_variable_metadata`, you could use the following for line 292:

```
cleaned_variable = create_variable_metadata(variable=tb_group[variable_name], **dims_id)
```

etl/steps/data/garden/ihme_gbd/2019/gbd_tools.py
```diff
@@ -152,12 +153,177 @@ def add_share_of_population(df: pd.DataFrame) -> pd.DataFrame:
...
+        tb_group.metadata.short_name = underscore(f"{dims_id['cause']} - {dims_id['sex']} - {dims_id['age']}")
+        variables = tb_group.columns.drop(dims + ["country", "year"])
+        for variable_name in variables:
+            tb_group[variable_name] = Variable(tb_group[variable_name])
+            cleaned_variable = create_variable_metadata(variable=tb_group[variable_name], cause=cause, sex=sex, age=age)
+            tb_group[cleaned_variable.name] = cleaned_variable
+            tb_group = tb_group.drop(columns=variable_name)
+        tb_group = tb_group.set_index(["country", "year"] + dims, verify_integrity=True)
+        ds_garden.add(tb_group)
+    ds_garden.save()
```
lucasrodes, line 297: maybe add some comments to help navigate this code?

etl/steps/data/grapher/ihme_gbd/2019/gbd_mental_health.py
```diff
@@ -0,0 +1,13 @@
+from owid import catalog
+
+from etl.helpers import PathFinder
+
+from .gbd_tools import run_wrapper
+
+N = PathFinder(__file__)
```
lucasrodes, line 7: maybe change to more standard naming `paths`?

etl/steps/data/garden/ihme_gbd/2019/gbd_tools.py
lucasrodes, line 297: I would change this function so it only adds the metadata and does not perform the save operation.

That is:

```
def add_metadata(dest_dir: str, ds_meadow: Dataset, df: pd.DataFrame, dims: List[str]) -> Dataset:
  ...
  return ds_garden
```

And then perform the `.save()` operation. Feels easier if we wanted to use the function to debug its output.

lucasrodes (approved):
Minor suggestions, LGTM!

etl/steps/data/garden/ihme_gbd/2019/gbd_tools.py
spoonerf: Oh wow, that's new python wizardry for me!

PR #1392, opened 2023-07-31
🐛 war.ucdp: fix missing values

lucasrodes:
- Improve and add docstrings
- Improve function names
- Add new function: `fix_extrasystemic_entries`, to help deal with `conflict_type=extrasystemic` conflicts after 1989 (prior to this PR values for these would be missing, with this PR we just zero-fill values after 1989(

### DIFF
--- a/etl/steps/data/garden/war/2023-06-22/ucdp.py
+++ b/etl/steps/data/garden/war/2023-06-22/ucdp.py
@@ -49,7 +49,7 @@ def run(dest_dir: str) -> None:
     #
     # Process data.
     #
-    log.info("ucdp: sanity checks")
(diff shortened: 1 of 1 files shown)


lucasrodes:
thanks for the review, Fiona!

Yeah, that can happen. Sometimes, a conflict occurs in multiple regions (e.g. WWII). Therefore, this conflict s counted in more than one region. Consequently, if adding the values for all regions, you may count a conflict more than once.

In this particular case, for instance, we have the conflict "IS - civilians" in 2015, which ー according to UCDP ー took place in regions: 'Asia', 'Africa', 'Middle East', 'Europe', 'Americas'.

PR #1394, opened 2023-07-31
:hammer: shortUnit should never be null in display

danyx23:
The grapher config schema specifies that the unit and shortUnit of the display section should not be null. This was checked in the ETL for unit but the same check was missing for shortUnit. This PR fixes this.


pabloarosado (approved):
Looks good.

PR #1405, opened 2023-08-01
📊 adding CSET investment, articles and patents

veronikasamborska1994:


etl/steps/data/garden/artificial_intelligence/2023-07-25/cset.py
```diff
@@ -0,0 +1,83 @@
...
+    df_total["country"] = "Total"
+    df_total.set_index(["year", "country", "field"], inplace=True)
+    merged_total = pd.concat([df_total, tb])
+    # List of columns to include for conversion to millions (investment values)
+    _investment_cols = [col for col in merged_total.columns if "investment" in col]
+    # Convert all other columns to million
+    merged_total.loc[:, _investment_cols] *= 1e6
+
+    # Import US CPI data from the API
+    df_wdi_cpi_us = us_cpi.import_US_cpi_API()
```
spoonerf, line 49: Wow this is super cool, you're adjusting the investment variables to account for inflation?

etl/steps/data/garden/artificial_intelligence/2023-07-25/cset.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+    publication_year: 2023
+    published_by: Emerging Technology Observatory Country Activity Tracker, Artificial Intelligence (Center for Security and
+      Emerging Technology, 2023)
+tables:
+  cset:
+    variables:
+      disclosed_investment:
+        title: Disclosed investment
+        description:  Investments in AI-related companies across various industries where the investment value has been publicly disclosed.
+        unit: US dollars
```
spoonerf, line 49: I think these are the variables that have been adjusted for inflation? It would be good to note that you have done this here, following the guide [here](https://docs.google.com/document/d/1tcknTryJAxOR0yv8x38s3jkONq-TY3T4HNCEcZQ_c-s/edit#heading=h.58dyovrzqydc) for the unit.

etl/steps/data/garden/artificial_intelligence/2023-07-25/cset.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+          numDecimalPlaces: 0
+          name: Applications
+
+      num_patent_granted_summary:
+        title: Total patent applications granted
+        description: Total number of patents granted between 2010-2020.
+        unit: granted applications
+        display:
+          numDecimalPlaces: 0
+          name: Granted
```
spoonerf, line 113: Perhaps making the display name slightly more specific would be useful, so here 'Patents granted'.

etl/steps/data/garden/artificial_intelligence/2023-07-25/us_cpi.py
```diff
@@ -0,0 +1,63 @@
...
+from structlog import get_logger
+
+log = get_logger()
+
+# change number of years to go back if neccessary (used up to 2000 in 2023)
+
+YEARS = 23
+
+
+def import_US_cpi_API(years_back=YEARS):
```
spoonerf, line 14: This is very cool that you are using this data! I think it might be better to get it from the WB dataset in the etl like you did for [the UNWTO dataset](https://github.com/owid/etl/blob/57113df3bd20f04de04aafd5aa37b8d08a60ed2d/etl/steps/data/garden/tourism/2023-05-05/unwto.py). There's just the slight advantage that we'll definitely know what version we are using and can be more sure it won't break unexpectedly.

spoonerf (approved):
Looks great, just a couple of comments!

etl/steps/data/garden/artificial_intelligence/2023-07-25/cset.meta.yml
veronikasamborska1994: actually I need to remove this altogether because I'd be expanding everything in grapher. Do you by any chance know if the only way to assign custom names for the variables that are created based on the index is manually in grapher?

etl/steps/data/garden/artificial_intelligence/2023-07-25/cset.meta.yml
veronikasamborska1994: yes, thank you!

etl/steps/data/garden/artificial_intelligence/2023-07-25/us_cpi.py
veronikasamborska1994: The reason why I am doing this here is because our WB dataset only goes up to 2021 whereas we can get the data for 2022 too via the API . So this is just a workaround I've used until our WB dataset is updated for my AI datasets in general :) Though it looks like the data in this one only goes up until 2021 anyway so I can also just use the WB dataset for now.

spoonerf: Ahhh that makes sense!

etl/steps/data/garden/artificial_intelligence/2023-07-25/cset.meta.yml
spoonerf: I have started doing it in the garden step like with the add_metadata() and create_variable_metadata() functions [here](https://github.com/owid/etl/blob/fabe2f1ea52276824755976eaf3fc7debda8fbca/etl/steps/data/garden/who/2023-08-01/mortality_database.py). I think you could also do something similar in the grapher step?

veronikasamborska1994: hey this is pretty cool! I've tried to do something similar for my education datasets but I don't think it will work in the grapher step without modifying some hardcoded functions :( I think Mojmir will take a look at whether it's possible though and I'll let you know if we find a way in case this will also be useful for you :)

spoonerf: That would be great - thanks!

veronikasamborska1994 (commented):
thanks so much again! :) I think this might end up being quite useful for other datasets as well! tagging @spoonerf in case that's something that's also useful for you!

PR #1413, opened 2023-08-02
Add uk egg statistics

pabloarosado:
Add Defra data on UK egg statistics.
Additionally, install a library to be able to open .ods files.

### DIFF
--- a/etl/steps/data/garden/animal_welfare/2023-08-01/uk_egg_statistics.meta.yml
+++ b/etl/steps/data/garden/animal_welfare/2023-08-01/uk_egg_statistics.meta.yml
@@ -1,0 +1,22 @@
+dataset:
+  title: "UK egg statistics (Defra, 2023)"
+
+tables:
+  uk_egg_statistics:
+    variables:
+      number_of_eggs_from_barns:
+        title: Number of eggs from hens in barns
+        unit: eggs
+        short_unit: ''
+      number_of_eggs_from_enriched_cages:
+        title: Number of eggs from hens in (enriched) cages
+        unit: eggs
+        short_unit: ''
+      number_of_eggs_from_non_organic_free_range_farms:
+        title: Number of eggs from hens in non-organic, free-range farms
+        unit: eggs
+        short_unit: ''
+      number_of_eggs_from_organic_free_range_farms:
+        title: Number of eggs from hens in organic, free-range farms
+        unit: eggs
+        short_unit: ''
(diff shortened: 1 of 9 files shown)


etl/steps/data/garden/animal_welfare/2023-08-01/uk_egg_statistics.meta.yml
```diff
@@ -0,0 +1,22 @@
+dataset:
+  title: "UK egg statistics (Defra, 2023)"
+
+tables:
+  uk_egg_statistics:
+    variables:
+      number_of_eggs_from_barns:
+        title: Number of eggs from hens in barns
```
spoonerf, line 8: Probably good to set decimal places to 0 for all of these.

snapshots/animal_welfare/2023-08-01/uk_egg_statistics.ods.dvc
```diff
@@ -0,0 +1,24 @@
...
+  license_url: 
+    https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1173898/egg-packers-27july23.ods
+  license_name: Public domain
+  date_accessed: 2023-08-01
+  is_public: true
+  description: |
+    DEFRA runs a quarterly survey of registered UK egg packing stations. It is a voluntary sample survey of 27 respondents that collects information on throughput by production type and prices of graded eggs and sales of ungraded eggs. The response rate is typically 100 per cent and the survey accounts for 75 per cent of eggs packed in the UK. The survey figures are raised up to give UK estimates using information on the number of commercial laying hens, average egg yields, average mortality rates, the proportion of UK eggs that go through packing stations. Throughput by egg type for packing stations not surveyed is calculated using data provided by packing stations responding to the survey. The raised figures are published in this statistics notice and the associated datasets. The figures in this notice therefore represent all Class A eggs passed over a grader in the UK, including seconds.
```
spoonerf, line 19: Maybe not important but I'm not sure all of this is relevant and it's not totally clear to me what the 'seconds' in the last sentence refers to.

spoonerf (approved):
Looks good. Would it also be interesting to calculate the total number of eggs and then the % of each type of egg shown here?

snapshots/animal_welfare/2023-08-01/uk_egg_statistics.ods.dvc
pabloarosado: I don't understand this explanation either. I think "seconds" means "[Class B](https://www.farminguk.com/content/knowledge/Grading%20your%20eggs.pdf)". But the sentence doesn't make sense to me.

pabloarosado:
> Looks good. Would it also be interesting to calculate the total number of eggs and then the % of each type of egg shown here?

Thanks Fiona! That's a good idea, although this data should be used to replace [this chart](https://ourworldindata.org/grapher/egg-production-system), in which case we can just use the "relative" option.
Additionally, I may have to come back to this dataset, given that there are a few issues, so I'll not polish the metadata just yet.

PR #1418, opened 2023-08-03
:hammer: refactor walkthrough

Marigold:
- Delete JSON files for cookiecutter templates
- Only keep YAML file in garden step
- Collapse instructions
- Fix minor bugs


pabloarosado (commented):
I'm a bit confused about the `dummy` metadata file `snapshots/dummy/2020-01-01/dummy.csv.dvc`, it seems like the new version of the file uses the old metadata.

Marigold:
> I'm a bit confused about the dummy metadata file snapshots/dummy/2020-01-01/dummy.csv.dvc, it seems like the new version of the file uses the old metadata.

This is just refactoring, origins will be added in [this PR](https://github.com/owid/etl/pull/1369) (together with new dummy metadata).

snapshots/ggdc/2020-10-01/ggdc_maddison.xlsx.dvc
```diff
@@ -1,5 +1,4 @@
 meta:
-  name: Maddison Project Database (Bolt and van Zanden, 2020)
```
Marigold, line 2: @pabloarosado snapshot name and description are not used anymore in the new metadata framework. Currently when you call
```
snap = Snapshot("...")
ds = create_dataset(dest_dir, tables=[...], default_metadata=snap.metadata)
```
it'll try to fill dataset metadata from snapshot's metadata [like this](https://github.com/owid/etl/blob/master/etl/steps/data/converters.py#L49)
```
# dataset title and description are filled from origin
title=snap.origin.dataset_title_owid,
description=snap.origin.dataset_description_owid,
```
(which can be overwritten by setting `dataset.title` in YAML file)

The thing is that `dataset_title_owid` might be sometimes missing. Should we try to fall back to `dataset_title_producer`? Perhaps making either of the titles should be mandatory?

pabloarosado: I think the easiest solution is that `dataset_title_owid` should be mandatory, and `dataset_title_producer` should be optional (as very often the data doesn't really have an official title). Neither of the descriptions should be mandatory. Do you agree?

Marigold: Yes, I'll add it to this PR. GGDC will look like this
```
meta:
  origin:
    dataset_title_owid: Maddison Project Database
    dataset_title_producer: Maddison Project Database
```
(We could DRY it, but it's probably ok. I haven't included author and year.)

pabloarosado (approved):
All good!

Issue #1403, opened 2023-08-01
wizard: wizard charts fails with large datasets


spoonerf:
### One-liner
When running `walkthrough charts` on 5593 -> 6161 it takes a long time refreshing when selecting new variables to match. I have also found it's crashed twice when using it, at different points, once when I was half way through 😢 . It gives the error below:

![Screenshot 2023-08-01 at 14 47 37](https://github.com/owid/etl/assets/10774728/2f7bb574-c8cb-49ed-9582-948138197d1a)

I've used it successfully on other datasets, this time it kind of felt like it was struggling with the size?

### Environment

- `python -V`: 3.11
- Operating system: Mac OS
- Working with local, staging or live: live

Marigold:
@lucasrodes do you have the capacity or should I take a look?

spoonerf:
I was thinking maybe I don't need to use the UI - is it possible to run chart revisions without it?

lucasrodes:
Hi! I remember that happened to me as well, with large datasets, the tool could be more stable.

@Marigold I am a bit low on time and unsure why `get_connection` raises an error. Most of the code for walkthrough charts is in `walkthrough/charts_v2`.

@spoonerf I think you could use the terminal version in the meantime.

1. First need to create the variable mapping dictionary. For this, you can use the command line tool `etl-match-variables`.
2. Need to create the charts and submit to admin. For this, use the command line `etl-chart-suggester`.

If you encounter any problems with these issues, let me know by raising another issue!

spoonerf:
Thanks @lucasrodes!

Is the correct way to use `etl-match-variables` like so: 

`ENV=.env.prod etl-match-variables -f ~/ -old IHME - Global Burden of Disease - Risk Factors - Institute for Health Metrics and Evaluation  (2022-04) -new Institute for Health Metrics and Evaluation - Global Burden of Disease (2019) - Risk factors - Risk factors` 

I don't think I am using it quite right 😅

spoonerf:
@lucasrodes With the `etl-match-variables` could it be possible that the function checks that `-output-file` is valid before the mappings are done? Just got burnt by using a directory instead of a file 😭

lucasrodes:
Gotcha, @spoonerf. Opened https://github.com/owid/etl/pull/1423

pabloarosado:
I'm having the same problem with the Global Carbon Budget (both local and on my staging server).
[...]

PR #1423, opened 2023-08-03
✨ match-variables: check validity of output_file

lucasrodes:
Command line tool `etl-match-variables` has a required input argument `output-file`, which points to the file where the variable mapping will be saved. If this file is not `JSON`, the export will fail.

This is currently not being checked, hence if the user of the tool was to insert an invalid value for `output-file` they will only realise at the very end (after investing time in building the variable mapping).

Hence, this PR sanity checks the value of `output-file` at first.

First raised in #1403


spoonerf (approved):
Works for me! Thanks @lucasrodes really appreciate you fixing this so quickly 💪🏼

PR #1428, opened 2023-08-04
:sparkles: make all metadata classes hashable

Marigold:
Make all metadata classes hashable (i.e. enable `hash(obj)`) to allow usage of `set` or `unique` on those objects.


pabloarosado (approved):
Thanks, it looks good!

PR #1430, opened 2023-08-05
📊  Adding data on ICD code version reported to WHO Mortality Database

spoonerf:


etl/steps/data/garden/who/2023-08-04/icd_codes.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # Calculate sum of number of countries using each ICD code type each year
+    tb_sum = tb.groupby(["year", "icd"]).count().reset_index()
+    tb_sum = tb_sum.rename(columns={"icd": "country", "country": "countries_using_icd_code"})
+
+    # More of a traditional grapher dataset. For each country-year, which ICD code is being used.
+    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path)
+
+    # Combine the datasets
+    tb_combined = pr.concat([tb_sum, tb])
```
paarriagadap, line 36: This function also includes a `short_name` parameter, so you can add it here instead of doing it two lines below.

etl/steps/data/garden/who/2023-08-04/icd_codes.py
```diff
@@ -0,0 +1,45 @@
...
+    tb = ds_meadow["icd_codes"]
+    tb = tb.reset_index()
+    icd_codes_map = {"Icd10": "ICD-10", "Icd9": "ICD-9", "Icd8": "ICD-8", "Icd7": "ICD-7"}
+    tb["icd"] = tb["icd"].replace(icd_codes_map)
+
+    # Calculate sum of number of countries using each ICD code type each year
+    tb_sum = tb.groupby(["year", "icd"]).count().reset_index()
+    tb_sum = tb_sum.rename(columns={"icd": "country", "country": "countries_using_icd_code"})
+
+    # More of a traditional grapher dataset. For each country-year, which ICD code is being used.
```
paarriagadap, line 32: It seems that this comment was referring to a different code

etl/steps/data/garden/who/2023-08-04/icd_codes.countries.json
```diff
@@ -0,0 +1,163 @@
...
+  "United Arab Emirates": "United Arab Emirates",
+  "United Kingdom": "United Kingdom",
+  "United States of America": "United States",
+  "Uruguay": "Uruguay",
+  "Uzbekistan": "Uzbekistan",
+  "Venezuela": "Venezuela",
+  "Zimbabwe": "Zimbabwe",
+  "China: Province of Taiwan only": "Taiwan",
+  "Czechoslovakia, Former": "Czechoslovakia",
+  "Germany, Former Democratic Republic": "Germany, Former Democratic Republic",
```
paarriagadap, line 149: I think `East Germany` is the official entity at OWID

etl/steps/data/garden/who/2023-08-04/icd_codes.countries.json
```diff
@@ -0,0 +1,163 @@
...
+  "Germany, Former Federal Republic": "Germany, Former Federal Republic",
```
paarriagadap, line 150: I think `West Germany` is the official entity at OWID

etl/steps/data/garden/who/2023-08-04/icd_codes.py
paarriagadap, line 36: Also, why don't you just create two different tables for the ICD totals and the country ICDs? It might be confusing to see two different types of entities in `country`

paarriagadap (approved):
Hi @spoonerf! Looks good! I added some minor comments and I wonder if it's better to have this in two tables to be clearer with the entities.

etl/steps/data/garden/who/2023-08-04/icd_codes.py
spoonerf: Ah nice, thanks!

spoonerf: Ah good point, I think I tried it and had some issues in only being able to upload one table to grapher, but you are right I can have two in garden and then combine them in grapher.

paarriagadap: And how about uploading two tables in Grapher? I think it should be fine

spoonerf: I shall try, for some reason it seemed not to work before but perhaps that was because of another issue.

spoonerf: You were right :)

PR #1435, opened 2023-08-07
:tada: upload data & metadata JSON in backporting

Marigold:
Add option `--data-metadata` to `backport` CLI that syncs backported data & metadata JSON to R2. This lets us reconstruct our data & metadata files from backported datasets rather than from MySQL as was done by `datasync` service. This PR deprecates `datasync` too.

## TODO
* [x] merge https://github.com/owid/etl/pull/1425 first
* [x] validate parity with existing data & metadata files
* [ ] backport everything again, including archived datasets
* [x] turn off old datasync service


pabloarosado (approved):
I don't fully understand this PR, sorry. If you need a more thorough review, I'd need a bit more context. The one thing that puzzled me is the fact that you are backporting archived datasets.

Marigold:
> I don't fully understand this PR, sorry. If you need a more thorough review, I'd need a bit more context.

It's fine, there's actually not much to review. 

> The one thing that puzzled me is the fact that you are backporting archived datasets.

Very good question (and still open). The reason I'm backporting them is to preserve their data somewhere (i.e. in backport snapshots) since we're getting rid of `data_values` table in MySQL. If I didn't backport them, we'd lose the data forever.

I was thinking about hiding archived backported datasets in catalog, but since no one is using backported channel anymore, it's probably a waste of time.

(I'm open to ideas!)

PR #1438, opened 2023-08-08
Add farmed fish data

pabloarosado:
* Add data about the number of farmed finfishes used for food worldwide.
* Add a helper function to print metadata (to easily create a yaml file in a garden step).
* Add a function `read_from_records` (analogous to `pd.DataFrame.from_records()`) that assigns metadata to each column.


snapshots/animal_welfare/2023-08-08/farmed_finfishes_used_for_food.zip.dvc
```diff
@@ -0,0 +1,26 @@
+meta:
```
spoonerf, line 1: For some reason the metadata has not passed through to my local grapher, probably a me problem but thought I would flag it anyway.

snapshots/animal_welfare/2023-08-08/farmed_finfishes_used_for_food.zip.dvc
```diff
@@ -0,0 +1,26 @@
...
...level where possible). We relate these numbers to knowledge on humane slaughter, animal welfare law, and certification schemes. Since 1990, farmed finfish numbers killed annually for food have increased nine-fold, to 124 billion (1.24 × 1011, range 78–171 billion) in 2019. This figure does not represent the total number farmed (due to mortalities during rearing and non-food production) and is expected to increase as aquaculture expands. Our estimates indicate that farmed finfishes now outnumber the 80 billion farmed birds and mammals killed globally each year for food. The majority are produced in Asia. Inhumane slaughter practices cause suffering for most farmed finfishes. Most, 70–72%, have no legal welfare protection, and less than 1% have any fish-specific legal protection, at slaughter. The main global certification schemes in 2013–2015 accounted for 2% of slaughtered farmed finfishes. Fishes for which species-specific parameters for automated humane stunning are published comprise 20–24%. As the dominant taxa of farmed vertebrates, finfishes would benefit from better welfare if species-specific humane slaughter was defined and incorporated into laws and certification schemes.
```
spoonerf, line 7: Maybe good to note what a finfish is, it's not immediately clear.

etl/steps/data/meadow/animal_welfare/2023-08-08/farmed_finfishes_used_for_food.py
```diff
@@ -0,0 +1,98 @@
...
+        records,
+        columns=[
+            "year",
+            "production_kilotonnes",
+            "n_fish_lower_billions",
+            "n_fish_upper_billions",
+            "n_fish_midpoint_billions",
+            "production_relative_to_1990_pct",
+            "n_fish_relative_to_1990_pct",
+            "n_birds_and_mammals_billions",
```
spoonerf, line 83: I find the bird and mammal comparison interesting, I think fish are quite different in that they can produce thousands of eggs and offspring, which in aquaculture are presumably bred with quite a high success rate. However, with birds and especially mammals, there is a pretty low limit on the number of offspring they can produce.

spoonerf (approved):
Looks good! The metadata has not come through to my local grapher, but perhaps that is a problem in my local setup. Does it work for you?

snapshots/animal_welfare/2023-08-08/farmed_finfishes_used_for_food.zip.dvc
pabloarosado: Hmm I do see the metadata (the old metadata fields still, if that's what's puzzling you).

pabloarosado: Ah yes I see the same thing. I'm looking only at metadata at the variable level, I don't really trust the admin (which is going to change very soon).

Issue #1437, opened 2023-08-08
Chart revisions: unable to validate schema


spoonerf:
### One-liner
When running `etl-chart-suggester` with [ihme_variable_match.json.zip](https://github.com/owid/etl/files/12292656/ihme_variable_match.json.zip) I get an error regarding the VariableId.

### Environment

- `python -V`: 3.11
- Operating system: Mac
- Working with local, staging or live: live

### Context & details
I can't see an issue with the affected variables - they both look okay in their respective datasets. 

* [old variable](https://owid.cloud/admin/variables/417751)
* [new variable](https://owid.cloud/admin/variables/775095)

The error: 
![Screenshot 2023-08-08 at 15 45 50](https://github.com/owid/etl/assets/10774728/fe155a03-0a12-48f0-a0c6-bb2348351cb9)

danyx23:
@spoonerf this was a recent schema cleanup - we removed variableId from the .map key of the grapher config. If you remove that field it should work. I'll be off for the rest of the week but please reach out in the developers channel if you have further questions!

Marigold:
@spoonerf thanks for nice bug report (with sample data!). [The fix](https://github.com/owid/etl/pull/1441) is ready and will be merged soon.

PR #1441, opened 2023-08-08
:bug: use map.columnSlug rather than map.variableId

Marigold:
Chart config field `map.variableId` was recently renamed to `map.columnSlug` and changed the type to string. All entries in MySQL have been migrated, so there's no `map.variableId`.

Update this in `etl-chart-suggester`. This was tested on the [IHME data](https://github.com/owid/etl/issues/1437#issue-1841475289) from the parent issue.


lucasrodes (approved):
LGTM! Thanks for fixing.

PR #1446, opened 2023-08-09
Improve metadata propagation

pabloarosado:
Various improvements on metadata propagation:
* Propagate metadata when doing operations like `tb[["a", "b"]].sum(axis=1)` (idem for `prod`).
* Idem for operations like `tb[["a", "b"]] *= 2` and other similar operations.
* Add tests.


Marigold (approved):
Looks good.

PR #1453, opened 2023-08-10
:honeybee: unify flake8 across projects

Marigold:
Unify flake8 config & version across our projects. This is a preparation for switching to a ruff linter.

This removes various `flake8-*` extensions specific for datautils lib. I've tried to add them to other projects, but they're too strict. @lucasrodes let me know if you want to keep them there.


lucasrodes (approved):
Thanks for doing this, makes sense.

Issue #1454, opened 2023-08-11
Unexpected behaviour when modifying variables metadata


pabloarosado:
If you create a new variable as a copy of an existing variable, using `.copy()` and modify the metadata of the new variable, the old variable's metadata is also affected.

This happens even when doing a deep copy of the variable, or using `.copy_metadata()`.

Example:
```
from owid.catalog import Table
tb1 = Table({"a": [1, 2, 3]})
tb1["a"].metadata.title = "Original title"
tb1["a_copy"] = tb1["a"].copy().copy_metadata(from_variable=tb1["a"])
tb1["a_copy"].metadata.title = "New title"
# Now the title of the original variable becomes "New title"
tb1["a"].metadata.title
```

This is unexpected and a bit concerning. At the very least, I think some of the tests in `owid.catalog` may rely on the assumption that a copied variable does not affect the original variable.

Issue #1269, opened 2023-06-27
Merge functions `copy_metadata` and `copy_metadata_from` in `catalog.tables`


pabloarosado:
Currently (after mergin [this PR](https://github.com/owid/etl/pull/1251)), `Table` has both a `copy_metadata_from` and a `copy_metadata` method.

* `copy_metadata_from` is fast (it avoid deep copies) but doesn't copy variables metadata, and does changes in place.
* `copy_metadata` copies variables metadata and has an `inplace` argument, but it's slow (it uses deep copies).

Both functions would need to be merged, combining features and avoiding issues.

Marigold:
@pabloarosado are you working on this? I could implement this pretty easily after we merge https://github.com/owid/etl/pull/1455. Any preference from `copy_metadata_from` and `copy_metadata` names?

pabloarosado:
@Marigold I'm not working on that, but I'm changing a few things on `variables` and `tables` at the moment. I'll create a PR very soon. Maybe after that, if you want, you can look into that.
I'd prefer the shorter version, `copy_metadata` which has been used more times than the other.

PR #1455, opened 2023-08-11
:sparkles: add .copy methods to Variable and VariableMeta

Marigold:
Add `Variable.copy(deep=True)` method. This fixes https://github.com/owid/etl/issues/1454. The method can be reused for other dataclasses (we could even add it to all by default with a decator, similarly to `@pruned_json`).

@pabloarosado I've noticed you implemented `Table.copy_metadata` method. There's already `Table.copy_metadata_from` that I believe does the same (and doesn't use `copy.deepcopy` which is quite slow). Do you think it'd be possible to unify them?

### DIFF
--- a/lib/catalog/owid/catalog/variables.py
+++ b/lib/catalog/owid/catalog/variables.py
@@ -303,6 +303,9 @@ class Variable(pd.Series):
             inplace=inplace,
         )
 
+    def copy_metadata(self, from_variable: "Variable", inplace: bool = False) -> Optional["Variable"]:
+        return copy_metadata(to_variable=self, from_variable=from_variable, inplace=inplace)
+
     def copy(self, deep: bool = True) -> "Variable":
         new_var = super().copy(deep=deep)
         if deep:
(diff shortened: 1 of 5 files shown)


pabloarosado:
Thanks a lot @Marigold! Regarding `copy_metadata`, there is already an issue for that: https://github.com/owid/etl/issues/1269

lib/catalog/owid/catalog/variables.py
```diff
@@ -303,8 +303,11 @@ def amend_log(
             inplace=inplace,
         )
 
-    def copy_metadata(self, from_variable: "Variable", inplace: bool = False) -> Optional["Variable"]:
-        return copy_metadata(to_variable=self, from_variable=from_variable, inplace=inplace)
+    def copy(self, deep: bool = True) -> "Variable":
```
pabloarosado, line 306: Why are you removing `copy_metadata`? It is useful to have both options at the variable level: `copy` to copy the data (and optionally the metadata), and `copy_metadata` (to copy only the metadata). The latter is useful, for example, after doing an operation that does not propagate metadata, e.g. `tb1["b"] = tb1.groupby("a").transform("sum").copy_metadata(tb1["a"])`

Marigold: My bad, I didn't mean to remove it.

pabloarosado (approved):

PR #1469, opened 2023-08-14
:bug: fix missing VariableMeta

Marigold:
* In some cases we created `_fields` as dict, but it should have been defaultdict
* replace `copy.deepcopy` by existing `.copy` methods


pabloarosado (approved):

PR #1471, opened 2023-08-14
:hammer: deprecate data_values table

Marigold:
Remove leftover use of `data_values` table from ETL. Should be merged after https://github.com/owid/owid-grapher/pull/1923.


pabloarosado (approved):
Looks good :-)

PR #1478, opened 2023-08-14
🐛 Fixing the share of deaths variables in IHME GBD

spoonerf:
- In response to the recent issue with GBD prevalence percentages being the % of total cases rather than % of the population, I added a function to convert the percent variables to be 1000x the rate per 100,000 (where the denominator is population). 
- However, this doesn't make sense for the `gbd_child_mortality`, `gbd_risk` or `gbd_cause` datasets, in these cases, the denominator was DALYs and deaths, which is what we want. 
- Fix is just only using the `add_share_population()` on the `gbd_prevalence` and `gbd_mental_health` datasets.


lucasrodes (approved):
Thanks for doing this!

PR #1481, opened 2023-08-15
Add farmed fish data

pabloarosado:
Add data from Fishcount on the number of farmed animals for 2015, 2016 and 2017 (unfortunately they come from different snapshot files).
NOTE: I may need to make small improvements on the metadata in the coming days, but for now it should be the minimum metadata needed.

I also made a small modification of `create_dataset`, so that, if no `default_metadata` is ingested, the new dataset will have the title and description of the included table. And, if there are multiple tables, a title/description will be used only if there aren't multiple titles/descriptions among the tables.


etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.py
```diff
@@ -0,0 +1,117 @@
...
+            countries_that_must_have_data=[],
+            # Here we allow aggregating even when there are few countries informed.
+            # However, if absolutely all countries have nan, we want the aggregate to be nan, not zero.
+            frac_allowed_nans_per_year=0.999,
+            num_allowed_nans_per_year=None,
+        )
+    return tb
+
+
+def run_sanity_checks_on_outputs(tb: Table) -> None:
```
paarriagadap, line 59: Maybe a check like this should be added to geo, or inside the `add_region_aggregates` function. We would need to specify whihc regions form the world, though.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
```diff
@@ -0,0 +1,36 @@
+definitions:
+  common: |
+    This study estimates the annual number of farmed fish slaughtered for food using data on aquaculture production tonnages (from the Food and Agriculture Organization of the United Nations) and mean weights estimated from available fish size data (from a range of internet sources).
+
+    Note that this estimate does not include:
```
paarriagadap, line 5: Should "this study" be better?

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
```diff
@@ -0,0 +1,36 @@
...
+    - Non fish species, e.g. lobsters.
```
paarriagadap, line 6: `Non-fish`?

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
```diff
@@ -0,0 +1,36 @@
...
+
+          {definitions.common}
+      n_farmed_fish_high:
+        title: Estimated number of farmed fish (upper bound)
+        unit: 'fish'
+        short_unit: ''
+        description: |
+          Upper bound estimate of the number of farmed fish.
+
+          {definitions.common}
```
paarriagadap, line 36: Remove the line below this, as it gets reflected on the metadata

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
```diff
@@ -0,0 +1,36 @@
+definitions:
+  common: |
```
paarriagadap, line 2: I would add `|-` here, because using it remove an additional line at the bottom of each description. There's no indentation, so it wouldn't affect anything else, I think

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
paarriagadap: Also, I don't see this filtered out in the dataset. (column `Fish species?`)

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
```diff
@@ -0,0 +1,36 @@
...
+    This study estimates the annual number of farmed fish slaughtered for food using data on aquaculture production tonnages (from the Food and Agriculture Organization of the United Nations) and mean weights estimated from available fish size data (from a range of internet sources).
+
+    Note that this estimate does not include:
+    - Non fish species, e.g. lobsters.
+    - Fish deaths arising in fish farms prior to harvest, in the capture of wild fish for feed or re-stocking, or from environmental impacts on surrounding wildlife.
+    - Fish farmed for use as bait, whose small production tonnages may comprise large numbers of fish due to their very small size of just a few grams.
+    - Species without estimated mean weight (which could be an additional 17% of total fish production).
```
paarriagadap, line 9: I also don't see this filtered out. Please let me know if I'm wrong

paarriagadap (approved):
Looks good! I added some minor comments and questions

paarriagadap (commented):
Looks good! I added some minor comments and questions

paarriagadap (approved):

snapshots/animal_welfare/2023-08-14/number_of_farmed_fish_2017.xlsx.dvc
```diff
@@ -0,0 +1,21 @@
...
+      Estimated number of individual fish in annual aquaculture production.
+    producer: Fishcount
+    citation_producer: Fishcount.org.uk based on data from the Food and Agriculture
+      Organization of the United Nations.
+    dataset_url_main: http://fishcount.org.uk/
+    dataset_url_download: http://fishcount.org.uk/studydatascreens2/2017/Farmed-fishes-2017-v2.xlsx
+    date_published: '2018'
+    date_accessed: 2023-08-14
+  license:
+    name: CC BY
```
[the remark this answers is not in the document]
pabloarosado: I couldn't find any mention of license anywhere on the website. I would assume it's public domain, but to be conservative I used CC BY. Do you have a better suggestion?

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.py
```diff
@@ -0,0 +1,117 @@
...
+    "Oceania",
+    "Low-income countries",
+    "Upper-middle-income countries",
+    "Lower-middle-income countries",
+    "High-income countries",
+]
+
+
+def add_region_aggregates(tb: Table, ds_regions: Dataset, ds_income_groups: Dataset) -> Table:
+    tb = tb.copy()
```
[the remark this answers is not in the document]
pabloarosado: This is something I commonly do. It's not a great solution, but it ensures that, whatever you do to `tb` within the function, will not affect the original `tb`. Sometimes, while debugging or doing some analysis, you may do something like:
```python
tb = Table(...)
test = add_region_aggregates()
...
```
Maybe you now want to compare `tb` and `test`. But if you are not careful, the function `add_region_aggregates` may have modified the original `tb` object. In other words, the function has unintended side effects.
One way to solve that is to be very careful while creating the function, ensuring that there are no side effects. Or another is to simply make a copy of the input object, `tb`, so that whatever happens inside the function stays inside the function. For big tables, this is a very inefficient solution (since it creates a copy of the input table). But for small tables, it's a simple trick.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.py
pabloarosado: Yes, that could be an optional check. It may not always be fulfilled (because of rounding errors, or if you use an aggregate that is not `sum`). But it most cases it would be useful.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.py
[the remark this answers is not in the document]
paarriagadap: Ah, now I see, thanks. It will depend on the name I assign when I call the function. I assumed it would always be tb

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: If you select `Fish species? = N` you'll see that none of those rows have an estimate on mean weights or numbers. I'm now adding data for decapod crustaceans (in a separate PR).

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: Given your next point, I've realised that it's better to keep using "estimate". The study (or the dataset) does actually include non-fish species. However, estimates on the number of fish is not included for non-fish species.

snapshots/animal_welfare/2023-08-14/number_of_farmed_fish_2017.xlsx.dvc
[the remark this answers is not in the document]
paarriagadap: I also struggle to find any mention of these on some of my datasets. I would prefer to leave it empty if we don't have more info

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.py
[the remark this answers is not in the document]
pabloarosado: If at the beginning you do `tb_new = tb` that may still modify the original `tb`. If instead you make a copy, it doesn't matter how you call the new object. For simplicity, I just call it the same thing.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: Hmm, I've just noticed that the three indicators have an additional line at the end, not just this one. I don't know how to remove it, but it doesn't bother me.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: Ah right, that solves the previous issue about the additional line :-)

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: Ah, ok, this is fixed by using `|-` in the common definition, as you suggest below.

snapshots/animal_welfare/2023-08-14/number_of_farmed_fish_2017.xlsx.dvc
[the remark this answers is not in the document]
pabloarosado: I think in such circumstances (or in general) we should contact the data providers. Both to check on licenses, but also to let them know that we are using their data. I'll contact Mood and Brooke (who seem to be the authors of this dataset) once this is published.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: Species without estimated mean weight, by construction, will not have an estimate on the number of individual fish, as far as I understand. At the top of the spreadsheet they take the sum of all countries' numbers of fish, and add a 17% to account for species without estimated mean weight. That's why I mention it here. But I'm not sure about this, and the documentation is a bit spread around the web site and some papers.

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
paarriagadap: I think there is still one additional line, but in the case of the last indicator there was another one after doing this change.

pabloarosado: Where do you see the additional line? I don't see it either in the grapher step's variable descriptions or in [the sources tab](http://staging-site-prosado/admin/datasets/6174).

paarriagadap: Ah, sorry, forget about it. I saw it when I pushed it to my staging server, because I saw an additional line at the bottom of this file. I see it's not there now

PR #1491, opened 2023-08-16
:hammer: merge functions copy_metadata and copy_metadata_from

Marigold:
* merge both functions into `copy_metadata`
* remove unused arguments from `copy_metadata` to keep it simple
* make shallow copy by default for performance reasons (shouldn't affect any existing use)


pabloarosado (approved):
Thanks, it looks good! Are you sure we don't want to have the option to do `copy_metadata(inplace=True)`? I don't have a strong opinion, but most likely we don't need it.

PR #1492, opened 2023-08-16
📊 animal_welfare: add data on wild fish killed for food

pabloarosado:
Add snapshots, meadow, garden and grapher steps for the dataset on the number of wild fish killed for food


etl/steps/data/garden/animal_welfare/2023-08-16/number_of_wild_fish_killed_for_food.meta.yml
```diff
@@ -0,0 +1,38 @@
...
+        unit: 'fish'
+        short_unit: ''
+        description: |
+          Lower bound estimate of the number of wild-caught fish killed for food.
+
+          {definitions.common}
+        display:
+          numDecimalPlaces: 0
+      n_wild_fish:
+        title: Estimated number of wild-caught fish
```
spoonerf, line 20: I am not sure we should calculate this variable - I think the data producers have purposely left a lot of uncertainty in the data because of the difficulty in estimating such a variable. I think producing a chart or map with this variable would give a false sense of accuracy.

spoonerf (commented):
Code looks good. I have some reservations about whether we should be estimating the mid-point when there is so much uncertainty in the data itself. I think the upper and lower bounds are meaningful enough to get the point across without adding a variable that may end up being misunderstood.

snapshots/animal_welfare/2023-08-16/number_of_wild_fish_killed_for_food.xlsx.dvc
```diff
@@ -0,0 +1,22 @@
+meta:
+  origin:
+    dataset_title_owid: Number of individual wild fish killed for food
+    dataset_title_producer: |-
+      Estimated numbers of individuals in annual global recorded capture (FAO) of fish species
+    dataset_description_owid: |-
+      Estimated numbers of individual fish killed annually in wild fish capture.
+    producer: Fishcount
```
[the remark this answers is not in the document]
pabloarosado: Yes, we have already noticed a few issues like this. That's mostly in Daniel's territory, so next week he'll probably look into that.

etl/steps/data/garden/animal_welfare/2023-08-16/number_of_wild_fish_killed_for_food.meta.yml
pabloarosado: True, I'm also unsure. But I created it for consistency with other similar datasets where we also have lower and upper limits.

PR #1497, opened 2023-08-17
📊 Updating the UN SDG dataset to the latest version

spoonerf:
I've tried to just highlight the relevant changes for this update, rather than all of the new code from the duplicated steps.


snapshots/un/2023-01-24/un_sdg.feather.dvc
```diff
@@ -19,6 +19,6 @@ meta:
 
 wdir: ../../../data/snapshots/un/2023-01-24
 outs:
-- md5: 4eec76d904280209ba5a635b21e28b54
-  size: 684540210
+- md5: cdb00e3b6f2244b497794c4f7c080fbc
```
pabloarosado, line 22: This seems to be an old snapshot, why did this one change?

pabloarosado (approved):
Looks good, thanks for showing only the changes, it makes the review much easier!

snapshots/un/2023-01-24/un_sdg.feather.dvc
spoonerf: Ah I accidentally had the version hardcoded in the snapshot script, so when running the snapshot script it was saving the old version. I will remove this dvc file from the PR.

PR #1498, opened 2023-08-17
✨ dag: add step to specific yaml

lucasrodes:
Add a dropdown to select the DAG file to include the Meadow, Garden, Explorers and Grapher steps.

dropdownyaml


walkthrough/garden.py
```diff
@@ -96,10 +97,10 @@ def app(run_checks: bool) -> None:
                 value=state.get("version", str(dt.date.today())),
                 help_text="Version of the meadow dataset (by default, the current date, or exceptionally the publication date).",
             ),
+            pi.select("Add to DAG", utils.ADD_DAG_OPTIONS, name="dag_file"),
```
Marigold, line 100: If you change it to
```
pi.select("Add to DAG", utils.ADD_DAG_OPTIONS, name="dag_file", value=state.get("dag_file")),
```
it will use the selected DAG from meadow step. Same for grapher and explorers.

Marigold (approved):
Nice one, thanks for doing this instead of me :).

PR #1505, opened 2023-08-17
📊  econ: create World Bank PIP data

paarriagadap:
owid-issue: https://github.com/owid/owid-issues/issues/1074

## Summary by CodeRabbit

- **New Features**
	- Enhanced data availability with new data steps for Luxembourg Income Study, World Inequality Database, and World Bank PIP explorers.
	- Improved data comparison with functionality to merge data from different explorers.
	- Implemented scripts for detailed percentile distributions and snapshot management of World Bank Poverty and Inequality Platform data.
	- Upgraded data processing capabilities for better insights and visualization.

- **Chores**
	- Established a new cache directory setup to improve data handling efficiency.


snapshots/wb/2023-08-14/pip_api.py
```diff
@@ -315,74 +342,186 @@ def get_region_data(povline, ppp_version, versions):
...
+            df_query_region = pd.read_csv(file_path_region)
+        else:
+            # Run the main function to get the data
+            log.warning(f"We need to come back to the extraction! regions, {povline}, {ppp_version} PPPs)")
+            get_percentiles_data_region(povline, versions, ppp_version)
+            df_query_region = pd.read_csv(file_path_region)
+
+        return df_query_region
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
```
paarriagadap, line 405: Hi @Marigold! I turned this file concatenation into a concurrent process, but I see there's not much difference compared to a normal loop. Is it because the process is quick on each step and there's not much to optimize?

Marigold: The issue here is that you **only run** `get_query_country` and `get_query_region` in parallel. The for-loops are still run synchronously because you're calling `.result()` inside them. You can try changing it to this:

```python
futures = []
with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
    for ppp_version in [2011, 2017]:
       for povline in povlines:
           # start the process, but don't call `.result()` which would wait for its completion
           future_country = executor.submit(get_query_country, povline, ppp_version)
           futures.append(future_country)
           
# now that all futures have been started, wait for them all
dfs = [f.result() for f in futures]
df_country = pd.concat(dfs, ignore_index=True)
```

If CPU is the bottleneck, you could also try ProcessPoolExecutor that parallelizes it with cores and not threads.

paarriagadap: That was it; now it does it in no time. Thanks!

snapshots/wb/2023-08-14/pip_api.py
```diff
@@ -501,11 +501,14 @@ def generate_consolidated_percentiles(df):
     percentiles = range(1, 100, 1)
     df_percentiles = pd.DataFrame()
 
-    # Estimate percentiles using multiprocessing
-    with concurrent.futures.ProcessPoolExecutor() as executor:
-        futures = [executor.submit(calculate_percentile, p, df) for p in percentiles]
+    # Estimate percentiles using concurrency
+    futures = []
+    with concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
```
paarriagadap, line 506: Here I applied the same principle, @Marigold, but the `calculate_percentile` function takes a lot. I tried replacing with `ProcessPoolExecutor` and the log went everywhere, to previous steps. How can this be even faster?

snapshots/wb/2023-08-14/pip_api.py
```diff
@@ -469,7 +469,7 @@ def calculate_percentile(p, df):
     df["distance_to_p"] = abs(df["headcount"] * 100 - p)
     df_closest = (
         df.sort_values("distance_to_p")
-        .groupby(["ppp_version", "country", "year", "reporting_level", "welfare_type"], as_index=False)
+        .groupby(["ppp_version", "country", "year", "reporting_level", "welfare_type"], as_index=False, sort=False)
```
paarriagadap, line 472: @Marigold This change was recommended by Copilot, to add `sort=False` for it to be faster

snapshots/wb/2023-08-14/pip_api.py
paarriagadap: For now I have set `max_workers` to 1, because it's too much RAM for the machine. That was probably the reason for this not working properly.

Marigold: Yeah, memory could be a problem. You can use smaller types in your dataframe. Either you do it automatically with
```
from owid.repack import repack_frame

def generate_consolidated_percentiles(df):
  ...
  df = repack_frame(df)
  ...
  # Estimate percentiles using concurrency
  ...
```
or you manually convert all columns `"ppp_version", "country", "year", "reporting_level", "welfare_type"` to categoricals with
```
df = df.astype({
  "ppp_version": "category",
  ...
})
```
Repacking / converting to categoricals should also speed up the groupby which probably takes the most time.

I can also try running the script myself and check for potential improvements. Should I run it on staging or download the files from staging and then run locally?

paarriagadap: Thanks @Marigold. Actually, waiting some hours for this is not a big deal, given the duration of the rest of the execution.

With my modifications to the code, I recommend you not to run it, because additional downloads from the API are necessary, so PIP wants to do that on their side. I will apply these changes and see if I need to change the code further afterward.

paarriagadap:
Don't close it 🙏🏼

snapshots/wb/2023-08-14/pip_api.py
```diff
@@ -0,0 +1,1093 @@
...
+import time
+from multiprocessing.pool import ThreadPool
+from pathlib import Path
+
+import click
+import numpy as np
+import pandas as pd
+import requests
+from botocore.exceptions import ClientError
+from joblib import Memory
```
paarriagadap, line 11: @Marigold By the way, we need to add joblib to `pyproject.toml`

Let me know all the things I need to set up after you have your tests ready!

Marigold: Oh, right, thanks! Added that. There's nothing else you have to setup, just run `poetry install` and then you should be able to run the script with `python snapshots/wb/2023-08-14/pip_api.py`.

etl/steps/data/garden/wb/2023-08-14/shared.py
```diff
@@ -0,0 +1,764 @@
...
+    if var in ["mean", "median"]:
+        meta = VariableMeta(
+            title=f"{var_dict[var]['title']} {inc_cons_dict[welfare_type]['name']}",
+            description_short=var_dict[var]["description"],
+            description_key=[
+                ppp_description,
+                inc_cons_dict[welfare_type]["description"],
+                non_market_income_description,
+            ],
+            description_processing=f"""
```
paarriagadap, line 536: @Marigold Hi! Do you know the reason why `description_processing` is not being saved in here or in any od the other functions below? All the others are saved correcly. I even changed it for a `"Hello"` here and still nothing.

Marigold: It works for me. Maybe you could try changing this line to "Hello" and then description_processing under `else:` branch below to "World" and see what you get.

paarriagadap: Right. I saw it dissapears because I do further processing to the table, extending it (I didn't know the ETL does that). But there's still a weird behavior: it's saving the wrong `description_processing` for the tables `income_2011` or `income_2017`. At least in my machine it's showing
```
To construct a global dataset, the World Bank combines estimates based on income data and estimates based on consumption data. Here we only include the estimates based on CONSUMPTION data.

You can find the data with all available income and consumption data points in our complete dataset of the World Bank PIP data.
```

When instead of CONSUMPTION it should show INCOME. Do you have it as well?

Marigold: I think it must be something with copying tables. Somewhere you get a subset (slice) of the table and don't call `.copy()` and then both tables have the same metadata object which overwrites itself.

[This place is suspicious](https://github.com/owid/etl/pull/1505/files#diff-a7cc65423e4ef280f1bfe78bffd3143180ca2061a17375fa2ab3d5e44150cddaR686). You should change it to this
```
tb_inc = tb[tb["welfare_type"] == "income"].reset_index(drop=True).copy()
tb_cons = tb[tb["welfare_type"] == "consumption"].reset_index(drop=True).copy()
```
it's possible there are other places.

paarriagadap: That was the one! I corrected it in some other places. So the recommended action to take is always to copy the table in a function or only when I am creating a different table?

Marigold: Great, I'm glad it worked :). You need it only when you create a different table.

paarriagadap:
Hi @Marigold! I wanted to rebase this but I didn't know what to keep in the case of the `poetry.lock` file, since it says `# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.`

Marigold:
@paarriagadap yeah, that one is a bit tricky. When rebasing, it doesn't really matter what code do you keep, but after you rebase, you have to run `poetry lock --no-update` which will automatically update `poetry.lock` to a good state. Ping me on slack if it doesn't work for any reason.

etl/steps/data/garden/wb/2024-01-17/shared.py
```diff
@@ -0,0 +1,855 @@
...
+            short_unit=var_dict[var]["short_unit"],
+            origins=origins,
+        )
+        meta.display = {
+            "name": meta.title,
+            "numDecimalPlaces": var_dict[var]["numDecimalPlaces"],
+            "tolerance": TOLERANCE,
+        }
+
+        meta.presentation = {
```
Marigold, line 541: Nit - to fix the linting error, you should use proper object, i.e.
```
from owid.catalog import VariablePresentationMeta
...
meta.presentation = VariablePresentationMeta(title_public=meta.title)
```

etl/steps/data/garden/wb/2024-01-17/world_bank_pip.py
```diff
@@ -0,0 +1,1047 @@
...
+    tb.columns = ["_".join(col).strip() for col in tb.columns.values]
+
+    # Reset index
+    tb = tb.reset_index()
+
+    return tb
+
+
+def create_stacked_variables(tb: Table, povlines_dict: dict, ppp_version: int) -> tuple([Table, list, list]):
+    """
```
Marigold, line 349: The output type should be `Tuple[Table, list, list]` (use `from typing import Tuple`).

etl/steps/data/garden/wb/2024-01-17/world_bank_pip.py
```diff
@@ -0,0 +1,1047 @@
...
+def identify_rural_urban(tb: Table) -> Table:
+    """
+    Amend the entity to reflect if data refers to urban or rural only
+    """
+
+    # Make country and reporting_level columns into strings
+    tb["country"] = tb["country"].astype(str)
+    tb["reporting_level"] = tb["reporting_level"].astype(str)
+
+    tb.loc[(tb["reporting_level"].isin(["urban", "rural"])), "country"] = (
```
Marigold, line 486: Nit - when I use an index multiple times in an expression, I tend to DRY it with
```
ix = tb["reporting_level"].isin(["urban", "rural"])
tb.loc[ix, 'country'] = tb.loc[ix, 'country'] + " (" + tb.loc[ix, 'reporting_level'] + ")"
```
It's a bit more readable

etl/steps/data/garden/wb/2024-01-17/world_bank_pip.py
```diff
@@ -0,0 +1,1047 @@
...
+
+    if not tb_error.empty:
+        log.warning(
+            f"""There are {len(tb_error)} observations with missing poverty values and will be deleted:
+            {tabulate(tb_error[['country', 'year', 'reporting_level', 'welfare_type'] + col_headcount], headers = 'keys', tablefmt = TABLEFMT)}"""
+        )
+        tb = tb[~mask].reset_index(drop=True)
+
+    ############################
+    # Missing median
```
Marigold, line 589: Nit - these functions are pretty cool and could be reused in the future. We could have functions like `check_missing_values(tb, col)` or `check_sum_to_100(tb, col)` in `shared.py` module (and eventually move them to `etl/checks.py` if we want to reuse them elsewhere). No need to do it now, you can do it when you realize you could use it elsewhere.

snapshots/wb/2024-01-17/pip_api.py
```diff
@@ -0,0 +1,1568 @@
...
+
+
+@retry(wait=wait_random_exponential(multiplier=1), stop=stop_after_attempt(MAX_REPEATS))
+def _get_request(url: str) -> requests.Response:
+    response = requests.get(url, timeout=TIMEOUT)
+    if response.status_code != 200:
+        log.info("fetch_csv.retry", url=url)
+        raise Exception("API timed out")
+
+    response = requests.get(url, timeout=TIMEOUT)
```
Marigold, line 240: Coderabbit was right! There's identical code, we can remove one block (it was probably me when I was working on this).

Marigold (approved):
Thanks for this MASSIVE effort. I ran it locally and went through the code. Code is pretty fast and looks good! I had a couple of nits, but nothing preventing this from merging. I haven't checked metadata or data very thoroughly (the easiest way to do this would be to visually check charts).

(There are a couple of errors from `make check`. Let me know if you need help fixing them.)

etl/steps/data/garden/wb/2024-01-17/shared.py
paarriagadap: Thanks! By the way, I don't have that linting error

Marigold: Oh weird, could you check if you have `Pylance` extension enabled in VSCode? (It's not obligatory, but it's pretty useful)

paarriagadap: It's enabled. Should I check a specific place in settings as well?

PR #1512, opened 2023-08-18
:sparkles: improve grapher step performance

Marigold:
Speed up grapher upsert by about ~20%.


pabloarosado (approved):
Amazing! Was the main bottleneck the use of `pd.read_sql`? Looks good.

Marigold:
> Amazing! Was the main bottleneck the use of `pd.read_sql`? Looks good.

There's some overhead when using `pd.read_sql` - it needs to get a new connection, and it creates a dataframe. This can be a bottleneck for datasets with tons of variables.

PR #1514, opened 2023-08-21
📊  war: define selected entities by default

lucasrodes:
Different indicators have data for different regions. We use the field `presentation.grapher_config.selectedEntityNames` to define the default entities selected for each indicator.

### DIFF
--- a/etl/steps/data/garden/war/2023-06-22/ucdp.meta.yml
+++ b/etl/steps/data/garden/war/2023-06-22/ucdp.meta.yml
@@ -45,6 +45,14 @@ tables:
           Active conflicts are those resulting into at least 25 deaths each year (based on most plausible estimate).
         display:
           numDecimalPlaces: 0
+        presentation:
+          grapher_config:
+            selectedEntityNames:
+              - Africa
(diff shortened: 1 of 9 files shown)


paarriagadap:
Hi @lucasrodes! I was wondering what's the best way to check if the default entities are working, because I pushed some datasets to my staging server and while I see the non-World entities selected by default in the indicators page I don't see them when I create a chart. I updated my server in case there's anything extra to update there and should be operational again soon.

Also, I was curious to know how you define a couple of indicators in the yaml file and then there is metadata for all the variants in the dataset. Just because of the propagation integrated with the functions?

lucasrodes:
@paarriagadap, thanks for your comments.

I just tested it and it seems that my charts do have the entities selected by default. If you don't mind, I'd push it to master and see what happens there.

Regarding the handling of metadata, I am not sure I understand what you mean. I usually have a dimension "conflict_type" or "hostility_level", which breaks down the indicators. Hence, in the description of the indicators I explain this breakdown (e.g. "this metric is broken down by conflict type ...")

PR #1517, opened 2023-08-21
:bug: don't delete dataset with the same prefix

Marigold:
Running `publish ...` doesn't handle prefixes properly. We have two datasets `excess_mortality` and `excess_mortality_economist` and this bug was deleting `excess_mortality_economist`.


lucasrodes (approved):
LGTM! Thanks for the quick fix

PR #1521, opened 2023-08-22
Create dummy full dataset

pabloarosado:
I let metaplay now use either `dummy` or `dummy_full`. The latter is a new dataset that includes (almost) all metadata fields with their names (when possible). I've also tried to express the fallback options, e.g.
```yaml
title: "[grapher-config.title > indicator.title]"
```
I'm not sure if all of them are true (in fact I know that `subtitle: "[grapher-config.subtitle > indicator.description-short]"` is not true, but I think it should be), but precisely for this reason I created this dataset, to be able to play around with the metadata fields and see how the resulting grapher chart/data page changes.

### DIFF
--- a/apps/metadata_playground/app.py
+++ b/apps/metadata_playground/app.py
@@ -17,28 +17,28 @@ from etl import config
 # Initial configuration ###########################
 ###################################################
 # Set page config
-# st.title("Metadata v2 preview")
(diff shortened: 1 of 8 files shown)


lucasrodes (approved):
LGTM ー  Thanks, @pabloarosado!

Issue #1451, opened 2023-08-10
Population dataset has multiple origins which take much space in per capita charts


pabloarosado:
# Description
Many of our etl steps depend on our population dataset, which is used, for example, when creating per capita indicators. Since we are moving into a workflow where origins are propagated automatically, all these per capita indicators end up having many origins. Then, all charts that use these indicators end up having too many origins shown (mostly irrelevant).
Additionally, our population dataset is used in 513 charts, of which only a few (e.g. [Population](https://owid.cloud/admin/charts/4546/edit)) are actually about population. The rest use population "indirectly", e.g. in scatter charts.

In all those cases where population origins have little importance, it would be good to either hide those origins, or collapse them into a generic one, e.g. "Population is based on various sources".

# Proposed solutions
To achieve this, we [discussed](https://www.notion.so/owid/2023-Data-architecture-8b5492227e6d4911b913ac29de44a263?pvs=4#42769dcda3c249d39153051ebf53f117) various options:
1. Flagging all population origins as "auxiliary". This implies adding at least a new metadata field to origins (and possibly other to grapher configs and presentation or display). This way, in data pages or charts, we could decide to hide or show auxiliary origins. But we may want to cite those origins somehow, with a "Population is based on various sources", which we would need to do manually.
2. Adding a special kind of replacement origin to the population dataset. Then in data pages or charts we could choose whether to show all origins, or only the replacement ones. This implies adding possibly a few new metadata fields. It also needs some further logic. For example, if you have an indicator that uses population, then you have multiple origins including the replacement one. But then it's unclear how to identify the origins that need to be hidden.
3. Creating a new population dataset that has only one origin, combination of all others. This dataset would depend on the original population dataset. All other steps would depend on the new combined population dataset. This implies having more steps with duplicated data, and refactoring all charts so that they depend on the new dataset. This dataset could be hosted within the same namespace `demography` or in a new namespace for combined origins, e.g. `owid`.
4. Creating a new table in the current population garden step, with the same data as the existing one, but with one combined origin. We could call this new table `population_combined_origins`. However, for simplicity, we can call the new table `population`, and rename the existing table to, e.g. `population_original`. This way, all steps that currently depend on `population` would not need to be edited. And, for those steps that have automatic metadata propagation, all unnecessary origins (e.g. Gapminder, Hyde, UN) will be hidden from charts (and shown in the sources tab). This seems to be the better option.

# Chosen solution

[...]

PR #1528, opened 2023-08-24
:bar_chart: Combined origin in population table

Marigold:
Implementation of https://github.com/owid/etl/issues/1451

* Have two tables `population_original` and `population` with the same data, but collapsed origin
* Use `population_original` in grapher population and garden population density (population density isn't reused by other datasets, so expanded origins in garden step are ok)
* [Collapsed Origin](https://github.com/owid/etl/compare/refactor-population...1451-population-dataset-condensed?expand=1#diff-f9d2ff0d8e86c9309ad0676931c890a21a88672ba19cd51ae925c30480b55d3cR1) could be improved for sure


etl/steps/data/garden/demography/2023-03-31/population/population.meta.yml
```diff
@@ -1,3 +1,9 @@
+collapsed_origin:
```
pabloarosado: `collapsed_origin` sound like a Christopher Nolan movie. Shall we call it `origin_combined`? If it's annoying, leave it like that.

pabloarosado (approved):
Thanks, it looks good! Before merging, let me check how it looks on different cases, to see if we should tweak the new metadata a bit. Otherwise, we can merge it and tweak it afterwards (I suppose the former option is better, given that it will trigger a huge ETL job).

Marigold:
> Thanks, it looks good! Before merging, let me check how it looks on different cases, to see if we should tweak the new metadata a bit. Otherwise, we can merge it and tweak it afterwards (I suppose the former option is better, given that it will trigger a huge ETL job).

Agree. Once it is good to go, we should merge this PR into #1536 and then that one to trigger the rebuild.

Marigold:
@pabloarosado I rebased again and updated all YAML files with the new metadata. Could you take a look at the [combined population origin metadata](https://github.com/owid/etl/pull/1528/files#diff-f9d2ff0d8e86c9309ad0676931c890a21a88672ba19cd51ae925c30480b55d3cR108) and let me know if it can be improved?

Marigold:
There's an issue when combining sources and origins (e.g. [this chart](http://staging-site-mojmir/admin/charts/4466/edit)) - only origin's full citation in `Data published by` and sources are missing. Below is a version with only sources
image
and [here is](http://staging-site-mojmir/admin/charts/4466/edit) version with sources and origins
image

Marigold:
@pabloarosado @lucasrodes here are charts using new population

## Population charts
http://staging-site-mojmir/grapher/population
http://staging-site-mojmir/grapher/population-density

## Charts using combined population (20 random)
http://staging-site-mojmir/admin/charts/4556/edit
http://staging-site-mojmir/admin/charts/4381/edit
http://staging-site-mojmir/admin/charts/4451/edit
http://staging-site-mojmir/admin/charts/6935/edit
http://staging-site-mojmir/admin/charts/6271/edit
http://staging-site-mojmir/admin/charts/4254/edit
http://staging-site-mojmir/admin/charts/4487/edit
http://staging-site-mojmir/admin/charts/4983/edit
http://staging-site-mojmir/admin/charts/486/edit
http://staging-site-mojmir/admin/charts/483/edit
http://staging-site-mojmir/admin/charts/4460/edit
http://staging-site-mojmir/admin/charts/4639/edit
http://staging-site-mojmir/admin/charts/4466/edit
http://staging-site-mojmir/admin/charts/6364/edit
http://staging-site-mojmir/admin/charts/4372/edit
http://staging-site-mojmir/admin/charts/4465/edit
http://staging-site-mojmir/admin/charts/3188/edit
http://staging-site-mojmir/admin/charts/4370/edit
http://staging-site-mojmir/admin/charts/5161/edit
http://staging-site-mojmir/admin/charts/4458/edit

There's a problem with missing **Data published by** for sources mentioned [above](https://github.com/owid/etl/pull/1528#issuecomment-1733741544). Do you know if this is a known issue? I could try fixing it in grapher and show both origins and sources.

pabloarosado:
Hi @Marigold thanks for making the list. I have checked all of them and have the following comments:
* In all charts using per capita variables, I think the source reads well (sometimes you have to go to the "text" tab -> "Footer"  -> "Source" and click on the link button, to see the automatically generated source. Once you do that, you see the added "Various sources - Population (2023)". I will ask on slack if people agree with this way of quoting the population dataset, and whether we should always show it.
* I see that, when there are many sources, the footer automatically changes into "Multiple sources", which I think is a good approach (and probably with the new grapher redesign, this will look even better). This happens for example on [this chart](http://staging-site-mojmir/admin/charts/2028/edit). Although here maybe we have room for all sources, given that they are rather short. Maybe (**TODO in grapher**) we could make the logic a bit smarter, so that you get "Multiple sources" only if there's more than, say, 2 sources **and** we exceed a number of characters.
* It would be better if the "Various sources - Population (2023)" was always the last source listed. But I don't see an easy way to achieve this (do you?), so it's not a big deal.
* As you mentioned, the full citation is not properly combining sources and origins (**TODO in grapher**).
* I see that on [this chart](http://staging-site-mojmir/admin/charts/6935/edit), we see both the "Various sources - Population (2023)" and the individual population sources. This may be the only case where something seems to be wrong.

So, all in all, apart from the last small issue, it looks good!

pabloarosado:
One additional point, @Marigold: As I mentioned on Slack, I'm not sure if scatter plots should also cite "Various sources - Population (2023)". Currently, I see that population is not cited anywhere on scatter plots (e.g. [this chart](http://staging-site-mojmir/admin/charts/4909/edit)). But I think we should probably cite population (no strong opinion). I suppose this is a **TODO in grapher** and may not affect this PR. What do you think?

pabloarosado:
One minor change suggested by Ed (after asking Hannah and Max). On the `2023-03-31/population/population.meta.yml` file:
```
    attribution: Population based on various sources (2023)
```

pabloarosado:
Hi @Marigold, the grapher issue of not combining sources and origins is probably affecting other charts. But after merging this PR, certainly hundreds of charts will be affected. So it would be good fix the grapher issue first. However, if you think it's easy to fix, we could merge this PR and fix the grapher issue soon after. I leave it up to you.

PR #1529, opened 2023-08-24
:hammer: speed up ETL in DEBUG mode

Marigold:
* When running `DEBUG=1 etl ...` don't fork new processes, but keep it all in the same env. That subtracts about 2s from each ETL step
* Run `etl-metaplay` in DEBUG mode (rendering goes down from ~30s to ~5s)
* Check `apps` for linting errors

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 
 include default.mk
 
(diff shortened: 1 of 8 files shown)


lucasrodes (approved):
LGTM! Fantastic, thanks @Marigold

PR #1539, opened 2023-08-25
✨ metadata: schema, walkthrough

lucasrodes:
- Fill the schema for Snapshot metadata with descriptions, examples, etc.
  - Not all has been written, as there are still some discussions going on (see [notion](https://www.notion.so/owid/Metadata-guidelines-29ca6e19b6f1409ea6826a88dbb18bcc)).
- Work on a v2 for `walktrhough snapshot` that relies on (i) the info from the schema and (ii) `streamlit`

**NOTE:** There is no logic yet implemented.

Run it with:

```
poetry run streamlit run apps/walkthrough/snapshot_v2.p
```


Marigold:
Cool work! Couple of thoughts (might add more later)

1. Do we have the appetite to rewrite walkthrough with streamlit? If yes, it'd be good to verify that these things work
  - passing state between channels & "go to next channel" buttons
  - form inputs validation
2. While you're at it, you can finally rename it to **wizard**

Marigold:
@lucasrodes heads up that there have been [some renames](https://github.com/owid/etl/pull/1554).

pabloarosado:
Awesome work!
Probably a small bug: At the meadow step, if I untick the option to generate a playground and submit I get:
```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/prosado/Documents/owid/repos/etl/etl/steps/data/meadow/test/2023-08-30/playground.ipynb'
```
The error does not occur if I leave the "Generate playground notebook" option selected (by the way, in my opinion, it should be unselected by default).

I remember there was a similar issue with walkthrough, but I thought it was fixed.

PR #1540, opened 2023-08-25
📊 Education datasets: adding weighted mean aggregation custom function + small edits to combined enrollment data

veronikasamborska1994:
hi @Marigold ! could you take a look at the etl/data_helpers/geo.py file? I am adding a couple of lines of code which will compute weighted population means if aggegation parameter is set to  _mean_. I need this for computing regional aggregates in my education dataset but this might end up being useful for others too.


etl/data_helpers/geo.py
```diff
@@ -323,17 +323,36 @@ def add_region_aggregates(
...
-            frac_allowed_nans=frac_allowed_nans_per_year,
-        ).reset_index()
+        # If aggreggate is mean then do weighted average using population data, replacing `aggregations[variable]` with a lambda function
+        if aggregations[variable] == "mean":
+            df_added = groupby_agg(
+                df=df_countries,
+                groupby_columns=year_col,
+                aggregations={
+                    country_col: lambda x: set(countries_that_must_have_data).issubset(set(list(x))),
+                    variable: lambda x: np.ma.average(
```
Marigold, line 336: Nice, you could DRY it a bit like this
```python
if aggregations[variable] == "mean":
    variable_agg = lambda x: np.ma.average(
                    np.ma.masked_invalid(x.astype("float64")),
                        weights=np.ma.masked_invalid(df_countries.loc[x.index, "population"].astype("float64")),
                    )
else:
    variable_agg = aggregations[variable]
    
df_added = groupby_agg(
                df=df_countries,
                groupby_columns=year_col,
                aggregations={
                    country_col: lambda x: set(countries_that_must_have_data).issubset(set(list(x))),
                    variable: variable_agg
                },
                num_allowed_nans=num_allowed_nans_per_year,
                frac_allowed_nans=frac_allowed_nans_per_year,
            ).reset_index()
```

Marigold (approved):
Looks good!

etl/data_helpers/geo.py
```diff
@@ -323,13 +323,24 @@ def add_region_aggregates(
     df_region = pd.DataFrame({country_col: [], year_col: []}).astype(dtype={country_col: "object", year_col: "int"})
     # Select data for countries in the region.
     df_countries = df[df[country_col].isin(countries_in_region)]
+    # Add population for doing weighted average aggregations
+    df_countries = add_population_to_dataframe(df_countries)
```
Marigold, line 327: I was checking why tests are failing and adding population to a dataframe is the reason. It might be better to add extra argument `weight` to the function and call it like this

```python
def add_data_for_regions(tb: Table, regions: List[str], ds_regions: Dataset, ds_income_groups: Dataset) -> Table:
    tb_with_regions = tb.copy()
    aggregations = {column: "mean" for column in tb_with_regions.columns if column not in ["country", "year"]}
    tb_with_regions = add_population_to_dataframe(tb_with_regions)

    for region in REGIONS:
        # Find members of current region.
        members = geo.list_members_of_region(
            region=region,
            ds_regions=ds_regions,
            ds_income_groups=ds_income_groups,
        )
        tb_with_regions = geo.add_region_aggregates(
            df=tb_with_regions,
            region=region,
            countries_in_region=members,
            countries_that_must_have_data=[],
            num_allowed_nans_per_year=None,
            frac_allowed_nans_per_year=0.2,
            aggregations=aggregations,
            weights=tb_with_regions['population']
        )
```

This is more explicit and reusable than adding population.

Then the function becomes
```python
def weighted_mean(x, w):
    return np.ma.average(
        np.ma.masked_invalid(x),
        weights=np.ma.masked_invalid(w),
    )

if weights is not None:
    assert aggregations[variable] == "mean", "Weights only work mean aggregation"
    variable_agg = lambda x: weighted_mean(x, weights.loc[x.index])
```

etl/steps/data/garden/education/2023-07-17/education_lee_lee.py
```diff
@@ -29,7 +28,7 @@
 
 def add_data_for_regions(tb: Table, regions: List[str], ds_regions: Dataset, ds_income_groups: Dataset) -> Table:
```
Marigold, line 29: You're not using `regions` argument here. You might want to either remove it and use `REGIONS` or pass it to the function.

etl/data_helpers/geo.py
veronikasamborska1994: @Marigold  thanks a lot! it seems like my unit tests are still failing because of using a lambda function here, but more annoyingly I am not sure how to best resolve an issue where I can't get a World aggregate here, because population is added before regions that the World aggregate depends on. I actually remember having this issue before which is why I moved the population step addition inside the add_regions function. Do you by any chance have an idea for a quick fix for this? No worries if not ofc! I'll keep trying to fix it in the mean time

Marigold: Oh right, so if your previous approach worked well, perhaps it'd be best to copy `add_region_aggregates` to your step and rewrite it to calculate the weighted mean (while adding population).

(Another approach would be running `add_region_aggregates` first for `population * grade`, then for `population` and divide them to get the weighted mean.)

veronikasamborska1994: perfect, thanks! I think this is the easiest solution for now

Issue #1548, opened 2023-08-28
enhance: metadata for indicator + dimensions


lucasrodes:
It is often the case that we have indicators that have some dimensions. The way to code these dimensions is by adding them to the index of the dataset table. 

This approach, however, prevents us from adding (indicator + dimension) specific metadata.

### Sample example
Based on [this chart](https://ourworldindata.org/grapher/countries-democracies-autocracies-row) and this [slack discussion](https://owid.slack.com/archives/C05BLQ567H8/p1692894633056479).

A solution could be to change from long-format to wide-format to have metadata for each column.

_Wide-format_

```
| year | country | number_closedaut_row_owid | number_electaut_row_owid |
|------|---------|---------------------------|--------------------------|
| 1991 | World   |                           |                          |
| 1992 | World   |                           |                          |
| 1993 | World   |                           |                          |
```

Instead, I think it is cleaner to have **long-format**:

```
| year | country | regime_type | number_countries |
|------|---------|-------------|------------------|
| 1991 | World   | closedaut   |                  |
| 1991 | World   | electaut    |                  |
| 1992 | World   | closedaut   |                  |
| 1992 | World   | electaut    |                  |
| 1993 | World   | closedaut   |                  |
| 1993 | World   | electaut    |                  |
```

### Real world example
- [History of war (Gibler and Miller, 2020)](https://owid.cloud/admin/datasets/6192)
- URI: [`garden/war/2023-07-20/brecke`](https://github.com/owid/etl/tree/master/etl/steps/data/garden/war/2023-07-20)

### Challenge

Where and how do we define different descriptions for indicator `number_countries` depending on the dimension value `regime_type`?

PR #1551, opened 2023-08-28
:tada: enable Jinja2 templating in metadata

Marigold:
Fix for https://github.com/owid/etl/issues/1548

This PR adds support for Jinja2 templating (using `<%` tags instead of standard `{{` because it would collide with `dynamic_yaml` that we use for placeholders). This is especially helpful for long tables where we want to construct titles or descriptions on the fly and the combinations grow exponentially (see [these descriptions](https://docs.google.com/spreadsheets/d/1Z3fWwUJhafUBc75YY2KsO0Cj8VW3XpzwA9P8tHqDfPA/edit#gid=0)).

Dimensions with values are passed to the template and we can use ifs & variables from Jinja2 to get what we want without repetition.

It's a bit annoying to get whitespaces right in Jinja. I recommend setting `DEBUG=1` flag to speed up ad-hoc etl builds and then check back and forth with grapher.

@lucasrodes I've implemented it for some variables of one dataset. Do you think it would work as a solution? If yes, I can try it on all war datasets that use descriptions from the spreadsheet.

## TODO
* [x] Enable it for other fields like `title` etc.
* [ ] Add it to docs if this get approved
* [ ] ~Add unit tests~
* [ ] ~Add a helper function for rendering YAML that we can use from a notebook~

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -154,31 +154,49 @@ def _yield_wide_table(
                     }
                 }
 
+            dim_dict = dict(zip(dim_names, dim_values))
+
             # Add dimensions to title (which will be used as variable name in grapher)
-            title_with_dims: Optional[str]
             if tab[short_name].metadata.title:
-                title_with_dims = _title_column_and_dimensions(tab[short_name].metadata.title, dim_values, dim_titles)
+                # We use template as a title
+                if _uses_jinja(tab[short_name].metadata.title):
+                    title_with_dims = _expand_jinja_template(tab[short_name].metadata.title, dim_dict)
+                # Otherwise use default
+                else:
+                    title_with_dims = _title_column_and_dimensions(
+                        tab[short_name].metadata.title, dim_values, dim_titles
+                    )
+
                 tab[short_name].metadata.title = title_with_dims
-            else:
-                title_with_dims = None
 
             # expand metadata with Jinja template
-            if dim_values and "<%" in tab[short_name].metadata.description:
-                try:
-                    tab[short_name].metadata.description = jinja_env.from_string(
-                        tab[short_name].metadata.description
-                    ).render(dict(zip(dim_names, dim_values)))
-                except jinja2.exceptions.TemplateSyntaxError:
-                    log.warning(
-                        "yield_wide_table.jinja_syntax_error",
-                        column=column,
-                        dims=dim_values,
-                        description="\n" + tab[short_name].metadata.description,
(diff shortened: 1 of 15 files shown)


Marigold:
@lucasrodes finished adding all [technical texts](https://docs.google.com/spreadsheets/d/1kTswJvy_Fvazxv9UfOXPQBV_Jz8Lk-n7_vrC6yeTul4/edit#gid=0) we have. There might be some repetition and a lot could be potentially templated more, but that'd be time consuming. I suggest we DRY it only when we have to modify those texts again.

@bastianherre take a look at how we deal with dimensions and [use templates](https://github.com/owid/etl/pull/1551/files#diff-eef0f076e6f6f539f673cdbb43f366cecf4474e010602c17ccca74fceb841f88). You don't need to understand it now, but if we do something similar in the future, we might use those templates right away.

New texts have been uploaded to my [staging server](http://staging-site-mojmir/admin/datasets) if you want to randomly check a couple of them.

lucasrodes (approved):
Once again, thanks for doing this, Mojmir!

I think this is ready to go. I can iterate over this if changes are required!

lucasrodes:
As per adding this to our docs, and for some content, I am working on some docs improvements in this branch: 
- https://github.com/owid/etl/pull/1531

In that PR, I want to add all the documentation regarding the new metadata structure. It's on hold now, as we are still having final discussions.

I suggest creating a new markdown file focusing on "multi-dimensional tables/datasets". That could go under `docs/architecture/metadata`. We can later think if it should go elsewhere or combine it with the new entries in the branch mentioned above.

bastianherre:
Thanks for adding the new dataset and indicator descriptions, @Marigold!

The dataset descriptions — in addition to my curated descriptions — seem to include notes from the publishers directly. I am not sure we need them, and think they could even be confusing due to their rather technical nature. At the least, I would separate them more clearly from our curated descriptions.

Beyond that, I did some random checks on indicator descriptions. Most looked good to me, but some indicators still seem to have the old descriptions: see [here](http://staging-site-mojmir/admin/variables/783602), [here](http://staging-site-mojmir/admin/variables/783614), [here](http://staging-site-mojmir/admin/variables/791346), and [here](http://staging-site-mojmir/admin/variables/791345). May it take more time to reflect all changes?

Marigold:
Thank you Bastian, you were right, I forgot to upload those missing datasets. All should be good now. I've also removed the source descriptions.

Issue #1552, opened 2023-08-29
walkthrough: dag_path not correctly set in apps/walkthrough/garden.py


spoonerf:
### One-liner
The `dag_path` seems to be set incorrectly in [apps/walkthrough/garden.py](https://github.com/owid/etl/blob/master/apps/walkthrough/garden.py). I think perhaps `CURRENT_DIR` needs to be changed?

### Environment

- `python -V`: 3.11.1
- Operating system: Mac OS
- Working with local, staging or live: local

### Context & details
When running `walkthrough garden` and selecting an option for "Add to DAG" I get this error: 

![Screenshot 2023-08-29 at 09 32 22](https://github.com/owid/etl/assets/10774728/9e4e5aee-3cc4-4f58-a184-e345883802d7)

lucasrodes:
Thanks for raising this, and sorry! Fix in https://github.com/owid/etl/pull/1553

PR #1553, opened 2023-08-29
🐛 walkthrough: add to dag path

lucasrodes:
Fixes https://github.com/owid/etl/issues/1552


spoonerf (approved):
Works for me - thanks for the quick fix!

Issue #1550, opened 2023-08-28
Small metadata changes


pabloarosado:
After a discussion earlier today we decided that the following changes should be implemented (both in ETL and grapher):
* [ ] Move `presentation.key_info_text`-> `indicator.description_key`.
* [ ] The field `description_short` should populate chart subtitles by default (currently, no subtitle appears unless passed in grapher config).
* [ ] Move `presentation.processing_info` -> `indicator.description_processing`.
* [ ] (This should probably be a separate issue in owid-grapher) Show `unit` in the chart, somewhere.
* [ ] Currently, the automatic attribution format is `producer year`, but it should be `producer (year)`.
* [ ] Rename `version` -> `version_producer`.

PR #1554, opened 2023-08-29
:sparkles: small metadata changes

Marigold:
This implements the metadata updates tracked in #1550. 

* [x] Move `presentation.key_info_text`-> `indicator.description_key`.
* [x] Move `presentation.processing_info` -> `indicator.description_processing`.
* [x] Rename `version` -> `version_producer`.


etl/steps/data/garden/dummy/2020-01-01/dummy_full.meta.yml
```diff
@@ -9,6 +9,9 @@ tables:
         description_short: "[indicator.description-short]"
         description_from_producer: "[indicator.description-from-producer]"
         processing_level: "[indicator.processing-level]"
+        description_key:
+          - "[presentation.key-info-text example 1]"
```
pabloarosado, line 13: This should be:
```yaml
- "[indicator.description-key example 1]"
- "[indicator.description-key example 2]"
```

pabloarosado (approved):
Looks good! Just a minor change.

etl/steps/data/garden/dummy/2020-01-01/dummy_full.meta.yml
```diff
@@ -9,6 +9,9 @@ tables:
...
+          - "[presentation.key-info-text example 2]"
         license:
```
Marigold, line 15: @pabloarosado from what I remember, the license should preferably be set at the origin level. Do we want to also include it at the indicator level?

danyx23: Yes, the license field is the license that we give to the indicator. We want to get rid of `licenses` at the VariableMeta class at some point because it is just the set union of the `Origin.license` of all indicators) but that depends on how much work it is.

lucasrodes:
Once this is merged, I'll update how the fields are rendered in `wizard snapshot` in https://github.com/owid/etl/pull/1539.

PR #1559, opened 2023-08-30
📊 Adding UN IGME dataset

spoonerf:


spoonerf:
@Marigold Random question - my VScode seems to have reformatted `dag/main.yml` so that the steps for each recipe are indented. Is that a big problem or is it okay to merge it like this? Not sure why it has happened..

Marigold:
@spoonerf Go for it, leading spaces before `-` are apparently the best practice for YAML files. VSCode would format it similarly for me.

Marigold (approved):
Looks good! Couple of nits, but nothing serious.

PR #1564, opened 2023-08-31
:sparkles: Improve grapher upserts performance

Marigold:
Improve performance of uploads to R2 by only uploading if checksums of R2 file and uploaded file differ. This is done by setting metadata of R2 object.

This could be further improved by calculating checksum of a dataframe using [hash_pandas_object](https://pandas.pydata.org/docs/reference/api/pandas.util.hash_pandas_object.html) rather than converting to dictionary, then JSON and then checksum from the string.

UPDATE: It's not making such difference, typically reduces grapher upsert speed by 30% or so.


Marigold:
The speedup is really minimal. Uploading files is not a bottleneck, but comparing checksums is.

PR #1565, opened 2023-08-31
:bar_chart: migrate living planet index to ETL

Marigold:
Migrated with
```
ENV=.env.prod backport-migrate --dataset-id 5175 --namespace owid --version 2022 --short-name living_planet_index
```

Migrate living planet index dataset from grapher to ETL so that we can change its metadata through YAML files. This is a precursor for generating a [living planet index data page](https://ourworldindata.org/grapher/global-living-planet-index) from the new metadata.

`owid` is a bad namespace, would `wwf` be a better one?


spoonerf:
> 

@Marigold  biodiversity is probably the easiest namespace to move to

etl/steps/data/garden/owid/2022/living_planet_index.meta.yml
```diff
@@ -0,0 +1,26 @@
...
+        description: The Living Planet Index (LPI) is a measure of the state of global biological diversity based on population
+          trends of vertebrate species from around the world. The index value is measured relative to species' populations
+          in 1970 (i.e. 1970 = 1).
+        unit: (1970 = 1)
+        display:
+          includeInTable: true
+          name: Living Planet Index
+      living_planet_index_lower:
+        title: living_planet_index_lower
+        unit: (1970 = 1)
```
spoonerf, line 17: Will this overwrite the chart? There is an conversion factor of 100 on the chart to turn it into a %, I'm not sure whether the etl or chart edits take precedence in cases like this?

spoonerf (approved):
Looks good - just a small query about units

etl/steps/data/garden/owid/2022/living_planet_index.meta.yml
Marigold: Chart config always takes precedence, so we should be good here.

Issue #1563, opened 2023-08-30
wizard: collect bugs / improvements


lucasrodes:
After #1539 you can now use the new tool `wizard` to generate templates for your ETL steps.

Use this issue to report bugs that you may encounter. If the issue is very complex, feel free to create a separate - and more extense - issue.

pabloarosado:
I'm using wizard now for the first time, so I'll write here all issues I find along the way:
* When I open wizard, I get the warning:
```
2023-09-01 08:37:55.275 WARNING streamlit.runtime.caching.cache_data_api: No runtime found, using MemoryCacheStorageManager
['streamlit', 'run', '/Users/prosado/Documents/owid/repos/etl/apps/wizard/app.py', '--server.port', '8053', '--', '--phase', 'all', '--run-checks']
```

lucasrodes:
These issues should be solved in https://github.com/owid/etl/pull/1567

lucasrodes:
@Marigold @pabloarosado From our discussions and the documentation on Notion, I had assumed that we agreed that the field `license` should be under `$.meta.origin`. 

However, from what Pablo is saying, it sounds like this is raising an error. As a solution for now, I've set wizard to export the field `license` both at `$.meta.origin.license` and `$.meta.license` levels.

Do you think this is fine?

Marigold:
> It took me a while to realise that the produced .dvc file had license indented too many spaces, hence becoming an item of of origin. I think they are meant to be separate fields in the snapshot metadata.

The original idea was to keep it **under origin**, not separate from it. `license` is there only for backward compatibility. Ideally, snapshot meta would have either `source & license` or `origin & origin.license` (we could refactor old files if this causes too much confusion).

lucasrodes:
walkthrough bug that might be occurring here too: https://github.com/owid/etl/issues/1571

paarriagadap:
I see that `update_period_days` is mandatory in the guideline, but it is not available by default in any of the YAML files

paarriagadap:
- [ ]

paarriagadap:
Two related comments to the ones above:
- [ ] `description_processing` is not shown by default when walkthrough garden generates the yaml, and it's a recommended field
- [ ] `description` is automatically generated in the yaml file, when I see it is not used anymore

lucasrodes:
Hey! Should I just reopen for a small issue then?

- [ ] The create playground notebook option in wizard doesn't generate a Jupyter Notebook

[...]

PR #1567, opened 2023-09-01
wizard: fixes

lucasrodes:
From https://github.com/owid/etl/issues/1563

- [x] Unselect "Generate playground notebook" and submit error: `FileNotFoundError: [Errno 2] No such file or directory: '.../etl/steps/data/meadow/animal_welfare/2023-09-01/playground.ipynb'`
  - ✅ Looks as if it was solved, also in `master`. However, Pablo reported that this issue persists in `add-chick-culling-laws-data` even after merging with master recently.
- [x] After submitting a meadow/garden step, the new files are mentioned on the new page (under "Generated files"). But I think the `dag` file should also be mentioned, if it has been modified.
  - ✅ Same as issue before, working on my end.
- [x] Field citation_producer should be expandable, given that it can be very long (as dataset description fields are).
    - ✅ Done 
- [x] When I open wizard, I get the warning:
  ```
  2023-09-01 08:37:55.275 WARNING streamlit.runtime.caching.cache_data_api: No runtime found, using MemoryCacheStorageManager
  ['streamlit', 'run', '/Users/prosado/Documents/owid/repos/etl/apps/wizard/app.py', '--server.port', '8053', '--', '--phase', 'all', '--run-checks']
  ```
    - Not ideal, but this warning is expected due to the nature of how we are running `streamlit` under the hood.
- [x] When loading the resulting snapshot, I was getting a strange error. I realised that `snapshot.metadata.license` was None, even if `license` was defined in the `.dvc` file. It took me a while to realise that the produced `.dvc` file had `license` indented too many spaces, hence becoming an item of of `origin`. I think they are meant to be separate fields in the snapshot metadata. Once I removed the indent of `license` (and its items), the issue was fixed. So, I suppose it's a bug, and wizard should not store `license` inside `origin` (we can discuss if that should be indeed the correct structure).
    - ✅ We now export the license metadata in snapshot both as `$.meta.license` and `$.meta.origin.license`.

Additional issues:

- [x] Improved management of default values in input fields:
  - First option is if a value has been set to the field in this step.
  - Second option is the value from the previous step (if it was given)
  - Third option is to look in the defaults defined in `wizard.utils.AppState.defaults`
  - Last option is the value given by `default_last` when calling `AppState.st_widget()`
- [x] Change the default command from `wizard` to `etl-wizard` for consistency.


pabloarosado (approved):
I haven't run it, but the code looks good.

Issue #1571, opened 2023-09-05
walkthrough snapshot not including --path-to-file when setting "Import dataset from local file"


paarriagadap:
### One-liner
The "Import dataset from local file" option, when selected, doesn't change the code to include `--path-to-file` instructions

### Environment

- `python -V`: Python 3.11.1
- Operating system: macOS Ventura 13.5.1
- Working with local, staging or live: NA

### Context & details
The code generated seems to be the same as the one with the unselected option. For now, I have been copying old codes that were set with the option.
[Post in Slack](https://owid.slack.com/archives/C46U9LXRR/p1693906385799679).

PR #1572, opened 2023-09-05
Fix walkthrough and wizard local file issue

pabloarosado:
Let walkthrough and wizard create a snapshot file using a path to a local file.
Fixes https://github.com/owid/etl/issues/1571


lucasrodes (approved):
Thanks, Pablo!

PR #1589, opened 2023-09-11
:tada: script for moving archived steps to their own folder

Marigold:
Move all archived steps to `etl/steps/archive` directory to separate them from our active steps. Once we have them there, we exclude them from type checks, linting and from **VSCode search and its file explorer**.

I find archived steps living next to the active ones pretty annoying when refactoring, and it also messes up a quick search in VSCode when I often find myself editing an archived version by mistake. It's still nice to have for me though, so if anyone objects, I'm ok with closing this.

This PR doesn't move archived snapshots as they're not refactored as often, but we could consider archiving them too if this gets approved.

If this gets approved and merged, moving those files will be done in a separate PR. The script could then run after every step archival (I'm not sure if we have a function for it already).

(By the way, there was a bug when running archived steps, which means that probably no one has ever run an archived step.)


pabloarosado (approved):
Looks good, I think it's a good solution. FAOSTAT scripts may need to be changed a bit, since they may rely on files and not the dag (I'd need to check). And I think the VersionTracker won't be affected. So, I think it's safe.

lucasrodes (approved):

PR #1597, opened 2023-09-12
Animal welfare bug bash

pabloarosado:
Small improvements on metadata of various animal welfare steps, to fix issues found during the latest bug bash.


etl/steps/data/garden/animal_welfare/2023-08-01/global_hen_inventory.py
```diff
@@ -31,30 +31,65 @@
...
+    tb_fixed.loc[uruguay_mask, "available_at"] = uruguay_url
+    tb_fixed.loc[uruguay_mask, "source"] = tb_fixed.loc[uruguay_mask, "source"].item().split(expected_text)[0]
+
+    # TODO: Malta has no link, and the source seems to be wrong.
+
+    # Fix some typos in the metadata, and ensure sources end in a period.
+    tb_fixed["source"] = (
+        tb_fixed["source"]
+        .str.replace("husbndry", "husbandry")
+        .replace("(Ministry of Agriculture). '", "(Ministry of Agriculture)'")
```
lucasrodes, line 56: maybe you can do `tb_fixed["source"].str.strip()` too, in case this were to happen for other values.

lucasrodes: Unless this string is something that appears within the text and is not the end of it.

etl/steps/data/garden/animal_welfare/2023-08-03/share_of_cage_free_eggs.py
```diff
@@ -1,14 +1,41 @@
...
+
+    # Add the share of UK eggs in cages and cage-free.
+    tb_uk["share_of_eggs_in_cages"] = 100 * tb_uk["number_of_eggs_from_enriched_cages"] / tb_uk["number_of_eggs_all"]
+    tb_uk["share_of_eggs_cage_free"] = 100 - tb_uk["share_of_eggs_in_cages"]
+
+    # Update their units.
+    tb_uk["share_of_eggs_cage_free"].metadata.unit = "%"
+    tb_uk["share_of_eggs_cage_free"].metadata.short_unit = "%"
+    tb_uk["share_of_eggs_in_cages"].metadata.unit = "%"
+    tb_uk["share_of_eggs_in_cages"].metadata.short_unit = "%"
```
lucasrodes, line 23: in case this were to grow...

```python
columns = ["share_of_eggs_cage_free", "share_of_eggs_in_cages"]
for column in columns:
  tb_uk[column].metadata.unit = "%"
  tb_uk[column].metadata.short_unit = "%"

File: etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml [hunk 3]
```diff
@@ -1,42 +1,44 @@
...
-    This study estimates the annual number of farmed fish slaughtered for food using data on aquaculture production tonnages (from the Food and Agriculture Organization of the United Nations) and mean weights estimated from available fish size data (from a range of internet sources).
+  attribution: &attribution Fishcount (2018)
+  description_key: &description_key
+    - This indicator uses data on aquaculture production tonnages and mean weights estimated from available fish size data.
+    - |
+      Note that this study does not include:
+      - Non-fish species, e.g. lobsters.
+      - Fish deaths arising in fish farms prior to harvest, in the capture of wild fish for feed or re-stocking, or from environmental impacts on surrounding wildlife.
+      - Fish farmed for use as bait, whose small production tonnages may comprise large numbers of fish due to their very small size of just a few grams.
+      - Species without estimated mean weight (which could be an additional 17% of total fish production).
```
lucasrodes, line 10: Just wondering here: is the second element of the list a string with bullet points? If so, I guess you were trying to nest some bullet points? One thing to write down for the future: "nested lists"

snapshots/animal_welfare/2023-08-01/global_hen_inventory.csv.dvc
```diff
@@ -1,18 +1,16 @@
...
-  source_data_url:
-  file_extension: csv
-  license_url: https://welfarefootprint.org/research-projects/laying-hens/
-  license_name: Creative Commons BY
-  date_accessed: 2023-08-01
-  is_public: true
-  description: |
+  origin:
+    dataset_title_owid: Global hen inventory
+    dataset_title_producer: Global hen inventory
```
lucasrodes, line 4: I guess we'll need to change these names (and others) after the renamings? Might have to do that as well for other snapshots :D

lucasrodes (approved):
LGTM! Nice work 🐮

etl/steps/data/garden/animal_welfare/2023-08-01/global_hen_inventory.py
pabloarosado: This was an odd case, where the text was between quotation marks, strip would not fix this.

etl/steps/data/garden/animal_welfare/2023-08-03/share_of_cage_free_eggs.py
pabloarosado: Same number of lines, not worth it :-). But one thing to keep in mind is that this issue may happen more often in the future: You construct a "share" variable, and need to change the units accordingly. Maybe we could create a simple function to create "share" variables (that would update the units automatically).

etl/steps/data/garden/animal_welfare/2023-08-14/number_of_farmed_fish.meta.yml
pabloarosado: Yes, I thought the key descriptions in data pages would be able to handle this, but it doesn't. However, since there's no expectation to have a data page for this indicator, so, for now, I'll leave it like this.

snapshots/animal_welfare/2023-08-01/global_hen_inventory.csv.dvc
pabloarosado: Yes, these metadata files are in limbo...

PR #1603, opened 2023-09-14
:bug: fix MySQL deadlocks

Marigold:
An attempt to fix MySQL deadlocks when inserting into `origins_variables`.


Marigold:
I tried fixing it with https://github.com/owid/etl/commit/594e05fd930aa431a1401def2d8d84c7bbd9def3, which doesn't decrease performance.

This PR unfortunately slows upserts down by about 30%, which is why I'm reluctant to merge it. Replicate it with
```
ENV=.env.mojmir etl grapher://grapher/ilostat/2023-09-19/unemployment --grapher --force --only
```

Marigold:
We started getting MySQL lock timeouts. Trying to fix upserts of variables-origins in [this PR](https://github.com/owid/etl/pull/1653) now.

Marigold:
The new approach seems to have worked.

PR #1604, opened 2023-09-14
:hammer: walkthrough wipeout

Marigold:
Deprecate walkthrough in favour of the new wizard.


apps/walkthrough/utils.py
lucasrodes, line 1: feels like we may want to remove this file too?

Marigold: You're right. I did it thoroughly this time and wiped out all walkthrough mentions (removing `apps/walkthrough` as part of it).

lucasrodes:
@Marigold I haven't looked at it, so I think it might not work in `wizard`. I think we should definitely have that, though. I can add it in the upcoming weeks (unless you do that first).

Marigold:
> I think we should definitely have that, though. I can add it in the upcoming weeks (unless you do that first).

No worries, I don't need it now. I just though you could find it useful in case you find yourself filling metadata over and over again.

Marigold:
@lucasrodes are we ready for this?

lucasrodes:
hey @Marigold, I think we are almost good to go. Thanks for the reminder.

I'd say let me ask early next week if other people in the data team are still using walkthrough for some reason/miss something from it.

Issue #1609, opened 2023-09-18
wizard: 'version' was unexpected error when submitting snapshot metadata


paarriagadap:
### One-liner
There is an error when I submit the metadata for a snapshot in wizard

### Environment

- `python -V`: Python 3.11.1
- Operating system: macOS Ventura 13.5.2
- Working with local, staging or live: N/A

### Context & details
After filling of the necessary fields, I submit the information and get this error:
```
Exception: Unknown error type additionalProperties with message Additional properties are not allowed ('version' was unexpected)
Traceback:
File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 548, in _run_script
    self._session_state.on_script_will_rerun(rerun_data.widget_states)
File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/streamlit/runtime/state/safe_session_state.py", line 68, in on_script_will_rerun
    self._state.on_script_will_rerun(latest_widget_states)
File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/streamlit/runtime/state/session_state.py", line 486, in on_script_will_rerun
    self._call_callbacks()
File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/streamlit/runtime/state/session_state.py", line 499, in _call_callbacks
    self._new_widget_state.call_callback(wid)
File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/streamlit/runtime/state/session_state.py", line 251, in call_callback
    callback(*args, **kwargs)
File "/Users/parriagadap/etl/apps/wizard/templating/snapshot.py", line 496, in update_state
    form = SnapshotForm.from_state()
           ^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/parriagadap/etl/apps/wizard/utils.py", line 388, in from_state
    return cls(**data)
           ^^^^^^^^^^^
File "/Users/parriagadap/etl/apps/wizard/templating/snapshot.py", line 96, in __init__
    super().__init__(**data)
File "/Users/parriagadap/etl/apps/wizard/utils.py", line 376, in __init__
    self.validate()
File "/Users/parriagadap/etl/apps/wizard/templating/snapshot.py", line 106, in validate
    self.validate_schema(SNAPSHOT_SCHEMA, ["meta"])
File "/Users/parriagadap/etl/apps/wizard/utils.py", line 421, in validate_schema
[...]

PR #1611, opened 2023-09-18
🐛 wizard: fix key name error

lucasrodes:
Fixes https://github.com/owid/etl/issues/1609


paarriagadap (approved):
I was able to generate the snapshot files now, thanks @lucasrodes!

PR #1614, opened 2023-09-18
:bar_chart: Create Colonial Dates Dataset (COLDAT)

paarriagadap:
Uploads COLDAT to the ETL, as part of the [State Capacity topic page](https://github.com/owid/owid-issues/issues/925).

Staging version of the dataset here: http://staging-site-parriagadap/admin/datasets/6218


etl/steps/data/garden/harvard/2023-09-18/colonial_dates_dataset.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Create an additional summarized colonizer column, replacing the values with " - " with "More than one colonizer"
+    # I add the "z." to have this at the last position of the map brackets
+    tb["colonizer_grouped"] = tb["colonizer"].apply(
+        lambda x: "z. Multiple colonizers" if isinstance(x, str) and " - " in x else x
+    )
+
+    # Create last_colonizer column, which is the most recent non-null colonizer for each country and year
+    tb["last_colonizer"] = tb.groupby(["country"])["colonizer"].fillna(method="ffill")
+    tb["last_colonizer_grouped"] = tb.groupby(["country"])["colonizer_grouped"].fillna(method="ffill")
```
paarriagadap, line 51: When I apply these two lines above I lose the origins, but when I add `.copy_metadata(tb)` I have an `AttributeError: 'TableMeta' object has no attribute 'description_short'` error. What can I do to solve it?

paarriagadap, line 51: When I apply these two lines above I lose the origins, but when I add `.copy_metadata(tb)` I have an `AttributeError: 'TableMeta' object has no attribute 'description_short'` error. What can I do to solve it? @Marigold @pabloarosado

Marigold: The problem is that `groupby` operation throws away metadata. Since you overwrite `tb` by that groupby, `.copy_metadata(tb)` tries to get metadata from non-table.

I'll see if I can groupby make work with Tables tomorrow. In the meantime you can try this:
```python
tb2 = tb.copy()

# Create last_colonizer column, which is the most recent non-null colonizer for each country and year
tb2["last_colonizer"] = tb2.groupby(["country"])["colonizer"].fillna(method="ffill")
tb2["last_colonizer_grouped"] = tb2.groupby(["country"])["colonizer_grouped"].fillna(method="ffill")

tb2 = tb2.copy_metadata(tb)
for col in ("last_colonizer", "last_colonizer_grouped"):
    tb2[col].metadata = tb.colonizer.copy()

# Merge both tables
tb = pr.merge(tb2, tb_count, on=["country", "year"], how="left", short_name="colonial_dates_dataset")
```

PR #1638, opened 2023-09-20
:sparkles: discourage use of variable.licenses

Marigold:
Don't raise a warning when `variable.licenses` is missing. Using `origin.license` is preferred. Also show linter error in YAML files when the field is present.

### DIFF
--- a/apps/walkthrough/garden.py
+++ b/apps/walkthrough/garden.py
@@ -242,7 +242,6 @@ def _fill_dummy_metadata_yaml(metadata_path: Path) -> None:
         variable_meta = {
             "title": "Dummy",
             "description": "This is a dummy indicator with full metadata.",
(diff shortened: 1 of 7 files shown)


pabloarosado (approved):
Looks good! I think we may need to also change `owid.catalog.variables`, so that licenses are not automatically propagated.

Marigold:
I removed `get_unique_licenses_from_variables`. I hope that's what you meant.

pabloarosado:
> I removed `get_unique_licenses_from_variables`. I hope that's what you meant.

Yes, that should be it!

Marigold:
I had to put `get_unique_licenses_from_variables` back. Removing it messed up unit tests. It doesn't really hurt to have it there, and we can remove it when we decide to deprecate the field completely.

PR #1640, opened 2023-09-20
🐛 wizard: remove licenses field

lucasrodes:
Remove `licenses` from `$.meta`.

This complements https://github.com/owid/etl/pull/1638, and fixes the wizard bug:

```
File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/streamlit/runtime/scriptrunner/script_runner.py", line 548, in _run_script
    self._session_state.on_script_will_rerun(rerun_data.widget_states)
File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/streamlit/runtime/state/safe_session_state.py", line 68, in on_script_will_rerun
    self._state.on_script_will_rerun(latest_widget_states)
File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/streamlit/runtime/state/session_state.py", line 486, in on_script_will_rerun
    self._call_callbacks()
File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/streamlit/runtime/state/session_state.py", line 499, in _call_callbacks
    self._new_widget_state.call_callback(wid)
File "/home/lucas/repos/etl/.venv/lib/python3.10/site-packages/streamlit/runtime/state/session_state.py", line 251, in call_callback
    callback(*args, **kwargs)
File "/home/lucas/repos/etl/apps/wizard/templating/snapshot.py", line 526, in update_state
    form = SnapshotForm.from_state()
File "/home/lucas/repos/etl/apps/wizard/utils.py", line 391, in from_state
    return cls(**data)
File "/home/lucas/repos/etl/apps/wizard/templating/snapshot.py", line 116, in __init__
    super().__init__(**data)
File "/home/lucas/repos/etl/apps/wizard/utils.py", line 378, in __init__
    self.validate()
File "/home/lucas/repos/etl/apps/wizard/templating/snapshot.py", line 126, in validate
    self.validate_schema(SNAPSHOT_SCHEMA, ["meta"])
File "/home/lucas/repos/etl/apps/wizard/utils.py", line 424, in validate_schema
    raise Exception(f"Unknown error type {error_type} with message '{error.message}'")
```

PR #1641, opened 2023-09-20
✨ wizard: limit options of attribution

lucasrodes:
This PR changes how the attribution field is field in Snapshot.

To test it, you can run

```
etl-wizard --dummy-data
```

Then, play with the input field `origin.attribution`. In particular, you can:
- Choose first format option: `origin.attribution` will not be added to the metadata yaml file.
- Choose the second format option: `origin.attribution` with the desired format.
- Choose the custom option and input it yourself. This value will be added to `origin.attribution` in the metadata.

---
This PR includes the fix from https://github.com/owid/etl/pull/1640


pabloarosado (approved):
I just tried it and it's a great idea. This will ensure much more consistency. One minor thing is that the attribution field should not be editable. For example, when choosing "Custom attribution...", I first tried to edit that text. It's not a big deal because you see the other editable field appear below. But if it's easy to make that attribution field non editable, that would be better.

lucasrodes:
hi @pabloarosado, thanks for your feedback. I may not completely follow you there. Do you mean that the user should not be able to edit the select box field?

That is, unfortunately not possible, as this is the standard behaviour of such input field in streamlit. However, that only serves one purpose, and this is to search among the options in the dropdown, not to actually customise the value of the field.

(this is the same behaviour as with `origin.license.name`.

PR #1642, opened 2023-09-21
:hammer: Various ergonomics improvements

Marigold:
Various improvements taken from [processing log work](https://github.com/owid/etl/pull/1574)

* **Add `Table.groupby` method**
* Add `.m` alias for `.metadata`
* Improve catalog structure by cleanly separating `utils.py`
* Fix metadata propagation in some datasets
* Minor bugfixes

I tested `Table.groupby` by rebuilding steps that use `Table.groupby` (probably not all, but a lot) and made sure I got the same results (here's [datadiff example](https://buildkite.com/our-world-in-data/etl-build-public-datasets-branch/builds/4688#018abcc1-ceb8-4d61-a051-0ecbff1063d0), country / entity from index now keeps origins).

I haven't tested every groupby method, the question is whether to invest the time now to do all of them or add them on the fly.

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, List, Optional, TypeVar, Union, cast
 
 import numpy as np
 import pandas as pd
-from owid.catalog import Dataset, Table
+from owid.catalog import Dataset, Table, Variable
 from owid.datautils.common import ExceptionFromDocstring, warn_on_list_of_entities
 from owid.datautils.dataframes import groupby_agg, map_series
 from owid.datautils.io.json import load_json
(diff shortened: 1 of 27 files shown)


pabloarosado (approved):
This is fantastic! Thanks for implementing metadata propagation on groupby operations!
I've played around with it a bit and it works well (except that the processing log doesn't get updated, but we can look into that in the future).
Maybe, for safety, you could include some tests on groupby operations to check not just the metadata, but also if the data is as expected (up to you).

Marigold:
> ...except that the processing log doesn't get updated, but we can look into that in the future

I already have that in a different PR for the processing log, but I wanted to merge this first. 

> Maybe, for safety, you could include some tests on groupby operations to check not just the metadata, but also if the data is as expected (up to you).

Makes sense! Adding a couple of asserts.

PR #1648, opened 2023-09-21
📊 plastic pollution

veronikasamborska1994:
hey @spoonerf! it looks like a lot more code than it actually is since the datasets all come from one OECD Plastic Outlook database :) but I had to download each of the files individually. I guess before you go over the code I wanted to ask what you'd do in this case? Would you combine all of the smaller datasets into 1 large dataset and if so would you do it in meadow/garden step? Keeping them separate helped me track things a bit better but it also means there is quite a lot of repetition... Thanks a lot and hope you are feeling better this week! :) and definitely let me know if this is not a good time!


spoonerf:
Hey, @veronikasamborska1994! 

I think how you've handled this makes complete sense and is the most natural fit for the ETL. 
The datasets are separate enough and have different enough metadata (e.g. different URLs) for it to make sense to handle them separately. I think combining them into one dataset with multiple tables would potentially be quite a lot of effort for no real gains in terms of clarity? But I would also be curious to hear what someone more experienced with the new metadata thinks! 

I had some thoughts about how you could potentially reduce some replication but don't feel like you have to do any:

* I think all your `etl/steps/data/garden/oecd/2023-09-21/*countries.json` files are the same? You could condense them into one `countries.json` file. I've done this before in `etl/steps/data/garden/ihme_gbd/2019`, where I had a similar situation.
* Some code steps are used in multiple of the garden steps. You could consider creating a `etl/steps/data/garden/oecd/2023-09-21/shared.py` file and creating shared functions that would help reduce replication slightly. However, this might be a bit of effort for minimal gains, considering that many datasets need to be handled slightly differently.
* I noticed there are quite a few variables in the "Global Plastics Outlook - Plastics waste in 2019 by region, polymer and application" dataset where the only value is 0t, you could consider dropping these.


Hope that's helpful - let me know when is a good time to review the code more closely!

veronikasamborska1994:
hi @spoonerf ! thanks a lot for these suggestions! v useful! I've deleted repeated country files as per your suggestion but other than this I think there isn't much to add to shared.py file so it's probably unnecessary here. I've also added a line to remove variables that turn out to be 0 after pivoting. I think given that there were > 200 variables there is almost no chance we'd be listing the ones that don't produce any plastic waste here! 

I think it's ready to be reviewed now, but no rush! :)

etl/steps/data/garden/oecd/2023-09-21/plastic_aquatic_leakage_projections.meta.yml
```diff
@@ -0,0 +1,10 @@
+tables:
+  plastic_aquatic_leakage_projections:
```
spoonerf, line 2: It's kind of surprising to me that the `plastic_aquatic_leakage` and `plastic_aquatic_leakage_projections` don't have the same variables. I think it's out of scope of this PR but I would be interested to know why the projections are such higher values - are they counting things not included in `plastic_aquatic_leakage`?

etl/steps/data/garden/oecd/2023-09-21/plastic_use_application.meta.yml
```diff
@@ -0,0 +1,100 @@
...
+  plastic_use_application:
+    variables:
+      building__and__construction:
+        title: Building and construction
+        unit: 'tonnes'
+        short_unit: 't'
+        display:
+          numDecimalPlaces: 0
+      consumer__and__institutional_products:
+        title: Consumer and institutuional products
```
spoonerf, line 12: Tiny typo here in 'institutional'

etl/steps/data/garden/oecd/2023-09-21/plastic_waste_2019.meta.yml
```diff
@@ -0,0 +1,10 @@
+tables:
+  plastic_waste_2019:
+    variables:
+      plastic_waste:
+        title: Plastic waste
+        unit: 'tonnes'
+        short_unit: 't'
+        display:
+          isProjection: true
```
spoonerf, line 9: Is this one a projection?

etl/steps/data/meadow/artificial_intelligence/2023-06-21/epoch.py
```diff
@@ -58,7 +58,6 @@ def run(dest_dir: str) -> None:
     #
     tb = Table(df, short_name=paths.short_name, underscore=True)
     ds_meadow = create_dataset(dest_dir, tables=[tb], default_metadata=snap.metadata)
-
```
spoonerf, line 61: Maybe an accidental commit here? Think you can remove it with:

`git checkout origin/master etl/steps/data/meadow/artificial_intelligence/2023-06-21/epoch.py`

Not a big deal either way, I guess just better practice to remove

etl/steps/data/meadow/plastic_waste/2023-09-26/geyer_2017.py
```diff
@@ -0,0 +1,64 @@
...
+    # Process data.
+    #
+
+    # Initialize an empty list to hold the table rows
+    all_rows = []
+    page_start = 7
+    page_end = 8
+
+    # Open the PDF file
+    with pdfplumber.open(file_path) as pdf:
```
spoonerf, line 35: Nice!

snapshots/oecd/2023-09-21/plastic_aquatic_leakage.csv.dvc
```diff
@@ -0,0 +1,36 @@
...
+        - Leakage from mismanaged waste and litter to aquatic environments
+        - Transport to oceans
+        - Accumulated stock of plastics in rivers and lakes
+        - Accumulated stock of plastics in oceans
+
+      Plastic leakages to aquatic environments and the subcategory trasport to oceans are estimated by applying the methodology adapted from Lebreton and Andrady (2019), on OECD ENV-Linkages model outputs and plastic leakage from mismanaged and litter. The accumulated stock of plastics leakages in rivers and lakes corresponds to the net cumulative sum of leakages in rivers and lakes from 1951 onwards. The accumulated stock of plastics leakages in oceans corresponds to the net cumulative sum of leakages to oceans from 1951 onwards.
+    date_published: 2022-02-09
+    # Citation
+    producer: OECD
+    citation_full: OECD (2022), Global Plastics Outlook , https://stats.oecd.org/viewhtml.aspx?datasetcode=PLASTIC_LEAKAGE_5&lang=en,     accessed
```
spoonerf, line 20: The URL in the citation looks slightly unusual to me - is that how they've requested it be cited? Also some slightly off spacing in the text here.

snapshots/oecd/2023-09-21/plastic_aquatic_leakage.csv.dvc
```diff
@@ -0,0 +1,36 @@
...
+
+    # Files
+    url_main: https://stats.oecd.org/viewhtml.aspx?datasetcode=PLASTIC_LEAKAGE_5&lang=en
+    date_accessed: 2023-09-21
+
+    # License
+    license:
+      name: CC BY 4.0
+      url: 
+        https://www.oecd.org/termsandconditions/#:~:text=You%20must%20give%20appropriate%20credit,(accessed%20on%20(date)).
```
spoonerf, line 31: I think this can probaly be just: https://www.oecd.org/termsandconditions/

spoonerf (approved):
Looks good! Just had some minor suggestions.

etl/steps/data/garden/oecd/2023-09-21/plastic_aquatic_leakage_projections.meta.yml
veronikasamborska1994: agree with you! it's quite a big discrepancy and I've checked that this is because of a mistake I've made. will need to dig into it more! thanks for pointing it out! :)

snapshots/oecd/2023-09-21/plastic_aquatic_leakage.csv.dvc
veronikasamborska1994: yep, they ask to include a link but I fixed the spacing now :) thanks!

PR #1656, opened 2023-09-22
:bug: replace . by _ in snapshot names

Marigold:
@lucasrodes don't use `.` in filenames, please. I'm not sure how the dots got there, you should get an assertion error if you run the snapshot step with dots.


lucasrodes:
Thanks, Mojmir!

Yup, I think this did not raise an error in the past. I recently detected this and decided to go and create new versions: https://github.com/owid/etl/pull/1495. So, the versions that you modified are not in use anymore.

PR #1659, opened 2023-09-22
✨ wizard: minor enhancements

lucasrodes:
- Use `|-` for `citation_full` in `origin` metadata template. This prevents weird errors (e.g. when citation full contains `:`).
- Accept numeric values for `version_producer`. Sometimes the version might just be a number (e.g. `3.2`) or a year (`2016`). Let's accept these and stop showing error messages when this is the case.


pabloarosado (approved):
Looks good!

PR #1664, opened 2023-09-25
📊 war: peace diehl et al

lucasrodes:
Add "Peace data" dataset to ETL.

Additionally, added `read_fwf` method for `Table` object. This was done together with @pabloarosado, so I guess it was already seen by a second pair of eyes.


etl/steps/data/garden/war/2023-09-25/peace_diehl.meta.yml
```diff
@@ -39,27 +39,37 @@ tables:
         unit: countries
         processing_level: minor
         description_processing: *description_processing
+        display:
+          numDecimalPlaces: 1
```
spoonerf, line 43: Decimal places should be 0 in all cases? I think they are all counts of countries?

spoonerf (approved):
Looks good - just a decimal places query

etl/steps/data/garden/war/2023-09-25/peace_diehl.meta.yml
lucasrodes: totally! nice catch

Issue #1569, opened 2023-09-04
Per capita variables use display name `population`


Marigold:
I've noticed that some variables like [Per capita CO2 consumption](http://staging-site-mojmir/admin/variables/738081) use `Population` as a display name.

This happens because dividing by population combines metadata of both variables and if the numerator doesn't have `display` metadata set, it'll inherit it from `Population`.

Maybe we could stop (mis?)using `display.name` and focus on setting the proper `title` instead. @pabloarosado what do you think?

```
v1 = Variable([1,2,3], name='co2')
v1.metadata.title = 'co2'
v2 = Variable([1,2,3], name='population')
v2.metadata.display = {
    'name': 'Population',
    'numDecimalPlaces': 0
}
per_capita = v1 / v2
per_capita.metadata
```
<img width="276" alt="image" src="https://github.com/owid/etl/assets/1550888/28c22efd-9a54-4426-8bdd-68e4db758dce">

pabloarosado:
Hi @Marigold thanks for finding this, I had not thought about this possibility. But in the example you give I don't see `Population` showing anywhere. Can you screenshot or pinpoint an example where this happened?

We could implement some logic on the function that combines metadata (for `title`, `display.name`, `presentation.title_public`, and maybe others). If the operation is `/`, then ignore the title of the second parent. But this might be annoying when doing temporary operations between variables, I'd need to think about it.

Marigold:
Here's an [example](https://owid.cloud/admin/variables/738081) that shows "Population" in the Preview chart. This is not a big problem now, but it becomes more annoying for our indicator-centered future with data pages.

pabloarosado:
After merging https://github.com/owid/etl/pull/1676 now divisions where the numerator has no display and the denominator has display, should yield no display. This should fix the issue.

PR #1672, opened 2023-09-26
Avoid propagating metadata on division

pabloarosado:
Avoid propagating metadata fields (titles, units and descriptions) of the denominator, when dividing two variables.

When dividing a variable by another, it only makes sense to keep the metadata values of the first variable. For example, if we have energy (without description) and population (with a description), when calculating energy per capita, the result shouldn't have the description of population. It should have no description.

This PR would fix for example [this chart](https://owid.cloud/admin/charts/4451/edit) of electricity per capita. The sources tab shows the description of population.

In a separate PR we can add similar logic to display and presentation.

### DIFF
--- a/lib/catalog/owid/catalog/variables.py
+++ b/lib/catalog/owid/catalog/variables.py
@@ -345,8 +345,18 @@ for k in VariableMeta.__dataclass_fields__:
 
 
 def _get_metadata_value_from_variables_if_all_identical(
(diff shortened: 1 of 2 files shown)


Marigold (approved):
Agree! Once we fix it for `display`, we'll be able to close https://github.com/owid/etl/issues/1569.

PR #1674, opened 2023-09-27
:bar_chart: backport literacy rates and public expenditure on education to ETL

Marigold:
Backport [literacy rates](https://owid.cloud/admin/datasets/2762) with
```
ENV=.env.prod backport-migrate --dataset-id 2762 --namespace education --version 2018-04-18 --short-name literacy_rates
```

Backport [public expenditure on education](https://owid.cloud/admin/datasets/242) with
```
ENV=.env.prod backport-migrate --dataset-id 242 --namespace education --version 2017-09-30 --short-name public_expenditure
```

Once these datasets are merged, we should run `ENV=.env.prod.write walkthrough charts` to migrate their charts and delete old datasets.

## Issues
1. Markdown in `snapshots/education/2018-04-18/literacy_rates.feather.dvc` doesn't show list properly, not sure what am I doing wrong?
2. Source fields `Variable geographic coverage` and `Data publisher's source` are not available in ETL (intentionally), we might want to move them to description

image

### DIFF
--- a/snapshots/education/2018-04-18/literacy_rates.feather.dvc
+++ b/snapshots/education/2018-04-18/literacy_rates.feather.dvc
@@ -4,7 +4,6 @@ meta:
     description: >-
       Additional Information
 
-      </br>
 
       This long run cross-country dataset combines data from a number of sources.
       We took the estimates from the World Bank’s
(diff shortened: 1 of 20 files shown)


veronikasamborska1994:
thanks so much again @Marigold🥇! The list in `snapshots/education/2018-04-18/literacy_rates.feather.dvc` looks good to me when I ran it on my staging site! Not sure we need `Variable geographic coverage `and` Data publisher's source` tbh as these aren't particularly useful! 

I've also added some code in my original WB dataset to create combined indicators with historical and more recent data. One thing I wasn't sure about is if there was a 'proper' way of propagating the description of the "dataset" for these indicators to these new combined indicators? Both of these have the old metadata structure so I am guessing the best way would be to just access the `metadata.source.description` and then use it at the indicator level in the WB dataset ?

snapshots/education/2018-04-18/literacy_rates.feather.dvc
```diff
@@ -1,26 +1,78 @@
...
-      - Data for 1820 and 1870 (except for the US): Broadberry and O'Rourke (2010) – The Cambridge Economic History of Modern Europe: Volume 1, 1700-1870
-      - Data for the US: National Center for Education Statistics, online [here](nces.ed.gov/naal/lit_history.asp).
-      - Global estimates for 1820-2000: van Zanden, J.L., et al. (eds.) (2014), How Was Life?: Global Well-being since 1820, OECD Publishing. Online [here](http://www.oecd.org/statistics/how-was-life-9789264214262-en.htm).
-      - Historical estimates for Latin America: OxLAD – Oxford Latin American Economic History Data Base, online [here](http://moxlad-staging.herokuapp.com/home/en).
-      - Most recent estimates for high-income countries, as well as any available estimates for 2016: [CIA World Factbook](https://www.cia.gov/library/publications/the-world-factbook/).
+      This long run cross-country dataset combines data from a number of sources.
```
Marigold, line 9: I think we've recently switched to Markdown for Sources. HTML still gets rendered, but we should write all new metadata in Markdown (I'm like 95% sure...)

Marigold:
> One thing I wasn't sure about is if there was a 'proper' way of propagating the description of the "dataset" for these indicators to these new combined indicators? Both of these have the old metadata structure so I am guessing the best way would be to just access the metadata.source.description and then use it at the indicator level in the WB dataset ?

You're right, that's probably the best approach at the moment. If we start doing it often, we should come up with a more ergonomic approach.

snapshots/education/2018-04-18/literacy_rates.feather.dvc
veronikasamborska1994: ah ok I see! I've added some code to do the conversion from html to markdown in the snapshot step and I think it looks ok!

Marigold: Oh, nice one! I might add it to the backporting template.

veronikasamborska1994: sounds good! I am happy to merge whenever now. This will trigger a rebuild of a pretty huge WB education dataset so I usually do it at the end of the day :) but not sure if this is actually something I needed to worry about

Marigold: Merge at will!

PR #1676, opened 2023-09-27
Improve the logic of display and presentation propagation on divisions

pabloarosado:
Improve the logic of display and presentation propagation on division operations.

I applied the analogous logic to other metadata fields (titles, descriptions and units). Given c = a / b:
* If a has no metadata, c has no metadata (regardless of b).
* If a has metadata and b has no metadata, c has metadata from a.
* If a and b have identical metadata, c has that metadata.
* If a and b have different metadata, c has no metadata.

I'm not totally sure about the last point. Alternatively, we could adopt "If a and b have different metadata, c has metadata from a". But I think the current logic is a bit safer.


Marigold (approved):
Your choice "If a and b have different metadata, c has no metadata." sounds solid. We can always change it if it becomes annoying.

PR #1673, opened 2023-09-26
📊 hmd: iteration 2

lucasrodes:
- Update snapshot
- Add new metrics:
  - `life_expectancy_fm_ratio`: Life expectancy ratio female to male.
  - `life_expectancy_fm_diff`: Life expectancy difference, female minus male.
- Add variable descriptions.
- Add `deaths` table in meadow, integrate table in single main table in garden. (required for https://github.com/owid/etl/pull/1680)

PR #1680, opened 2023-09-27
📊 demography: survivorship ages

lucasrodes:
Adding HMD-derived indicator.

- More context: https://github.com/owid/owid-issues/issues/1184
- goes after https://github.com/owid/etl/pull/1673

This dataset will power charts like this one:

survivorship-ages

Note: This adds `scipy` as a dependency to the project.


etl/steps/data/garden/demography/2023-09-27/survivor_percentiles.py
```diff
@@ -0,0 +1,164 @@
...
+def obtain_survivorship_ages(tb_group: Table, start_age: int = 0, end_age: int = 110) -> pd.DataFrame:
+    """Get survivorship ages given a life and deaths table.
+
+    Output dataframe has a column for each percentile of survivorship age.
+
+    tb_group is expected to be a subset of the compelte table. It should only concern a particular (country, year, sex) triple.
+    """
+    # Step 1: Apply splines, get Mx for each (country, year, sex, age)
+    ## Define splines
+    spline_deaths = UnivariateSpline(tb_group["age"], tb_group["deaths"], s=0, k=3)
```
spoonerf, line 132: Cool! I wonder if it's just worth noting what the s and k mean here. From [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.UnivariateSpline.html):

If 0, spline will interpolate through all data points. This is equivalent to [InterpolatedUnivariateSpline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.InterpolatedUnivariateSpline.html#scipy.interpolate.InterpolatedUnivariateSpline).

And: 
k = 3 is a cubic spline.

I was wondering if the function would be equivalent across R and Python and I think this is a key point of clarification on that.

spoonerf (approved):
Looks great! Just a couple of small queries :)

PR #1683, opened 2023-09-28
:sparkles: enable Jinja for description_short

Marigold:
Enable Jinja templating for `description_short` in addition to `description` and `title`. @lucasrodes do you have a use case for any other fields? We could do it for any field, but the fewer the better (those nested fields or lists are slightly trickier and I don't have a good test case).


lucasrodes (approved):
LGTM, thanks

PR #1686, opened 2023-09-28
Archive unused natural disasters steps

pabloarosado:
I've moved all natural disasters steps from the old version into the archive, using `move_steps_to_archive.py`.

However, I couldn't find any snapshots archive. Isn't there one?


Marigold (approved):
Snapshots are not archived (DVC complains when we move files).

PR #1690, opened 2023-09-28
📊 demography: gini life expectancy inequality

lucasrodes:
I am adding an HMD-derived indicator.

- More context: https://github.com/owid/owid-issues/issues/504
- Goes after 
  - https://github.com/owid/etl/pull/1673
  - #1680 

It would enable charts like
gini-coefficient-of-lifespan-inequality-hmd-aburto-et-al-2023


paarriagadap (changes requested):
Hi @lucasrodes! It looks very good, but I would like to have the functions `AKm02a0` and `gini_from_mx` documented in more detail because there are several ways to calculate Gini and I would like to have clarity on all the variables being used

lucasrodes:
Thanks, @paarriagadap! I've added some clarifications.

Issue #1679, opened 2023-09-27
metadata: Find a way to identify where metadata was not propagated


pabloarosado:
### One-liner
We need a way to identify where metadata was lost within an ETL step.

### Context & details
While working on an ETL step, sometimes you do an operation that did not respect propagation of metadata. And if you don't notice it, you may reach the end of the step and then need to go back and figure out where metadata was lost.

We need to avoid that situation. I can think of a few options:
* The easiest is to move the function `owid.catalog.tables.check_all_variables_have_metadata` to `etl.helpers` (and possibly rename it to something shorter). This function checks if all columns have at least one origin. This function can be used (manually) every now and then while debugging.
* Add a method to `Table` that checks the same thing. It could be something like `tb.check_metadata()`.
* Create a new tool that executes an etl step and goes line by line, checking if at any point an indicator loses its origins. Not sure how this would look like.

Marigold:
Ideas 1. and 2. sound good (2. is more slick in my opinion). I do this quite often, too, and I tend to add breakpoints `__import__('ipdb').set_trace()` and print metadata. Having such a method would save me some time.

@pabloarosado there are too many assignees :), pick one and let him do this.

Marigold:
Just added `tb.check_metadata()` (feel free to modify it as you want @pabloarosado, I didn't notice you assigned yourself)

pabloarosado:
Hey @Marigold thanks for the heads-up. I'm trying something here, if it works out, I'll do a PR soon.

Marigold:
Thanks to @pabloarosado, we can now set `Table.DEBUG = True` and see where it fails. See https://github.com/owid/etl/pull/1697

PR #1697, opened 2023-09-29
🎉 add a way to debug when having metadata issues

pabloarosado:
Possible fix for https://github.com/owid/etl/issues/1679

I was trying different things that didn't work. But I think this approach may be enough for now:

While working on an ETL step, manually define `Table.DEBUG = True`, and then:
* When running a line of code that creates a table column without an origin, a warning is raised.
* When displaying a table that has any column without an origin, a warning is raised.
* When loading a table that has indicators without origins, a warning is raised.

@Marigold please let me know if you have other ideas, thanks!


Marigold (approved):
Nice, let's give it a try! I wonder if we could use env variable `DEBUG` and have it turned on all the time when working locally (I haven't tried it, curious whether it'd be spamming me too much or not). Alternatively have both, but have Table.DEBUG default to environ.DEBUG.

pabloarosado:
Hi @Marigold, thanks. Yes, I initially thought of using the DEBUG variable from .env (or creating another one called META_DEBUG). But then I realised that lib/catalog doesn't read from ETL .env, right?

pabloarosado:
Also, @Marigold I was about to merge and noticed that CI was running 222 ETL steps. Is that expected? If so, I'd wait for the end of the day to merge this PR.

Marigold:
> But then I realised that lib/catalog doesn't read from ETL .env, right?

Yes, exactly, I got confused by it at first, too. Perhaps we can just load it from `os.environ['DEBUG']`.

> I was about to merge and noticed that CI was running 222 ETL steps. Is that expected? If so, I'd wait for the end of the day to merge this PR.

It's unrelated, feel free to merge this. (I'm working on fixing ETL)

PR #1707, opened 2023-10-02
:sparkles: use Jinja for description_key and other fields

Marigold:


etl/grapher_helpers.py
```diff
@@ -174,6 +174,8 @@ def _yield_wide_table(
             for k in (
                 "description",
                 "description_short",
+                "description_from_producer",
+                "description_processing",
```
lucasrodes, line 178: any chance that `description_key` is missing here?

lucasrodes (commented):
LGTM ー but support for `description_key` might be missing?

etl/grapher_helpers.py
Marigold: `description_key` is a list which needs a special treatement and is right below.

lucasrodes: ah, my bad ー I scanned rapidly, and didn't see it

lucasrodes (approved):
LGTM! Thanks for doing this

PR #1717, opened 2023-10-03
:bug: fix aother_subtype_details column in flunet

Marigold:
`who/latest/flunet` step was failing because column `aother_subtype_details` contains mix of everything (floats, nans, strings, ...) and repack wasn't able to convert it to category. Converting it to string (while preserving nans) fixes it.

@spoonerf I'm gonna merge this now to fix ETL, but could you still take a look?


spoonerf (commented):
Looks good! We don't need some of these potentially problematic ones, so we could also drop out a few with:

`tb[tb.columns.drop(list(tb.filter(regex='comment|details')))]`

Issue #1700, opened 2023-09-29
metadata: Ensure origins are propagated on steps that used `to_table_metadata()`


pabloarosado:
After removing `origins` from `DatasetMeta`, now neither tables nor datasets have origins. This means that `snap.to_table_metadata()` doesn't have origins. Therefore, when this function is used (in various meadow steps), we need to ensure that also `origin` is passed. Otherwise, the indicators in the resulting table will have no origins.

Marigold:
@pabloarosado let me know if you need help. I didn't realise we might be using `.to_table_metadata()` outside `snap.read` method.

We have a few options for how to pass the origin:

1. Pass it explicitly as `pr.read_csv(snap.path, metadata=snap.to_table_metadata(), origin=snap.origin)`
2. Pass `SnapshotMeta` instead of `TableMeta` as `pr.read_csv(snap.path, snapshot_metadata=snap.metadata)`
3. Add `origins` to `Table` class (this would be similar to what we had before, but it would be somehow "hidden" from users)

I guess it doesn't matter much whether we choose 1 or 2, most steps load data using snapshot methods anyway.

Marigold:
We went with the first option that explicitly passes origin. It's not perfect, but it works.

PR #1718, opened 2023-10-03
:bug: propagate origins on steps that use to_table_metadata()

Marigold:
Fix for https://github.com/owid/etl/issues/1700.

* Added `snap.ExcelFile` object similar to how `snap.read_csv` works
* Fixed ggdc_maddison (thank you @pabloarosado)

Implementing `snap.ExcelFile` made me realise how annoying is it to have `Snapshot` in `etl` and not in `owid.catalog`. I'm not sure how hard would it be to move it there (it uses a few things from etl and isn't really needed for owid.catalog), but perhaps would be worth a try. @pabloarosado were you as annoyed as I when you were working on processing functions?


etl/steps/data/meadow/shift/2023-07-10/energy_production_from_fossil_fuels.py
```diff
@@ -16,10 +11,10 @@ def run(dest_dir: str) -> None:
     # Load inputs.
     #
     # Retrieve snapshot.
-    snap = cast(Snapshot, paths.load_dependency("energy_production_from_fossil_fuels.csv"))
+    snap = paths.load_snapshot("energy_production_from_fossil_fuels.csv")
 
     # Load data from snapshot.
-    tb = pr.read_csv(snap.path, metadata=snap.to_table_metadata(), underscore=True)
+    tb = snap.read_csv(snap.path, underscore=True)
```
pabloarosado, line 17: Here `snap.path` is unnecessary.

pabloarosado (approved):
Looks beautiful, thanks a lot!

PR #1721, opened 2023-10-03
:sparkles: use observed=True by default in groupby

Marigold:


lib/catalog/owid/catalog/tables.py
```diff
@@ -912,8 +912,12 @@ def __ipow__(self, other: Union[Scalar, Series, variables.Variable]) -> "Table":
...
         return super().sort_index(*args, **kwargs)  # type: ignore
 
-    def groupby(self, *args, **kwargs) -> "TableGroupBy":
-        return TableGroupBy(pd.DataFrame.groupby(self.copy(deep=False), *args, **kwargs), self.metadata, self._fields)
+    def groupby(self, *args, observed=True, **kwargs) -> "TableGroupBy":
+        """Groupby that preserves metadata. Uses `observed` by default to avoid
+        fix behavior with categoricals."""
+        return TableGroupBy(
+            pd.DataFrame.groupby(self.copy(deep=False), *args, observed=observed, **kwargs), self.metadata, self._fields
+        )
```
pabloarosado, line 920: I'm not sure if this is a good idea. Usually, we try to respect the default pandas behaviour. It could be bad training for us (in case we encounter a similar situation when working with true dataframes). A safer option would be to raise a warning if the dtype of any of the columns in `self.by` (or `self.by` itself if it's a string) is "category". That's the only case where this is relevant, as far as I understand.
But if you don't agree, I don't have a strong opinion.

pabloarosado (approved):
Looks good, although not sure about your proposed solution (see comment).

lib/catalog/owid/catalog/tables.py
Marigold: I don't have a strong opinion either. I wonder if "users" are even aware of this behaviour and if we there's a case where we really need this. Even though I think `observed=True` would be harmless, I'll convert it to a warning just to see cases where we use it.

PR #1723, opened 2023-10-04
:left_arrow: backport share of agriculture in GDP

Marigold:
Backported with

```
ENV=.env.prod backport-migrate --dataset-id 3009 --namespace agriculture --version 2017-03-08 --short-name share_of_agriculture_in_gdp
```

### DIFF
--- a/dag/migrated.yml
+++ b/dag/migrated.yml
@@ -16,3 +16,7 @@ steps:
   - data://garden/education/2017-09-30/public_expenditure
   data://garden/education/2017-09-30/public_expenditure:
   - snapshot://education/2017-09-30/public_expenditure.feather
+  data://grapher/agriculture/2023-10-04/share_of_agriculture_in_gdp:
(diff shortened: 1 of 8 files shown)


dag/migrated.yml
```diff
@@ -16,3 +16,7 @@ steps:
   - data://garden/education/2017-09-30/public_expenditure
   data://garden/education/2017-09-30/public_expenditure:
   - snapshot://education/2017-09-30/public_expenditure.feather
+  data://grapher/agriculture/2017-03-08/share_of_agriculture_in_gdp:
```
pabloarosado, line 19: I think it would be better to use the version "2023-10-04", for consistency. Would that be a problem?

pabloarosado (approved):
Thanks, it looks good. I suggested to change the version, but it's up to you, feel free to merge it like this (and I guess moving from sources to origins is not worth it).

PR #1727, opened 2023-10-04
We should only raise a warning on groupby if observed is False

pabloarosado:
We should only raise a warning on groupby if observed is False.


Marigold (commented):
Ah, sorry 🤦

PR #1732, opened 2023-10-04
📊 population: estimate former USSR population using estimates on current countries

lucasrodes:
A user reported an inconsistency in the population data from former soviet unions and the USSR. In particular, the sum from the former states differed from the population reported for the USSR.

I've opened this issue with more details: https://github.com/owid/owid-issues/issues/1232

In the long term, we should stop using Gapminder's dataset. For now, I've changed the code for this case to estimate the USSR's population based on the aggregated sum from the former soviet states' estimates.

I had to adapt some parts of the codebase to comply with the new checks (e.g. index).

### USSR with old data
population (1)

### USSR with new data

population (2)


etl/steps/data/garden/demography/2023-03-31/population/__init__.py
```diff
@@ -212,20 +222,50 @@ def add_world(df: Table) -> Table:
...
+        year_filter = year_filter[year_filter].index.tolist()
+        tb_ = tb_[tb_["year"].isin(year_filter)]
+        # Perform operations
+        tb_ = tb_.groupby("year", as_index=False).agg(
+            {"population": sum, "source": lambda x: "; ".join(sorted(set(x)))}
+        )
+        tb_["country"] = former_country_name
+        # Add to main table
+        tb = pr.concat([tb, tb_], ignore_index=True)
+    return tb
```
pabloarosado, line 268: The result is roughly consistent with the population from [wikipedia](https://en.wikipedia.org/wiki/Demographics_of_the_Soviet_Union#Population).

pabloarosado (approved):
Looks good. Some minor suggestions below. This is quite a significant change to the population data. Should we consider creating a new version? It would be good to keep the old version for comparison, but it's quite a big undertaking, probably not worth it. At least, it would be good to add a chart to [the issue](https://github.com/owid/owid-issues/issues/1232) showing the old and new populations for the USSR, for the records (since we may detect big changes in a bunch of per capita indicators).
If we simply overwrite the current version, please make sure to merge this at the end of the day, since it will trigger a big ETL run.

lucasrodes:
@pabloarosado I think creating a new version for cases like this, might be a bit of an overkill. Several charts will need to be reviewed. Plus, it is not really a new version, but a "hot-fix" like update, where we have fixed the values for a particular entity: USSR.

However, I agree that it'd be nice to graphically compare both trends and ensure that that's the only change.

lucasrodes:
@pabloarosado, FYI, I suspect that I will be working on a new version soon in https://github.com/owid/owid-issues/issues/1232. I think the population dataset needs a careful review for old data, where we combine different sources which can lead to some inconsistencies.

pabloarosado:
Hey @lucasrodes if you don't mind, could you remove the "." in all imports after merging https://github.com/owid/etl/pull/1741 ?

PR #1735, opened 2023-10-05
🐛 war: fix bug in number of new conflicts with hostility=all

lucasrodes:
Fix bug that prevents `number of new conflicts` to be correctly estimated. At the moment, it is an all-zero timeseries.

See this bug in the chart in the data page: https://owid.cloud/admin/datapage-preview/812693

image

Corrected version in staging: http://staging-site-lucas/admin/datapage-preview/816451


spoonerf (approved):
Looks good to me!

PR #1736, opened 2023-10-05
:scroll: Add docs about metadata YAML structure and workflow

Marigold:
Lucas made me do it...

(Feel free to edit/rewrite anything and commit back to this PR)


docs/architecture/metadata/structuring-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+    variables:
+      my_var_1:
+        # Final description will be First, Second, Third line
+        description_key:
+          - Third line.
+        presentation:
+          # Final tags will be Internet, Energy
+          topic_tags:
+            - Internet
+```
```
pabloarosado, line 119: This logic makes sense as long as we also have the option to override common values. We could do that maybe using the override operator `<<:`?

docs/architecture/metadata/structuring-yaml.md
```diff
@@ -0,0 +1,171 @@
...
+tables:
+  my_table:
+    variables:
+      my_var_1:
+        description: |-
+          This is a description.
+          { additional_info }
+```
+
+There are also special variables like `{ TODAY }` that can be used for automatically updated datasets.
```
pabloarosado, line 139: I think we should never use `TODAY` (except for scripts that run literally every day), because every random day we run the step it will change the file and make the file and its dependencies dirty, right? If this is correct, and we should discourage this use, maybe we should remove this from the docs.

docs/architecture/metadata/workflow.md
```diff
@@ -0,0 +1,29 @@
+## Live Reloading of a Data Page
+
+When modifying a YAML file for a Data page, executing the `etl ... --grapher` command to visualize updates can be inefficient. The following steps outline a streamlined approach for rapid development and testing.
+
+### Example Step
+For illustration, the step `grapher/gcp/2023-07-10/global_carbon_budget` is considered. The corresponding YAML file, `global_carbon_budget.meta.yml`, is assumed to be edited within the grapher channel and previewed on a [staging Data page](http://staging-site-mojmir/admin/datapage-preview/738081).
+
+### Initial Configuration
+- Add `DEBUG=1` to your `.env.myname` file and any other `.env.*` files in use to expedite ETL operations.
```
pabloarosado, line 9: What does DEBUG do? I've just tried with and without, and the difference is small:
```
WITH DEBUG=1
--- Detecting which steps need rebuilding...
OK (5.1s)
--- Running 3 steps:
--- 1. data://garden/gcp/2023-09-28/global_carbon_budget...
OK (17.3s)

--- 2. data://grapher/gcp/2023-09-28/global_carbon_budget...
OK (0.7s)

--- 3. grapher://grapher/gcp/2023-09-28/global_carbon_budget...
...
OK (29.3s)

WITHOUT DEBUG=1
--- Detecting which steps need rebuilding...
OK (4.7s)
--- Running 3 steps:
--- 1. data://garden/gcp/2023-09-28/global_carbon_budget...
OK (18.9s)

--- 2. data://grapher/gcp/2023-09-28/global_carbon_budget...
OK (2.1s)

--- 3. grapher://grapher/gcp/2023-09-28/global_carbon_budget...
...
OK (28.8s)
```

Maybe the global carbon budget is a bad example. I often wanted to see the result of some metadata edits in the data page, but the ETL step was halfway uploading to S3, and I had to wait for it to finish and be executed again automatically. Therefore I stopped using `--watch` and executed the steps always manually. I think `--watch` will be much more useful for shorter steps.

Also, I prefer working on my local grapher because the grapher step runs way faster:
```
WITHOUT DEBUG=1 but on local grapher
--- Detecting which steps need rebuilding...
OK (4.1s)
--- Running 3 steps:
--- 1. data://garden/gcp/2023-09-28/global_carbon_budget...
OK (18.5s)

--- 2. data://grapher/gcp/2023-09-28/global_carbon_budget...
OK (1.9s)

--- 3. grapher://grapher/gcp/2023-09-28/global_carbon_budget...
...
OK (12.4s)
```

docs/architecture/metadata/workflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+### File Monitoring
+- Execute the ETL command with the `--watch` flag. This monitors the YAML file for changes and automatically re-executes the corresponding step.
+
+### Data Filtering
+- If the dataset contains numerous indicators, consider using `GRAPHER_FILTER=consumption_emissions_per_capita` to filter the data to only relevant variables. This is optional for smaller datasets.
```
pabloarosado, line 15: This is really useful!! But still, the garden step alone takes 19 seconds, so this dataset is not a good example.

docs/tutorials/backport.md
```diff
@@ -1,21 +1,33 @@
 !!! warning "This tutorial might be partial or incomplete. Please check with the team if you have questions."
 
-Before the existance of the ETL, datasets were directly uploaded to our Grapher admin site. Sometimes, we want to use these datasets in ETL. However, they are not available by default as they were never imported via the ETL. To make them available in ETL we use _backport_.
+Before ETL came along, datasets were uploaded directly to our Grapher admin site. If you want to leverage these datasets in the ETL, they won't be available out of the box, since they weren't initially imported via ETL. You have two main options to bring them into ETL:
 
+1. Migrate them into the ETL pipeline.
+2. Move them to Fast-Track.
 
-The first step is getting them to Snapshot using:
+Choose Fast-Track if the dataset requires author edits. Otherwise, it's more straightforward to migrate to ETL.
```
pabloarosado, line 8: This sounds as if fast-track gives you more freedom than migrating to ETL, which is not true. It's not clear to me what the benefit of fast-track is. Maybe "Fast-Track is recommended for users who prefer working on Google sheets rather than the ETL environment." (e.g. authors). Is this right?

docs/tutorials/backport.md
```diff
@@ -1,21 +1,33 @@
...
+
+
+## How to Migrate to ETL
+
+Automatically generate ETL steps for a specific dataset (xxx) with the following command:
 
 ```
-bulk_backport
+ENV=.env.prod backport-migrate --dataset-id xxx --namespace your_namespace --version your_version --short-name your_short_name
```
pabloarosado, line 16: I'd suggest renaming `backport-migrate` -> `etl-backport-migrate`. It's easier at least for me to be in the ETL poetry shell and type `etl-` and tab, to remember the options available. But up to you.

pabloarosado: Can you please explain what `xxx` and `short_name` mean?

pabloarosado (approved):
Thanks a lot for doing this, I've learned a lot of useful things! I've also committed some small changes.

docs/architecture/metadata/workflow.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Optional Optimization
+- Include the `--only` flag in the command to further improve performance by avoiding dependency checks.
+
+### Command Summary
+```bash
+ENV=.env.myname GRAPHER_FILTER=consumption_emissions_per_capita etl grapher/gcp/2023-07-10/global_carbon_budget --grapher --watch --only
+```
+NOTE: Working on your local grapher instead of a remote grapher will also reduce upload times. One option is to work on local grapher to begin with, and move to your staging server at a later point to be able to share the result with other colleagues.
+
```
lucasrodes, line 25: Use the admonition instead

```suggestion

!!! note
    Working on your local grapher instead of a remote grapher will also reduce upload times. One option is to work on local grapher to begin with, and move to your staging server at a later point to be able to share the result with other colleagues.
```

docs/architecture/metadata/workflow.md
```diff
@@ -0,0 +1,30 @@
...
+### Usage
+After initiating the above command, changes made to the YAML file can be reviewed by refreshing the [staging Data page](http://staging-site-mojmir/admin/datapage-preview/738081). The updates should be reflected promptly.
+
+!!! info
+    Note: If editing the YAML file in the **garden channel** and the step execution time is long, refresh latency will be dependent on the step's runtime. In such cases, consider developing the YAML file in the **grapher channel** before moving it into the garden channel (if the table and indicator names are identical).
```
lucasrodes, line 30: Unsure if this should be an info or note admonition? If info, let's remove the "Note: " bit. Otherwise, set it to `!!! note`.

You can find a complete list of available admonitions here: https://squidfunk.github.io/mkdocs-material/reference/admonitions/

docs/architecture/metadata/workflow.md
```diff
@@ -0,0 +1,30 @@
+## Live Reloading of a Data Page
```
lucasrodes, line 1: It looks like it only covers data-page specific use case? If so, I'd change the entry name from Workflow to something else, and level up the headings.

Or if you think this should cover other use cases in the future, I'd clarify at the top of the page that we only talk about data-page but will include other types of workflows in the future.

lucasrodes (changes requested):
Thanks for doing this, Mojmir! This is a great resource.

I've left some suggestions. 

Also, a feature you may want to use if you wanted to add comments and more fine-grained descriptions to the yaml templates: https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#code-annotations

docs/architecture/metadata/workflow.md
lucasrodes, line 25: Use admonition instead

```suggestion
!!! note 
    Working on your local grapher instead of a remote grapher will also reduce upload times. One option is to work on local grapher to begin with, and move to your staging server at a later point to be able to share the result with other colleagues.

```

lucasrodes (changes requested):
additional suggestion

lucasrodes:
Additionally, I'd suggest adding the list of metadata fields that support each feature (especially Jinja, since I believe the rest is supported in all fields).

docs/architecture/metadata/workflow.md
Marigold: `DEBUG=1` gets rid of 2s overhead per executed step (updated docs), that's all. It's true that `--watch` won't really work for steps with long execution time. I was using it mostly for editing grapher YAML files.

> Also, I prefer working on my local grapher because the grapher step runs way faster:

Weird, it makes no difference to me. JSON files are uploaded to S3 even if you work locally, so I guess it must be database latency in your case (and it's strange that it's so much).

docs/architecture/metadata/structuring-yaml.md
Marigold: > it will change the file and make the file and its dependencies dirty, right?

Not really. The file checksum is calculated using the string "TODAY", not the actual date, so TODAY will only change if dependencies get updated. We use this for covid automatic updates.

docs/architecture/metadata/workflow.md
Marigold: True. I was using this on grapher steps. I've added a note that you could test your metadata in the grapher step and then move it to garden.

docs/tutorials/backport.md
Marigold: You're right, I've clarified that.

docs/tutorials/backport.md
Marigold: Yeah, the naming isn't great. I'd rather do it as part of a different PR though (we could even use subcommands like `etl backport-migrate`).

> Can you please explain what xxx and short_name mean?

Clarified that. It's just dataset id and choosing short_name is up to you.

docs/architecture/metadata/workflow.md
Marigold: Good point. Fixed.

docs/architecture/metadata/structuring-yaml.md
Marigold: Mentioned `<<:` there and also updated that metadata is overwritten instead of merging.

lucasrodes (approved):
LGTM, thanks @Marigold!

Issue #1733, opened 2023-10-04
Add YAML section `definitions.common` that merges its content with all indicators


Marigold:
Contents of section `definitions.common` in YAML should be merged with all indicators.

Some ideas:

1.  I suppose common could have as many inputs as indicator metadata fields. I was wondering if we may want a common for each table...
```yaml
definitions:
  common:
    global_emissions: [the name of one table]
      description_key: [common description key for all indicators in the table]
    land_use_emissions: [another table]
      description_key: [common description key for all indicators in the table]
...
```

2. Would be nice if we could choose between two behaviours: (1) anything in commons is propagated to indicators’ fields, or (2) it is appended to the indicators fields’ content (e.g. string fields are added at the end, description_key bullet points are added to those already present, etc.)

3. It would be super cool if we could somehow choose to which indicators the definitions apply. It is nice that by default they apply to all of the indicators, but maybe sometimes we want to apply them to a subset of the indicators.

PR #1737, opened 2023-10-05
✨ add definitions.common to YAML metadata

Marigold:
Implementation of https://github.com/owid/etl/issues/1733. Its behaviour is best seen on [unit tests](https://github.com/owid/etl/pull/1737/files#diff-2b4afe7f7246c478125031da2d141d6d30a6c20b0aa63d6384bf4c8daab7502fR31).

### DIFF
--- a/lib/catalog/tests/test_yaml_metadata.py
+++ b/lib/catalog/tests/test_yaml_metadata.py
@@ -79,29 +79,23 @@ tables:
 
     ym.update_metadata_from_yaml(t, path, "test")
 
-    assert t.a.m.sources == []
-    assert t.b.m.sources == []
-    assert len(t.a.m.origins) == 2
-    assert len(t.b.m.origins) == 1
-    assert t.a.m.description_key == ["C", "D"]
-    assert t.b.m.description_key == ["D"]
-    assert t.a.m.description_processing == ["A", "B"]
-    assert t.b.m.description_processing == ["A", "B"]
-    assert t.a.m.description_short == "A desc short"
-    assert t.b.m.description_short == "Default desc short"
-    assert t.a.m.display["numDecimalPlaces"] == 1
-    assert t.b.m.display["numDecimalPlaces"] == 0
+    assert t.a.m.to_dict() == {
+        "description_short": "A desc short",
+        "description_key": ["C", "D"],
+        "origins": [{"producer": "Origin2", "title": "Title2"}, {"producer": "Origin1", "title": "Title1"}],
+        "display": {"numDecimalPlaces": 1},
+        "presentation": {
+            "grapher_config": {"subtitle": "A subtitle", "selectedEntityNames": ["France"]},
+            "attribution": "A presentation attribution",
+        },
+        "description_processing": ["A", "B"],
+    }
 
-    # display is completely overwritten, not merged
-    assert "conversionFactor" not in t.a.m.display
-    assert t.b.m.display["conversionFactor"] == 2
-
-    # merging presentation
-    assert t.a.m.presentation.attribution == "A presentation attribution"
-    assert t.b.m.presentation.attribution is None
-
-    # merging grapher_config
-    assert t.a.m.presentation.grapher_config["selectedEntityNames"] == ["France"]
-    assert t.b.m.presentation.grapher_config["selectedEntityNames"] == ["France"]
-    assert t.a.m.presentation.grapher_config["subtitle"] == "A subtitle"
-    assert "subtitle" not in t.b.m.presentation.grapher_config
+    assert t.b.m.to_dict() == {
+        "description_short": "Default desc short",
+        "description_key": ["D"],
+        "origins": [{"producer": "Origin1", "title": "Title1"}],
+        "display": {"conversionFactor": 2, "numDecimalPlaces": 0},
+        "presentation": {"grapher_config": {"selectedEntityNames": ["France"]}},
+        "description_processing": ["A", "B"],
+    }
--- a/lib/catalog/tests/test_yaml_metadata.py
+++ b/lib/catalog/tests/test_yaml_metadata.py
@@ -74,6 +74,7 @@ tables:
         f.write(yaml_text)
 
     t = Table({"a": [1, 2, 3], "b": [1, 2, 3]})
+    t.a.metadata.description_short = "Will be overwritten"
     t.a.metadata.sources = [Source()]
 
     ym.update_metadata_from_yaml(t, path, "test")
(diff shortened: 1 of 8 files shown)


lib/catalog/tests/test_yaml_metadata.py
```diff
@@ -0,0 +1,107 @@
...
+            subtitle: A subtitle
+""".strip()
+
+    path = tmp_path / "test.yaml"
+    with open(path, "w") as f:
+        f.write(yaml_text)
+
+    t = Table({"a": [1, 2, 3], "b": [1, 2, 3]})
+    t.a.metadata.description_short = "Will be overwritten"
+    t.a.metadata.sources = [Source()]
```
pabloarosado, line 78: I'd recommend adding
```
t.b.metadata.origins = [Origin(producer="Producer", title="Title")]
```
And then asserting that origins get appended also to `b` (that had origins prior to the application of the yaml file). And instead of asserting with the number of origins, assert with the actual origins (the order is also relevant).

pabloarosado (commented):
Thanks a lot for this! It looks fine, but I think it's a bit confusing. Sometimes the common metadata replaces existing values (e.g. sources, description_short) and sometimes the common metadata is appended (e.g. origins and description_key). I think this can cause unexpected results.

Maybe we could have three sections:
```
definitions_prepend:
    ...

tables:
  ...

definitions_replace:
    ...

definitions_append:
    ...
```
This could apply to sources, origins, lists of strings and even strings. I haven't thought it very deeply, but maybe we can figure out something along these lines.

lucasrodes:
on Pablo's suggestion, wb having all of those inside definitions? E.g definitions.prepend, definitions.append, etc.

pabloarosado:
> on Pablo's suggestion, wb having all of those inside definitions? E.g definitions.prepend, definitions.append, etc.

Yes, totally. What I suggested doesn't make much sense. It could simply be:
```
definitions:
  common_prepend:
    ...
  common_replace:
    ...
  common_append:
    ...

tables:
  ...
```

Marigold:
> Sometimes the common metadata replaces existing values (e.g. sources, description_short) and sometimes the common metadata is appended (e.g. origins and description_key).

It has a consistent logic:
- lists are appended (origins, sources, description_processing, description_key)
- non-lists overwrite (but only if they're defined under `tables:...`, definitions in `common:...` don't overwrite)
- setting [] explicitly removes (this is really only useful for getting rid of sources)

I'd rather not make this more complex than it needs to be. Are you sure you'll need all prepend / append / replace combinations?

lib/catalog/tests/test_yaml_metadata.py
Marigold: Good idea. I rewrote it and I am comparing dictionaries now which easier on the eyes.

pabloarosado:
> > Sometimes the common metadata replaces existing values (e.g. sources, description_short) and sometimes the common metadata is appended (e.g. origins and description_key).
> 
> It has a consistent logic:
> 
> * lists are prepended (origins, sources, description_processing, description_key)
> * non-lists overwrite (but only if they're defined under `tables:...`, definitions in `common:...` don't overwrite)
> * setting [] explicitly removes (this is really only useful for getting rid of sources)
> 
> I'd rather not make this more complex than it needs to be. Are you sure you'll need all prepend / append / replace combinations? For instance the parameter `if_origins_exist` has been added by me, but I've never used it in the end. (And was thinking about removing it)

Hi @Marigold, I'd say that the logic currently implemented is already too complex. We need to be totally sure of what's been replaced and what isn't.

Also, if I understand correctly, with your current logic, as soon as you have one "special indicator" (that requires its own set of `description_key` bullet points) you can't define `definitions.common.description_key`: Those common definitions will be imposed to all indicators, including the special one.

I think the most transparent option should be to apply the logic of **always replace**, as follows:
* Whatever you write in the `definitions.common` section of the yaml file **always replaces** the metadata that was on the table already.
* Whatever you write in the `tables` section of the yaml file **always replaces** what was written in the `definitions.common` section of the yaml file.
With this logic there is no appending and no prepending, and it lets you have common definitions that apply to all indicators, and then overwrite for special cases.

We can then consider adding the options to prepend or append, which would also be useful, especially for `description_key` (origins in general should never be touched). Maybe we can figure out some other way to achieve that, with special YAML operators like `<<`.

pabloarosado:
With the **always replace** logic, you can still prepend and append things for special indicators:
```
definitions:
  common:
    description_key: &common-description-key
      - D 1
      - D2

tables:
  table_1:
    var_1:  # This variable will have the common description_key.
      title: Var 1
    var_2:  # This variable will have the common description_key, plus an additional point.
      description_key: 
        - *common-description-key
        - An additional point
    var_3:  # This variable will have a totally different description_key.
      description_key:
        - Special point.
    var_4:  # This variable will have no description key.
      description_key: []

```
On a related topic, I keep thinking that having `description_key` as a list of points instead of a string causes way too many inconveniences. It should instead be a string of bullet points. Then we would apply the same logic to all fields (except sources and origins, which should mostly be untouched), and figure out common ways to combine texts.

lucasrodes:
Nice work! This is going to be sooooo helpful!

> On a related topic, I keep thinking that having description_key as a list of points instead of a string causes way too many inconveniences. It should instead be a string of bullet points. 

I am not fully convinced by this. What's the motivation for having it as a free-text field? IMO, having a list, forces researchers/data to synthesise the text into bits of key info, which I think is the goal of this field.

pabloarosado:
> Nice work! This is going to be sooooo helpful!
> 
> > On a related topic, I keep thinking that having description_key as a list of points instead of a string causes way too many inconveniences. It should instead be a string of bullet points.
> 
> I am not fully convinced by this. What's the motivation for having it as a free-text field? IMO, having a list, forces researchers/data to synthesise the text into bits of key info, which I think is the goal of this field.

We should all follow the guidelines, which could explicitly say that bullet points are mandatory (at least for curated indicators).
The motivation for having free-text is that we would treat all metadata fields in the same way, which implies less headache. But anyway, we can discuss this elsewhere, for now we can keep `description_key` as a list.

Marigold:
@pabloarosado All right, you've changed my mind! I've changed it to overwriting, it sounds quite intuitive. Let's see how it works in practice.

Marigold:
@pabloarosado I've changed it to overwriting. Is it ok to get merged? (We can iterate later)

pabloarosado:
> @pabloarosado I've changed it to overwriting. Is it ok to get merged? (We can iterate later)

@Marigold feel free to merge, unless you want me to review it again.

PR #1741, opened 2023-10-05
:bug: add proper path to pythonpath if step is a module

Marigold:
With this, you'll be able to import files without relative imports. Absolute imports let you use VSCode code interpreter.


pabloarosado (approved):
Awesome, thanks! Maybe revert the changes in population, or otherwise it will trigger a bit ETL run. Lucas is making changes on that dataset in a separate PR: https://github.com/owid/etl/pull/1732

Marigold:
> Maybe revert the changes in population, or otherwise it will trigger a bit ETL run

Thanks for spotting it, that was close 😅

PR #1746, opened 2023-10-09
Fix wrong URL when using etl-metaplay on local grapher

pabloarosado:


apps/metadata_playground/app.py
```diff
@@ -57,6 +57,9 @@ def run_steps() -> None:
 def get_data_page_url() -> str:
     """Get data page URL"""
     HOST = config.DB_HOST
+    if HOST == "127.0.0.1":
+        # The following port is defined in one of owid-grapher's config files.
+        HOST = "localhost:3030"
```
lucasrodes, line 62:
```suggestion
        HOST = f"localhost:{config.DB_PORT}"
```

pabloarosado: That `DB_PORT` is a different parameter, as far as I understand. I think the actual port (3030) is not defined anywhere in ETL, but in owid-grapher.

lucasrodes: you are right!

Issue #1753, opened 2023-10-10
wizard: Running a snapshot silently fails


pabloarosado:
### One-liner
Wizard Snapshot fails to execute a snapshot but on the browser it says "Snapshot uploaded!".

### Context & details
Creating the snapshot .py and .dvc files goes well, but when clicking on the button to execute the snapshot, on the browser I see "Snapshot uploaded!", however the snapshot was not properly executed. When I look at the terminal, it shows an error (for an example error, see this other issue: https://github.com/owid/etl/issues/1752).

lucasrodes:
hi @larsyencken,
In https://github.com/owid/etl/pull/1755 I improved the message after running the snapshot step.

However, the error is not caught from the web app at the moment, and the user still needs to check the terminal (that's what the updated message says). 

I can look into improving this.

lucasrodes:
Improvement shipped in

- https://github.com/owid/etl/pull/1809

lucasrodes:
Should be fixed now

Issue #1752, opened 2023-10-10
wizard: Wizard snapshot cannot handle citation full with line breaks


pabloarosado:
### One-liner
Wizard snapshot cannot handle citation full with line breaks.

### Context & details
The .dvc file that is created when the `citation_full` entered through the browser contains multiple paragraphs is not properly indented. This issue probably affects other fields, not just `citation_full`.

lucasrodes:
hi @pabloarosado,
Can you provide your example? snapshot / how-to-reproduce?

pabloarosado:
> hi @pabloarosado, Can you provide your example? snapshot / how-to-reproduce?

I couldn't give an example since the code wasn't yet pushed. [Here's the dvc file](https://github.com/owid/etl/blob/add-net-zero-targets-data/snapshots/emissions/2023-10-10/net_zero_tracker.xlsx.dvc) that failed (after correcting `citation_full`).

To reproduce, run `etl-wizard --dummy-data` and then, in `citation_full`, add a line break. For example:
```
Dummy description for a dummy snapshot.
Another sentence.
```
This will generate a broken dvc file.

PR #1755, opened 2023-10-10
🐛 wizard: minor fixes

lucasrodes:
- Fixes indentation in `origin.citation_full` (see #1752).
- Improves uploading message in snapshot step (see #1753).
- Update docs and schema to stop encouraging the usage of `date_published: latest`.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,21 +1,19 @@
 {
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
         {
             "name": "Python: bulk_packport",
             "type": "python",
             "request": "launch",
-            "program": "debug.sh",
+            "program": "/home/daniel/.local/bin/poetry",
+            "args": ["run", "bulk_backport", "--dry-run"],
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -1,19 +1,21 @@
...
             "type": "python",
             "request": "launch",
-            "program": "/home/daniel/.local/bin/poetry",
-            "args": ["run", "bulk_backport", "--dry-run"],
+            "program": "debug.sh",
             "console": "integratedTerminal",
             "justMyCode": true,
-            "python": "/usr/bin/python",
-
+            "python": "/home/lucas/repos/etl/.venv/bin/python"
```
pabloarosado, line 11: I've never understood the point of this file...

apps/wizard/templating/snapshot.py
```diff
@@ -806,4 +806,6 @@ def run_snap_step() -> None:
     # Run step
     with st.spinner(f"Running snapshot step... {command_str}"):
         subprocess.call(args=commands)
-    st.write("Snapshot uploaded!")
+    st.write(
```
pabloarosado, line 809: Could you not catch the response from the subprocess call and create the message accordingly? If not, this solution is fine.

pabloarosado (approved):
I think you forgot to commit the changes on `schemas/definitions.json`

PR #1765, opened 2023-10-11
:bar_chart: migrate PIP to ETL

Marigold:
Migrate World Bank Poverty and Inequality Platform (PIP) to ETL (see the dataset [on staging](http://staging-site-mojmir/admin/datasets/6262)). 

We need to migrate it because we have JSON-based data pages based on this dataset and we need it in ETL in order to migrate JSON-based data pages to ETL-based. 

I ran chart revisions on staging and all looked good.


paarriagadap (approved):
Thank you!

PR #1778, opened 2023-10-13
Create dummy monster dataset

pabloarosado:
This dataset is a playground to test how things look in the sources tab and in data pages. To play around with it we just need to add/edit/remove indicators in the YAML file of the garden step (no need to change any other file).

I added a few normal, interesting, and extreme cases. We can add more in the future.


Marigold (approved):
This is really cool! This is sort of a "visual" integration test that will be very helpful for development. (I'm a daily user of the other dummy dataset that shows where fields end up)

PR #1784, opened 2023-10-13
Fix to_csv failing when having 'index' as an argument

pabloarosado:
`Table.to_csv()` failed when passing `index=False`. This PR should fix that.

### DIFF
--- a/lib/catalog/owid/catalog/tables.py
+++ b/lib/catalog/owid/catalog/tables.py
@@ -192,9 +192,11 @@ class Table(pd.DataFrame):
             raise ValueError(f'filename must end in ".csv": {path}')
 
         df = pd.DataFrame(self)
-        # if the dataframe uses the default index then we don't want to store it (would be a column of row numbers)
-        save_index = self.primary_key != []
-        df.to_csv(path, index=save_index, **kwargs)
+        if "index" not in kwargs:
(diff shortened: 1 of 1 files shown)


lucasrodes:
I wonder if we actually want to force Tables to have a primary ID always.

Table object was conceived for ETL, where we enforce index usage.

But I can dee how a user would load the table from the catalog, totally oblivious to the ETL project, process the table like a dataframe and then export ad a CSV.

I guess we might want to support index=False for this latter case.

pabloarosado:
> I wonder if we actually want to force Tables to have a primary ID always.
> 
> Table object was conceived for ETL, where we enforce index usage.
> 
> But I can dee how a user would load the table from the catalog, totally oblivious to the ETL project, process the table like a dataframe and then export ad a CSV.
> 
> I guess we might want to support index=False for this latter case.

Hi @lucasrodes, I think you are worried that a table, after some processing, may end up having an index that is different from its `primary_index`. If so, keep in mind that `primary_index` is calculated every time it's called as `[n for n in self.index.names if n]`, so I think there's no risk that `primary_index` becomes outdated.

The question is whether we want to assume `index=True` (as pandas does) when the index is dummy. Currently, we assume `False`, which is the most convenient and meaningful option, but it's different to pandas' default.

Marigold (approved):
Looks good. I don't know about any side effects this might cause.

PR #1785, opened 2023-10-16
Bump owid catalog version

pabloarosado:
Bump owid catalog version and update the changelong.

For context, I'm trying to use `catalog.find` in the `co2-data` repos, and it raises an error, which doesn't happen when using the same command from the current ETL. I haven't looked into it, but I suppose that upgrading `owid-catalog` should fix the issue.


Marigold (approved):
Looks good!

PR #1809, opened 2023-10-18
🐛 error message when failing to upload to snapshot fails

lucasrodes:
Improve error when snapshot is uploaded and fails.


pabloarosado (approved):
Thanks, looks good!

PR #1814, opened 2023-10-18
:hammer: disable unused open numbers datasets

Marigold:
We only use two datasets from [Open numbers initiative](https://github.com/open-numbers), but we still download and republish all the other ones. Those unused datasets use up **almost 28gb** in `.owid/git/`, and they slow down all `etl` runs because we have to compare etags against their github repos.

Disabling these datasets would considerably improve performance of our nightly and full ETL runs as well as automated ETL servers. The only reason for keeping them here was if someone was using them through `owid-catalog`, but I don't think anyone does that.

## TODO after merging
* [ ] remove `.owid/git/` on local and in `owid@etl-prod-1`

### DIFF
--- a/dag/open_numbers.yml
+++ b/dag/open_numbers.yml
@@ -1,197 +1,200 @@
 steps:
-  data://open_numbers/open_numbers/latest/bp__energy:
-    - github://open-numbers/ddf--bp--energy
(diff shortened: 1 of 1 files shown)


pabloarosado (approved):
Looks good! I copied them to the archive dag and ran `version_tracker` to check if there were missing dependencies, but there weren't. So it looks safe.

PR #1824, opened 2023-10-19
:bar_chart: Create Moatsos historical poverty dataset

paarriagadap:
Move the Moatsos historical poverty dataset pipeline to the ETL and include some smoothing suggestions by Max.

owid-issue: https://github.com/owid/owid-issues/issues/1208


etl/steps/data/garden/moatsos/2023-10-09/moatsos_historical_poverty.py
```diff
@@ -0,0 +1,107 @@
...
+    ].round()
+
+    tb = geo.harmonize_countries(
+        df=tb,
+        countries_file=paths.country_mapping_path,
+    )
+    tb = tb.set_index(["country", "year"], verify_integrity=True)
+
+    # Add metadata by code
+    tb = add_metadata_vars(tb)
```
paarriagadap, line 105: @pabloarosado Hi! I think when adding metadata by code I lose origins. How can I avoid that?

etl/steps/data/garden/moatsos/2023-10-09/shared.py
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    return meta
+
+
+def var_metadata_between(var, povline1, povline2) -> VariableMeta:
+    meta = VariableMeta(
+        title=f"{povline_dict[povline1]['title_between']}-{povline_dict[povline2]['title_between']} - {var_dict[var]['title']}",
+        description_short=f"""{var_dict[var]['description']}""",
+        description_key=f"""
```
paarriagadap, line 170: @pabloarosado Another quick question: When adding metadata by code, how can I add `description_key` fields to see them as a list?

etl/steps/data/garden/moatsos/2023-10-09/moatsos_historical_poverty.py
pabloarosado: There are a few things I don't understand in the code, but I can quickly see that `var_metadata_absolute` (could you please use a function name that describes the purpose of the function?) creates a new variable that does not have origins, so I suppose that's where you are losing them.

PR #1834, opened 2023-10-20
📊  Anthromes dataset

spoonerf:


spoonerf:
Hey @pabloarosado - very much not urgent but I am getting a bit stuck with the metadata/origin issues here. 

I _think_ it is because I am not using `snap.read()` in `etl/steps/data/meadow/papers/2023-10-20/anthromes_input.py` as it's an unusual file and the origins are lost at this point, but I'm not sure. 

There are still some other changes I will make to the code (so won't mark this ready for review yet) but I wanted to try and figure out the metadata issues first.

pabloarosado:
Hi @spoonerf the issue was indeed that you were creating a `Table` without any metadata or origin. You could just do a for loop over all columns and add the origin to each variable.
But instead, I've fixed that by creating a new `pr.read_df` function, analogous to `read_csv`, but for cases like this, where the input data is already a dataframe. Please let me know if you think of a better solution.

spoonerf:
Ah thanks @pabloarosado! That's very helpful!

Issue #1820, opened 2023-10-19
Create app etl-archive-steps


pabloarosado:
## Motivation

We are accumulating new versions of datasets, and may want to make it easy to archive old ones to keep the DAG clean.

## Proposal

Create a script/web UI that makes it easy to archive a series of steps, and double-checks that they're not used.

UPDATE: either it should be part of the step updater automatically, or it should be a one-click option from that UI

## Technical notes

- This would be similar to the current `scripts/move_steps_to_archive.py`, but it would also attempt to move steps from the active to the archive dag (and fail if `version_tracker` complains).
- This could be implemented as part of `StepUpdater`

pabloarosado:
Also, I've noticed that the current `scripts/move_steps_to_archive.py` doesn't move all files to the archive. For example, I see that any `shared.py` files are left behind. I suppose this make sense for safety, in case there were other steps in the same folder that shouldn't be archived.
But we could check if there are other steps in the same folder (which is in most cases not the case), and if not, archive everything.

larsyencken:
We note that we will want to make it easy to archive things, but we don't necessarily want to archive every step immediately after update; keeping old versions around in the active DAG is helpful for QA. We may still want helpers for archiving steps that aren't working any more.

larsyencken:
Based on this, suggesting we wait until archiving is more painful to automate it more.

Marigold:
I'd have found this really useful when migrating to pandas 2. There are like three faostat versions we keep in DAG, which means three times more refactoring. Archiving old versions would also make the ETL faster (especially for faostat steps). Ideally, we'd have "Archive" button in the dashboard that would move the steps automatically and warn about any leftover dependencies. (We could also have "Unarchive" just in case, but that might not be as useful in the end)

PR #1840, opened 2023-10-23
Move unused step files to the archive

pabloarosado:
As mentioned in https://github.com/owid/etl/issues/1820#issuecomment-1775314947, the script `scripts/move_steps_to_archive.py` left behind some unused files in `etl/steps/data` that should have been archived. This PR moves them to the archive.


pabloarosado:
> @pabloarosado please try to follow our team PR practices:
> 
> * [GitHub ettiquette](https://www.notion.so/owid/Git-GitHub-etiquette-68e824bed1a747e1a01f2b771c695bf7) for PR titles. For this case, you could use "🔨" .
> * Use "squash and merge" to ensure a cleaner public commit history.
> 
> Thanks

Thanks for the reminder, I've renamed this PR. I use the default PR merge operation for this project, which is "create a merge commit". If you think we should change it to squash and merge, then you could propose it in the data arch channel, and we can change the default.

Marigold:
@pabloarosado I don't think you can select a default one [in settings](https://github.com/owid/etl/settings), but it should always show the option you used last time.

lucasrodes:
> Thanks for the reminder, I've renamed this PR.

Thanks! Note that this is not really needed in terms of cleaning the commit history, as the commit message was already created.

> I use the default PR merge operation for this project, which is "create a merge commit". If you think we should change it to squash and merge, then you could propose it in the data arch channel, and we can change the default.

I've shared the details on Slack!

PR #1844, opened 2023-10-24
📊 Create IPCC CO2 emission factors dataset

pabloarosado:
Add snapshots, meadow, garden and grapher steps for CO2 emission factors.
This is a very small dataset that we used to have [here](https://owid.cloud/admin/datasets/970). I thought we would find more updated emission factors, but the best data I've found from [the official source](https://www.ipcc-nggip.iges.or.jp/EFDB/) comes from the same document, from 2006. However, I think those factors are still used as reference values.

### DIFF
--- a/snapshots/emissions/2023-10-24/emission_factors.py
+++ b/snapshots/emissions/2023-10-24/emission_factors.py
@@ -1,8 +1,8 @@
 """Script to create a snapshot of dataset.
 
 To create a snapshot of the dataset, follow these steps:
-* Go to https://www.ipcc-nggip.iges.or.jp/EFDB/
-* Click on the button "Export to XLS".
+* Go to https://www.ipcc-nggip.iges.or.jp/EFDB/find_ef.php
+* Search for "Export to XLS" in the page, and click on that button.
 * Run this snapshot using the flag "--path-to-file" followed by the path to the downloaded file.
 
 """
--- a/snapshots/emissions/2023-10-24/emission_factors.py
+++ b/snapshots/emissions/2023-10-24/emission_factors.py
@@ -1,0 +1,32 @@
+"""Script to create a snapshot of dataset.
+
+To create a snapshot of the dataset, follow these steps:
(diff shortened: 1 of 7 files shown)


snapshots/emissions/2023-10-24/emission_factors.xlsx.dvc
```diff
@@ -0,0 +1,36 @@
...
+      The Intergovernmental Panel on Climate Change (IPCC) Emission Factor Database (EFDB) is a library of emission factors and
+      parameters that can be used for estimation of national greenhouse gas
+      emissions/removals. For more details, see [the User Manual](https://www.ipccnggip.iges.or.jp/EFDB/documents/EFDB_User_Manual.pdf).
+    date_published: 2023-10-24
+
+    # Citation
+    producer: Intergovernmental Panel on Climate Change
+    citation_full: |-
+      The Intergovernmental Panel on Climate Change (IPCC) Emission Factor Database (EFDB) contains default data from IPCC Guidelines and data
+      from other sources (e.g., peer-reviewed papers) with background information.
```
spoonerf, line 17: Just an odd line break here which is showing up on the sources tab.

snapshots/emissions/2023-10-24/emission_factors.py
```diff
@@ -0,0 +1,32 @@
+"""Script to create a snapshot of dataset.
+
+To create a snapshot of the dataset, follow these steps:
+* Go to https://www.ipcc-nggip.iges.or.jp/EFDB/
+* Click on the button "Export to XLS".
```
spoonerf, line 5: I cannot see this button

spoonerf: Is this the link instead? https://www.ipcc-nggip.iges.or.jp/EFDB/find_ef.php?reset=

spoonerf (approved):
Looks good - just a couple of tiny points

PR #1852, opened 2023-10-24
:sparkles: add topic tags as enum entries to schema for autocomplete

danyx23:
This add the currently in the DB existing topic tags to the schema so you get autocomplete when editing tags in yaml files.

@Marigold can merging this cause issues if some yaml files currently violate the schema?


lucasrodes:
Thanks for doing this, Daniel!

Do you have any thoughts on how to maintain this aligned with the database table? I don't think this is a big deal, since tags aren't added frequently, but just wondering. 

We probably need some warning message whenever this list and the db table is miss aligned.

Marigold (approved):
Nice, I wish I had it when I was migrating data pages. I think this is good to go even if these tags might go out of sync one day (there's a link to datasette with tags, so updating them should be straightforward).

danyx23:
Yeah I think a little python script that fetches tags from datasette would be the way to go - but I think topic tags are added pretty rarely so let's do it manually for now

PR #1867, opened 2023-10-26
📊 AI: Epoch dataset update

veronikasamborska1994:
hey @Marigold ! This is going to be a very small update but I just wanted to see what you think would be the best approach in this situation :) 

I update this dataset every month or so and so far I haven't actually created new directories each time I've updated it as everything pretty much stayed the same and it seemed unnecessary. Now, I am going to create a couple of datapages based on this dataset so will be updating the garden step with the new metadata structure. Would you create a new directory for each step with the latest date ? or just for the steps that actually change dramatically? e.g.,  It seems like the grapher step could always stay the same so that I don't need to update the charts manually after uploading the dataset? 

p.s. I haven't updated the garden step metadata yet


Marigold:
> It seems like the grapher step could always stay the same so that I don't need to update the charts manually after uploading the dataset?

Yeah, definitely. There's no point in running chart revisions every month. You could start using `latest` as a version, i.e.
```
data://meadow/artificial_intelligence/latest/epoch:
- snapshot://artificial_intelligence/2023-10-26/epoch.csv
data://garden/artificial_intelligence/latest/epoch:
- data://meadow/artificial_intelligence/latest/epoch
data://grapher/artificial_intelligence/latest/epoch:
- data://garden/artificial_intelligence/latest/epoch
```
and only change the snapshot version.

You could even use `latest` for the snapshot.

(Versioning files is mainly helpful for our internal purposes for e.g. comparing versions. Anyone could in theory go back in git history and execute the step, which is already a _versioning_. It's just that keeping multiple versions in repo is more convenient.)

Marigold (approved):
Looks good!

Issue #1868, opened 2023-10-26
Add at least one dataset level tag to all datasets that have new metadata


danyx23:
For the new datapages it would be extremely useful if we could guarantee that every indicator has at least one `topic_tag` assigned. This will ensure that we can link our topic pages with our data pages and help in showing the most relevant content in various places.

The easiest way to ensure this is to set one (or more) tag at the dataset level. This is not always appropriate but works in many cases. Tags have to match a [topic tag in the database](http://datasette-private/owid?sql=select+name%2C+isTopic+from+tags+where+%22isTopic%22+%3D+%3Ap0+order+by+id+desc+limit+101&p0=1) (case sensitive). The schema has been updated to include the current topic tags so you should get autocomplete in the yaml file for this field. The only non-topic that that can be used in "Uncategorized" which can be used for cases where you explicitly don't want to assign a topic (e.g. useful in cases like the WDI where every indicator needs a human or AI assigned tag and you don't have the time to do it yet).

The easiest way to assign a tag to all indicators in a dataset is to set it via the common section in the yaml file:
```yaml
definitions:
  common:
    presentation:
      topic_tags:
      - Poverty
```

As of today we have about 100 datasets that have at least one indicator with new metadata. The list below has a checkmark to indicate if a tag has been added. Then it states the number of indicator the dataset has, the dataset name and the name is linked to the probable location of the metadata yaml file that you need to edit with the above snippet.

### From Lucas
(working on #1869)
- [x] [2 - Population density (various sources, 2023.2)](https://github.com/owid/etl/tree/master/etl/steps/data/garden/demography/2023-06-12/population_density.meta.yml)
- [x] [7 - Population (various sources, 2023.1)](https://github.com/owid/etl/tree/master/etl/steps/data/garden/demography/2023-03-31/population.meta.yml)
- [x] [682 - Human Mortality Database](https://github.com/owid/etl/tree/master/etl/steps/data/garden/hmd/2023-09-19/hmd.meta.yml)
- [x] [168 - Life tables](https://github.com/owid/etl/tree/master/etl/steps/data/garden/demography/2023-10-04/life_tables.meta.yml)
- [x] [54 - Life Expectancy (period)](https://github.com/owid/etl/tree/master/etl/steps/data/garden/demography/2023-10-10/life_expectancy.meta.yml)
[...]

PR #1874, opened 2023-10-27
✨ wizard: add tags as a requirement

lucasrodes:
Data managers now have to select a tag for all indicators as a requirement.

As part of https://github.com/owid/etl/issues/1868


pabloarosado (approved):
Looks good.

Issue #1864, opened 2023-10-26
Fill spurious missing nuclear data with zeros in energy datasets


pabloarosado:
There are indicators have have missing data where they should have zero. This happens, e.g.:
* Nuclear energy in Australia prior to 2000.
* Nuclear energy in Kazakhstan after 2000.

It's unclear where these fixes should be implemented (it could be in the Statistical Review garden step, or in the electricity mix, or somewhere else).

For more context and examples, see https://github.com/owid/owid-issues/issues/1257

PR #1877, opened 2023-10-30
📊 Fix spurious zeros in energy data and create generic solution

pabloarosado:
In this PR:
* I address https://github.com/owid/owid-issues/issues/1257 from the data side, and fix https://github.com/owid/etl/issues/1864
* I create a generic `grapher_helpers` function that will create columns for multiindicator charts.
  * The idea is that, for each multiindicator chart (e.g. a stacked area chart), this function will create columns for each indicator used in the chart. Those indicators will be nan if any of the indicators in the chart is nan (after optionally filling some of those indicators with zeros).
* I use the previous function to fix issues in charts in the electricity mix dataset.
* I went country by country to check if missing values in the Statistical Review of World Energy could be filled with zeros.

@spoonerf I'll merge this PR without waiting for a review to start fixing the above issues in the energy dataset (and to start adapting the charts to use the new indicators). But please have a review after the merge, and I'll implement any possible changes afterwards.

@spoonerf and @lucasrodes what do you think about the proposed solution in `etl/grapher_helpers.py`?


etl/grapher_helpers.py
```diff
@@ -587,3 +587,85 @@ def sanitize_numpy(obj: Any) -> Any:
...
+    columns_in_chart : List[str]
+        Column names in the relevant multi-indicator chart.
+    chart_slug : str
+        URL slug of the chart, which will be added to the name of the new columns.
+        By convention, the suffix added to columns will be "_chart_" followed by the chart's slug in snake case format.
+    suffix_for_titles: Optional[str]
+        Suffix to be added to the new columns' titles, to avoid having multiple indicators with the same title.
+    columns_to_fill_with_zeros : Optional[List[str]]
+        Subset of columns_in_chart whose nans should be filled with zeros.
+        Note: The original columns will not be affected.
```
spoonerf, line 637: I am finding that the original columns are affected, for example the garden variable `bioenergy_generation__twh` for Zimbabwe in 2022 has NaN, but in grapher, this is 0 and it looks a bit wrong in [this chart](https://ourworldindata.org/grapher/electricity-production-by-source?country=~ZWE).

spoonerf: Perhaps it would be good to add additional variables with a 'zero_filled' suffix, so that the original data is preserved?

spoonerf (commented):
Looks generally good, however, I think zero-filled columns should be additional rather than changing the original columns. 

For example, in [this chart](https://ourworldindata.org/grapher/electricity-production-by-source?country=~ZWE), the original variable for Bioenergy had NaN for 2022, but the `add_columns_for_multiindicator_chart()` changes this to zero, which I think is incorrect?

etl/grapher_helpers.py
spoonerf: Maybe the zero-filling could be split out to a different function as this wouldn't need to be done independently for each chart.

pabloarosado: Thanks for pointing this out. I think the original columns should definitely not be filled with zeros. The subset of new columns should be filled with zeros. I'll fix this.

etl/grapher_helpers.py
```diff
@@ -587,3 +587,85 @@ def sanitize_numpy(obj: Any) -> Any:
...
+    * This function assumes a wide-format table.
+    * In the grapher admin, ensure you select "Hide entities with missing data" in the corresponding chart.
+
+    Parameters
+    ----------
+    table : Table
+        Original table.
+    columns_in_chart : List[str]
+        Column names in the relevant multi-indicator chart.
+    chart_slug : str
```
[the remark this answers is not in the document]
pabloarosado: Yes, but I think it's also more convenient to see a slug in the variable names (so you know roughly what the chart is about) rather than an ID. If a chart slug changes, nothing will break, the only issue is that the variable names will be a little confusing.
So, both options can cause some confusion, but using grapher IDs seems to cause confusion more often, so I'll keep slugs for now.

PR #1878, opened 2023-10-30
📊 un: upload UN members database

paarriagadap:
Upload the historical list of members of United Nations, from @edomt's scrapping of [this website](https://www.un.org/en/about-us/member-states). 

This dataset is for the [State Capacity topic page](https://github.com/owid/owid-issues/issues/925).

Staging dataset: http://staging-site-parriagadap/admin/datasets/6282


snapshots/un/2023-10-30/web_scrapping.py
edomt, line 1: I didn't really polish the script before sending it to you, but for the ETL it's possibly worth simply wrapping this up in a `main` function?

paarriagadap: In this case I am just keeping that code to proceed with a local file, I don't see much problem doing that. I included a way to export it as a csv file to then use --path-to-file when running snapshots.

etl/steps/data/garden/un/2023-10-30/un_members.meta.yml
```diff
@@ -0,0 +1,72 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    presentation:
+      topic_tags:
+        - State Capacity
```
Marigold, line 6: Nit - You could add

```
definitions:
  common:
    processing_level: major
```

Marigold (approved):
Looks good!

etl/steps/data/garden/un/2023-10-30/un_members.py
```diff
@@ -0,0 +1,217 @@
...
+
+def add_missing_countries(tb: Table) -> Table:
+    """
+    Add countries not available in the list of UN members
+    """
+
+    # Load region dataset
+    tb_regions = paths.load_dependency("regions", channel="grapher")
+    tb_regions = tb_regions["regions"].reset_index()
+    tb_regions = tb_regions[
```
paarriagadap, line 87: There;s this error here, @Marigold:
```
/Users/parriagadap/etl/etl/steps/data/garden/un/2023-10-30/un_members.py
  /Users/parriagadap/etl/etl/steps/data/garden/un/2023-10-30/un_members.py:86:18 - error: "__getitem__" method not defined on type "Snapshot" (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations 
Completed in 8.308sec
make: *** [check-typing] Error 1
```

Should I just add `# type: ignore`? I needed to use load_dependency, because I am calling both garden and grapher steps for regions.

Marigold: Correct, you shouldn't be ashamed of using `# type: ignore`. I always do it when I'm lazy to fix type hints...

PR #1880, opened 2023-10-31
Fix function for multiindicator charts

pabloarosado:
Thanks @spoonerf for your suggestions in https://github.com/owid/etl/pull/1877
I've changed the function so that the original columns are not affected (and did a minor change in the electricity mix step to trigger the changes in production).

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -612,7 +612,7 @@ def add_columns_for_multiindicator_chart(
 
     This function can be used to create columns for all indicators used in a specific multi-indicator chart.
     It will:
-    * Optionally, fill missing data for certain indicators with zeros.
+    * Optionally, fill missing data for some of the new columns with zeros.
       In the example of issue 1, nuclear power could be filled with zeros.
     * Make nan rows in the new columns if any indicator used in the chart is nan.
       In the example of issue 2, Algeria's renewable electricity in 2022 would be nan.
@@ -620,6 +620,7 @@ def add_columns_for_multiindicator_chart(
     NOTES:
     * This function assumes a wide-format table.
     * In the grapher admin, ensure you select "Hide entities with missing data" in the corresponding chart.
+    * This function creates new columns and does not affect already existing ones.
 
     Parameters
     ----------
@@ -634,7 +635,7 @@ def add_columns_for_multiindicator_chart(
         Suffix to be added to the new columns' titles, to avoid having multiple indicators with the same title.
     columns_to_fill_with_zeros : Optional[List[str]]
         Subset of columns_in_chart whose nans should be filled with zeros.
-        Note: The original columns will not be affected.
+        Note: The original columns will not be affected, only the newly created ones.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


etl/grapher_helpers.py
```diff
@@ -644,16 +645,24 @@ def add_columns_for_multiindicator_chart(
...
-        table[columns_to_fill_with_zeros] = table[columns_to_fill_with_zeros].fillna(0)
-
-    # Create new columns.
-    new_columns = [f"{column}_chart_{underscore(chart_slug)}" for column in columns_in_chart]
-    table[new_columns] = table[columns_in_chart].copy()
+        # Fill nans with zeros (in new columns).
+        new_columns_to_fill_with_zeros = [
+            _rename_column(old_column_name=column, chart_slug=chart_slug) for column in columns_to_fill_with_zeros
+        ]
+        table[new_columns_to_fill_with_zeros] = table[new_columns_to_fill_with_zeros].fillna(0)
```
spoonerf, line 665: I'm not sure it is working as expected, variables that are in `columns_to_fill_with_zeros`, get their zeros here and then these zeros are then replaced with NaNs with `table.loc[table[new_columns].isnull().any(axis=1), new_columns] = np.nan`. I think the zero-filled variables need a different column name.

spoonerf (commented):
I'm not sure it is working as expected, variables that are in `columns_to_fill_with_zeros`, get their zeros and then these zeros are then replaced with NaNs with `table.loc[table[new_columns].isnull().any(axis=1), new_columns] = np.nan`. 

I think the zero-filled variables need a different column name, perhaps an additional argument in `_replace_names()` specifying the reason for the new names would do it.

etl/grapher_helpers.py
pabloarosado: Yes, this is the behaviour I needed:
* I first fill with zeros those variables that are not too important and often missing (e.g. "Other renewables"), to avoid having an empty chart where important variables (e.g. "Gas") were available but hidden because of the nans in less important variables.
* Then, if any variable is missing, then all variables should be missing in the same row. That way we ensure that *either we show data for all indicators on a specific country-year, or we show no data*. I think that's the key ingredient to avoid having strange jumps, or to have years where a country has, e.g. 100% renewables because fossil fuels were not informed.

If you simply wanted to fill some indicators with zeros, regardless of the other indicators, I suppose you wouldn't need this function. But if you think that could be the desired outcome in some circumstances, I could add an option to the function, so that the nans can be imposed only in `set(new_columns) - set(new_columns_to_fill_with_zeros)`.

pabloarosado: If filling with zeros is all you need for a certain chart, you could simply do fillna(0) and rename those indicators as "*_zero_filled", as we used to do in the past.
Alternatively, you could simply do nothing at all, because grapher will by default treat nans as zeros.

spoonerf: Ah, I see, I misunderstood but am with you now!

I am having some issues running it locally and am currently rebuilding my staging server so currently can't check out the output data just yet. However, feel free to merge so the electricity mix issues will get solved and I can double-check data once I am able.

PR #1889, opened 2023-11-01
📊 Cardiovascular diseases: custom dataset based on WHO mortality dataset

veronikasamborska1994:
hey @pabloarosado! this is a small dataset that's based on WHO mortality data where I am aggregating some categories mostly for visualisation purposes. It relies on a dataset with old metadata structure but I am including the new metadata structure at the indicator level. It's a small pull request so would be great if you can take a look over the next couple of days. thanks :)


etl/steps/data/garden/cardiovascular_diseases/2023-11-01/who_cvd.py
```diff
@@ -0,0 +1,101 @@
...
+    all_causes = pr.concat(all_causes, axis=1)
+
+    # Invert the mapping to facilitate grouping.
+    reverse_mapping = {col: category for col, category in CAUSE_MAPPING.items()}
+
+    # Group and sum columns based on broader categories.
+    summed_df = all_causes.groupby(reverse_mapping, axis=1).sum()
+
+    # Combine the aggregated data with specific columns, e.g., cardiovascular diseases.
+    df_with_cardiovascular_diseases = pd.concat(
```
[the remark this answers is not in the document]
pabloarosado: Then, instead of concatenating a table (summed_df) with a variable (all_causes["share..."]), make the latter a table, all_causes[["share..."]] (in other words, use double square brackets instead of single).

etl/steps/data/garden/cardiovascular_diseases/2023-11-01/who_cvd.py
```diff
@@ -0,0 +1,101 @@
...
+        [
+            summed_df,
+            all_causes["share_of_total_deaths_in_both_sexes_aged_all_ages_years_that_are_from_cardiovascular_diseases"],
+        ],
+        axis=1,
+    )
```
pabloarosado, line 86: I find this operation risky: What if they add a new category on the next update? Instead, I would gather the list of share columns from all tables, and then:
1. Assert that the column "share_of_total_deaths_in_both_sexes_aged_all_ages_years_that_are_from_all_causes" contains only 100.
2. Assert that the only remaining column, that is neither "share_of_total_deaths_in_both_sexes_aged_all_ages_years_that_are_from_all_causes" nor accounted for in the dictionary `CAUSE_MAPPING`, is "share_of_total_deaths_in_both_sexes_aged_all_ages_years_that_are_from_cardiovascular_diseases".

etl/steps/data/garden/cardiovascular_diseases/2023-11-01/who_cvd.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Compute the percentage of deaths with unidentified causes.
+    row_sums = df_with_cardiovascular_diseases.sum(axis=1)
```
pabloarosado, line 89: Assert that the result is between 0 and 100.

etl/steps/data/garden/cardiovascular_diseases/2023-11-01/who_cvd.py
pabloarosado: Alternatively, why not adding "share_of_total_deaths_in_both_sexes_aged_all_ages_years_that_are_from_cardiovascular_diseases" as another group in CAUSES_MAPPING, with just one cause? It would simplify the code, and make column names more consistent.
Currently all names are, e.g. "respiratory_diseases", and then all of a sudden "share_of_total_deaths_in_both_sexes_aged_all_ages_years_that_are_from_cardiovascular_diseases" is totally different.

etl/steps/data/garden/cardiovascular_diseases/2023-11-01/who_cvd.py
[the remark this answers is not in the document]
pabloarosado: Also, you will be working with tables, so it makes more sense to call the object `tb_...` instead of `df_...`.

etl/steps/data/garden/cardiovascular_diseases/2023-11-01/who_cvd.py
```diff
@@ -0,0 +1,101 @@
...
+    remaining_percentage = 100 - row_sums
+    df_with_cardiovascular_diseases["unidentified_cause"] = remaining_percentage
+
+    # Set metadata and format the dataframe for saving.
+    df_with_cardiovascular_diseases.metadata.short_name = paths.short_name
+    df_with_cardiovascular_diseases = df_with_cardiovascular_diseases.underscore()
```
pabloarosado, line 95: Here I'd suggest doing:
```
tb_with_cardiovascular_diseases = add_origins_to_mortality_database(tb_with_cardiovascular_diseases)
```
And define that function in `etl.data_helpers.misc`, similar to `add_origins_to_wdi`. This is a temporary solution to add origins to datasets that don't have origins. If you think of a better solution, please let me know.

pabloarosado (changes requested):
I made multiple suggestions, but please let me know if some of them don't make sense.

veronikasamborska1994:
> I made multiple suggestions, but please let me know if some of them don't make sense.

hey @pabloarosado! thanks a lot for taking a look. One final issue I have is that after adding origins to each column I think I am getting a conflict with the original metadata sources and the grapher on my staging shows both source names (WHO Mortality Database (2022); World Health Organisation (2023)). Do you know how to best fix this? Thanks again :)

veronikasamborska1994:
> I made multiple suggestions, but please let me know if some of them don't make sense.

hey @pabloarosado! thanks a lot for taking a look. One final issue I have is that after adding origins to each column I think I am getting a conflict with the original metadata sources and the grapher on my staging shows both source names (WHO Mortality Database (2022); World Health Organisation (2023)). Do you know how to best fix this? Thanks again :)

pabloarosado (approved):
It looks good, and I see no problems with origins. I've just checked a few indicators and they have just one origin.

etl/data_helpers/misc.py
```diff
@@ -119,4 +119,36 @@ def add_origins_to_energy_table(tb_energy: Table) -> Table:
...
+
+    # For each indicator, add an origin (using information from the old source) and then remove the source.
+    for column in data_columns:
+        tb_who[column].metadata.sources = []
+        error = "Remove temporary solution where origins where manually created."
+        assert tb_who[column].metadata.origins == [], error
+        tb_who[column].metadata.origins = [
+            Origin(
+                title="Mortality Database",
+                producer="World Health Organisation",
```
pabloarosado, line 138: If you want to change what's shown by default in the footer of charts, you could add here:
```
attribution="WHO Mortality Database (2023)"
```
or `World Health Organisation - Mortality Database (2023)` (whatever we use elsewhere).

PR #1892, opened 2023-11-01
📊 unu_wider: Create Government Revenue Dataset

paarriagadap:
Part of these topic pages: [Taxation](https://github.com/owid/owid-issues/issues/803) and [State Capacity](https://github.com/owid/owid-issues/issues/925)


pabloarosado (approved):
Looks good. I suggested some minor changes. And, another general suggestion that I've mentioned on other occasions: it would be better to use more descriptive short names for variables, instead of cryptic ones like `tax_g_s`. But if this is too much added work, then leave it.

pabloarosado:
Since you modified the `regions` dataset, it would be better to merge this PR by the end of the day, since it will trigger a big run (as you can see in [buildkite](https://buildkite.com/our-world-in-data/etl-build-public-datasets-branch/builds/5673#018b911b-b458-4484-b8b5-70a82c4c17a7)).

paarriagadap:
Thank you, @pabloarosado!

PR #1893, opened 2023-11-01
📊 Update greenhouse gas emissions by sector

pabloarosado:
**WARNING**: Do not merge until Mojmir has done the transfer of charts from the staging PR server to live.

Update data on greenhouse gas emissions by sector.

This data used to be under the `cait` namespace, but now it's under `climate_watch`. Many things have changed, so there's little point in comparing to the previous version.
There used to be four different grapher datasets (for [all gases](https://owid.cloud/admin/datasets/5684), [methane](https://owid.cloud/admin/datasets/5686), [nitrous oxide](https://owid.cloud/admin/datasets/5687) and [carbon dioxide](https://owid.cloud/admin/datasets/5685)), but now I've put all indicators together in one dataset, for simplicity.


etl/steps/data/garden/climate_watch/2023-10-31/emissions_by_sector.py
```diff
@@ -0,0 +1,224 @@
...
+    "Manufacturing/Construction": "Manufacturing and construction emissions",
+    "Other Fuel Combustion": "Other fuel combustion emissions",
+    "Total excluding LUCF": "Total emissions excluding LUCF",
+    "Total including LUCF": "Total emissions including LUCF",
+    "Transportation": "Transport emissions",
+    "Waste": "Waste emissions",
+}
+
+# Suffix to add to the name of per capita variables.
+PER_CAPITA_SUFFIX = " (per capita)"
```
paarriagadap, line 31: Why not just "per_capita"?

etl/steps/data/garden/climate_watch/2023-10-31/emissions_by_sector.py
```diff
@@ -0,0 +1,224 @@
...
+    # List columns with emissions data.
+    emissions_columns = [column for column in tb_gas.columns if column not in ["country", "year", "population"]]
+
+    # Add per capita variables.
+    for variable in emissions_columns:
+        tb_gas[variable + PER_CAPITA_SUFFIX] = MT_TO_T * tb_gas[variable] / tb_gas["population"]
+
+    # Remove rows and columns that only have nans.
+    tb_gas = tb_gas.dropna(how="all", axis=1)
+    tb_gas = tb_gas.dropna(subset=emissions_columns, how="all").reset_index(drop=True)
```
paarriagadap, line 140: Shouldn't this be before adding per capita variables? Some per capita columns could stay empty after the calculation

paarriagadap (approved):
Very good! I added minor comments

etl/steps/data/garden/climate_watch/2023-10-31/emissions_by_sector.py
pabloarosado: It doesn't make a difference (in fact there are no rows or columns with nans.

PR #1894, opened 2023-11-02
:tada: move snapshots to R2

Marigold:
We already host our ETL catalog on R2, which works very well and the purpose of this PR is to move the rest (snapshots and walden) there, too, to have a unified experience.

This PR removes all code for accessing DigitalOcean's Spaces and replaces it by R2. I've also done a bit of refactoring to DRY wrapper code around R2.

ETL users shouldn't need to change their config. It's enough to have `R2_ACCESS_KEY` and `R2_SECRET_KEY` in their `.env` or have valid credentials in `~/.aws/config`. 

S3 helper functions are still a bit scattered, but fixing it turned out to be more complex than expected, so I'm leaving it for future refactoring.

## Appendix

Clone Spaces to R2 with (run on lxc-manager)
```
rclone sync s3:owid-catalog/snapshots r2:owid-snapshots/ --verbose --transfers=128 --checkers=128 --fast-list --size-only
```
and
```
 rclone sync s3:walden r2:owid-walden/ --verbose --transfers=128 --checkers=128 --fast-list --size-only
```

```
(etl-py3.11) ➜  etl git:(staging-sync) rclone size s3:owid-catalog/snapshots
Total objects: 10.185k (10185)
Total size: 61.814 GiB (66372405761 Byte)

(etl-py3.11) ➜  etl git:(staging-sync) rclone size s3:walden/
Total objects: 4.001k (4001)
Total size: 19.791 GiB (21250252444 Byte)

```


lucasrodes (approved):
Looked at a high level and looked fine!

PR #1895, opened 2023-11-02
:hammer: add topic tag to WDI GDP indicators

danyx23:
Adds the topic tag "Economic Growth" to all GDP per capita indicators inside WDI

### DIFF
--- a/etl/steps/data/garden/worldbank_wdi/2023-05-29/wdi.meta.yml
+++ b/etl/steps/data/garden/worldbank_wdi/2023-05-29/wdi.meta.yml
@@ -2633,33 +2633,33 @@ tables:
         unit: current US$
         presentation:
           topic_tags:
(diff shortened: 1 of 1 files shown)


etl/steps/data/garden/worldbank_wdi/2023-05-29/wdi.meta.yml
```diff
@@ -2631,27 +2631,45 @@ tables:
         title: GDP per capita (current US$)
         short_unit: $
         unit: current US$
+        presentation:
+          topic_tags:
+            - "Economic Growth"
```
pabloarosado, line 2636: I think you don't need the `"` or the indentation, you could simply pass:
```
topic_tags:
- Economic Growth
```
But if this works well, then all good.

pabloarosado (approved):

PR #1905, opened 2023-11-06
Update global warming potential factors

pabloarosado:
Create a very simple dataset with the GWP-100 factors for a list of greenhouse gases, from the IPCC AR6 (to replace [this dataset](https://owid.cloud/admin/datasets/860) based on AR5).


spoonerf (approved):
Looks good!

PR #1906, opened 2023-11-06
:tada: add STAGING env for accessing staging-site-*

Marigold:
This lets you use
```
 STAGING=mojmir etl igme --grapher
```
which will load your `.env` and point environment variables to server `staging-site-mojmir`. You won't have to keep multiple `.env.*` files for staging servers anymore.


pabloarosado (approved):
Good idea!

PR #1920, opened 2023-11-07
Combine new metadata fields

pabloarosado:
* Propagate `description_key` when doing variable operations.
* Add tests.
* Some small improvements on natural disasters metadata (some require the propagation of `description_key` to show up in charts).


Marigold (approved):
Looks good!

PR #1928, opened 2023-11-08
Use description key as description short in population indicators

pabloarosado:
* I changed `description_key` -> `description_short` in all indicators in the population dataset.
* I also changed `·` with `-` in a list of bullet points (which should render properly as a list).

@lucasrodes @Marigold do you see any unintentional risks on these changes? And do you remember if there was any particular reason why we were using `description_key` for those short descriptions? Thanks.

**NOTE**: This PR should be merged at the end of the day, since it will trigger a big ETL run.


lucasrodes (approved):
It LGTM, thanks for creating the PR.

I just realised that we could use `definitions.commons.description_processing` instead of typing `*description-processing` for all indicators.

Marigold (approved):

PR #1930, opened 2023-11-08
Update owid co2 dataset to use the latest emissions by sector

pabloarosado:
* Create a new owid_co2 step that uses the latest emissions by sector by Climate Watch (previously by CAIT).
* Archive old CAIT steps.
* Update dag.
* Remove Walden ingest scripts for CAIT.

@Marigold there's actually just one file to review, namely `etl/steps/data/garden/emissions/2023-11-08/owid_co2.py`.


Marigold (approved):
Looks good

PR #1934, opened 2023-11-08
🔨 Fixing youth mortality rate

spoonerf:


Marigold:
I'm having a bit of a hard time with staging servers now, so yeah, feel free to merge it.

PR #1946, opened 2023-11-10
Fix WDI origins

pabloarosado:
Define origins only to the specific indicator that is used in a data page, and keep all other indicators with sources.


Marigold (approved):
You're right, thanks for fixing this!

PR #1948, opened 2023-11-10
📊 state_capacity: Create Information Capacity Dataset

paarriagadap:
For the [State Capacity topic page](https://github.com/owid/owid-issues/issues/925) 

Dataset in staging server: http://staging-site-parriagadap/admin/datasets/6295


etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # Merge population data.
+    tb_regions = tb_regions.merge(tb_pop[["country", "year", "population"]], how="left", on=["country", "year"])
+
+    # List of index columns
+    index_cols = ["censusgraded_ability", "ybcov_ability", "infcap_irt", "infcap_pca"]
+
+    for col in index_cols:
+        # Create new columns with the product of the index and the population
+        tb_regions[col] = tb_regions[col] * tb_regions["population"]
```
paarriagadap, line 76: This introduces the error
```
AssertionError: Column `censusgraded_ability` is using np.nan, but it should be using pd.NA because it has type Float64
```

How can I avoid it, @Marigold? I have never found it before, and I have proceeded similarly.

Marigold: Just for context - `np.nan` and `pd.NA` are two completely different nan values, and it's causing major headaches (ask PabloR...). The reason is appears there is that `tb_regions[col]` might have UInt8 type which is one of the types that use pd.NA and not the good old `int` type.

The fix is luckily easy - just convert it to the old "float" type (which is also used by population)
```
tb_regions[col] = tb_regions[col].astype(float) * tb_regions["population"]
```

Generally, if you keep having issues with the new types (like UInt, Float64, those starting with uppercase), just convert them to `int` or `float` as soon as possible in the pipeline.

paarriagadap: Thanks @Marigold! I haven't found that before. I will correct this way if I find it anywhere else

etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    # Add regional aggregates, by summing up the variables in `aggregations`
+    for region in regions:
+        tb_regions = geo.add_region_aggregates(
+            tb_regions,
+            region=region,
+            aggregations=aggregations,
+            countries_that_must_have_data=[],
+            frac_allowed_nans_per_year=0.1,
```
paarriagadap, line 105: @pabloarosado Do you know why the `frac_allowed_nans_per_year` is not working as expected? I would expect to see [here](http://staging-site-parriagadap/admin/variables/816330) more missing aggregations for Africa (too few countries covered), but I see it's not the case:
image

etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py
```diff
@@ -0,0 +1,126 @@
...
+
+    tb_regions = tb.copy()
+
+    # Load population data.
+    tb_pop = paths.load_dataset("population")
+    tb_pop = tb_pop["population"].reset_index()
+
+    # Merge population data.
+    tb_regions = tb_regions.merge(tb_pop[["country", "year", "population"]], how="left", on=["country", "year"])
+
```
pabloarosado, line 68: Instead of the previous, I'd suggest:
1. Loading all necessary datasets at the beginning of `run` (it's useful to clearly see what are the inputs of the step) instead of hidden inside the function. The function would take `tb_population` as an argument.
2. Use `data_helpers.geo.add_population_to_table` instead of manually merging your table with the population table.

etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py
pabloarosado: Hey @paarriagadap can you give me a very specific example of what you would expect to have and what you are getting instead?

paarriagadap: I would expect to not see aggregations for Africa, because too many countries from the continent are covered:
image

pabloarosado: For column "infcap_irt" and e.g. year 2000, I see that there are 7 African countries informed, which is more than 10% of all African countries (59). So, to me, it looks consistent. But if you have a specific example that sounds wrong, please let me know.

pabloarosado: Ah, you are right, what I said before would not explain the problem, sorry for the confusion.
I think here the problem is different: The function doesn't actually care about how many countries are in Africa. The function simply checks how many African countries appear in a given year. And, if 10% or more are nan, then it removes the aggregate. However, the data actually contains only 7 African countries, and all of them informed (at least for year 2000 column "infcap_irt", which is what I took as a reference). So actually 0% of the rows are nan, and therefore the aggregate is valid. Does that make sense?

paarriagadap: Ahh, that's it, thanks. I've always thought the parameter worked with all the countries, regardless if they are in the dataset (by looking at the regions table). I will try to find an alternative method then.

pabloarosado: I'm happy it's clear now. You could impose a list of countries that need to be informed, by passing `countries_that_must_have_data`. But maybe you can figure out a better solution. As I always mention, we should rethink our function to create aggregates at some point.

etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py
```diff
@@ -0,0 +1,167 @@
...
+    # Create a new garden dataset with the same metadata as the meadow dataset.
+    ds_garden = create_dataset(
+        dest_dir, tables=[tb], check_variables_metadata=True, default_metadata=ds_meadow.metadata
+    )
+
+    # Save changes in the new garden dataset.
+    ds_garden.save()
+
+
+def regional_aggregations(tb: Table, ds_pop: Dataset, tb_country_list: Table) -> Table:
```
paarriagadap, line 65: Isn't rolling included for Tables?
I have this error:
```
  File "/Users/parriagadap/etl/etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py", line 95, in add_new_indicators
    tb["census_10_years"] = tb.groupby("country")["census"].rolling(10, min_periods=1).sum()
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/lib/catalog/owid/catalog/tables.py", line 997, in func
    raise NotImplementedError()
NotImplementedError
```

What would the alternative be here, @Marigold?

pabloarosado: I'll check now if it's trivial to implement `rolling`, but for now, you can do the following (a big cumbersome) workaround:
```
Variable(pd.DataFrame(tb).groupby("country")["census"].rolling(10, min_periods=1).sum()).copy_metadata(tb["census"])
```
Does that make sense? (I couldn't test it because I couldn't find it anywhere in the code in your current branch).

pabloarosado: I've realised it's not trivial to implement (we may need to define a special `Rolling` object). So I'll leave it for now. Feel free to create an issue if you think this can be useful.

pabloarosado: I think maybe what you can do is:
```
tb["census_10_years"] = pd.DataFrame(tb).groupby("country", as_index=False)["census"].rolling(10, min_periods=1).sum()["census"]
tb["census_10_years"] = tb["census_10_years"].copy_metadata(tb["census"])
```
But I haven't looked into the code in depth, so I don't know what the intention is.

paarriagadap: Bah, sorry, now it's in the branch. I tried that and I had `TypeError: incompatible index of inserted column with frame index`. I included `.values` to each operation and now I have `ValueError: variable must be named to have metadata` when I apply `.copy_metadata(tb["census"])`

paarriagadap: Tried the second option and now I have lost the short name:
```
(etl-py3.11) parriagadap@Pablos-MacBook-Pro etl % ENV=.env.parriagadap etl information_capacity --grapher
--- Detecting which steps need rebuilding...
OK (1.4s)                                                             
--- Running 3 steps:
--- 1. data://garden/state_capacity/2023-11-10/information_capacity_dataset...
2023-11-22 14:47:30,681 - [bugsnag] WARNING - No API key configured, couldn't notify
  File "<string>", line 1, in <module>
  File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/etl/run_python_step.py", line 34, in main
    _import_and_run(path, dest_dir)
  File "/Users/parriagadap/etl/etl/run_python_step.py", line 54, in _import_and_run
    module.run(dest_dir)
  File "/Users/parriagadap/etl/etl/steps/data/garden/state_capacity/2023-11-10/information_capacity_dataset.py", line 61, in run
    ds_garden = create_dataset(
                ^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/etl/helpers.py", line 164, in create_dataset
    ds.add(table, formats=formats)
  File "/Users/parriagadap/etl/lib/catalog/owid/catalog/datasets.py", line 143, in add
    table_filename = join(self.path, table.metadata.checked_name + f".{format}")
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/parriagadap/etl/lib/catalog/owid/catalog/meta.py", line 444, in checked_name
    raise Exception("table has no short_name")
Exception: table has no short_name

COMMAND: poetry run run_python_step data://garden/state_capacity/2023-11-10/information_capacity_dataset /Users/parriagadap/etl/data/garden/state_capacity/2023-11-10/information_capacity_dataset
```

pabloarosado: That's a different problem. You lose the table short name when you merge the main table with another that doesn't have a short name (line 90, inside `add_new_indicators`). To fix it:
```
    # Merge this table with tb
    tb = pr.merge(tb_countries_years, tb, how="left", on=["country", "year"], short_name=tb.metadata.short_name)

```

paarriagadap: Thanks! I won't forget to add it for merges/concats

PR #1017, opened 2023-04-18
chore: switch from flake8 to ruff

larsyencken:
The ruff linter is a much faster linter written in rust.

PR #1954, opened 2023-11-20
:lipstick: format files by ruff

Marigold:
Formatting changes after switching to ruff from #1017

This also changes line length to 120 for datautils package (for consistency).


Marigold:
Formatting changes after switching to ruff from https://github.com/owid/etl/pull/1017

PR #1963, opened 2023-11-22
Update Global Carbon Budget

pabloarosado:
Update Global Carbon Budget data.

NOTE:
* This PR compares only the changes with respect to the current version.
* Some URLs of snapshots are private S3 links, since the data is not going to be published until 2023-12-05.

TODO: (@lucasrodes, @Marigold do you think the following workflow makes sense?)
* Before 2023-05-12:
  * Merge this PR as soon as it is reviewed.
  * Execute chart revisions in the temporary PR server of the reference branch.
* On 2023-12-05:
  * Remove all the instances of `-private` in the dag of the reference branch.
  * Update URLs in snapshot dvc files, remove `outs:` sections, and run snapshots again.
    * Run all steps locally and check that nothing changed, then push to reference branch.
  * Merge reference branch to master.
  * Execute the sync tool to move all charts to live.


pabloarosado:
> @pabloarosado sounds good! We should be a bit careful when using the sync tool; it still hasn't been battle-tested. On 2023-12-05 I'll first try to migrate it to a different staging server (not production) and if everything looks good then we merge & migrate to prod.

Thanks, if we are not yet very confident with this tool, I can just not use it and do the chart revision again on live. This dataset is big and important, so it's better not to risk errors on publication day.

Marigold:
No, let's use it! But with a test run first...

pabloarosado:
> No, let's use it! But with a test run first...

OK, feel free to edit the issue description above to explain how the workflow should be on that day.

snapshots/gcp/2023-12-05/global_carbon_budget.py
```diff
@@ -44,6 +52,7 @@ def main(upload: bool) -> None:
         snap = Snapshot(f"gcp/{SNAPSHOT_VERSION}/{data_file}")
 
         # Replace the full citation and description in the metadata.
+        snap.metadata.origin.attribution = ATTRIBUTION  # type: ignore
```
lucasrodes, line 55: (no action needed)
I was checking and I am not sure what the warning raised here actually means.

> "attribution" is not a known member of "None"Pylance[reportOptionalMemberAccess](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportOptionalMemberAccess)

pabloarosado:
> LGTM, nice work, Pablo.
> 
> Just a side comment on how you structure functions in a Python file: I prefer having the 'parent' functions (i.e. the main functions, like `run`) at the top of the file. I find it easier to understand the flow in the script, and don't have to scroll all the way down to get a first idea on what the script does.
> 
> I think we discussed this in the past, and you were keen on having it as you have it here (i.e. from detailed to more general functions), so it might just be a personal preference.

Thanks Lucas. I don't have a strong preference, so, given that you do have a preference, I'll change the order of the run function. I agree it may improve readability.

PR #1966, opened 2023-11-22
💄 Adding origins to historical literacy and expenditure datasets

veronikasamborska1994:
Because the two datasets on expenditure and literacy were fasttracked I am using a custom function in misc to define origins for the historical literacy and expenditure estimates. These are combined with the World Bank EdStats metadata (already updated to a new format) for the subsequent data pages (specifically [this one which is on the list of the most important ones](https://ourworldindata.org/grapher/cross-country-literacy-rates?tab=map)).


veronikasamborska1994:
hey @pabloarosado I've now moved the metadata reformatting to snapshot and it looks good on grapher!

snapshots/education/2018-04-18/literacy_rates.feather.dvc
```diff
@@ -1,82 +1,27 @@
...
+      This dataset combines long-run cross-country literacy rates from several sources
+      including the World Bank's WDI, CIA Factbook, and other historical data series.
+      Key sources include Buringh & Van Zanden (2009) for data before 1800, Broadberry
+      and O'Rourke (2010) for 1820 and 1870, and van Zanden et al. (2014) for global
+      estimates from 1820-2000. It also includes recent estimates for high-income
+      countries and available 2016 data from the CIA World Factbook. Note that the
+      sources may have discrepancies and different underlying methodologies, and thus
+      year-to-year changes should be interpreted cautiously.
+    # Citation
+    producer: Our World in Data
```
pabloarosado, line 17: As explicitly mentioned in [the guidelines](https://docs.owid.io/projects/etl/architecture/metadata/reference/origin/#originproducer), and discussed multiple times, the producer is (almost) never OWID. The only case where this is true is for the `regions` dataset. If you have very strong reasons to choose OWID as a producer, it would be good to discuss it and change the guidelines.
In this case, I suppose the producer could be either "World Bank" or "Various sources".

pabloarosado (changes requested):
Hi @veronikasamborska1994, there are a few things I would change in this PR:
1. It's much better (for reviewers and for future you) to explicitly load all datasets at the beginning of the run function, instead of "secretly" loading them inside `combine_historical_literacy_expenditure`.
2. A minor improvement is to load datasets using: `ds_meadow = paths.load_dataset("education")` (and the same for the other two datasets). This simplifies the code.
3. I don't understand why you use `copy_metadata` multiple times. This is used mostly for operations that do not propagate metadata. But `merge` does propagate origins. When you use `copy_metadata` you are actually **removing** origins. And because of this later you are forced to manually combine origins. Why not letting `merge` propagate origins automatically? So, I'd suggest to remove all instances of `copy_metadata`.
4. The functions `decide_literacy` and `decide_expenditure` look to me like they could be deleted. Instead, you can define `combined_df["combined_literacy"] = combined_df["SE.ADT.LITR.ZS"].fillna(combined_df["literacy_rates__world_bank__cia_world_factbook__and_other_sources"])` (and similar for expenditure), isn't that the intention? Using a fillna would also propagate origins automatically.
5. After you have done the previous changes, I believe you won't need to manually change origins.
6. Be careful when doing `metadata = tb.metadata`, if you change `metadata`, it may alter `tb.metadata` (not totally sure). A safer option would be to do `metadata = tb.metadata.copy()`.
7. I've spotted a `Normalised` somewhere. We should use American spelling, `Normalized`.

I'd suggest to implement those changes, unless I misunderstood something.

veronikasamborska1994:
thanks @pabloarosado! and thanks for spotting the redundancy with .copy_metadata. I think when I first started using the new metadata the pr.merge wasn't working properly and propagating the metadata correctly in every case and this was my solution and looks like I continued just using this as a default. I've made all of the changes you've suggested now. Could you take a look and approve if it all looks ok now (when you get a chance)?

veronikasamborska1994:
@pabloarosado all done now! thanks again for this feedback! also, useful for any future updates I'll have to do (old ---> new metadata format).

etl/steps/data/garden/wb/2023-07-10/education.py
```diff
@@ -73,21 +75,19 @@ def run(dest_dir: str) -> None:
...
     # Combine recent literacy estimates and expenditure data with historical estimates from a migrated dataset
-    tb, combined_literacy_description, combined_expenditure_description = combine_historical_literacy_expenditure(tb)
+    tb = combine_historical_literacy_expenditure(tb, tb_literacy, tb_expenditure)
 
     # Compare two columnst that seem to have identical indicies (if values are the same then remove)
     if tb["SE.XPD.TOTL.GD.ZS"].equals(tb["SE.XPD.TOTL.GD.ZS."]):
         # If they are the same, drop one of the columns
         tb.drop("SE.XPD.TOTL.GD.ZS.", axis=1, inplace=True)
     else:
         print("The columns are not the same.")
```
pabloarosado, line 85: Instead of printing, you should assert things. Assert that the columns are identical, then drop one of them. If it changes in the next update, you will not have any way to miss it.

etl/steps/data/garden/wb/2023-07-10/education.py
```diff
@@ -100,22 +100,34 @@ def run(dest_dir: str) -> None:
     ds_garden.save()
 
 
-def combine_historical_literacy_expenditure(tb):
+def combine_historical_literacy_expenditure(tb, tb_literacy, tb_expenditure):
     """
-    Combines historical and recent literacy and expenditure data into a single DataFrame.
+    Merges historical and recent literacy and expenditure data into a single DataFrame.
```
pabloarosado, line 105: Style: In Python, the first line in a docstring should be imperative, e.g. "Merge historical...".

pabloarosado: DataFrame -> Table?

snapshots/education/2018-04-18/literacy_rates.feather.dvc
```diff
@@ -1,82 +1,27 @@
...
-      there was data available.
-
-
-      - Sources for each country-year observation can be found in [this table.](http://ourworldindata.org/wp-content/uploads/2018/06/cross-country-literacy-sources-final.csv)
-    url: '- World Bank: https://data.worldbank.org/indicator/SE.ADT.LITR.ZS </br>
-      - CIA Factbook: https://www.cia.gov/library/publications/the-world-factbook/
-      </br> - Links to other sources below'
+  origin:
+    # Data product / Snapshot
+    title: Cross-country literacy rates - World Bank, CIA World Factbook, and other Sources
```
pabloarosado, line 6: The title should not include the producer.

snapshots/education/2018-04-18/literacy_rates.feather.dvc
```diff
@@ -1,82 +1,27 @@
...
+    description: >
```
pabloarosado, line 7: The sources should be specified in `citation_full` (although there can be some overlap with the description, if it helps describe the dataset).

snapshots/education/2018-04-18/literacy_rates.feather.dvc
```diff
@@ -1,82 +1,27 @@
...
+      and O'Rourke (2010) for 1820 and 1870, and van Zanden et al. (2014) for global
+      estimates from 1820-2000. It also includes recent estimates for high-income
+      countries and available 2016 data from the CIA World Factbook. Note that the
+      sources may have discrepancies and different underlying methodologies, and thus
+      year-to-year changes should be interpreted cautiously.
+    # Citation
+    producer: Various sources
+    citation_full: 'Literacy rates (World Bank, CIA World Factbook, and other sources)'
+    # Files
+    url_main: 'https://data.worldbank.org/indicator/SE.ADT.LITR.ZS, https://www.cia.gov/library/publications/the-world-factbook/, and other sources.'
```
pabloarosado, line 20: `url-main` as far as I know should be just one URL (otherwise I can imagine that it will cause issues in grapher). Any additional ones should be in `description` or `citation_full` or elsewhere.

pabloarosado (approved):
I added some minor style and metadata corrections. After that, feel free to merge!

PR #1967, opened 2023-11-23
:sparkles: Speedup add_regions_aggregates

Marigold:
Speed up add_regions_aggregates by avoiding for loop over variables. This is very noticeable for datasets with 100+ variables.

Also add unit test to clarify what happens if mandatory country has NaN values (@pabloarosado  this is not a but, it works in the end)

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -320,26 +320,25 @@ def add_region_aggregates(
     variables = list(aggregations)
 
     # Initialise dataframe of added regions, and add variables one by one to it.
-    df_region = pd.DataFrame({country_col: [], year_col: []}).astype(dtype={country_col: "object", year_col: "int"})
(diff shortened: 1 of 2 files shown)


pabloarosado (approved):
Thanks for doing this, it looks good, and right now I can't remember why I imposed going variable by variable when aggregating. But to give old Pablo R the benefit of the doubt, I'd appreciate if you could run a datadiff on all ETL to ensure that absolutely nothing changes. Would that be possible? Cause this function is used in many places, and issues with regions are always tricky to find and to fix.

Marigold:
> Would that be possible? Cause this function is used in many places, and issues with regions are always tricky to find and to fix.

Yep. I added comment `# DELETEME` after every `add_regions_aggregate` which triggered build for all these datasets. Datadiff looked clean.

PR #1981, opened 2023-11-27
:bug: fix overriding table's short name

Marigold:
The problem is in [these lines](https://github.com/owid/etl/blob/master/etl/steps/data/meadow/rff/2023-10-19/emissions_weighted_carbon_price.py#L40-L45)
```
# Read economy data.
tb_economy = pr.read_csv(z.open(DATA_PATH_ECONOMY), metadata=metadata, origin=origin)
tb_economy.metadata.short_name = "emissions_weighted_carbon_price_economy"

# Read coverage data.
tb_coverage = pr.read_csv(z.open(DATA_PATH_COVERAGE), metadata=metadata, origin=origin)
tb_coverage.metadata.short_name = "emissions_weighted_carbon_price_coverage"
```
`metadata` refers to the same object and the second `tb_coverage.metadata.short_name` updates the short name of the first table, too. Function `create_dataset`, which accepts two tables, then overwrites the tables.

I've added a uniqueness check for table short names and also fixed this by copying `metadata` object inside `pd.read_csv` function.

(Found this bug in our [full rebuild](https://buildkite.com/our-world-in-data/etl-full-private-rebuild-every-2nd-night/builds/379#018c0e13-c5de-4775-b67c-c104f1478958).)

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -158,9 +158,13 @@ def create_dataset(
     ds = catalog.Dataset.create_empty(dest_dir, metadata=default_metadata)
 
     # add tables to dataset
+    used_short_names = set()
     for table in tables:
         if underscore_table:
             table = catalog.utils.underscore_table(table, camel_to_snake=camel_to_snake)
+        if table.metadata.short_name in used_short_names:
+            raise ValueError(f"Table short name `{table.metadata.short_name}` is already in use.")
+        used_short_names.add(table.metadata.short_name)
         ds.add(table, formats=formats)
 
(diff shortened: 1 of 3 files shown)


pabloarosado (approved):
Looks good! The only thing I don't understand is why this happened now, and not when the step was created, less than two months ago.

Marigold:
That's a good question. I don't know either. It could be related to the plog changes (I were doing a lot of changes to `tables.py`) or it could be that the table somehow stayed  cached.

PR #1984, opened 2023-11-27
🔨 Flunet fix: Loosen filter on removing sparse years of data to allow for cases of rare strains e.g. AH5

spoonerf:


lucasrodes (approved):
LGTM!

PR #2003, opened 2023-11-28
📊 survey: Create consolidated surveys from barometers

paarriagadap:
Create trust questions dataset from the European Social Survey, rounds 9 and 10. The data is extracted from Stata code, which estimates the percentages of trust selecting scores 6 to 10 in a scale from 0 (no trust) to 10 (complete trust). These percentages are estimated using survey weights and stratification.

Part of this issue: https://github.com/owid/owid-issues/issues/1127


etl/steps/data/garden/survey/2023-08-04/trust_surveys.meta.yml
```diff
@@ -0,0 +1,46 @@
+# NOTE: To learn more about the fields, hover over their names.
```
paarriagadap, line 1: Hey @pabloarosado @Marigold ! How should I set this up to run without errors? I am concatenating three different tables and I get:
AssertionError: Column `trust` must have either sources or origins

I saw sources is deprecated and origins is not in the schema for the yaml, so how should I proceed?

etl/steps/data/meadow/ess/2023-08-02/ess_trust.py
```diff
@@ -0,0 +1,107 @@
...
+
+def run(dest_dir: str) -> None:
+    #
+    # Load inputs.
+    #
+    # Retrieve snapshot.
+    snap = cast(Snapshot, paths.load_dependency("ess_trust.csv"))
+
+    # Load data from snapshot.
+    df = pd.read_csv(snap.path)
```
Marigold, line 91: You should use `pr.read_csv` instead of `pd.read_csv` to preserve the origin.

etl/steps/data/meadow/ess/2023-08-02/ess_trust.py
```diff
@@ -0,0 +1,107 @@
...
+
+    #
+    # Process data.
+    df = rename_columns(df)
+
+    # Create a new table and ensure all columns are snake-case.
+    tb = Table(df, short_name=paths.short_name, underscore=True)
```
Marigold, line 98: If load the table with `pr.read_csv(...)` you'll be working with `Table` all the time and this won't be necessary. Underscoring can be done with
```
tb = tb.underscore()
```

etl/steps/data/garden/survey/2023-08-04/trust_surveys.meta.yml
Marigold: @paarriagadap the problem is that origins are not propagated properly. You should be using `pr.read_csv` and `pr.merge` instead of their `pd.` versions. See my other comments.

paarriagadap: Bah, right, thanks. I tried to adapt this old code from August and I forgot that. I had only changed the metadata and parts of the code.

PR #2012, opened 2023-11-30
🔨  etl-staging-sync improvements

Marigold:
## Changes

* Match existing charts on chart id instead of slug (which can change)
* Don't use `chart.updatedAt` when comparing whether chart has been updated on live, but use staging server creation time
* Show metadata diff if a chart has been updated on live
* Improve CLI help
* Add docs


pabloarosado (approved):
Looks good and works well!

PR #2020, opened 2023-12-01
ChatGPT tool for filling out metadata automatically 

veronikasamborska1994:
Features atm:
- at the moment can work with old snapshot dvc files and should work with @Marigold's automatically generated metadata grapher yaml files for automating the creation of data pages
- creates a new metadata file based on the new format in the same directory (with updated_with_gpt_ prefix) 

Possible changes/additions:

- We can change it so that a user can select their preferred destination directory (e.g., if we are updating a dataset this might be useful but might want to leave some fields explicitly blank - e.g., date updated) 
- It won't work super well with the garden step's yaml atm because it doesn't yet have a more precise prompt --> can think of how to make this more useful depending on what we might want to use it for (e.g., if we want to use it for future updates then maybe integrating it with the tool that we used to have for automatically generating a yaml file based on the garden step might be the best option) 

@Marigold what do you think? 

p.s. the garden step will definitely be trickier to implement and we might need to do something along the lines of the country standardising tool we have atm --> will need to break down each prompt into smaller ones (maybe just feed the background dataset metadata and the indicator name (needs to be informative) and any other information we already have as one prompt) to avoid running into issues with our prompts being too long..


Marigold:
Sweet! I've tried it on Data Pages migration and [created a PR](https://github.com/owid/etl/pull/2021) to easily see suggestions. It's a pretty good starting point. I can imagine this being used internally in `etl-metadata-migrate` - flag `--metagpt` would do AI post-processing of YAML and replace those "TBD - ..." comments with "AI - ..." comments (so the user would only check the output and remove `AI -` prefix).

A couple of observations:
- It truncates the output when YAML is too long. Maybe we could ask it to only fill metadata for a **single indicator**.
- We can ask for returning **just the updated fields** and then update the YAML ourselves (I can give you snippets how to do that). It could also be faster & cheaper.
- It changes `origin.description` (would be fixed by only returning changed fields for an indicator)
- We don't need to fill topic tags or `update_period_days` or `processing_level`, these are easy.
- There might be some overlap with FAQs. Perhaps we could add FAQ entries to the input and tell it not to repeat itself. (This might be trickier to implement, so probably not worth doing atm)
- I think you can set `json_object` [response format](https://github.com/openai/openai-python#nested-params). That could be helpful? (It would go well together with only returning changed fields) (You might have to upgrade openai with `poetry add openai@latest`)
- It would be helpful to print openai query cost (it's available in the response somewhere)

Overall, this looks very promising. I have high hopes of merging it with `etl-metadata-migrate` and having automatic data page migrations.

apps/metagpt/cli.py
```diff
@@ -0,0 +1,198 @@
...
+    - URL to data product's main site.
+
+15. version_producer (string or number, recommended if existing)
+    - Use producer's version naming.
+"""
+
+
+# Main CLI command setup with Click
+@click.command()
+@click.option("--path-to-file", prompt=True, type=str, help="Path to the metadata file.")
```
Marigold, line 96: It'd be nice to have the option for overwriting input file, i.e.
```
@click.option(
    "--overwrite",
    default=False,
    type=bool,
    help="Overwrite input file if set to True. Otherwise create a new file with prefix `updated_with_gpt_`.",
)
```

apps/metagpt/cli.py
```diff
@@ -0,0 +1,198 @@
...
+    except Exception as e:
+        log.error("Metadata update process failed.", error=str(e))
+
+
+def read_metadata_file(path_to_file: str) -> str:
+    """Reads a metadata file and returns its content."""
+    try:
+        with open(path_to_file, "r") as file:
+            return file.read()
+    except IOError as e:
```
Marigold, line 114: It's unnecessary to catch the error, print error message and then reraise it. You could just use
```
with open(path_to_file, "r") as file:
    return file.read()
```
for simplicity.

veronikasamborska1994:
> created a PR



> * We can ask for returning **just the updated fields** and then update the YAML ourselves (I can give you snippets how to do that). It could also be faster & cheaper.

regarding this point I think there are two potential solutions here depending on whether we are only solving this particular issue or if we want to later adapt it to infer metadata in the garden step for working with new datasets as well. For this particular problem we can keep the prompt at the metadata file level and only ask it to generate the changed fields. Alternatively, if we anticipate these files to be pretty big then we'd want to create a separate prompt for each indicator and fill it out one by one (I think the cumulative cost will end up being higher for this though). What do you think?

Marigold:
Your point about garden metadata files and flexibility in general makes sense, I'll leave it up to you then. 

> any others you think will be particularly useful? I have a sense that asking it to generate titles hasn't been working super well for me.

I don't work with metadata on a daily basis, so your call ;).

Marigold:
@veronikasamborska1994, do you think this is already in a state to be merged and started being used by others? (e.g. to help with data pages migrations)

veronikasamborska1994:
> @veronikasamborska1994, do you think this is already in a state to be merged and started being used by others? (e.g. to help with data pages migrations)

hey @Marigold! sorry I've been away for a week and have a couple of things to catch up on first but I'll take a look tomorrow and see if it's ready :)

veronikasamborska1994:
> This looks super useful, @veronikasamborska1994!
> 
> Maybe you could add an entry to the docs explaining how to use this tool? If you go for it, I'd suggest creating a file like `docs/tutorials/metadata_gpt.md` and put there all the docs. In order to be correctly listed in the page, you should add this new file to [mkdocs.yml](https://github.com/owid/etl/blob/master/mkdocs.yml). Find more details on how to add a new entry to the docs here: https://docs.owid.io/projects/etl/dev/docs/

thanks @lucasrodes! will add the docs later today! I'll merge this today and will probably have more time this week to look into whether we can do something similar for the garden step!

PR #2027, opened 2023-12-04
📊 Adding TB disaggregated estimate data

spoonerf:


snapshots/tuberculosis/2023-11-27/burden_estimates.csv.dvc
```diff
@@ -3,11 +3,11 @@
 meta:
   origin:
     # Data product / Snapshot
-    title: Global tuberculosis report
+    title: Global Tuberculosis Report
     description: |-
       WHO has published a global tuberculosis (TB) report every year since 1997. The report provides a comprehensive and up-to-date assessment of the TB epidemic, and of progress in prevention, diagnosis and treatment of the disease at global, regional and country levels.
     date_published: 2023-11-07
-    title_snapshot: Global tuberculosis report - burden estimates
+    title_snapshot: Global Tuberculosis Report - Burden Estimates
```
pabloarosado, line 10: A couple of comments on the rest of this file:
* The license should not appear in citation_full.
* The license.url should be the place within the WHO website where they explain their license, not the explanation of the license itself.

pabloarosado: The same applies to the other dvc file.

etl/steps/data/meadow/tuberculosis/2023-12-04/burden_disaggregated.py
```diff
@@ -0,0 +1,36 @@
...
+def run(dest_dir: str) -> None:
+    #
+    # Load inputs.
+    #
+    # Retrieve snapshot.
+    snap = paths.load_snapshot("burden_disaggregated.csv")
+
+    # Load data from snapshot.
+    tb = snap.read()
+    tb = tb.drop(columns=["iso2", "iso3", "iso_numeric"])
```
pabloarosado, line 18: The last line should be part of Process data.

etl/steps/data/garden/tuberculosis/2023-11-27/burden_disaggregated.py
```diff
@@ -0,0 +1,183 @@
...
+                region=region,
+                ds_regions=ds_regions,
+                ds_income_groups=ds_income_groups,
+            )
+            gr_cal = gr[["country", "year", "best", "lo", "hi", "population"]]
+            # Add region aggregates.
+            gr_reg = geo.add_region_aggregates(
+                df=gr_cal,
+                region=region,
+                countries_in_region=countries_in_region,
```
pabloarosado, line 125: Unfortunately, you must pass here the argument `countries_that_must_have_data=[]`, otherwise it will impose that certain countries must be informed. This is one of the reasons we should create a new function to create region aggregates.

pabloarosado: Even worse than that, it will silently load `key_indicators` to achieve that, which is not listed as a dependency. So, to avoid this issue, simply add the `countries_that_must_have_data=[]`.

etl/steps/data/garden/tuberculosis/2023-11-27/burden_disaggregated.py
```diff
@@ -0,0 +1,183 @@
...
+    return tb_gr_out
+
+
+def add_population_column(tb: Table) -> Table:
+    """
+    Adding the population for each age-group, in rows where the risk factor is "all".
+    """
+    tb_pop = tb[tb["risk_factor"] == "all"]
+    tb_no_pop = tb[tb["risk_factor"] != "all"]
+    tb_pop = add_population(
```
pabloarosado, line 147: Hmm, this is another function we should rethink. Under the hood, it loads `Dataset(DATA_DIR / "garden" / "un" / "2022-07-11" / "un_wpp")`. So, if you want to use this function, ensure that dependency is listed in the dag.

etl/steps/data/garden/tuberculosis/2023-11-27/burden_disaggregated.py
```diff
@@ -0,0 +1,183 @@
...
+
+    return tb_gr_out
+
+
+def add_population_column(tb: Table) -> Table:
+    """
+    Adding the population for each age-group, in rows where the risk factor is "all".
+    """
+    tb_pop = tb[tb["risk_factor"] == "all"]
+    tb_no_pop = tb[tb["risk_factor"] != "all"]
```
pabloarosado, line 146: To avoid warnings, add `reset_index(drop=True)` after slicing dataframes/tables.

pabloarosado (approved):
Looks good, except for some minor changes to avoid dependency issues.

snapshots/tuberculosis/2023-11-27/burden_estimates.csv.dvc
spoonerf: Ah okay - the source includes the licence in their citation, so I wasn't sure. I should still remove it?

etl/steps/data/garden/tuberculosis/2023-11-27/burden_disaggregated.py
spoonerf: Ah I didn't realise this before - thanks!

snapshots/tuberculosis/2023-11-27/burden_estimates.csv.dvc
pabloarosado: If they are very explicit about how to cite them, then follow their guidelines. Otherwise, it's better to remove the license from there (since it's defined elsewhere). Given that this citation is propagated, and the license of the data can change down the line, it's better to not mention any license outside the appropriate fields.

PR #2028, opened 2023-12-04
Fix missing countries in crop yields explorer

pabloarosado:
Fix missing countries in crop yields explorer. They were missing because Mueller et al (2012) does not have data for many countries (and in some cases uses a definition that is different to ours).

This PR will add those missing countries. It will also add custom FAO regions. I may remove them if they become confusing in the explorer.


edomt (commented):
All good, thank you @pabloarosado!

PR #2031, opened 2023-12-05
:bug: remove isTopic from tags

Marigold:
Column `tags.isTopic` was removed from MySQL. Reflect that in ETL.

@ikesau I'm not very familiar with how tags work on our site. Is there a way now to tell whether a tag is a topic tag? Topic tags had unique names, which was necessary to get their IDs in ETL. I don't they are unique any more. Any ideas how to filter them?

### DIFF
--- a/apps/metadata_migrate/cli.py
+++ b/apps/metadata_migrate/cli.py
@@ -262,7 +262,7 @@ def cli(
 
         print("\n[bold yellow]Follow-up instructions:[/bold yellow]")
         print(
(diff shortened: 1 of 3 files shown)


danyx23:
@Marigold we now have a slug field that we use for this purpose. If the slug is not null then it is a topic page. @ikesau you said that the "full" version to check this is to join on posts and posts_gdocs and check if it is a topic page there but I think for now we could just say as a convention that slug should only be set for topic tags - agree?

PR #2036, opened 2023-12-05
Update owid_co2 dataset

pabloarosado:
This PR:
* Updates `owid_co2` dataset.
* Updates other dependencies (namely the `country_profile` dataset).
* Ensures old GCB dataset can safely be archived.

I did not move the old GCB dataset to the archive. I think in general it would be good to be able to execute old versions in ETL. I tried moving those steps to the archive dag and executing `etl data://garden/gcp/2023-09-28/global_carbon_budget --dag-path dag/archive/main.yml`, but that failed. So, ETL is currently not ready to execute archive steps. It's not urgent, but it would be good to find a generic solution for this at some point.

For now, I simply left old steps in the same active dag, in a section of the yaml file that lists steps that can safely be archived, if needed.

### DIFF
--- a/dag/emissions.yml
+++ b/dag/emissions.yml
@@ -45,11 +45,11 @@ steps:
   data://grapher/gcp/2023-12-05/global_carbon_budget:
     - data://garden/gcp/2023-12-05/global_carbon_budget
   #
-  # Emissions - CO2 dataset (2023-11-23).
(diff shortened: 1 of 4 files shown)


Marigold (approved):
Looks good. If you run into the archiving issue again, could you please copy that branch and ping me? It's hard to replicate it from scratch.

PR #2038, opened 2023-12-05
:sparkles: parallel ETL

Marigold:
Continuation of https://github.com/owid/etl/pull/1331

Add option `etl ... --workers` for parallel processing of steps. This should help when rebuilding ETL after regions updates or if you want to speed it up locally. It still uses 1 process by default, but this can be customised with `--workes` or with `RUN_STEPS_WORKERS`.

There's not much overhead when spawning multiple processes. Unfortunately, snapshots can't run in parallel due to some internal DVC locking issues (I had to add file lock to get rid of them). This is suboptimal, but I've tried to put grapher steps in front of the queue which should make it slightly better. (Another option would be to pull all dirty snapshot steps before running all the steps - that would be efficient and could happen in parallel. However, pulling snapshots is not typically a bottleneck as you only have to do it once).

I plan to run all CI etl runs with `RUN_STEPS_WORKERS=8` for a start and see whether the foundation can handle it (I was testing it with 32 workers, but that explodes memory usage).

## Appendix:

There are three types of parallelisation happening right now:
1. `DIRTY_STEPS_WORKERS` - number of threads when computing `is_dirty` (to parallelise etag requests)
2. `GRAPHER_INSERT_WORKERS` - number of threads when upserting to MySQL
3. `RUN_STEPS_WORKERS` - number of **processes** to use in ETL


etl/config.py
```diff
@@ -100,6 +100,12 @@ def variable_metadata_url(variable_id):
 # number of workers for grapher inserts
 GRAPHER_INSERT_WORKERS = int(env.get("GRAPHER_WORKERS", 40))
 
+# number of workers for checking ditry steps
+DIRTY_STEPS_WORKERS = int(env.get("DIRTY_STEPS_WORKERS", 5))
+
+# number of processes for running steps
+RUN_STEPS_WORKERS = int(env.get("RUN_STEPS_WORKERS", 1))
```
pabloarosado, line 107: I'm a bit confused: You define here some default values of workers (which by the way would be good to justify, why 1, 5 or 40), but then in other places the default values of functions can be either a hardcoded number, or taken from config, or cpu_counts(). I don't know if all those choices were intentional, but it would be good to explain them.

etl/command.py
```diff
@@ -270,22 +287,100 @@ def run_dag(
...
+    strict = _detect_strictness_level(step, strict)
+    with strictness_level(strict):
+        time_taken = timed_run(lambda: step.run())
+
+    print(f"--- Finished {step_name} ({time_taken:.0f}s)", flush=True)
+
+
+def enumerate_steps(steps: List[Step]) -> None:
+    for i, step in enumerate(steps, 1):
+        print(f"{i}. {step}...")
```
pabloarosado, line 383: Personally I have never liked the "..." at the end, cause it looks as if the name of the step is incomplete, and it also makes it less convenient to copy/paste.

etl/command.py
```diff
@@ -270,22 +287,100 @@ def run_dag(
...
+        # different attributes
+        exec_graph[str(step)] = {str(dep) for dep in step.dependencies if str(dep) in steps_str}
+
+    exec_graph_parallel(exec_graph, _exec_step_job, workers, dag=dag, strict=strict)
+
+
+def exec_graph_parallel(
+    exec_graph: Dict[str, Any], func: Callable[[str], None], workers: int, use_threads=False, **kwargs
+) -> None:
+    # Create a topological sorter and add tasks with their dependencies
```
pabloarosado, line 338: Could you please briefly explain the logic of this sorter?

pabloarosado (approved):
Amazing!! I couldn't follow the logic of the number of workers. I suggested some clarifications in the code. Apart from that, it looks good.
I tried running the steps from namespace `papers`, and I timed it as a function of workers (e.g. `etl papers --workers 1 --force):
* 1 worker -> 1m 04s
* 2 workers -> 42s
* 4 workeres -> 31s
* 8 workers -> 26s
* 12 workers -> 25s
So it works well! Of course, the log becomes impossible to follow on the terminal, but I don't know if there's an easy way to fix that (like storing them and printing them in order afterwards). Anyway, this speedup is going to be very useful!

etl/config.py
Marigold: I've added some explanations. Values in `config.py` are default ones which can be overridden by `--workers`. The other ones for threads have good defaults and shouldn't be changed unless you are debugging core ETL. Ideally, we'd have just a single variable for `--workers` and the rest would be handled async instead of threads.

etl/command.py
Marigold: Makes sense, fixed.

etl/command.py
Marigold: Good point, added.

etl/command.py
```diff
@@ -270,22 +288,109 @@ def run_dag(
...
+    # the load on MySQL
+    steps = sorted(steps, key=lambda x: "grapher://" not in str(x))
+
+    # create execution graph from steps
+    exec_graph = {}
+    steps_str = {str(step) for step in steps}
+    for step in steps:
+        # only add dependencies that are in the list of steps (i.e. are dirty)
+        # NOTE: we have to compare their string versions, the actual objects might have
+        # different attributes
```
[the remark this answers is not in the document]
Marigold: Oh right, that'll be much nicer. I'll add this in a sec.

etl/steps/__init__.py
```diff
@@ -656,7 +668,11 @@ def run(self) -> None:
 
         with _unignore_backports(Path(self._path)):
             try:
-                self._cached_dvc_repo().pull(self._path, remote="public-read", force=True)
+                repo = self._cached_dvc_repo()
+                # DVC must be locked across processes. This is pretty limiting as it allows us to only pull
+                # one snapshot at a time. One option is to pre-pull all snapshot steps.
+                with dvc_pull_lock:
+                    repo.pull(self._path, remote="public-read", force=True)
```
larsyencken, line 675: This turns out to be a big slow-down if you're building the ETL from scratch. If we wanted to speed this up as well, we'd need to preferentially execute non-snapshot steps whenever they're available.

This is more of a note for the future. See earlier comment about the ordering of steps into the topological sorter.

larsyencken:
Of note, `foundation-1` has half the swap of the other servers, and it's the swap space you need to handle the moments of max memory pressure that spill over.

etl/command.py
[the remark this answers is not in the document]
Marigold: I've tried implementing it, but the problem is that our steps are also defined as `dataclasses`, which override `__hash__` and `__eq__` methods. I'm unsure how much we benefit from having them as data classes. In any way, the code wouldn't get much simpler with `__eq__` and `__hash__`.

etl/command.py
```diff
@@ -270,22 +288,109 @@ def run_dag(
...
+        with strictness_level(strict):
+            time_taken = timed_run(lambda: step.run())
+            click.echo(f"{click.style('OK', fg='blue')} ({time_taken:.1f}s)")
+            print()
+
+
+def exec_steps_parallel(steps: List[Step], workers: int, dag: DAG, strict: Optional[bool] = None) -> None:
+    # put grapher steps in front of the queue to process them as soon as possible and lessen
+    # the load on MySQL
+    steps = sorted(steps, key=lambda x: "grapher://" not in str(x))
```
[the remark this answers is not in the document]
Marigold: > Is this actually honoured by the topological sorter?

I briefly checked the code, and it looks like it. I've added other channels as you suggested.

etl/steps/__init__.py
Marigold: True. I'm honestly a bit disappointed by DVC and its multiprocessing. In the future we might consider writing custom pull function (adapted from walden) or replace the entire DVC.

(In nightly builds when building ETL from scratch we `dvc pull` all files before running the actual ETL. That happens in parallel and is pretty fast, like 5 mins.)

PR #2047, opened 2023-12-06
Improve add region aggregates

pabloarosado:
We have been carrying issues with `geo.add_region_aggregates` for a while. Given that we never allocate time to properly define a function or class to properly handle region aggregates, this PR will, at least:
1. Ensure that the function we currently have uses good default values.
2. When silently loading datasets, we get a warning.

Main changes in this PR:
* Added warnings to all functions that silently load datasets. They should not be used in any new steps, and the warning will help us identify misuses in old steps.
* Changed `owid.datautils.dataframes.groupby_agg`, so that by default it behaves like pandas `groupby().agg()`. Also, it now receives an argument `min_num_values`, which is the minimum number of non-nan values that the group needs to have in order to generate a non-nan aggregate.
* Changed `etl.data_helpers.geo.add_region_aggregates` so that, by default, it doesn't impose any conditions on the countries that must be informed, or the number of allowed nans. It also now has the option to impose a minimum number of non-nan values.

I searched for all current uses of those functions and ensured that they keep the old default values of the function (although, in some cases, the default values may not make a difference).

TODO: Check that datadiff shows no difference in any of the affected datasets.


etl/data_helpers/population.py
```diff
@@ -57,9 +65,12 @@ def add_population(
     pd.DataFrame
         Dataframe with extra column `population`.
     """
-    # Load granular population dataset
-    ds = Dataset(DATA_DIR / "garden" / "un" / "2022-07-11" / "un_wpp")
-    pop = ds["population_granular"].reset_index()
+    if ds_un_wpp is None:
+        ds_un_wpp_path = DATA_DIR / "garden" / "un" / "2022-07-11" / "un_wpp"
```
Marigold, line 69: Super Nit - when I search for steps (e.g. for refactoring), I usually their URI like `garden/un/2022-07-11/un_wpp`. If this was written as
```
ds_un_wpp_path = DATA_DIR / "garden/un/2022-07-11/un_wpp"
``` 
it would be easier to find. (Feel free to ignore this!)

Marigold:
> I have a vague memory that sometimes indexes where changing order arbitrarily, but the data was not affected (specifically in the war dataset). Do you remember or understand why this happens?

I checked that and you're right, only index ordering has changed.

PR #2053, opened 2023-12-08
:bug: Fix sources of UN SDG

Marigold:
There was an issue with UN SDG with sources defined in meadow file as
```
sources:
  - name: United Nations
  - url: https://sdgs.un.org/goals
```

This PR fixes it plus a few other enhancements.


spoonerf (approved):
Looks good to me!

PR #2056, opened 2023-12-08
Create `add_regions_to_table` that should be the new default function to create region aggregates

pabloarosado:
Create `add_regions_to_table` that should be the new default function to create region aggregates.


pabloarosado:
Hi @Marigold can you please help me understand the result of datadiff?
1. I see many metadata changes like:
```
= Dataset meadow/nasa/2023-03-06/ozone_hole_area
--
  | = Table ozone_hole_area
  | ~ Column max_hole_area (changed metadata)
  | - title: max_hole_area
  | + {}

```
Do you understand what this means?

2. There seems to be an error, that takes a big chunk of the log:
```
~ Dataset meadow/un/2023-01-24/un_sdg
--
  | - licenses:
  | -   - {}
  | = Table un_sdg
  | 2023-12-11 19:46:38 [error    ] boolean value of NA is ambiguous

```
3. Given that the log is too long, buildkite doesn't show the whole thing, so I downloaded the full log. But the file is hard to parse, it has many unusual characters, I guess due to the coloring of the log. Do you know a simple way to open it and be able to visualize it properly? Thanks!

Marigold:
1. The following
```
= Dataset meadow/nasa/2023-03-06/ozone_hole_area
--
  | = Table ozone_hole_area
  | ~ Column max_hole_area (changed metadata)
  | - title: max_hole_area
  | + {}
```
means that there is a change in metadata for column `max_hole_area`. Title `max_hole_area` got removed and the new title is None. So, something is not preserving titles.

`etl-datadiff` has an option `--snippet` which gives you a python snippet you can paste into the notebook and dig in the data. For instance
```
etl-datadiff REMOTE data/ --include meadow/nasa/2023-03-06/ozone_hole_area --cols max_hole_area --verbose --snippet
```
returns
```python
from owid.catalog import RemoteCatalog, Dataset

ta = RemoteCatalog(channels=["meadow"]).find_one(table="ozone_hole_area", dataset="ozone_hole_area", version="2023-03-06", namespace="nasa", channel="meadow")
tb = Dataset(DATA_DIR / "meadow/nasa/2023-03-06/ozone_hole_area")["ozone_hole_area"]
```
then you immediately see that `ta.max_hole_area.m.title, tb.max_hole_area.m.title` are different.

pabloarosado:
Hi @Marigold, thanks. I see that those changes have nothing to do with this PR. I think that some time ago, when creating a `Table(df)`, the column name was used as an indicator's title, but then we decided to avoid that (because, e.g. `max_hole_area_date` would not be a good title).
This means that datasets in production haven't been rebuilt for a while. Then I suppose I don't have a good reference to compare the output of this PR with.
I think this PR should not affect metadata, only data. So I'd like to see in the log if any of the data changed (regardless of the metadata). After a quick look, I didn't see any changes in data, but as I said before, the log is poorly formatted, so it's hard to tell.

Marigold:
1. I'm looking into it.
2. I've added a fix to your PR for the following
```
~ Dataset meadow/un/2023-01-24/un_sdg
--
  | - licenses:
  | -   - {}
  | = Table un_sdg
  | 2023-12-11 19:46:38 [error    ] boolean value of NA is ambiguous
```
Data has changed its type from category to int for some reason. Probably worth checking what's going on (it might make sense to wait after I figure our the metadata issue above)
```
(etl-py3.10) ➜  etl2 git:(create-add-regions-to-table) ✗ etl-datadiff REMOTE data/ --include meadow/un/2023-01-24/un_sdg --cols baseperiod --verbose
~ Dataset meadow/un/2023-01-24/un_sdg
    - licenses:
    -   - {}
  = Table un_sdg
    ~ Column baseperiod (changed data)
        - Changed values: 2216310 / 2216310 (100.00%)
        - country: 343 affected
        - year: 64 affected
        - Removed values: 2015.0, nan, 2011.0, 2020.0, 2017.0, 2019.0
        - New values: 2011, 2020, 2017, 2015, 2019
```

Marigold:
3. Those weird escape characters in logs are ANSI colors and symbols. You can try installing this [VSCode extension](https://marketplace.visualstudio.com/items?itemName=iliazeus.vscode-ansi) and opening it in VSCode. That could help a bit, though when I tried it now, it didn't get much better. Let me know if you find a better way.

(Ideally, we'd log less into buildkite)

Marigold:
@pabloarosado you were right that some datasets in production have outdated metadata, because we didn't increment epoch after changing titles (which I think was a fine decision). I'll wait for datadiff and refresh all problematic datasets.

I also plan to move datadiff from `etl-build-public-datasets-branch` to automated staging server. The reason is that the former rebuilds ETL from scratch on every commit, while data on staging server persists between runs and is resumable.

etl/data_helpers/geo.py
```diff
@@ -851,3 +900,314 @@ def list_members_of_region(
     countries = sorted(countries_set)
 
     return countries
+
+
+def detect_overlapping_regions(
```
Marigold, line 905: It'd be cool to have type hints on this one.

etl/data_helpers/misc.py
```diff
@@ -152,3 +155,140 @@ def add_origins_to_mortality_database(tb_who: Table) -> Table:
 
 
 ########################################################################################################################
+
+
+def compare_tables(
```
Marigold, line 160: Having a shared module with QA functions, especially visualisations, is a great idea.

Marigold (commented):
I haven't checked every line, but it looks great at high level. I haven't found any performance or structural red flags. Let's start using this and see how it works.

Marigold (approved):
I haven't reviewed every line, but it looks great at high level. I haven't found any performance or architecture red flags. Let's start using this and improve it on the go.

PR #2058, opened 2023-12-11
:bug: Replace DVC pull with direct fetch

Marigold:
UPDATED 2024-01-02

## Problem

DVC has become a bit annoying in some aspects:

1. DVC doesn't interact well with multiprocessing (we got around this by using various hacks, and we can only pull a single snapshot at a time).
2. DVC is slow with the large number of DVC files we have (the current workaround is to dynamically use `.dvcignore` for backported datasets... it works, but is prone to errors).
3. DVC is spamming our logs because of these workarounds.
4. The new major version of DVC doesn't support our file model (which involves having DVC YAML files next to data files).
5. We can't add comments to `*.dvc` files.

We only use a small fraction of DVC's capabilities (just storing files on remote S3 storage), so it's worth considering whether dealing with these issues is justifiable. Initially, we thought we might leverage DVC more extensively, but that hasn't materialized.

## Solution

While we could consider replacing DVC, most of the problems mentioned above are solved by reimplementing `dvc pull` from `Snapshot.pull` and the MD5 checksum comparison in `Snapshot.is_dirty`. The only annoying thing is that **DVC uses its own MD5 function**.

Checksum comparison is only done for files < 10MB. Larger snapshots are compared based on size only. (This significantly speeds up ETL)

We could similarly implement `dvc add`, but that isn't critical and would only slightly improve performance.

### DIFF
--- a/etl/files.py
+++ b/etl/files.py
@@ -11,7 +11,7 @@ import time
 from collections import OrderedDict
 from pathlib import Path
 from threading import Lock
-from typing import Any, Dict, Generator, List, Optional, Set, TextIO, Union
+from typing import Any, Dict, Generator, List, Literal, Optional, Set, TextIO, Union
 
 import ruamel.yaml
 import yaml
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -11,7 +11,7 @@
   "[python]": {
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true
+      "source.organizeImports": "explicit"
```
larsyencken, line 14: Seems like this got added by accident, want to strip it from this PR?

(...and probably, we do still want format on save for imports too?)

etl/steps/__init__.py
```diff
@@ -666,13 +665,12 @@ def _cached_dvc_repo(self) -> Any:
     def run(self) -> None:
         from dvc.exceptions import CheckoutError
 
-        with _unignore_backports(Path(self._path)):
+        with dvc_pull_lock, _unignore_backports(Path(self._path)):
             try:
                 repo = self._cached_dvc_repo()
```
larsyencken, line 670: So, this PR is basically about putting this line inside the lock. Can you explain what that does? (both here, and in a comment?)

Marigold:
I went a bit further and tried replacing DVC pulling with direct fetching from R2. The code is a bit simpler, but I worry about ending up replacing a lot of DVC functionality. Parking this for now.

Marigold:
@larsyencken I requested a review, but feel free to reassign if you're busy.

Marigold:
@larsyencken I looked into why DVC and our checksum differ, and it's because they replace `\r\n` from Windows with `\n`. This affects about 10 csv files in snapshots. Doing it by default in our checksum calculation would simplify the code a lot and it wouldn't affect anything else (since no one is using Windows). In addition, I think it's a pretty reasonable thing to do. Can I go ahead and do it?

larsyencken:
> @larsyencken I looked into why DVC and our checksum differ, and it's because they replace `\r\n` from Windows with `\n`. This affects about 10 csv files in snapshots. Doing it by default in our checksum calculation would simplify the code a lot and it wouldn't affect anything else (since no one is using Windows). In addition, I think it's a pretty reasonable thing to do. Can I go ahead and do it?

For CSV, JSON, etc, go ahead. Do they also do that for arbitrary, large binary files?

etl/files.py
```diff
@@ -50,27 +50,63 @@ def clear(self) -> None:
...
     "Return the md5 hex digest of the file without using cache."
     chunk_size = 2**20
     _hash = hashlib.md5()
     with open(filename, "rb") as istream:
         chunk = istream.read(chunk_size)
         while chunk:
+            if typ == "dvc":
+                if istextblock(chunk[:DEFAULT_CHUNK_SIZE]):
+                    chunk = dos2unix(chunk)
+
```
larsyencken, line 96: Looks reasonable 👍

etl/snapshot.py
```diff
@@ -117,55 +119,56 @@ def _cached_dvc_repo(self) -> Any:
...
+        # update metadata
+        with open(self.metadata_path, "r") as f:
+            m = ruamel_load(f)
+
+        m["outs"][0]["md5"] = checksum_file(self.path)
+        m["outs"][0]["size"] = self.path.stat().st_size
+        m["outs"][0]["path"] = self.path.name
+
+        with open(self.metadata_path, "w") as f:
+            f.write(ruamel_dump(m))
```
larsyencken, line 153: I hadn't come across `ruamel` before. It might be more readable if the `ruamel` methods are `yaml_load` and `yaml_dump` methods instead, but it's only a soft preference.

Marigold:
> Do they also do that for arbitrary, large binary files?

Nope, they do it only for text files (there's `istextblock` function I borrowed from them).

larsyencken (approved):
Minor feedback only, looks good!

etl/snapshot.py
```diff
@@ -116,47 +118,56 @@ def _cached_dvc_repo(self) -> Any:
...
+
+        assert len(self.metadata.outs) == 1
+        file_size = self.path.stat().st_size
+        # Compare file size if it's larger than 10MB, otherwise compare md5
+        # This should be pretty safe and speeds up the process significantly
+        if file_size >= 10 * 2**20:  # 10MB
+            return file_size != self.m.outs[0]["size"]
+        else:
+            # NOTE: DVC uses different md5 checksum than md5sum
+            return checksum_file(self.path.as_posix(), typ="dvc") != self.m.outs[0]["md5"]
```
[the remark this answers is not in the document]
Marigold: I've been thinking about that, but then I decided that managing an on-disk cache would be too much of a hassle. We already use a runtime cache that gets us most of the way there. I thought that size comparison could be pretty safe for large files.

etl/snapshot.py
Marigold: We already have both `ruamel_dump` and `yaml_dump`. ruamel tries its best to minimize modifications to the yaml file, while `yaml_dump` strips all comments and style. Maybe we could unify them one day.

PR #2069, opened 2023-12-12
:bug: fix chart revisions for edge cases

Marigold:
There were two issues with chart revisions:
1. Some charts don't have `config["id"]` - Hotfixed by creating it when loading the chart, but the plan is to fix all charts like that in MySQL
2. Some suggested revisions don't have `version` - this was caused by function that was deleting properties with default values (where `version = 1`) even though the property is required

I've tested it locally. If you run into an error that says chart revisions don't have a unique `chartId`, you'll have to delete that revision manually from DB (this happens because of previous failed attempts).


pabloarosado (approved):
Thanks a lot for fixing this, looks good!

PR #2075, opened 2023-12-13
Fix issue in tuberculosis step

pabloarosado:
The problem seems to happen because "country_col" is categorical, and it doesn't like using ".any()" when the column is empty.


Marigold (approved):
Looks good!

PR #2076, opened 2023-12-13
📊 Adding sum of TB antibiotic susceptible patients with disaggregated HIV status

spoonerf:
Also trying out the new `add_regions_to_table()` function


pabloarosado (approved):
Looks good!

PR #2085, opened 2023-12-13
Propagate metadata on fillna

pabloarosado:
When filling with nans using another table, we should combine the metadata of the two tables.
This change will also make `datautils.dataframes.combine_two_overlapping_dataframes` automatically propagate metadata of the two tables.

`fillna` is a very common operation (although it's unusual to fill with entire tables), and `combine_two_overlapping_dataframes` is also fairly commonly used. So it would be very good to check if this PR alters the metadata of existing steps.

One possible issue I can imagine happening is that, when doing tb1.fillna(tb2), one may expect that the title and description of tb1 should be preserved. But, if tb1 and tb2 have different titles and descriptions, the combination will have no title/description (which also makes sense).

### DIFF
--- a/lib/catalog/owid/catalog/tables.py
+++ b/lib/catalog/owid/catalog/tables.py
@@ -895,6 +895,20 @@ class Table(pd.DataFrame):
         tb = tb.set_index(column_idx_new)
         return tb
 
+    def fillna(self, value, **kwargs) -> "Table":
+        """Usual fillna, but, if the object given to fill values with is a table, transfer its metadata to the filled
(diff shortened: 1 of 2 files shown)


Marigold (approved):
Nice one!

pabloarosado:
* I've checked that all uses of `fillna` in ETL are filling with either a single value, or a variable, never a table. So this PR should not affect any of them.
* I've checked that all uses of `combine_two_overlapping_dataframes` yield the same, or a better result (with more complete metadata), with this PR.

PR #2092, opened 2023-12-14
Avoid metadata warning on division

pabloarosado:
Avoid a warning when dividing two variables with different units (which is very common, e.g. when computing per capita variables).

We may want to make this logic a bit more generic in the future, but for now this will significantly reduce the number of warnings.


Marigold (approved):
Looks good!

PR #2104, opened 2023-12-15
:hammer: warn if display.name is set, but not presentation.title

Marigold:


etl/helpers.py
```diff
@@ -98,6 +98,20 @@ def grapher_checks(ds: catalog.Dataset) -> None:
...
+            # the Table tab
+            # [title_public > display.name > title] - [title_variant] - [attribution_short]
+            # and chart heading
+            # [grapher_config.title > title_public > display.name > title] - [grapher_config.subtitle > description_short]
+            #
+            # Warn if display.name (which is used for legend) exists and there's no title_public set. This
+            # would override the indicator title in the Data Page.
+            if tab[col].m.display.get("name") and not tab[col].m.presentation.title_public:
+                log.warning(
+                    f"Column {col} uses display.name, but not presentation.title_public. It would override the indicator title in the Data Page.",
```
pabloarosado, line 112: To make the solution clearer, I'd say:
"Column {col} uses display.name but no presentation.title_public. Ensure the latter is also defined, otherwise display.name will be used as the indicator's title."

pabloarosado (approved):
Looks good!

etl/helpers.py
Marigold: Thanks! fixed

PR #2110, opened 2023-12-15
Adapt renewable energy patents steps

pabloarosado:
Adapt renewable energy patents steps.

Good news: This was the only remaining snapshot from walden!


Marigold (approved):
Looks good!

PR #2130, opened 2023-12-20
Improve format of war shared file

pabloarosado:
Hey @Marigold this file has been showing up every time I do `make test` (because it had an unused import). And `make check` doesn't notice it. I just created this PR in case this is a potential issue of `make check` (maybe it ignores `shared.py` files).


Marigold:
Thanks for the heads up! It looks like we've merged something (i.e. that unused import) that didn't pass `make check`. Since `make check` isn't checking files that are different from `master`, this never popped up again.

If you run into this again, just fix it in your PR (no need to create a special PR if the issue is so trivial).

Marigold (approved):

PR #2145, opened 2023-12-28
Fix attribution in World GDP dataset

pabloarosado:
I have changed the attribution of the World GDP dataset, since [Max pointed out](https://owid.slack.com/archives/C03NV9Z3YSV/p1703764800168289) that it was wrong. But I'm not sure if this is the correct attribution (I haven't worked on this dataset), @lucasrodes feel free to change it if this is not correct.


paarriagadap (approved):
Thanks @pabloarosado! Yes, this is the best fix for now, and I can expand the data page later.

PR #2147, opened 2023-12-29
📊 World Bank Gender Statistics 

veronikasamborska1994:


etl/steps/data/garden/wb/2023-12-29/gender_statistics.py
```diff
@@ -0,0 +1,219 @@
...
+            tb[column].metadata.description_from_producer = description_string
+            tb[column].metadata.title = name
+            tb[column].metadata.processing = "minor"
+            tb[column].metadata.display = {}
+
+            #
+            # Update metadata units, short_units and number of decimal places to display depending on what keywords the variable name contains
+            #
+
+            def update_metadata(table, column, display_decimals, unit, short_unit):
```
Marigold, line 181: Nit - we're updating just indicator's metadata, not a table. It would be more intuitive to write
```
def update_metadata(meta, display_decimals, unit, short_unit):
  meta.display["numDecimalPlaces"] = display_decimals
```

and call it like

```
meta = table[column].metadata

update_metadata(meta, 1, unit="%", short_unit="%")
...
```

Also using nested functions is a bit of a bad practice (sometimes it's useful, but it could lead to nasty bugs). It'd be better to have it on the same level as `add_metadata`.

etl/steps/data/garden/wb/2023-12-29/gender_statistics.py
```diff
@@ -0,0 +1,219 @@
...
+    tb["total_maternity_leave"].metadata.title = "Total days of leave available for the mother"
+
+    tb[
+        "total_maternity_leave"
+    ].metadata.description_processing = "The total days of leave available for the mother indicator was created by OWID by combining maternity leave and parental leave mother quota indicators. For more details on each indicator see the section on **How producer describes this data**."
+    tb[
+        "total_paternity_leave"
+    ].metadata.description_processing = "The total days of leave available for the father indicator was created by OWID by combining paternity leave and parental leave father quota indicators. For more details on each indicator see **How producer describes this data**."
+
+    for leave_col in ["total_paternity_leave", "total_maternity_leave"]:
```
Marigold, line 108: Nit - is there a reason why is this done programatically and not saved in YAML file?

snapshots/wb/2023-12-29/gender_statistics.py
```diff
@@ -0,0 +1,134 @@
...
+    # Rename the 'code' column to 'wb_seriescode' to match the key in the input DataFrame
+    metadata_df.rename(columns={"code": "wb_seriescode"}, inplace=True)
+
+    wb_codes_to_extract = metadata_df["wb_seriescode"].unique().tolist()
+
+    # Fetch data from the World Bank API.
+    wb_df = get_data(wb_codes_to_extract)
+
+    # Merge the input DataFrame with the metadata DataFrame
+    enriched_df = pd.merge(wb_df, metadata_df, on="wb_seriescode", how="left")
```
Marigold, line 74: The final dataframe is quite a monster with its 8gb. Could you try adding optimal dtypes to it with
```
from owid.repack import repack_frame
enriched_df = repack_frame(enriched_df)
```
and then saving it in `.feather` format rather than as `.csv`?

Marigold (approved):
Couple of nits, though one with the size of the snapshot is pretty important. I think we can do much better than 8gb snapshot. Other than that looks good.

snapshots/wb/2023-12-29/gender_statistics.py
veronikasamborska1994: good point! thanks :)

etl/steps/data/garden/wb/2023-12-29/gender_statistics.py
veronikasamborska1994: good point! I think it's because I wanted to combine `description_from_producer` from the two sources and then just kept going with the other fields.

PR #2151, opened 2024-01-02
:hammer: stop using display.name to prevent warnings spam

Marigold:
We get huge number of warnings from `ihme_gbd` like this
```
2024-01-02 11:09:53 [warning  ] Column share_of_total_dalys_that_are_from_varicella_and_herpes_zoster__in_females_aged_7_27_days uses display.name but no presentation.title_public. Ensure the latter is also defined, otherwise display.name will be used as the indicator's title.
```

The `display.name` is the same as the indicator's title, so there's no point in keeping both.


spoonerf (approved):
Looks good!

Issue #2154, opened 2024-01-03
UN SDG metadata links stopped working


Marigold:
We run into the following error with step `grapher/un/2023-08-16/un_sdg`
```
File "/var/lib/buildkite-agent/builds/etl-build-2-20/our-world-in-data/etl-full-private-rebuild-every-2nd-night/etl/steps/data/grapher/un/2023-08-16/un_sdg.py", line 53, in run
--
  | df_tab = add_metadata_and_prepare_for_grapher(df_var, ds_garden, source_desc)
  | File "/var/lib/buildkite-agent/builds/etl-build-2-20/our-world-in-data/etl-full-private-rebuild-every-2nd-night/etl/steps/data/grapher/un/2023-08-16/un_sdg.py", line 122, in add_metadata_and_prepare_for_grapher
  | source_desc_out = create_metadata_desc(
  | File "/var/lib/buildkite-agent/builds/etl-build-2-20/our-world-in-data/etl-full-private-rebuild-every-2nd-night/etl/steps/data/grapher/un/2023-08-16/un_sdg.py", line 105, in create_metadata_desc
  | source_url = get_metadata_link(indicator)
  | File "/var/lib/buildkite-agent/builds/etl-build-2-20/our-world-in-data/etl-full-private-rebuild-every-2nd-night/etl/steps/data/grapher/un/2023-08-16/un_sdg.py", line 251, in get_metadata_link
  | assert ctype_a == "application/pdf", url_a + "does not link to a pdf"
  | AssertionError: https://unstats.un.org/sdgs/metadata/files/Metadata-01-0a-02a.pdfdoes not link to a pdf
  |  
  | COMMAND: prlimit --as=34359738368 poetry run run_python_step data://grapher/un/2023-08-16/un_sdg /var/lib/buildkite-agent/builds/etl-build-2-20/our-world-in-data/etl-full-private-rebuild-every-2nd-night/data/grapher/un/2023-08-16/un_sdg
```

It looks like some (or maybe all?) links like https://unstats.un.org/sdgs/metadata/files/Metadata-10-01-01b.pdf have stopped working.

As a hotfix, we should consider disabling that URL until we find a workaround and build SDG without it.

Ideally, we shouldn't be making external HTTP requests in the grapher step, as we should be able to rebuild everything from snapshots (this might have been overkill for this case?). We could also fix this by getting those URLs from an old dataset and **adding them as a snapshot**. @spoonerf I'm happy to do that if you think that's the way to go.

[...]

PR #2156, opened 2024-01-03
🐛 Hotfix for UN SDG Metadata issue

spoonerf:
Hotfix for https://github.com/owid/etl/issues/2154. 

I will work towards moving the metadata to snapshot and grabbing them from there, rather than doing http requests in the grapher step.


Marigold (approved):
Nice, thanks for the quick fix!

PR #2172, opened 2024-01-09
:bar_chart: Global Health Observatory dataset

Marigold:
## Overview

Get data from Global Health Observatory to ETL. It matches the original from [importers repository](https://github.com/owid/importers/tree/master/who_gho) pretty closely (it scrapes metadata from HTML, creates the same OMMs), but it uses a different endpoint for fetching data that doesn't use entity codes (honestly, this might have been a mistake, the endpoint is not very reliable).

I tried to match indicator-based charts from importers as closely as possible (but using origins and new metadata fields).

### Snapshot step

Snapshot step downloads all data and metadata and bundles it into zip file. Every indicator has its **own table**. It supports caching and retries.

### Meadow step

Does some basic cleanup and moves regions to `country` column.

### Garden step

Fills metadata, especially `description_from_producer` (contains what used to be `description` in importers) and `description_short` (=their "Definition"). 

## Notes

* The list of indicators in GHO also contains non-health indicators like "Percentage of individuals using the Internet", but they don't have a page with metadata for these indicators. I'm including them to be consistent with GHO from importers
* I exclude archived indicators (i.e. those with `archived` in their name)
* I exclude **low and high estimates** - I initially worked with them but decided not to overcomplicate things. If we decide to add them, we just uncomment a few lines and it should work
* Regions can come from a variety of sources (UN, WHO, ...) - add source as a suffix to region name (e.g. `Africa (WHO)`). We don't do any aggregation; hence no regions without suffix
* OMMs code was copied from importers
* I updated [GHO suicides dataset](https://github.com/owid/etl/pull/2358) to use this dataset rather than using a custom snapshot script.

## How to review

Rather than reviewing the code, it'd be more useful to check the output:
1. Take a look at [chart revisions on the staging server](http://staging-site-gho/admin/suggested-chart-revisions/review) (I already went through all of them once and it seems that we should be able to revise all of them as they are or with minor updates)
2. Compare some indicators from the old and then new dataset (see Examples section below)
3. Ad-hoc check of [indicator names](http://staging-site-gho/admin/datasets/6400) in a dataset (are dimensions ok?)
4. Check origin metadata - I've created it to be consistent with importers, but perhaps we should update it?

## Issues

* There's one chart with [pneumonia symptoms receiving healthcare](http://staging-site-gho/admin/charts/3875/edit) where updated data doesn't have "Total" value (only split by dimensions like sex). We could either ignore that chart or delete it (it has no pageviews) or compute it ourself (e.g. from averaging male & female values).

## TODO
- [x] Add OMMs from [importers](https://github.com/owid/importers/blob/master/who_gho/core.py#L845)
- [x] Remove changes to Wizard before merging

### Measles
* [metadata](https://www.who.int/data/gho/indicator-metadata-registry/imr-details/2)
* [old variable](http://staging-site-gho/admin/variables/470724)
* [new variable](http://staging-site-gho/admin/variables/828106)

### Population without reliance on clean fuels

This is OMM. Note that the old indicator shows negative values, but the new one caps them to zero.

* [old variable](http://staging-site-gho/admin/variables/471855)
* [new variable](http://staging-site-gho/admin/variables/830781)


snapshots/who/2024-01-03/gho.zip.dvc
```diff
@@ -0,0 +1,32 @@
...
+    description: |-
+      The GHO data repository is WHO's gateway to health-related statistics for its 194 Member States. It provides access to over 1000 indicators on priority health topics including mortality and burden of diseases, the Millennium Development Goals (child nutrition, child health, maternal and reproductive health, immunization, HIV/AIDS, tuberculosis, malaria, neglected diseases, water and sanitation), non communicable diseases and risk factors, epidemic-prone diseases, health systems, environmental health, violence and injuries, equity among others.
+    date_published: 2024-01-03
+
+    # Citation
+    producer: World Health Organization
+    # NOTE: this is what we had in the previous version, perhaps we should change it?
+    attribution: WHO, Global Health Observatory
+    citation_full: |-
+      World Health Organization. 2023. Global Health Observatory data repository. http://www.who.int/gho/en/. Accessed on 2024-01-03
```
spoonerf, line 16: I think this is okay, I certainly can't find anything better online. Maybe we should bump the year to 2024?

snapshots/who/2024-01-03/gho.zip.dvc
```diff
@@ -0,0 +1,32 @@
+# Learn more at:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/origin/
+meta:
+  origin:
+    # Data product / Snapshot
+    title: WHO Global Health Observatory
```
spoonerf, line 6: Maybe just 'Global Health Observatory'

etl/steps/data/garden/who/2024-01-03/gho.py
```diff
@@ -0,0 +1,350 @@
...
+# only keep those sections from GHO metadata
+HEADINGS_TO_USE = [
+    "Rationale",
+    "Definition",
+    "Method of measurement",
+    "Method of estimation",
+]
+
+# remove values that are considered as NaN
+NAN_VALUES = [
```
spoonerf, line 39: I think there are also some "..."

spoonerf (approved):
This looks awesome, thanks so much @Marigold!

I've followed your suggestions on the 'How to review' section and I think this is good to merge, I just have some small comments below. 

1. The chart revisions all look good, or just minor changes as you suggest. Some of the `description_short` are quite long, but I think we can just deal with this on a case-by-case basis. I don't think there is a better scalable solution than the one you have.
2. The new indicators and metadata look great and make sense.
3. The dimensions look good, some of the variable names are not totally informative, but again I think that is the source being a little unhelpful and something we can deal with case-by-case, if needed. Related to the issue below, I noticed that the 'total' dimension is missing from some indicators e.g. the ones starting with 'children aged < 5 years', is it the case that is missing from the data?
4. I've made some small suggestions to change the origin metadata


For the issue with [this chart](http://staging-site-gho/admin/charts/3875/edit) it would be good to have it for both sexes if poss. Mostly because we should update[this chart](https://owid.cloud/admin/charts/3605/edit) to be using it. I'm happy to do that in a separate PR at a later date to prevent holding up this monster from being merged.

PR #2174, opened 2024-01-10
Add nuclear warhead inventories data

pabloarosado:
Add data from [the Federation of American Scientists](https://fas.org/initiative/status-world-nuclear-forces/), on stockpiles of nuclear weapons over time, and on 2023 inventories.
- [Old dataset](https://owid.cloud/admin/datasets/6203) -> [New dataset](http://staging-site-add-nuclear-warhead-inventories-data/admin/datasets/6347) (2 charts).
- [Old dataset](https://owid.cloud/admin/datasets/6204) -> [New dataset](http://staging-site-add-nuclear-warhead-inventories-data/admin/datasets/6346) (1 chart). Here, the old dataset used to have data for 2022 and 2023, whereas the new one only has data for 2023. This is because the 2022 data has been removed from their page. We will consider fetching it from somewhere else, if there is time (in a separate PR).

### DIFF
--- a/dag/war.yml
+++ b/dag/war.yml
@@ -136,3 +136,11 @@ steps:
     - data://meadow/war/2024-01-09/nuclear_weapons_inventories
   data://grapher/war/2024-01-09/nuclear_weapons_inventories:
     - data://garden/war/2024-01-09/nuclear_weapons_inventories
+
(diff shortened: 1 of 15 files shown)


paarriagadap (approved):
All good!

PR #2178, opened 2024-01-11
:tada: streamlit app for etl-staging-sync

Marigold:
UI for command `etl-staging-sync` for migrating charts from staging servers to production. 

It will be run on `owid@etl-prod-2` together with fast-track (on a different port), and there'll be a link to it from admin.

The current version needs **active staging server** to sync charts from there, and the branch **must be merged in ETL**. I've increased the time for killing staging servers from 1 day to **3 days after merging**.

Last time, we talked about snapshotting the staging server somewhere so that we can kill the server right away and reconstruct updated charts from the snapshot. I gave it a try, and it turned out to be unnecessarily complicated. Moreover, having an active staging server when syncing charts could be useful (e.g. if you realise you still have to make some chart edits there or if you find that the chart has been updated on live and you want to make the same changes on your staging before syncing the chart).

The streamlit app is very basic, though one think that's pretty cool is `_run_command` that runs a CLI command and shows its output (with colours!) as HTML in the app.

I haven't used streamlit for a while, so any tips are welcome!

## How to test it

1. `poetry install`
2. `etl-staging-sync-app`
3. Use source `gho` and target `etl-staging-sync-app` (I've updated two charts)
4. Follow instructions (feel free to do anything to both staging servers)

### DIFF
--- a/apps/staging_sync/app.py
+++ b/apps/staging_sync/app.py
@@ -1,0 +1,132 @@
+import os
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+import streamlit as st
+import streamlit.components.v1 as components
+from git.exc import GitCommandError
+from git.repo import Repo
+from rich.ansi import AnsiDecoder
+from rich.console import Console
+from rich.terminal_theme import MONOKAI
+
+from etl.paths import BASE_DIR
+
+CURRENT_DIR = Path(__file__).resolve().parent
+
+
+def main():
+    st.title("etl-staging-sync")
+    st.markdown(
+        """
+    Sync charts and revisions from staging server to production.
+    """
+    )
+
+    # SIDEBAR
+    with st.sidebar:
+        with open(CURRENT_DIR / "instructions.md", "r") as f:
+            st.markdown(f.read())
+
+    st.markdown("### Config")
+    source = st.text_input(
+        "Source",
+        placeholder="my-branch",
+        help="Name of the branch to sync from (with existing `staging-site-mybranch` server).",
+    )
+    target = st.text_input("Target", value="live", help="Using `live` uses DB from local `.env` file as target.")
+    publish = st.checkbox(
+        "Automatically publish new charts", value=False, help="Otherwise you'd have to publish new charts manually."
+    )
+    approve_revisions = st.checkbox(
+        "Automatically approve chart revisions for edited charts",
+        value=False,
+        help=" This still creates a chart revision if the target chart has been modified.",
+    )
+    dry_run = st.checkbox("Dry run", value=True)
+
+    # Live uses `.env` file which points to the live database in production
+    if target == "live":
+        target = ".env"
+
+    # Button to show text
+    if st.button("Sync charts", help="This can take a while."):
+        if not _is_valid_config(source, target):
+            return
+
+        # Open the local repository
+        repo = Repo(BASE_DIR)
+
+        # Fetch the specific branch from the remote
+        try:
+            repo.git.fetch("origin", source)
+        except GitCommandError:
+            st.error(f"Branch {source} not found in owid/etl repository.")
+            sys.exit(1)
+
+        cmd = ["poetry", "run", "etl-staging-sync", source, "master"]
+        if dry_run:
+            cmd.append("--dry-run")
+        if publish:
+            cmd.append("--publish")
+        if approve_revisions:
+            cmd.append("--approve-revisions")
+
+        _run_command(cmd)
+
+
+def _is_valid_config(source: str, target: str) -> bool:
+    if source.strip() == "":
+        st.warning("Please enter a valid source.")
+        return False
+    if target.strip() == "":
+        st.warning("Please enter a valid target.")
+        return False
+    return True
+
+
+def _run_command(cmd: list[str]):
+    console = Console(record=True)
+
+    # Run bash command
+    my_env = os.environ.copy()
+    my_env["FORCE_COLOR"] = "1"
+    result = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=my_env)
+    stdout, stderr = result.communicate()
+
+    stdout = stdout.decode()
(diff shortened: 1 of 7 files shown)


apps/staging_sync/instructions.md
```diff
@@ -0,0 +1,28 @@
+# Instructions
+
+1. Merge your ETL branch to master and wait for the ETL to rebuild your dataset.
+2. Fill `Config` section and tick `Dry run` first to see what charts will be updated and hit `Sync charts`.
+3. Check the output, look for warnings that might indicate unexpected changes to charts in production.
```
lucasrodes, line 5: Maybe I got it wrong, but shouldn't it say

> 3. Check the output, look for warnings that might indicate unexpected changes to charts in `Target` server?

apps/staging_sync/instructions.md
```diff
@@ -0,0 +1,28 @@
...
+4. Untick `Dry run` and hit `Sync charts` again to apply changes to production.
+5. New charts were created as drafts, make sure to **publish them in production**.
```
lucasrodes, line 7: Again here, instead of "production", shouldn't it refer to whatever `Target` server the user specified?

apps/staging_sync/app.py
```diff
@@ -0,0 +1,132 @@
...
+    # SIDEBAR
+    with st.sidebar:
+        with open(CURRENT_DIR / "instructions.md", "r") as f:
+            st.markdown(f.read())
+
+    st.markdown("### Config")
+    source = st.text_input(
+        "Source",
+        placeholder="my-branch",
+        help="Name of the branch to sync from (with existing `staging-site-mybranch` server).",
```
lucasrodes, line 37: I am not sure I understand this.

Does this mean that the branch in ETL should have its corresponding staging server created, otherwise this won't work?

As I was reading, I was wondering when this is not the case. As long as I create a PR and wait, the corresponding server should exist, shouldn't it?

Maybe a clarification here could be helpful. Or a link to our docs if we have a section where we explain how we create staging servers from PRs.

apps/staging_sync/app.py
```diff
@@ -0,0 +1,132 @@
...
+from etl.paths import BASE_DIR
+
+CURRENT_DIR = Path(__file__).resolve().parent
+
+
+def main():
+    st.title("etl-staging-sync")
+    st.markdown(
+        """
+    Sync charts and revisions from staging server to production.
```
lucasrodes, line 24: I am not sure I understand how the app operates.

Isn't it designed to get charts from an arbitrary branch (source) and sync them toan arbitrary server (target), i.e. not necessarily production?

If so, I think that talking about "target" and "production" can be confusing. We should probably just use "target server" or something like that. Maybe explain somewhere that usually we will have target=live

apps/staging_sync/instructions.md
```diff
@@ -0,0 +1,28 @@
...
+# Notes
+
+Staging servers are **destroyed 3 days after merging to master**, so this app should be
+run before that, but after the dataset has been built by ETL in production.
+
+
+## Charts:
+- Only **published charts** from staging are synced.
+- New charts are synced as **drafts** in target (unless you set the checkbox).
+- Existing charts (with the same slug) are added as chart revisions in target (unless you set the checkbox).
```
lucasrodes, line 20: This is the correct general nomenclature, right? Using "target"

apps/staging_sync/instructions.md
```diff
@@ -0,0 +1,28 @@
...
+- You get a warning if the chart **has been modified on live** after staging server was created.
```
lucasrodes, line 21: Along my previous comments, maybe use "target" or "target server" instead of "live"?

apps/staging_sync/instructions.md
```diff
@@ -0,0 +1,28 @@
...
+- Deleted charts are **not synced**.
+
+## Chart revisions:
+- Approved chart revisions on staging are automatically applied in target, assuming the chart has not been modified.
+
+## Tags:
+- Tags are synced only for **new charts**, any edits to tags in existing charts are ignored.
```
lucasrodes, line 28: When you say "new charts", do you mean new in the source server?

lucasrodes (approved):
Thanks for doing this, @Marigold. This looks great and I think this will be super helpful.

Very nice that you implemented `_run_command`. I think I may move it at some point to sth like `apps.utils` so we can use in other apps.


I added some comments. Some general thoughts below:

### 1. Using terms "target", "live", "production"
I was a bit confused when reading the instructions and following the form with the terms "target", "live", and "production". 

In the example you mention in this PR description, we are using `etl-staging-sync-app` instead right?
I am unsure I got it right, so ignore me if I am way off. My understanding is that this app works for any source and target servers, and not necessarily just production.
From the instructions, it seems like you are talking from the case where target=production. I think that harmonising the word used here could make it more understandable. Maybe just use "target" or "target server", and mention that usually we will want to use target=live.

### 2. Motivation/purpose of the app
I think the app could benefit from a short description on what it does and why is it helpful.

### 3. Integrating it with `etl-wizard`
I think we could have this app integrated with `etl-wizard`, under "Other tools". Same as we did with other tools (Charts, MetaGPT, etc.). What do you think? I am happy to do it at some point.

apps/staging_sync/instructions.md
Marigold: That's a really good point. I changed everything to source -> target and made it clear that source is typically staging and target is typically live.

Marigold:
> 1. Using terms "target", "live", "production"

You're totally right. I changed it to source and target with clarification of what they typically refer to. I have this problem in other places too, it might make sense to clean it up at one point.

> 2. Integrating it with etl-wizard

That was surprisingly easy!

apps/staging_sync/app.py
Marigold: > As long as I create a PR and wait, the corresponding server should exist, shouldn't it?

That's right. I tried to rephrase it to make it more clear. If you find a better wording, feel free to update it, please.

PR #2190, opened 2024-01-15
:tada: ETL API

Marigold:
Add ETL API that lets us edit metadata of ETL indicators. There's a [PR in owid-grapher](https://github.com/owid/owid-grapher/pull/3106) that calls the API on metadata updates in Admin.

Please read `api/README.md` for a summary. Here are some additional notes:

* I used our old [ETL Data API](https://github.dev/owid/data-api) as a FastAPI template for this project
* It sends notifications about every request / response. This is to get a sense how will it be used.
* The API runs on `uvicorn` in **debug mode** (which is not really production ready, but I think it's ok given that it's behind Tailscale and we'll get 1 request / week)
* ETL API will be started automatically on the staging server, but it's unclear whether it should be doing commits to the branch or just making changes to YAML files (which would be overwritten)

## How to test

1. Go to http://staging-site-admin-edit-api/admin, find any indicator you like and change its metadata
2. Hit refresh to verify that updated values persist in admin
3. Verify that YAML file was committed to this branch
4. Verify that this PR builds cleanly then

(I'm not 100% sure whether commits should be turned on staging servers, but they are on `staging-site-admin-edit-api`)

## After merging

* Add this as a service to `etl-prod-2` next to Fast-track
* Add envs `ETL_API_COMMIT=1`, `ETL_API_URL=http://etl-prod-2:8000/v1/`, `SLACK_API_TOKEN` to secrets env
* Add `ETL_API_URL` to `owid-live`

## Summary by CodeRabbit

- **New Features**
	- Added an ETL API for editing metadata for indicators.
	- Integrated observability features (Slack and Bugsnag) for ETL API monitoring and error reporting.

- **Enhancements**
	- Implemented new API endpoints for updating indicator metadata.
	- Included CORS support and custom Slack middleware for the FastAPI application.

- **Documentation**
	- Provided detailed instructions and considerations for using the new ETL API.

- **Configuration**
	- Updated remote storage configuration with new profile settings.
	- Introduced global variables for API tokens and keys related to Slack and Bugsnag.

- **Refactor**
	- Enhanced metadata merging capabilities and handling in ETL processes.


api/README.md
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "dataApiUrl": "https://api-staging.owid.io/mojmir/v1/indicators/",
+  "triggerETL": true,
+  "dryRun": false
+}' | http PUT http://127.0.0.1:8000/v1/indicators/540251
+
+```
+
+`dataApiUrl` points to the Data API that is used by Admin. If ETL API uses different Data API, it'll
+raise an error (to prevent confusion when querying the wrong ETL API).
```
larsyencken, line 46: At first I didn't understand this, but if I've understood you right, this is to avoid any misunderstandings/misconfiguration on the Grapher side. Grapher expects to be making the change in e.g. prod, then we don't want to lie to grapher and only make the change in staging.

That seems sensible, though I'll give feedback on it when reviewing the other PR to understand how things are configured there.

api/v1/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+
+v1 = APIRouter()
+
+
+@v1.get("/v1/health")
+def health() -> dict:
+    return {"status": "ok"}
+
+
+@v1.put("/v1/indicators/{indicator_id}")
```
larsyencken, line 42: The ETL should probably be working on ETL paths by now, which is something we already know on the `owid-grapher` side. Let's do an update pattern based on those paths rather than based on MySQL ids.

Either the path can go into the URL or it can go into the PUT body, I don't have a strong preference.

I understand the downside being that you want to do pushes to R2, which are still id-based. Let's chat about it.

api/v1/__init__.py
```diff
@@ -0,0 +1,180 @@
...
+    # ourselves running into race conditions.
+    if not update_request.dryRun:
+        with open(override_yml_path, "w") as f:
+            f.write(yaml_str)
+
+    if not update_request.dryRun:
+        _update_indicator_in_r2(db_indicator, update_request.indicator)
+
+    if config.ETL_API_COMMIT:
+        background_tasks.add_task(_commit_and_push, override_yml_path, ":robot: Metadata update by Admin")
```
larsyencken, line 83: It's not clear to me that this should be a background task. The whole request should probably fail if you can't commit the result and push it, perhaps it should even return the repo to a known good state on error so that the next attempt to edit is likely to succeed.

larsyencken: The commit should also happen before a push to R2, since we want to avoid pushing to R2 and a commit+push failing. The R2 push can be in the background.

api/v1/schemas.py
```diff
@@ -0,0 +1,68 @@
...
+    def to_meta_dict(self) -> Dict[str, Any]:
+        d = {
+            "title_public": self.titlePublic,
+            "title_variant": self.titleVariant,
+            "attribution_short": self.attributionShort,
+        }
+        return utils.prune_none(d)
+
+
+class Indicator(BaseModel):
```
larsyencken, line 26: Since we can mean a bunch of things by this, it's good to clarify in a comment what this represents. E.g. the JSON metadata served by the data API, or the MySQL model used by `owid-grapher`.

etl/steps/data/grapher/biodiversity/2023-01-11/cherry_blossom.meta.override.yml
larsyencken, line 1: Delete?

larsyencken (changes requested):
## Feedback on the overall approach

If we want to do this without leaving more tech debt behind us, our goal should be to only run one single web service for the `etl`, something that acts both as admin and API.

That suggests that to do this nicely you should refactor fast track and put these things together. That might mean moving more fast track config from command-line args to something else, and new guidelines on how to launch it.

If you agree, then we would move everything here to an `/api/` prefix. You would need to use something like `webio_handler` to embed PyWebIO into the FastAPI stack https://pywebio.readthedocs.io/en/v1.2.2/platform.html#application-deploy

api/v1/__init__.py
Marigold: > The commit should also happen before a push to R2

Yep, good point!

> perhaps it should even return the repo to a known good state on error so that the next attempt to edit is likely to succeed.

I'm not sure yet what a bad state would look like (which is why I'm logging all requests to slack, to find it out), so I'd rather implement it when this becomes a problem.

api/README.md
Marigold: Yeah, exactly. I've spent tons of time debugging something just to find out my env file is pointing to a different environment. The fact that we're doing fallbacks to producion in Data API makes it even more confusing.

api/v1/__init__.py
Marigold: I had to add it to the body because catalog path contains `/` and `#` (e.g. `grapher/agriculture/2023-04-21/uk_long_term_yields/uk_long_term_yields#oats_yield`). It's not a big deal now.

Marigold:
> If you agree, then we would move everything here to an /api/ prefix. You would need to use something like webio_handler to embed PyWebIO into the FastAPI stack https://pywebio.readthedocs.io/en/v1.2.2/platform.html#application-deploy

I totally agree, but it'd be better to do it with nginx and manage all apps separately (ETL API, Fast-track and Streamlit staging-sync). Integrating it into FastAPI was doable for Pywebio, but non-trivial for Streamlit and I needed a lot of custom code in ETL API. Better to manage them separately as pm2 processes.

@larsyencken I've implemented all the other things. Would it be ok to merge this now and do ops part in a new PR? (In ops repository)

Makefile
larsyencken, line 1: Add `make api` to the `help:` stanza

api/v1/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+        _commit_and_push(override_yml_path, ":robot: Metadata update by Admin")
+
+    if not update_request.dryRun:
+        _update_indicator_in_r2(db_indicator, update_request.indicator)
+
+    # trigger ETL in the background (this is usually fast, but for some datasets this could take a while)
+    if update_request.triggerETL:
+        background_tasks.add_task(_trigger_etl, indicator_short_name, db_indicator, update_request.dryRun)
+
+    return {"yaml": yaml_str}
```
larsyencken, line 90: For later / nit: this function would be more readable if is was 5 lines that read like pseudocode, calling out to smaller functions that do the work.

etl/config.py
```diff
@@ -123,9 +123,15 @@ def variable_metadata_url(variable_id):
 # any garden or grapher dataset after this date will have strict mode enabled
 STRICT_AFTER = "2023-06-25"
 
+SLACK_API_TOKEN = env.get("SLACK_API_TOKEN")
+
+# if True, commit and push updates to YAML files coming from admin
+ETL_API_COMMIT = env.get("ETL_API_COMMIT") in ("True", "true", "1")
```
larsyencken, line 129: Note: an alternative to checking for all these truthy values is just to do the behaviour if the key is present at all.

larsyencken (approved):
Thanks for addressing the feedback, you can merge it! :shipit:

api/v1/__init__.py
Marigold: I tried to break it down to functions. It's not exactly 5 lines, but better than before.

PR #2193, opened 2024-01-16
:honeybee: complete the merkle tree with checksums

larsyencken:
Calculate checksums all the way up the folder structure and all the way to the catalog level.

## What

Each time you save a dataset, you store its checksum in a `.md5` file at the root of its folder.

You likewise walk upwards `dataset < version < namespace < channel < catalog` and calculate the checksum of all the higher level folders, based on their constituents. Since now you are only combining checksums, these steps are cheap.

The net result is that you have a folder tree of checksums, including one top-level checksum that represents the state of the whole catalog.

## Why

- It allows quick comparison between two catalogs by comparing checkums along their trees
  - This can be done in fewer network calls, meaning faster diff between local and object storage
- It serialises the dataset checksums to disk
  - This means other codebases can use them without having to know the rules for computing them
- It allows you to build something secondary off of the ETL output, and know when you need to rebuild it


larsyencken:
This proof of concept is solid, but hard to merge because of the side-effects it adds to dataset saves.

Two thoughts on the next round of doing this:

- Instead of a `.md5` file, use a `.manifest` file that lists the checksums of every child file or folder, since this would make diffing against or mirroring an HTTP target easier
- Instead of doing this in the abstract, use it to create a faster data diff, then the value will be clearer

PR #2196, opened 2024-01-18
Improve calculation of the share of electricity in primary energy

pabloarosado:
We currently calculate the share of electricity in primary energy consumption (shown in [this chart](https://ourworldindata.org/grapher/electricity-as-a-share-of-primary-energy)) as simply dividing the total electricity generation by primary energy consumption.
However, primary energy consumption (from the Statistical Review of World Energy) is defined in "input-equivalents" (what we usually call [the substitution method](https://ourworldindata.org/energy-substitution-method)). This means that non-fossil electricity sources (nuclear, hydro, solar, wind, bioenergy and other renewables) are inflated to mimic the inefficiencies of fossil fuels. To achieve that, the Statistical Review divides the electricity generation of non-fossil sources by an efficiency factor (which is roughly between 0.3 and 0.5).

Therefore, currently, our share of electricity in primary energy is underestimated. We claim in the subtitle that we use the substitution method, but we only do this in the denominator, not in the numerator.

This PR does the calculation of the share of electricity in primary energy properly:
100 * (((electricity generated by non-fossil sources) / efficiency factor) + (electricity generated by fossil fuels) ) / (primary energy consumption)

An alternative would be to estimate the direct primary energy consumption (not using the substitution method), and then simply divide total generation by direct primary energy consumption. This is also possible, but since we use the substitution method in most charts showing primary energy, it makes sense to use it here too.

To clarify, [the affected chart](https://ourworldindata.org/grapher/electricity-as-a-share-of-primary-energy) (also included in [the energy explorer](https://ourworldindata.org/explorers/energy?facet=none&country=USA~GBR~CHN~OWID_WRL~IND~BRA~ZAF&Total+or+Breakdown=Total&Energy+or+Electricity=Electricity+only&Metric=Share+of+total)) will barely be affected for countries that rely mostly on fossil fuels. But for countries that have a big share of renewables and nuclear, the change can be significant. One of the most significant examples is Norway. Here's the comparison of the old (currently in production) and new (this PR) indicator:
newplot
And the missing points in the new curve happen because we don't have electricity generation disaggregated by source for all countries and years (for which we do have total electricity generation and primary energy consumption).

### DIFF
--- a/etl/steps/data/garden/energy/2023-12-12/electricity_mix.py
+++ b/etl/steps/data/garden/energy/2023-12-12/electricity_mix.py
@@ -52,6 +52,8 @@ def process_statistical_review_data(tb_review: Table) -> Table:
         "oil_electricity_generation_twh": "oil_generation__twh",
         "coal_electricity_generation_twh": "coal_generation__twh",
         "gas_electricity_generation_twh": "gas_generation__twh",
+        # Load efficiency factor to be able to convert from electricity generation into input-equivalent primary energy.
+        "efficiency_factor": "efficiency_factor",
     }
     tb_review = tb_review[list(columns)].rename(columns=columns, errors="raise")
     # New columns to be created by summing other columns.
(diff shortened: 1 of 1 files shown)


HannahRitchie:
Thanks a lot for this @pabloarosado.

I agree: the share here should be calculated as electricity's share of _direct primary energy_ (not adjusted for input equivalents).

The alternative way of correcting for this – that you've used – makes sense.

I'm wondering if the mention of the substitution method in the subtitle is confusing. Again, people might wonder whether both or just one variable has been adjusted.

Maybe just `Measured as a percentage of total primary energy from all sources.` works better?

Either way, I'm happy for this to go through. As Fiona mentions, we should be ready to answer any queries that we might get in about why some values have changed so much.

PR #2202, opened 2024-01-22
:tada: add non_redistributable field to dataset metadata

Marigold:
Set field `nonRedistributable` from ETL instead of manually setting it from admin (which could be non-transparent, e.g. when bumping version of a dataset).

## TODO
- [ ] make the field non-editable in Admin


lucasrodes (approved):
Clean and simple, thanks @Marigold ! LGTM

PR #2204, opened 2024-01-22
🐛 wizard charts: fix and improve experience with same-dataset-updates

lucasrodes:
There were minor issues when updating indicators from the same dataset (i.e. having `old dataset = new dataset`).

Pablo surfaced this in https://github.com/owid/etl/issues/1563#issuecomment-1903700162.

**Some fixes/improvements include**:

- **Mapping identically named variables**
  - By default, indicators with the same name were mapped automatically (with no edits allowed by the user) if there was a 100% name match. This option can be tweaked by the parameter "Mapped identically named variables". This flexible parameter is useful when old and new datasets differ. However, in this case, we prefer the default behaviour to be that no automatic mapping is done to allow for "variable migrations" (migrating a variable to a new one, which does not necessarily have the same name).
image
  - Added a help text to explain this option better to the user.
- Bug: a minor bug would give a variable in the new dataset a higher score than another with the same name as the old variable. For a certain variable, even if there is a variable with the same name in the new dataset, it wouldn't be suggested as the top mapping. This is now fixed.

### DIFF
--- a/apps/wizard/charts/search_config.py
+++ b/apps/wizard/charts/search_config.py
@@ -35,7 +35,11 @@ def build_dataset_form(df: pd.DataFrame, similarity_names: Dict[str, Any]) -> "S
     col0, _, _ = st.columns(3)
     with col0:
         with st.expander("Parameters"):
-            map_identical = st.checkbox("Map identically named variables", value=True)
+            map_identical = st.checkbox(
+                "Map identically named variables",
+                value=True,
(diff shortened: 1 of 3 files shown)


paarriagadap (approved):
Looks good! Now I was able to map them all.

One thing I noticed is that filtering doesn't work as expected when I use the same words in reverse order. For example, if i search for `number smoothed` there are results (because the words are in the same order they are saved) but when I search for `smoothed number` there are no results.

I can add it to the issue collection if it's not fixable in this one.

image
image

lucasrodes:
hi @paarriagadap, thanks for reporting.

I think this last issue that you mention has been there for a while, right? As in, it is not new from this PR.

I believe this is how search works with Streamlit, and it might be hard to change. We can look into this in the future, and see if it changes.

PR #2227, opened 2024-01-29
🐛 Remove double counting of one-year olds in age-standardized variables of global health estimates

spoonerf:
One-year olds were incorrectly being double-counted in the add_population() function, used to calculate age-standardised variables. 

This will fix that but I expect it will impact a lot of variables (although not many of those are used), so should probably be merged in the evening or overnight. 

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted the age range for the "YEARS0-1" group to ensure accurate data processing.

### DIFF
--- a/etl/steps/data/garden/who/2022-09-30/ghe.py
+++ b/etl/steps/data/garden/who/2022-09-30/ghe.py
@@ -19,7 +19,7 @@ log = get_logger()
 paths = PathFinder(__file__)
 AGE_GROUPS_RANGES = {
     "ALLAges": [0, None],
-    "YEARS0-1": [0, 1],
+    "YEARS0-1": [0, 0],
     "YEARS1-4": [1, 4],
     "YEARS5-9": [5, 9],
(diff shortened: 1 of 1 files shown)


pabloarosado (approved):
Looks good. If I understand correctly, these changes will only affect `data://grapher/who/2022-09-30/ghe`, right? So this PR can be merged at any time. That dataset, however, has 100 charts. So it would be good to check that the changes are not too significant (either offline, or by creating a new version of the dataset and running the chart revision tool). But it's up to you what you think is necessary.

spoonerf:
Thanks, @pabloarosado!  I checked a few charts, and it seems like there is just a moderate increase in the age-standardized rates, particularly for causes of death that impact children more, but this is expected. 

It's quite a large dataset, so I understood it would block up the etl a bit while the update runs. The grapher step takes about 15 mins locally and my experience is that this is longer on live. I'll merge in an hour or so, to be safe.

PR #2239, opened 2024-01-31
Use literal style when dumping yaml files

pabloarosado:
By default, when writing snapshot yaml files, we use the folded style (">") for multiline texts. I find this annoying, because metadata often needs to be manually edited, and the folded style introduces arbitrary line breaks. It's much easier to use the literal style ("|") and let the text editor wrap the code conveniently.

This PR imposes that, when writing text to a yaml file that either has line breaks, or contains long lines of text, the literal style will be used.

@Marigold this will affect all places where `yaml_dump` is used. Let me know if you think this can be dangerous or lead to unwanted changes. Also, I'll ask the data people if anyone is against the literal style. Thanks!

## Summary by CodeRabbit

- **Refactor**
	- Improved the logic for determining YAML string style in the ETL process, enhancing readability and simplification.

### DIFF
--- a/etl/files.py
+++ b/etl/files.py
@@ -140,12 +140,11 @@ class _MyDumper(Dumper):
 
 def _str_presenter(dumper: Any, data: Any) -> Any:
     lines = data.splitlines()
-    if len(lines) > 1:  # check for multiline string
-        max_line_length = max([len(line) for line in lines])
-        if max_line_length > 120:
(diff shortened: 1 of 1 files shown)


Marigold (approved):
Nice, it'll be nice to have it unified, I always forget what each symbol does. It's possible that there'll be more (legacy) places that work with YAML files, we can fix them on the go.

PR #2243, opened 2024-02-01
Avoid raising error when no steps are found; instead print closest matches

pabloarosado:
It happens often that, when trying to run an etl step, a ValueError is raised, with a long traceback message, simply because the name of the step was misspelled. However, I think this should not raise an error: The step was not found, so there's nothing to run, and therefore nothing failed.

Instead, with this PR, if no steps are found, a message is printed, showing the closest matches.

NOTES:
* I used a very unsophisticated function to find matches. We can use a better one if needed.
* I don't know if there were other good reasons to raise a ValueError.

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -3,6 +3,7 @@
 #  etl.py
 #
 
+import difflib
 import itertools
 import re
 import resource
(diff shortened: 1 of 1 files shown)


Marigold (approved):
Very cool! I don't think we rely on that `ValueError` anywhere.

PR #2257, opened 2024-02-05
💄 AI chess dataset

veronikasamborska1994:


Marigold (approved):
Looks good. If we plan to update the dataset frequently, it'd make sense to commit the data here (to make the process straightforward and not keeping uncommitted files on local disk).

PR #2266, opened 2024-02-07
:honeybee: add unit test that all DAG steps can run

larsyencken:
In light of some recent cases where the DAG included steps whose code had been archived, this change adds a unit test that checks for code for every data step in the DAG.

Cherry-picking it over the previous issues confirms it would have caught those errors in CI.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced data processing workflow with improved file type checking and error handling.
- **Tests**
	- Added a test to ensure the integrity and functionality of the workflow structure.


pabloarosado (approved):
Looks good.

pabloarosado:
I've just realised that the new `can_run` is very similar to another already existing method, called `can_execute` for which there is a unit test in `test_etl`. I guess it would be better to remove the former and edit the latter if they serve the same purpose.

larsyencken:
> I've just realised that the new `can_run` is very similar to another already existing method, called `can_execute` for which there is a unit test in `test_etl`. I guess it would be better to remove the former and edit the latter if they serve the same purpose.

Good catch! This is basically identical to an existing CI check.

The reason that the previous check didn't pick up your error is that it was too flexible and auto-detected the new archive location.

I'll modify the existing check and make sure they don't double up.

Issue #1821, opened 2023-10-19
Improvements on `version_tracker`


pabloarosado:
Currently, `version_tracker` has a few issues:
* [x] It checks if any active step depends on an archive step. If so, it raises an error. But if you simply delete dependencies of active steps (instead of moving them to the archive) it doesn't complain. A better logic would be to raise an error whenever an active step doesn't find its dependencies.
* [x] Additionally, the error raised is not easy to understand. If you have moved multiple steps to the archive, it can be cumbersome to find out which steps need to be recovered.
* [x] It checks if an active step can safely be archived. If so, it raises a warning. Once you have archived that step, it raises another warning if the dependencies of that step can be archived, and so on. A better logic would be to raise a warning that includes all steps that can safely be archived (so, the list of archivable steps should be calculated recursively).
* [x] It should include info from grapher DB (maybe when executed with a --grapher flag).

A data step is “archivable” if:
1. It has no forward grapher DB dataset with charts.
2. It has no forward grapher DB dataset with indicators that are used in indicator-based explorers. WARNING: Is there a way to track this from DB?
3. It has no active ETL explorer step. WARNING: We still need to ensure that tsv files cite the latest version in ETL (or ensure they all ETL explorer steps have version “latest”).
4. It is not used in external repos (e.g. energy, co2, poverty, covid).
5. (Maybe) Its version is older than a week. This is imposed so that we don’t consider archivable new steps for which we haven’t had time to create charts.

pabloarosado:
Don't close.

PR #2268, opened 2024-02-08
Fix version tracker issues and improve its capabilities

pabloarosado:
Fix some known issues of `etl.helpers.VersionTracker` and improve its capabilities.

More context in https://github.com/owid/etl/issues/1821


pabloarosado:
Hey @Marigold, walden unit tests [are failing](https://buildkite.com/our-world-in-data/walden-unit-tests/builds/6205#018d9c8d-c484-45c2-add8-4cd6c651c0fc) because of bad formatting of imports, but when I run `make format` locally nothing changes (and I don't know what the issue is). We could simply avoid checking the formatting of the ingest files. Do you have any suggestion? Thanks.

Marigold:
@pabloarosado Walden uses legacy formatting with isort / black and `make format` from `etl` directory doesn't affect it. You have to run `cd lib/walden && make format` to fix it.

etl/helpers.py
```diff
@@ -647,7 +647,7 @@ def list_all_steps_in_dag(dag: Dict[str, Any]) -> List[str]:
     return all_steps
 
 
-def get_direct_dependencies_for_step_in_dag(dag: Dict[str, Any], step: str) -> Set[str]:
+def get_direct_step_dependencies(dag: Dict[str, Any], step: str) -> List[str]:
```
Marigold, line 650: In the future we might wrap all this in its own DAG class (inherit from `Dict[str, Any]`) or even use some existing libraries. (Doing it now would be too much work, but we should keep it in mind)

Marigold (approved):
Very nice, thank you! I think we're reaching a point where deprecating walden could be worth the effort.

pabloarosado:
Hi @Marigold thanks! I was actually running `make format` within `lib/walden`. But I was doing it from the etl poetry shell. Now I deactivated it and ran `make format` again from `lib/walden`and... Now I'm having issues installing pyzmq and numpy together :-/ . Does `poetry install` work for you in walden? I'm trying a few things but I'm so far not succeeding.

PR #2271, opened 2024-02-09
:honeybee: update mysqlclient version

larsyencken:
This is in order to fix some MacOS issues when using the most recent version of `mysql-client` in Homebrew.


larsyencken:
@Marigold Heads up, since you had previously pinned this change back. I think the reason it was failing for you was the `pkg-config` issue.

As well as add `pkg-config` to multiple recipes, I also added it manually to every running staging server so that there'd be no surprises for anyone.

Marigold:
> I also added it manually to every running staging server so that there'd be no surprises for anyone.

Thanks! By the way how do you run a command on all staging servers? I have a Python script for that, but I haven't committed it yet. It'd be nice to have it somewhere in ops README (I can add it myself)

PR #2284, opened 2024-02-12
:bug: dynamic is_public from selected sheet

Marigold:
The old fast-track is setting the `Make dataset private` field dynamically if the user selects the existing sheet. This is especially important for updating existing private datasets. Without this, we'd have to tick "private" every time when doing an update.

Unfortunately, streamlit doesn't support `on_change` in forms, so I had to put it outside the form [using this hack](https://discuss.streamlit.io/t/updating-radio-inside-a-form/49076/4). It's not very nice.

Alternative solutions:
- select box with `Auto`, `Public` and `Private` (`Auto` would use existing sheet's visibility with public as default)
- disable setting private / public for existing sheets (we might face the same issue with dynamic forms here)

## Summary by CodeRabbit

- **Refactor**
	- Improved form handling and user input processes in the application, including new placeholders for selections and settings adjustments.

### DIFF
--- a/apps/wizard/pages/fasttrack/app.py
+++ b/apps/wizard/pages/fasttrack/app.py
@@ -110,6 +110,8 @@ for option, option_params in IMPORT_OPTIONS.items():
 # CREATE AND SHOW THE FORM
 ##########################################################
 with st.form("fasttrack-form"):
+    existing_google_sheet = None
+
     # Import field
     if import_method == IMPORT_GSHEET:
         st.text_input(
@@ -119,13 +121,7 @@ with st.form("fasttrack-form"):
         )
     elif import_method == UPDATE_GSHEET:
         options = load_existing_sheets_from_snapshots()
-        st.selectbox(
-            label="Existing Google Sheets",
-            options=options,
-            format_func=lambda x: x["label"],
-            help="Selected sheet will be used if you don't specify Google Sheets URL.",
-            key="dataset_uri",
-        )
+        placeholder_for_existing_google_sheet = st.empty()
     else:
         st.file_uploader(
             label="Upload Local CSV",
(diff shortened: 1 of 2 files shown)


lucasrodes:
Thanks for doing this, @Marigold.

This is particularly an annoying part of Streamlit I've encountered in other places, too. This is a well-established workaround at the moment.

I've tweaked some of the code to avoid running the first step without clicking the submit button. This happens if a user submits a dataset and then changes the state of an "out of the form" element (e.g. checks the "make dataset private" checkbox). This occurs because Streamlit re-executes the script every time there is a change in an element (elements within a form don't count), and then since the `st.session_state.to_be_submitted` is set to True, it just proceeds with submission.

To avoid this, I've added a `reset_states` call in the `on_change` parameter of the "out of the form elements" to ensure that `st.session_state.to_be_submitted` is set to False whenever the user changes the state of these elements.

Also, made sure to pass the formatting tests.

apps/wizard/pages/fasttrack/app.py
```diff
@@ -158,6 +164,36 @@
...
+        raise ValueError("placeholder_for_existing_google_sheet is None. This was not expected.")
+    else:
+        with placeholder_for_existing_google_sheet:
+            existing_google_sheet = st.selectbox(
+                label="Existing Google Sheets",
+                options=options,  # type: ignore
+                format_func=lambda x: x["label"],
+                help="Selected sheet will be used if you don't specify Google Sheets URL.",
+                key="dataset_uri",
+                on_change=reset_states,
```
lucasrodes, line 180: This is what I have added to ensure that code after line 201 is not executed whenever the user changes this state after having clicked on "Submit" earlier.

lucasrodes (commented):
LGTM!

I was getting some error messages because "?output=csv" is not in the snapshot URL. I suspect this happened for private datasets bc I don't have `FASTTRACK_SECRET_KEY` set, and therefore `_decrypt` function is not able to do its job?

I added some code to handle this situations and show a slightly more informative error message.

PR #2288, opened 2024-02-13
Include grapher db data in version tracker

pabloarosado:
Fixes https://github.com/owid/etl/issues/1821

# Summary
* Move `VersionTracker` from `etl.helpers` to its own module `etl.version_tracker`, and adapt tests and other uses.
* Create a function in `etl.db` to fetch some data from DB about datasets and their charts.
* Fix some minor issues with `VersionTracker`.
* Let `VersionTracker` use data from DB about charts. Now `steps_df` contains useful information about each step, like the number of charts that, in one way or another, are affected by the current step.
* One can run `version_tracker --connect_to_db` to get a list of errors if there is any public dataset with charts for which there is no active ETL step. Thanks to this, I found some inconsistencies ([a missing FAOSTAT step](https://github.com/owid/etl/pull/2290), and [two missing fasttrack steps](https://owid.slack.com/archives/C0193RW5E2J/p1707832336154589)).
* One can also add the `--warn_on_archivable` flag. However, this prints a long list of archivable steps. And some of them are not actually archivable, since they are used by explorers. So, for now, we can ignore it and run it only optionally.

# To be reviewed
* [x] etl/db.py
* [x] pyproject.toml   <- Here, should `version_tracker` by default connect to DB? It would help us detect missing steps, but maybe it causes issues when run locally. We could also let it try to connect to DB, and on failure run it without connecting.
* [x] etl/version_tracker.py <- Here, the only methods to review are `get_all_archivable_steps`, `_add_info_from_db`, `_create_steps_df`, `check_that_db_datasets_with_charts_are_not_archived`, and `check_that_db_datasets_with_charts_have_active_etl_steps`.

There is no need to review `etl/helpers.py`,  `tests/test_helpers.py`, `tests/test_version_tracker.py`, and `etl/scripts/faostat/archive_old_datasets.py`, as they have been affected just by the refactoring.


etl/db.py
```diff
@@ -173,3 +173,90 @@ def get_all_datasets(archived: bool = True, db_conn: Optional[MySQLdb.Connection
...
+    if len(unknown_channels) > 0:
+        log.error(
+            "Variables in grapher DB are expected to come only from ETL grapher channel, "
+            f"but other channels were found: {unknown_channels}"
+        )
+
+    # Create a column with the step name.
+    # First assume all steps are public (hence starting with "data://").
+    # Then edit private steps so they start with "data-private://".
+    df["step"] = ["data://" + "/".join(etl_path.split("#")[0].split("/")[:-1]) for etl_path in df["etl_path"]]
```
Marigold, line 259: I've been writing code for parsing steps in other places too. Perhaps it would be useful to have a class
```
class StepName:
  def __init__(self, name):
    ...
  @property
  def table_name(self):
    ...

  @property
  def indicator(self):
    ...
```
Something similar to how `pathlib.Path` works. (If you don't want to do it as part of this PR I can do it in a different one and refactor it)

Marigold:
> Here, should version_tracker by default connect to DB? It would help us detect missing steps, but maybe it causes issues when run locally. 

I'd be inclined to turn it on by default first and only turn it off if it causes issues.

> We could also let it try to connect to DB, and on failure run it without connecting.

I'd let it explicitly fail with a good error message instead of falling back.

etl/db.py
pabloarosado: Good idea. I'd prefer to leave it for a different PR. Feel free to do that yourself!

etl/db.py
```diff
@@ -173,3 +173,90 @@ def get_all_datasets(archived: bool = True, db_conn: Optional[MySQLdb.Connection
...
+    query = """\
+    SELECT
+        q1.datasetId AS dataset_id,
+        d.name AS dataset_name,
+        q1.etlPath AS etl_path,
+        d.isArchived AS is_archived,
+        d.isPrivate AS is_private,
+        q2.chartIds AS chart_ids
+    FROM
+        (SELECT
```
Marigold, line 197: Looks like having `catalogPath` in `datasets` table would be useful. I'll add it one day, I promise...

pabloarosado:
> > Here, should version_tracker by default connect to DB? It would help us detect missing steps, but maybe it causes issues when run locally.
> 
> I'd be inclined to turn it on by default first and only turn it off if it causes issues.
> 
> > We could also let it try to connect to DB, and on failure run it without connecting.
> 
> I'd let it explicitly fail with a good error message instead of falling back.

Thanks for the suggestion. In my last commit I've changed things a bit: By default, it tries to connect to DB. If it fails, it raises a warning and internally sets `self.connect_to_db=False`, therefore skipping the checks that need data from DB. Please let me know if you have a better idea, thanks.

lucasrodes:
Not major, but I agree on Mojmir's point. 

Maybe it is a bit clearer if the script just fails when it can't connect to the DB. I can see how we may let the warning pass and not see it. Then we would not consider Grapher datasets without us knowing.

PR #2297, opened 2024-02-14
📊 SDGs related to urbanization

veronikasamborska1994:
the following indicators seem to not have been processed correctly in the main grapher dataset (likely because there are values for multiple cities in each country): 
https://owid.cloud/admin/variables/812289 
https://owid.cloud/admin/variables/812286

I am adding a garden step that calculates averages across cities for each country specifically for these two indicators.


etl/steps/data/meadow/un/2024-02-14/sdgs_urbanization.py
```diff
@@ -0,0 +1,39 @@
...
+    ds_meadow = paths.load_dataset("un_sdg")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow["un_sdg"].reset_index()
+
+    urbanization_tb = tb[(tb["indicator"] == "11.7.1") | (tb["indicator"] == "11.2.1")]
+    cols = ["seriesdescription", "country", "year", "value", "cities"]
+    urbanization_tb = urbanization_tb[cols]
+    # Convert the seriesdescription to string to avoid using categorical data
+    urbanization_tb["seriesdescription"] = urbanization_tb["seriesdescription"].astype(str)
```
Marigold, line 24: Nit - this won't help you, calling `ds_meadow.save()` converts string to categorical type anyway (to make the dataset smaller on disk).

etl/steps/data/garden/un/2024-02-14/sdgs_urbanization.py
```diff
@@ -0,0 +1,53 @@
...
+
+    average_by_country = average_by_country.dropna(subset=["value"])
+    pivot_tb = average_by_country.pivot(
+        index=["country", "year"], columns="seriesdescription", values="value"
+    ).reset_index()
+    pivot_tb = pivot_tb.underscore().set_index(["country", "year"], verify_integrity=True)
+
+    pivot_tb.metadata = metadata
+
+    print(pivot_tb.columns)
```
Marigold, line 39: Forgotten print?

etl/steps/data/garden/un/2024-02-14/sdgs_urbanization.py
```diff
@@ -0,0 +1,53 @@
...
+    metadata = tb.metadata
+
+    #
+    # Process data.
+    #
+    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path)
+
+    average_by_country = tb.groupby(["country", "year", "seriesdescription"])["value"].mean().reset_index()
+
+    average_by_country = average_by_country.dropna(subset=["value"])
```
Marigold, line 31: Nit - grouping by categorical variables in pandas by default creates a group even for **non-observed categories**. That's why you end up with so many NaNs. It is ok since the dataset is small, but the dataframe can explode with bigger datasets.

If you add `groupby(..., observed=True)`, it'll only keep non-null values. (I don't know why isn't it default)

Issue #2170, opened 2024-01-09
:hammer: Replace DVC by custom code


Marigold:
DVC is starting to [become a burden](https://github.com/owid/etl/pull/2058#issue-2035441003). We've already replaced `dvc pull` with our own custom code (that does the same thing but is intuitive and faster). We could go even further and wipe it out altogether. This would involve:

1. Replacing `Snapshot.dvc_add` that uploads files to R2
2. Removing `dvc pull` used as pre-pulling from some Buildkite jobs

Both are pretty straightforward and would simplify our codebase. The only reason for not doing this would be if we wanted to leverage more DVC functionality, but I don't think that's gonna happen.

PR #2310, opened 2024-02-19
:hammer: deprecate DVC

Marigold:
Fixes https://github.com/owid/etl/issues/2170.

Remove DVC library from ETL and replace it by a custom code. DVC has been pretty slow lately, and since we haven't used any of its other features, it doesn't make sense to keep it here. Custom code for uploading files based on their checksums is way faster and more straightforward. We keep YAML format for `*.dvc` files, so reverting this decision in the future is certainly possible.

### DIFF
--- a/apps/fasttrack/cli.py
+++ b/apps/fasttrack/cli.py
@@ -299,7 +299,7 @@ def app(dummy_data: bool) -> None:
     po.put_markdown("""## Links""")
     po.put_markdown(
         f"""
-    * [Dataset in admin]({os.environ.get("ADMIN_HOST", "")}/admin/datasets/{_dataset_id(dataset.metadata)})
+    * [Dataset in admin]({config.ADMIN_HOST}/admin/datasets/{_dataset_id(dataset.metadata)})
     * [Commit in ETL]({github_link})
     """
     )
--- a/etl/config.py
+++ b/etl/config.py
@@ -132,6 +132,8 @@ ETL_API_COMMIT = env.get("ETL_API_COMMIT") in ("True", "true", "1")
 # if True, commit and push updates from fasttrack
 FASTTRACK_COMMIT = env.get("FASTTRACK_COMMIT") in ("True", "true", "1")
 
+ADMIN_HOST = env.get("ADMIN_HOST", "http://localhost:3030")
+
 BUGSNAG_API_KEY = env.get("BUGSNAG_API_KEY")
 
 OPENAI_API_KEY = env.get("OPENAI_API_KEY", None)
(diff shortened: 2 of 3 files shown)


etl/snapshot.py
```diff
@@ -111,26 +66,18 @@ def metadata_path(self) -> Path:
...
-            download_url = f"{https_url}/{md5[:2]}/{md5[2:]}"
+            download_url = f"{config.R2_SNAPSHOTS_PUBLIC_READ}/{md5[:2]}/{md5[2:]}"
             files.download(download_url, str(self.path), progress_bar_min_bytes=2**100)
         else:
-            s3_url = DVC_CONFIG['remote "private"']["url"]
-            download_url = f"{s3_url}/{md5[:2]}/{md5[2:]}"
+            download_url = f"{config.R2_SNAPSHOTS_PRIVATE}/{md5[:2]}/{md5[2:]}"
             s3_utils.download(download_url, str(self.path))
 
+        assert checksum_file(self.path) == md5, f"Checksum mismatch for {self.path}"
```
larsyencken, line 79: I wonder if it's appropriate to leave the file in place if the checksum is wrong. Should we rename it to `.bad` or something? The worry is that we forget or rerun and it doesn't get checked again.

Open question, I'm really not sure.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ click = ">=8.0.1"
 PyYAML = ">=6.0.1"
 papermill = ">=2.3.3"
 jupyterlab = ">=3.1.13"
-mysqlclient = ">=2.2.4"
+mysqlclient = "^2.2.4"
```
larsyencken, line 24: Is this on purpose?

larsyencken (approved):
Overall I think it's a good move 👍 

Two minor comments, you can decide whether to address them or not and then :shipit:

larsyencken:
I added `mkdir -p .dvc` to a BuildKite step to avoid a spurious failure: https://buildkite.com/our-world-in-data/etl-build-public-datasets-branch/settings/steps

etl/snapshot.py
Marigold: Good idea, better to remove it. It's worth saying that this should NEVER happen and I was surprised to find one snapshot and two walden datasets that failed this assert. This was due to my recent change in md5 calculation that fixed newlines compatiblity between Windows/Linux in csv files. I fixed thoes files and we should hopefully never again run into this error.

pyproject.toml
Marigold: Oh shit, no! Thanks for noticing.

PR #2312, opened 2024-02-20
:hammer: pin mysqlclient to 2.2.3 to avoid decimal.InvalidOperation

Marigold:


larsyencken (commented):
I'm running 2.2.4 and I don't get any error on `SELECT VERSION()`.

What's the code path that's giving you grief?

Issue #2315, opened 2024-02-20
🐛 chart-revision: improve handling of large datasets


lucasrodes:
@spoonerf reported problems when trying to use `etlwiz charts` for large dataset updates. 

### Recreate example

1. Run `etlwiz charts`.
2. Select datasets:
  i. Old dataset: **5599 IHME - Global Burden of Disease - Deaths and DALYs - Institute for Health Metrics and Evaluation (2022-04)**
  ii. New dataset: **5890 Institute for Health Metrics and Evaluation - Global Burden of Disease (2019) - Deaths and DALYs - Deaths and DALYs**
3. Click "Next (1/3)".
4. 🐛 Here, the code takes some time to find the mappings. Possible solutions:
  i. Add spinner so users know that something is going on
  ii. Optimize call to `find_mapping_suggestions`.
5. The app starts rendering the select boxes and crashes after some time.

PR #2316, opened 2024-02-20
🐛 chart-revision: limit number of variables to map

lucasrodes:
ref: 
- https://github.com/owid/etl/issues/2315


spoonerf:
Thanks @lucasrodes ! I will have to give this a go :)

PR #2328, opened 2024-02-22
Create ETL dashboard

pabloarosado:
# Overview
This PR creates an ETL dashboard, with two main usages:
1. Track datasets and updates (mostly for production), to be used by anyone, possibly helpful for Ed to prioritize updates.
2. Update specific steps (to be used locally).

# Changes
This PR:
* Creates an ETL dashboard, added to the wizard.
* Fixes some issues in `VersionTracker` and `StepUpdater`.
* Creates a temporary snapshot script `snapshots/un/2022-07-11/un_wpp.py` which does nothing (it actually raises an informative error), but contains the code from walden. This way on next update we will have something to start with.

# Reviewers
I don't think a thorough review is necessary (except for the changes in lib/catalog).
@Marigold and @lucasrodes feel free to play around with [the dashboard on staging](http://staging-site-create-etl-dashboard/etl/wizard/ETL%20Dashboard). If you want to try updating files, you'd need to fetch the code and run `etlwiz` on your local machine (with local grapher running). But that's up to you.

@Marigold could you please review the following files?
* [x] etl/db.py
* [x] etl/helpers.py
* [x] lib/catalog/owid/catalog/tables.py
* [x] etl/version_tracker.py

@lucasrodes could you please review the following files?
* [x] apps/step_update/cli.py
* [x] apps/wizard/config/config.yml
* [x] apps/wizard/pages/dashboard.py

No need to review:
* poetry.lock
* pyproject.toml
* snapshots/un/2022-07-11/un_wpp.py

If you find any issues or have any ideas, please add them to the following issue: https://github.com/owid/etl/issues/2365

Thanks a lot!


apps/wizard/pages/dashboard.py
```diff
@@ -0,0 +1,572 @@
...
+    GRAPHER_DATASET_BASE_URL = f"http://{GRAPHER_DATASET_BASE_URL}"
+
+# List of dependencies to ignore when calculating the update state.
+# This is done to avoid a certain common dependency (e.g. population) to make all steps appear as needing major update.
+DEPENDENCIES_TO_IGNORE = [
+    # "data://garden/demography/2023-03-31/population",
+    "snapshot://hyde/2017/general_files.zip",
+]
+
+# Define labels for update states.
```
Marigold, line 38: Nit - you could also use Enum to make it a bit more readable.
```
from enum import Enum

class UpdateState(Enum):
    UNKNOWN = "Unknown"
    UP_TO_DATE = "No updates known"
    OUTDATED = "Outdated"
    MINOR_UPDATE = "Minor update possible"
    MAJOR_UPDATE = "Major update possible"
    ARCHIVABLE = "Archivable"

...
print(UpdateState.UNKNOWN)
```

pabloarosado: Thanks, nice idea! However, I just tried, and unfortunately it seems that streamlit is not happy about this kind of object:
```_pickle.PicklingError: Can't pickle <enum 'UpdateState'>: it's not the same object as __main__.UpdateState```
So I've reverted the changes.

pabloarosado:
> Code looks good! I also played with it for a bit on staging and it looked useful (I guess zero pageviews in the dashboard come from the bug we had on Friday, right?)

Great, thanks for the review! Yes, the issue with analytics is beyond this PR.

apps/wizard/pages/dashboard.py
lucasrodes: This could be solved by `enum.StrEnum`, which is avaialble since Python 3.11. Currently ETL supports >=py3.10, so not an option right now.

Instead, could use `UpdateState.UNKNOWN.value` instead of `UpdateState.UNKNOWN`. I've added these changes.

lucasrodes (approved):
This is really promising, thanks for planting this seed, Pablo! And great integration with Wizard :)) It LGTM!

Minor comments below.

I've added some minor tweaks. Mainly: (i) to improve code readability, (ii) UX tweaks, (iii) fix the double-button-click requirement to clear the operations list.

I've also tweaked the logic so that the buttons to clear the operations list, or to submit it are _only_ shown if at least one step has been updated to the list. Before, the user could click on them (and the submit button would actually trigger an action). I also find it cleaner to only show them when they are actually usable.

Finally, I'd suggest improving the help text for the action buttons on "add direct dependencies", "add all dependencies". It took me some time to understand their differences. My interpretation was that "direct dependencies" = dependencies listed in the dag under the step, and "all dependencies" = direct dependencies + dependencies of dependencies (i.e. recursive dependencies)? Same applies for "usages".

lucasrodes:
Sorry for causing this, Pablo!

I'm fixing the bug. This is coming from the removal of the former `df` object. In the past,`steps_df` was preserved with all columns, and now I'm overwriting it only with the subset we want to display. Some functions were trying to access columns that were removed in this process.

PR #2335, opened 2024-02-22
:sparkles: speedup surface temperature

Marigold:
Speed up geometry clipping by running rough `clip_box` before the actual `clip` operation. I googled this recommendation by accident on some random forum. Loading gzip still takes a lot of memory (peak is ~10gb), but the actual processing is much faster now, and the whole step takes ~8min on the staging server. I visually checked the output and it looks the same.

@veronikasamborska1994 I am leaving the decision on whether to merge this or the snapshot version to you. Both are valid approaches.


veronikasamborska1994:
hey @Marigold! cool you found this! let's go with your solution! I also just checked the output and it looks basically the same except I think we lose Aruba (Neth.) and Saint Lucia data when we do it this way but tbh I am not sure whether we should be doing this clipping for such small countries anyway.

veronikasamborska1994 (approved):

PR #2336, opened 2024-02-22
📊 Weekly Fires - GWIS web scraping

veronikasamborska1994:


veronikasamborska1994:
hey @Marigold! this is actually not working properly yet but I thought I'd tag you in case you have time at some point to take a look at how we might be able to make the code more stable. Right now I can successfully add a year's worth of data from every country but then begin to get errors that are often not consistent between different runs (also it takes forever to run because I need to iterate through a unique URL for each country and year).

this is not urgent at all and don't worry if you don't have time for this! thanks a lot anyway :)

Marigold:
@veronikasamborska1994 I was checking [the page](https://gwis.jrc.ec.europa.eu/apps/gwis.statistics/seasonaltrend/HKG/2021/CO2) with Chrome Dev tools on the side and found that they're making calls to https://api2.effis.emergency.copernicus.eu/statistics/v2/gwis/weekly?country=HKG&year=2021 to get the data for charts. Would it be possible to use it instead of selenium? (I haven't checked whether the data is identical, but it looks similar)

veronikasamborska1994:
> @veronikasamborska1994 I was checking [the page](https://gwis.jrc.ec.europa.eu/apps/gwis.statistics/seasonaltrend/HKG/2021/CO2) with Chrome Dev tools on the side and found that they're making calls to https://api2.effis.emergency.copernicus.eu/statistics/v2/gwis/weekly?country=HKG&year=2021 to get the data for charts. Would it be possible to use it instead of selenium? (I haven't checked whether the data is identical, but it looks similar)

niceeee!!!!!! thank you!!! I've completely missed this and still can only see them making a call to the API for the download of the PDF file only. do you by any chance see how they are getting the CO2 emissions data?

Marigold:
> do you by any chance see how they are getting the CO2 emissions data?

I've tried to plot this
```
import requests
url = 'https://api2.effis.emergency.copernicus.eu/statistics/v2/emissions/weekly?country=HKG&year=2021'
response = requests.get(url)
df = pd.DataFrame(response.json()['emissionsweekly'])
df.query('plt == "CO2"')[['dt', 'avgv', 'minv', 'maxv']].set_index('dt').plot()
```

and I get something resembling their graph. Would it work?
image

veronikasamborska1994:
> > do you by any chance see how they are getting the CO2 emissions data?
> 
> I've tried to plot this
> 
> ```
> import requests
> url = 'https://api2.effis.emergency.copernicus.eu/statistics/v2/emissions/weekly?country=HKG&year=2021'
> response = requests.get(url)
> df = pd.DataFrame(response.json()['emissionsweekly'])
> df.query('plt == "CO2"')[['dt', 'avgv', 'minv', 'maxv']].set_index('dt').plot()
> ```
> 
> and I get something resembling their graph. Would it work? image

yeeesssssss that's perfect!! thank you so much!!! you are the best! <3

veronikasamborska1994:
hey @Marigold! thanks again for all your help with this. I think it's now ready for review :)

snapshots/climate/2024-02-22/weekly_wildfires_2003_2023.py
```diff
@@ -0,0 +1,408 @@
...
+# Initialize logger.
+log = get_logger()
+
+
+# Version for current snapshot dataset.
+SNAPSHOT_VERSION = Path(__file__).parent.name
+# Define a range of years for which data will be processed.
+YEARS = range(2003, 2024)
+
+COUNTRIES = {
```
Marigold, line 24: Nit - you could load countries from regions dataset like this
```
from owid.catalog import find
tb_regions = find(table="regions", dataset="regions").iloc[0].load()
```
(it's not by any means necessary, feel free to ignore it)

Marigold (approved):
All looks good!

snapshots/climate/2024-02-22/weekly_wildfires_2003_2023.py
veronikasamborska1994: thanks! that's actually a useful function to know

Issue #2334, opened 2024-02-22
ETL freezing on step `data://meadow/climate/2023-12-20/surface_temperature`


Marigold:
ETL tends to freeze on step `data://meadow/climate/2023-12-20/surface_temperature`. The snapshot has 1.6gb, so ETL may start swapping memory after decompressing it.

I can't run it locally as the memory spikes and starts swapping. It takes ~3min on @veronikasamborska1994 machine.

Marigold:
@veronikasamborska1994 I looked at the script and couldn't find any way how to make it faster. I tried to optimise memory usage a bit [in this PR](https://github.com/owid/etl/pull/2335/files), but the peak memory usage could still be >20gb (I've seen 40gb of memory on some staging servers). The memory spikes after creating `clip` object.

It takes ~2h to run the step on the server. I don't know why it is so long. It could be memory swapping, but it could be as well something else.

Is such a heavy computation expected? Is there any way to make it faster?

veronikasamborska1994:
@Marigold I'll take a look but there aren't that many ways in which this clipping can be done so I am not sure if I'll be able to improve it much. 

What I could definitely do though is move this processing step to snapshot? Would that help?

Marigold:
> What I could definitely do though is move this processing step to snapshot? Would that help?

That's a great idea! It's definitely easier than figuring out why it is so slow in production.

Marigold:
Also getting this from CI run `ETL - build public datasets (branch)` (I didn't know there's a memory cap for CI or Buildkite agents?)
```
File "/var/lib/buildkite-agent/builds/etl-build-2-15/our-world-in-data/etl-build-public-datasets-branch/etl/steps/data/meadow/climate/2023-12-20/surface_temperature.py", line 77, in run
  clip = da.rio.clip(data.geometry.apply(mapping), data.crs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/var/lib/buildkite-agent/builds/etl-build-2-15/our-world-in-data/etl-build-public-datasets-branch/.venv/lib/python3.11/site-packages/rioxarray/raster_array.py", line 911, in clip
  cropped_ds = _clip_xarray(
               ^^^^^^^^^^^^^
[...]

PR #2338, opened 2024-02-23
:bar_chart: temporarily disable climate/2023-12-20/surface_temperature

Marigold:
Hotfix until we fix the issue https://github.com/owid/etl/issues/2334


veronikasamborska1994 (approved):
sounds good! will try to make the change to snapshot ASAP

PR #2351, opened 2024-02-28
:sparkles: Various core ETL improvements needed for GHO dataset

Marigold:
I had to do various improvements to core ETL to make Global Health Observatory dataset work in ETL. The plan is to merge this first and then rebase [GHO PR](https://github.com/owid/etl/pull/2172).

## New env variable `SUBSET` for debugging

When working with a huge dataset like GHO or WDI, running a step can take minutes or more. In development, one typically focuses on a single indicator, debugs, fixes it, and pushes to grapher to see the output. To speed up this process, you can now load `SUBSET` from environment and filter your data by it (typically indicator name). With this, I can run GHO steps for indicator `sa_0000001547` in a few seconds
```
SUBSET=sa_0000001547 STAGING=gho etl run who/2024-01-03/gho --grapher
```

If you use `SUBSET`, ghost indicators are not removed from MySQL, hence you only add / edit that indicator without messing up the whole dataset. `SUBSET` is added to the checksum if used.

## Trimming long `short_name`

Sometimes, we run into 255 char limit for `short_name`, typically because of many dimensions with long values. Trimming to 255 chars doesn't help because prefix could be same and manually updating affected variables is a tedious process (and there isn't a good way how to do that).

What I'm doing now is that I take prefix with 245 characters and use the remaining 10 characters for `hash(short_name)`. This guarantees that my **short names are unique**.

## Other improvements
- Refactor dimensions when upserting to MySQL
- Don't exclude NaN values in dimensions
- A few minor fixes specific for GHO dataset

Adding unit tests.... ⌛

### DIFF
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -15,6 +15,7 @@ from owid.catalog.utils import underscore
 
 from etl.db import get_connection, get_engine
 from etl.db_utils import DBUtils
+from etl.files import checksum_str
 
 log = structlog.get_logger()
 
@@ -233,7 +234,7 @@ def _underscore_column_and_dimensions(column: str, dim_dict: Dict[str, Any], tri
 
     if len(short_name) > 255:
         if trim_long_short_name:
-            unique_hash = f"_{abs(hash(short_name))}"
+            unique_hash = f"_{checksum_str(short_name)}"
             short_name = short_name[: (255 - len(unique_hash))] + unique_hash
             log.warning(
                 "short_name_trimmed",
--- a/etl/grapher_helpers.py
+++ b/etl/grapher_helpers.py
@@ -69,8 +69,8 @@ def expand_dimensions(tb: catalog.Table) -> catalog.Table:
 def _yield_wide_table(
     table: catalog.Table,
     na_action: Literal["drop", "raise"] = "raise",
-    dim_titles: Optional[List[str]] = None,
     warn_null_variables: bool = True,
+    trim_long_short_name: bool = True,
 ) -> Iterable[catalog.Table]:
     """We have 5 dimensions but graphers data model can only handle 2 (year and entityId). This means
     we have to iterate all combinations of the remaining 3 dimensions and create a new variable for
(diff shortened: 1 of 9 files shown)


etl/steps/__init__.py
```diff
@@ -499,6 +499,13 @@ def checksum_input(self) -> str:
         for f in self._step_files():
             checksums[f] = files.checksum_file(f)
 
+            # if using SUBSET to process just a subset of the data, then add it to its checksum if
+            # the file contains it
+            if config.SUBSET:
+                with open(f) as istream:
+                    if "SUBSET" in istream.read():
+                        checksums[f] += config.SUBSET
+
```
larsyencken, line 508: This is nice, so that a different subset choice (or the lack of a subset) triggers a rebuild 👍

lib/catalog/owid/catalog/datasets.py
```diff
@@ -205,14 +206,23 @@ def update_metadata(
...
-                    table = self[table_name]
+                    try:
+                        table = self[table_name]
+                    except KeyError as e:
+                        if errors == "raise":
+                            raise e
+                        else:
+                            if errors == "warn":
+                                warnings.warn(str(e))
+                            continue
```
larsyencken, line 225: What's the situation where this comes up? Is it because with a subset some tables are not included? Can you leave a comment behind explaining why you're making it more flexible here?

etl/grapher_helpers.py
```diff
@@ -212,24 +214,38 @@ def _expand_jinja(obj: Any, dim_dict: Dict[str, str]) -> Any:
...
+            log.warning(
+                "short_name_trimmed",
+                short_name=short_name,
+                column=column,
+                dims=dim_dict,
+            )
+        else:
+            raise AssertionError(f"short_name {short_name} is too long for MySQL variables.shortName column")
+
+    return short_name
```
larsyencken, line 248: This makes total sense, and keeps us within a helpful max char length for `short_name`.

It's clear that what we're packing into `short_name` is not that helpful though. I wonder what alternatives there are for us in the long term. I'll bring it to a `#data-architecture` meeting.

Marigold: Cool, perhaps we can come up with a better solution. This happens extremely rarely, though, and only for massive datasets like WDI or GHO. I don't think it's worth spending too much time on it.

lib/catalog/owid/catalog/datasets.py
Marigold: > What's the situation where this comes up? Is it because with a subset some tables are not included?

Yes, this is useful if you work with `SUBSET` of dataset and have YAML metadata full of indicators. I'm updating the docs with motivation.

pabloarosado (approved):
Nice work!

PR #2358, opened 2024-02-29
:bar_chart: Update GHO suicides and reuse data from GHO dataset

Marigold:
Proof of concept that we can create other datasets from the primary GHO dataset.


spoonerf (approved):
Yeh it totally makes sense to me, especially for future updates!

PR #2378, opened 2024-03-06
Fix missing info in operations list of etl dashboard

pabloarosado:
I noticed that the information in the operations list was incomplete.

To replicate the issue:
* Tick “Select only grapher...”.
* Select step “data://grapher/irena/2023-12-12/renewable\_power\_generation_costs”.
* Click on "Add selected steps to the Operations list".
* Click on "Add direct dependencies". A garden step is added to the operations list.
* Click on "Add direct dependencies" of the garden step. The corresponding meadow step should appear, but instead, nothing happens.

If you repeat this without ticking on "Select only grapher...", the problem is solved, and you do see meadow as well.

So to perform operations, we need the full steps_df, not just the filtered version that is displayed.

That's why, in this PR, I define steps_df (as all the original data, used in operations), and steps_df_display (as the filtered data, used for the Steps table).

### DIFF
--- a/apps/wizard/pages/dashboard.py
+++ b/apps/wizard/pages/dashboard.py
@@ -96,7 +96,7 @@ if not can_connect():
 # LOAD STEPS TABLE
 ########################################
 @st.cache_data
-def load_steps_df_all() -> pd.DataFrame:
+def load_steps_df() -> pd.DataFrame:
     """Generate and load the steps dataframe.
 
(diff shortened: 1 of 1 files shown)


pabloarosado:
@lucasrodes I'll merge now since I wanted to present the dashboard this morning, but please have a look when you have some time, and let me know if you think there is a better solution.

PR #2379, opened 2024-03-06
year zero

lucasrodes:
This PR rolls back to have year zero in the indicator. I think that we should have astronomical year units in ETL.

This affects our population OMM and other indicators that rely on HYDE.

However, we should decide on how to present years on Grapher. I’d leave it to the editors to decide what’s the best way to communicate years. And do a conversion on the Grapher side from astronomical years to CE (or keep it astronomical). If editors decide to use CE-years, I think we should remove the year 0 in the display of the x-axis.


pabloarosado (approved):
I haven't checked their documentation, but I suppose that, if they included a year 0 and negative years, we should respect that. So I agree on these changes.

PR #2382, opened 2024-03-07
Improve etl dashboard

pabloarosado:
This PR improves the ETL dashboard in the following ways:
* Instead of showing by default grapher steps with charts, show all grapher steps by default (plus explorer steps).
* Operations list now has a button to remove steps that we most likely don't want to update (e.g. regions or population).
* Operations list now has a button to replace steps in the list by their corresponding updated versions (useful for after updating).
* There is a button to run snapshots and ETL steps.

Feel free to play around with [the resulting new dashboard on staging](http://staging-site-improve-etl-dashboard/etl/wizard/ETL%20Dashboard).

### DIFF
--- a/apps/wizard/config/config.yml
+++ b/apps/wizard/config/config.yml
@@ -76,12 +76,12 @@ sections:
 
   - title: "Dashboard"
     description: |-
-      Control panel for ETL updates.
+      Control panel for ETL steps.
     apps:
       - title: "ETL Dashboard"
         alias: dashboard
         entrypoint: pages/dashboard.py
-        description: Control panel for ETL updates.
+        description: Control panel for ETL steps.
         maintainer: "@pablo"
         emoji: "📋"
         image_url: "https://cdn.pixabay.com/photo/2018/09/04/17/02/indicator-3654257_960_720.jpg"
--- a/apps/wizard/pages/dashboard.py
+++ b/apps/wizard/pages/dashboard.py
@@ -677,121 +677,125 @@ with st.container(border=True):
 ########################################
 
 if st.session_state.selected_steps:
-    # Add an expander menu with additional parameters for the update command.
-    with st.container(border=True):
-        with st.expander("Additional parameters to update steps", expanded=False):
-            dry_run_update = st.toggle(
-                "Dry run",
-                True,
-                help="If checked, the update command will not write anything to the DAG or create any files.",
-            )
-            version_new = st.text_input("New version", value=TODAY, help="Version of the new steps to be created.")
+    cols_primary_buttons = st.columns([2, 2])
 
-        btn_submit = st.button(
-            f"Update {len(st.session_state.selected_steps)} steps",
-            help="Update steps in the _Operations list_.",
-            type="primary",
-            use_container_width=True,
-        )
-
-        # Button to execute the update command and show its output.
-        if btn_submit:
-            if ENV_IS_REMOTE:
-                st.error(
-                    "The update command is not available in the remote version of the wizard. Update steps locally."
+    with cols_primary_buttons[0]:
+        # Add an expander menu with additional parameters for the update command.
+        with st.container(border=True):
+            with st.expander("Additional parameters to update steps", expanded=False):
+                dry_run_update = st.toggle(
+                    "Dry run",
+                    True,
+                    help="If checked, the update command will not write anything to the DAG or create any files.",
                 )
-                st.stop()
-            else:
-                with st.spinner("Executing step updater..."):
-                    # TODO: It would be better to directly use StepUpdater instead of a subprocess.
-                    command = (
-                        "etl update "
-                        + " ".join(st.session_state.selected_steps)
-                        + " --non-interactive"
-                        + f" --step-version-new {version_new}"
+                version_new = st.text_input("New version", value=TODAY, help="Version of the new steps to be created.")
+
+            btn_submit = st.button(
+                f"Update {len(st.session_state.selected_steps)} steps",
+                help="Update steps in the _Operations list_.",
+                type="primary",
+                use_container_width=True,
+            )
+
+            # Button to execute the update command and show its output.
+            if btn_submit:
+                if ENV_IS_REMOTE:
+                    st.error(
+                        "The update command is not available in the remote version of the wizard. Update steps locally."
                     )
-                    if dry_run_update:
-                        command += " --dry-run"
-                    cmd_output = execute_command(command)
-                    # Show the output of the command in an expander.
-                    with st.expander("Command Output:", expanded=True):
-                        st.text_area("Output", value=cmd_output, height=300, key="cmd_output_area")
-                    if "error" not in cmd_output.lower():
-                        # Celebrate that the update was successful, why not.
-                        st.balloons()
-                        if not dry_run_update:
-                            # Reload steps_df to include the new steps.
-                            st.session_state["reload_key"] += 1
-                    # Add a button to close the output expander.
-                    st.button("Close and reload _Steps table_", key="acknowledge_cmd_output")
+                    st.stop()
+                else:
+                    with st.spinner("Executing step updater..."):
+                        # TODO: It would be better to directly use StepUpdater instead of a subprocess.
+                        command = (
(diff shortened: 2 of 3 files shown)


lucasrodes:
Thanks for working on this, Pablo! 


Some suggestions:

- As [mentioned before](https://github.com/owid/etl/pull/2328#pullrequestreview-1915002726), I think we could improve a bit the help messages for the direct/all usages/dependencies buttons. I think it can be a bit clearer.
- I think the tool is slowly getting packed with multiple buttons. That makes readability a bit harder. Some suggestions to improve that:
  - The layout of the last three buttons is a bit confusing. Why put them in the same arrow? Ideally we'd remove the unnecessary spacing between them, but that is not trivial with Streamlit. Instead, I'd suggest aligning them vertically (put them one after the other vertically).
  - Change button name "Remove non-updateable (e.g. population)" → "Drop auxiliary datasets". Improve the help message from this button. Currently, it makes a reference to the variable `NON_UPDATEABLE_IDENTIFIERS`. I think you could render those identifiers in the help text, e.g. `- "\n- ".join(NON_UPDATEABLE_IDENTIFIERS)`.
    - I am also wondering when someone would want to NOT drop these datasets from the operations list? If we are just going to be dropping them always, we can consider filtering them out by default? In that case, and if needed, you could have a toggle that could be used to 'add' and 'remove' those auxiliary datasets.
  - I find the the label of the third button ("Replace 1 steps by their latest versions") a bit confusing. What does that mean? It is also unclear from the help message what that button is for.
- I see that there are two columns at the end. One to update the steps, and a new one to run snapshots/etl steps. My impression is that we'd want to click on run etl / snapshot steps only after having edited them? If so, I think this button belongs somewhere else. Currently it has the same prominence as the "update etl steps" one. I'd (i) change the button type to "secondary" and/or (ii) put it below the "update etl steps" one. Maybe you can keep them in the first column so that they only take half the screen and are more readable.


Remove non-updateable (e.g. population)

pabloarosado:
Thanks @lucasrodes! I've implemented some of your suggestions. Let me address them:

> * As [mentioned before](https://github.com/owid/etl/pull/2328#pullrequestreview-1915002726), I think we could improve a bit the help messages for the direct/all usages/dependencies buttons. I think it can be a bit clearer.

Done.

> * I think the tool is slowly getting packed with multiple buttons. That makes readability a bit harder. Some suggestions to improve that:
>   
>   * The layout of the last three buttons is a bit confusing. Why put them in the same arrow? Ideally we'd remove the unnecessary spacing between them, but that is not trivial with Streamlit. Instead, I'd suggest aligning them vertically (put them one after the other vertically).

Done, although I think I prefer them horizontally aligned, since they take much less space. If you agree, I'd revert this.

>   * Change button name "Remove non-updateable (e.g. population)" → "Drop auxiliary datasets". Improve the help message from this button. Currently, it makes a reference to the variable `NON_UPDATEABLE_IDENTIFIERS`. I think you could render those identifiers in the help text, e.g. `- "\n- ".join(NON_UPDATEABLE_IDENTIFIERS)`.

What you suggest is not exactly right: You are not only dropping auxiliary datasets, but also any steps that can't be updated (e.g. because they have `update_period_days=0`). And the list of non-updateable identifiers is very long. I just rephrased the help message a bit.

>  * I am also wondering when someone would want to NOT drop these datasets from the operations list? If we are just going to be dropping them always, we can consider filtering them out by default? In that case, and if needed, you could have a toggle that could be used to 'add' and 'remove' those auxiliary datasets.

I'm not sure about this. I think it's also confusing if you are clicking on "Add selected steps to the Operations list" and nothing happens. You may also want to do other kind of operations on those steps, not just updating them (e.g. run ETL on them). We can change this behavior in a future PR, based on feedback.

>   * I find the the label of the third button ("Replace 1 steps by their latest versions") a bit confusing. What does that mean? It is also unclear from the help message what that button is for.

I have rephrased it a bit. The idea is that, after you update steps, you can try to run them. If they work out of the box, then you don't need to do anything, and simply create a PR and go to chart revisions. Otherwise, you need to go to the terminal and execute them manually, one or more steps (which can get messy when there are many, e.g. in the climate namespace).

> * I see that there are two columns at the end. One to update the steps, and a new one to run snapshots/etl steps. My impression is that we'd want to click on run etl / snapshot steps only after editing them. If so, I think this button belongs somewhere else. Currently it has the same prominence as the "update etl steps" one, and feels like they can be clicked anytime (and that is not like that, there is some sequentially).
>   
>   * I'd (i) change the button type to "secondary" and/or (ii) put it below the "update etl steps" one. Maybe you can keep them in the first column so that they only take half the screen and are more readable. It would also be helpful to add some help text stating that the user should go first to the generated new files and edit them!

That's how I placed them first. But then I realised that you may just want to execute the ETL on some steps without necessarily updating them, which is convenient. Hence the two actions are parallel, not necessarily sequential.

lucasrodes:
Thanks for the clarifications! This all makes sense. And for adding some changes.

Some follow-up comments

> Done, although I think I prefer them horizontally aligned, since they take much less space. If you agree, I'd revert this.

I think that looks good now. I agree it takes more vertical space, but aligning them horizontally created some sort of arbitrary spacing IMO.

> What you suggest is not exactly right: You are not only dropping auxiliary datasets, but also any steps that can't be updated (e.g. because they have update_period_days=0). And the list of non-updateable identifiers is very long. I just rephrased the help message a bit.

I see. In that case I guess it is fine to keep the title as you have now, and in the future, we can remove the `(e.g. population)` bit, once everyone understands what's going on.

Again, my understanding is that the list of non-updateable datasets is given by `NON_UPDATEABLE_IDENTIFIERS`. If so, it is of length ~20, so I think it is totally fine to list them in the help message. The user will be able to scroll down (or not) as they please. I suggest adding a `- "\n- ".join(sorted(NON_UPDATEABLE_IDENTIFIERS)` bit or so to the help message.

> I'm not sure about this. I think it's also confusing if you are clicking on "Add selected steps to the Operations list" and nothing happens. You may also want to do other kind of operations on those steps, not just updating them (e.g. run ETL on them). We can change this behavior in a future PR, based on feedback.

IMO we can find a good solution in the future:

- toggle to show/not show these non-updateable datasets. I think a toggle here is better than a unidirectional button. With a button, once you drop the datasets, there is no way back.
- if the toggle is enabled or not by default that's trickier. That may depend on the datasets selected by the user.

> I have rephrased it a bit. The idea is that, after you update steps, you can try to run them. If they work out of the box, then you don't need to do anything, and simply create a PR and go to chart revisions. Otherwise, you need to go to the terminal and execute them manually, one or more steps (which can get messy when there are many, e.g. in the climate namespace).

I see. So there is a workflow where the user picks datasets from the table and just runs them? If the user has already updated the dataset, wouldn't this new dataset (with the new dependencies) already appear in the table? I guess I am still wondering when would I use this button. Is it bc the newly created datasets aren't shown in the table or tracked in underlying variables? If so, we should solve this.

> That's how I placed them first. But then I realised that you may just want to execute the ETL on some steps without necessarily updating them, which is convenient. Hence the two actions are parallel, not necessarily sequential.

I see. So you are actually catering two uses (for now) in this page. I feel like the title and the description of the page are focussed on the "update" process. Maybe you can update these like:

- "Control panel for ETL updates" → "Control panel for ETL steps"
- "Explore all active ETL steps, and, if you are working on your local machine, update them." → "Explore all active ETL steps, and, if you are working on your local machine, perform some actions".
- "🔨 To update some steps, select them from the Steps table and add them to the Operations list below." → "🔨 To perform actions on some steps, select them from the Steps table and add them to the Operations list below."
- "💡 If you want to perform an action on, say, a specific grapher dataset, you can select just that step to the Operations list, then click on "Add dependencies", and bulk-update them all in one go."
- Maybe add a clarification like "Current available actions are: 'update step', 'run step'.

pabloarosado:
Hey @lucasrodes thanks for the suggestions. I've implemented those changes, and added a little step-by-step tutorial at the beginning. Please let me know if you think it's good to ship.

lucasrodes:
Thanks @pabloarosado! LGTM! Let's go

PR #2385, opened 2024-03-07
Remove new unused population steps

pabloarosado:
Move unused new (experimental) population steps to the archive dag, and comment them.

Given that this experimental dataset is currently not used, and we do not advise using for now, this PR moves the steps to the archive (and comments them). This way we avoid population to appear as outdated (hence making all datasets that use population be outdated too) in the ETL dashboard. We also avoid confusion and the risk of using this dataset by mistake.

Once we decide to use this dataset, we can just move the steps from the archive to the active dag (the steps code has not moved).


lucasrodes (approved):

lucasrodes:
thanks for doing this, Pablo!

PR #2388, opened 2024-03-08
✨ Add switch to stop Table warnings about metadata issues

pabloarosado:
When combining variables with different units (and short units) it is useful to have a warning. For example:
```
tb = Table({"index": [1, 2, 3], "a": [4, 5, 6], "b": [7, 8, 9]})
tb["a"].metadata.unit = "terawatt-hours"
tb["b"].metadata.unit = "megawatt-hours"
tb["c"] = tb["a"] + tb["b"]
```
will raise a warning because "a" and "b" have different units.

When doing a melt operation,
```
tb.melt(id_vars="index", value_vars=["a", "b"])
```
you also get a warning if you are melting together variables with different units. In this case, it would also be useful. But if you are consciously melting together variables with different units, you may want to hide the warning (which is what started [this conversation](https://owid.slack.com/archives/C6C0MCXU0/p1709829759694139)).

For that reason, this PR lets you do that in two ways:
(A) Having a general switch `Table.WARN = False` that turns off all warnings.
(B) Having a function argument `melt(..., warn=False)`.

Similarly, I created a class-level WARN attribute to Variable, in case we want to turn off warnings for other operations (e.g. `tb["a"] + tb["b"]`).

@Marigold what do you think about this approach? Please let me know if you think there are better solutions.

### DIFF
--- a/lib/catalog/owid/catalog/tables.py
+++ b/lib/catalog/owid/catalog/tables.py
@@ -774,7 +774,7 @@ class Table(pd.DataFrame):
         variable_name = variables.UNNAMED_VARIABLE
         variable = variables.Variable(super().sum(*args, **kwargs), name=variable_name)
         variable.metadata = variables.combine_variables_metadata(
-            variables=[self[column] for column in self.columns], operation="+", name=variable_name
+            variables=[self[column] for column in self.columns], operation="+", name=variable_name, warn=self.WARN
         )
 
         return variable
@@ -783,7 +783,7 @@ class Table(pd.DataFrame):
         variable_name = variables.UNNAMED_VARIABLE
         variable = variables.Variable(super().prod(*args, **kwargs), name=variable_name)
         variable.metadata = variables.combine_variables_metadata(
-            variables=[self[column] for column in self.columns], operation="*", name=variable_name
+            variables=[self[column] for column in self.columns], operation="*", name=variable_name, warn=self.WARN
         )
 
         return variable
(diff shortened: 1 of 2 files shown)


Marigold:
I recently run into a similar issue where I needed to silence a (different type) of warning. Your approach would work, but I'm not sure how well will it scale with more types of warnings. Perhaps we could go with native python warnings instead and silence them with context managers like this

```python
import warnings

class MetadataWarning(UserWarning):
    pass

class DifferentValuesWarning(MetadataWarning):
    pass


def foo():
    warnings.warn("Different values detected", DifferentValuesWarning)


# ignore warnings with context manager
with warnings.catch_warnings():
    warnings.simplefilter("ignore", MetadataWarning)
    foo()


# we could create our own context manager to make it more concise
# with ignore_metadata_warnings():
#     foo()

# Alternatively, control warnings at the module level
# warnings.simplefilter('ignore', MetadataWarning)
```

That way, we wouldn't need to propagate `warn` all the way down.

(My bad for using `log.warning` which doesn't have a good mechanism for silencing them except of setting log level.)

What do you think about this alternative?

Marigold:
@pabloarosado let me know if you want me to implement it. I want to also look at step `data://grapher/lis/2023-08-30/luxembourg_income_study` which raises tons of warnings that should be silenced (or fixed?).

pabloarosado:
Hi @Marigold, yes, feel free to implement those changes you suggested. I was a bit unsure about using `warning`, since the output is much less nice than what we currently have (though I suppose there may be a way to improve it).

Marigold:
Yeah, that's precisely why I went for `log.warning` instead. I'll think about this again and reconsider whether it's worth it or not.

pabloarosado:
Closing this, since @Marigold already implemented a better solution in https://github.com/owid/etl/pull/2418

PR #2389, opened 2024-03-08
✨ accept List[str] in Table.sort_values

lucasrodes:
Typing `Table.sort_values` should also list `List[str]` as a possible type for argument `by`.


pabloarosado (approved):
Good catch!

PR #2403, opened 2024-03-11
Update World Bank income groups

pabloarosado:
Update World Bank income groups.

This PR will replace the old datasets:
* [Income Classification - World Bank (2021)](https://owid.cloud/admin/datasets/5741).
* [Income Classification - World Bank (2017)](https://owid.cloud/admin/datasets/944).
* [Income classifications (World Bank, 2022)](https://owid.cloud/admin/datasets/5997).

By the new [World Bank's historical income classification](http://staging-site-update-wb-income-groups/admin/datasets/6406).

For more context: https://github.com/owid/owid-issues/issues/1179

### DIFF
--- a/etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
+++ b/etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
@@ -25,12 +25,12 @@ dataset:
 
 tables:
   income_groups:
-    title: World Bank's historical income classification
+    title: World Bank's income classification
     variables:
       classification:
-        title: World Bank's historical income classification
+        title: World Bank's income classification
         unit: ""
-        description_short: Historical income classification based on the country's income each year.
+        description_short: Income classification based on the country's income each year.
   income_groups_latest:
     title: World Bank's latest income classification
     variables:
--- a/etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
+++ b/etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
@@ -3,16 +3,36 @@ definitions:
     presentation:
       topic_tags:
         - Economic Growth
-    processing_level: minor
+    processing_level: major
+    description_from_producer: |-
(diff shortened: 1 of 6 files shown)


etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
```diff
@@ -1,7 +1,40 @@
...
+  title: World Bank's income classification
+  update_period_days: 0
+
 tables:
   income_groups:
+    title: World Bank's historical income classification
     variables:
       classification:
-        title: World Bank income classification
+        title: World Bank's historical income classification
```
lucasrodes, line 31: I think having "World Bank's income classification" would be fine, without the historical. I find it a bit redundant since all our indicators are "historical" by this criterium.

I understand that the purpose of this word is to differentiate the indicator title from the other one "World Bank's latest income classification", but I think that it is not necessary.

etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
```diff
@@ -1,7 +1,40 @@
...
+
+      Regions in this dataset include economies at all income levels. The term country, used interchangeably with economy, does not imply political independence but refers to any territory for which authorities report separate social or economic statistics. For more information about how the World Bank classifies countries, check [their documentation](https://datahelpdesk.worldbank.org/knowledgebase/articles/378834-how-does-the-world-bank-classify-countries).
+    description_key:
+      - The World Bank creates a yearly classification of countries by income, for all countries with population over 30,000.
+      - This classification stays the same throughout the fiscal year (from July 1 to June 30) even if the income data for a country changes.
+      - Low-income countries are those with a gross national income (GNI) per capita of $1,135 or less in 2022.
+      - Lower-middle-income countries are those with a GNI per capita between $1,136 and $4,465 in 2022.
+      - Upper-middle-income countries are those with a GNI per capita between $4,466 and $13,845 in 2022.
+      - High-income countries are those with a GNI per capita of $13,846 or more in 2022.
+
```
lucasrodes, line 21: Interesting.

I'd have put this in the snapshot description. I can see why you want to put it there. Is that something we want to do more often?

lucasrodes (approved):
LGTM! Thanks for updating this.

etl/steps/data/garden/wb/2024-03-11/income_groups.meta.yml
pabloarosado: I initially had this info only in the snapshot description. Then I copied it to key description. Then I removed it from snapshot description (not sure if this was a good decision). I think it would be ok to have this information in both places, but the key description is certainly more important, as it's surfaced in more prominent places.

PR #2409, opened 2024-03-12
Improve etl dashboard

pabloarosado:
Some small improvements of the ETL dashboard I thought of, as I was using it to update the climate namespace:
* Add a "Details list" with more details about steps selected in the Steps table (e.g. list of updateable dependencies).
* Add more parameters on how to run the ETL.
* Minor speed-up of the code.

@lucasrodes there's no need for a thorough code review, but you can simply run it, to check that the changes make sense, thanks!


lucasrodes (approved):
LGTM! Nice improvements, Pablo! Very cool to see an example right there (in the collapsable area)! I think this tool will be very valuable. Also, the "details list" seems like a useful addition to the app.

Some comments
- The instructions use some colours that are not visible if using "light mode". Could be nice if these worked for both themes.
   image
- The columns `7-day views` and `365-day views` are all zero. Is that something intended? Does it only work from live?
- Very minor: There might be a better way of displaying the details list. A more compact one that works with a large list of datasets. Maybe using a grid (instead of just rows). But I think that is a "very nice to have" and we should just go with the current approach and think of it later on as we use the tool!


(btw, I merged with master, i think nothing should've changed from your code)

pabloarosado:
Hi @lucasrodes thanks for the suggestions! I've made the background color of the buttons in the tutorial transparent, which should work for both light and dark layouts.
The issue with analytics is not related to the dashboard. You need to run `make refresh.pageviews` in your grapher to fetch analytics. This is a bit annoying, but it's a grapher issue.
I agree that the Details list can get very long, which is inconvenient. Let's improve it in a separate PR!

PR #2415, opened 2024-03-14
📊 UN WPP population five year age-groups for 0-100, instead of 0-20

spoonerf:
* Adding 5-year age-groups for population across the whole age range (0-100) - for [this chart](http://staging-site-five-year-pops/admin/charts/7691/edit)
* Removing the new 5-year age-groups from the explorer step, so the explorer is unchanged
* Adding an experimental dataset to show the populous age-group for each country-year - for[ this chart](http://staging-site-five-year-pops/admin/charts/7690/edit)

### DIFF
--- a/etl/steps/data/explorers/un/2022/un_wpp.py
+++ b/etl/steps/data/explorers/un/2022/un_wpp.py
@@ -29,7 +29,33 @@ def _keep_relevant_rows(df: pd.DataFrame) -> pd.DataFrame:
     df = df.reset_index()
     df = df.loc[
         df.variant.isin(["estimates", "low", "medium", "high"])
-        & -(df.metric.isin(["net_migration", "net_migration_rate"]) & (df.location == "World"))
+        & -(
+            df.metric.isin(["net_migration", "net_migration_rate"])
+            & (df.location == "World")
+            & -(df.metric == "population")
+            & (
+                df.age.isin(
+                    [
+                        "20-24",
+                        "25-29",
+                        "30-34",
+                        "35-39",
+                        "40-44",
+                        "45-49",
+                        "50-54",
+                        "55-59",
+                        "60-64",
+                        "65-69",
+                        "70-74",
+                        "75-79",
+                        "80-84",
+                        "85-89",
+                        "90-94",
+                        "95-99",
+                    ]
+                )
+            )
+        )
     ].reset_index(drop=True)
     return df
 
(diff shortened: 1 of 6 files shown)


etl/steps/data/explorers/un/2022/un_wpp.py
```diff
@@ -29,7 +29,33 @@ def _keep_relevant_rows(df: pd.DataFrame) -> pd.DataFrame:
     df = df.reset_index()
     df = df.loc[
         df.variant.isin(["estimates", "low", "medium", "high"])
-        & -(df.metric.isin(["net_migration", "net_migration_rate"]) & (df.location == "World"))
+        & -(
```
lucasrodes, line 32: very minor, I generally prefer using `~` than `-` as the negation operator. I think is better practice, but I just can't remember why : D

etl/steps/data/garden/un/2024-03-14/un_wpp_most.meta.yml
```diff
@@ -0,0 +1,15 @@
+tables:
+  population:
+    variables:
+      age:
+        title: Age-group with the highest population
+        unit: ""
+        description_short: |-
+          Age of the five-year population age-group with the highest population
```
lucasrodes, line 8: I find it a bit redundant to use "age" twice in the sentence.

Maybe something like

> Five-year age group with the highest population.

maybe that's too simple?

Another example:

> The largest five-year population age group.

lucasrodes (approved):
LGTM! Nice that you are working on these charts.

I was wondering, [this chart](http://staging-site-five-year-pops/admin/charts/7690/edit) would really benefit from having brackets instead of individual boxes. Maybe we could have that for ordinal indicators, like this one.

etl/steps/data/explorers/un/2022/un_wpp.py
spoonerf: Ah okay - I will update it to that!

etl/steps/data/garden/un/2024-03-14/un_wpp_most.meta.yml
spoonerf: Good point - yours is much better!

spoonerf:
Thanks @lucasrodes! 

Because the UN WPP dataset has so many downstream dependencies should I wait until this evening to merge it? Also is it worth waiting for the ordinal map brackets to be a feature before merging this?

lucasrodes:
> Because the UN WPP dataset has so many downstream dependencies, should I wait until this evening to merge it?
Good call! Would be nice to have a scheduler.

> Also is it worth waiting for the ordinal map brackets to be a feature before merging this?

Nah! Let's take a note and improve this once that's out.

PR #2418, opened 2024-03-15
:tada: custom warnings module

Marigold:
This was heavily inspired by https://github.com/owid/etl/pull/2388.

## Motivation

Warnings are sometimes unavoidable, or it's just more convenient to ignore them (especially when you are not the maintainer of the data and need to silence them to have readable logs).

## Solution

Implement our own `warnings.py` module with custom warnings that can be selectively ignored. The module lets us use context manager to ignore warnings which is the cleanest way.

```python
from owid.catalog import warnings

with warnings.ignore_warnings([warnings.NoOriginsWarning, warnings.DisplayNameWarning]):
        ds_garden = create_dataset(
            dest_dir, tables=[tb_garden], check_variables_metadata=True, default_metadata=ds_garden.metadata
        )
```

Warnings module also overwrites the formatting function so that warnings are printed by colourful `structlog.warning` rather than the ugly default.

image

I've only updated warnings from tables.py and variables.py, but we can add more warnings on the fly.

This has already been useful in silencing noisy warnings from `data://grapher/lis/2023-08-30/luxembourg_income_study`.

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -16,7 +16,7 @@ import jsonref
 import structlog
 import yaml
 from owid import catalog
-from owid.catalog import CHANNEL, DatasetMeta, Table
+from owid.catalog import CHANNEL, DatasetMeta, Table, warnings
 from owid.catalog.datasets import DEFAULT_FORMATS, FileFormat
 from owid.catalog.meta import SOURCE_EXISTS_OPTIONS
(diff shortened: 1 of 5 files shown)


pabloarosado (approved):
Really nice solution! :-) Thanks for doing this.

PR #2420, opened 2024-03-15
🎉 wizard: add "the Expert"

lucasrodes:
Generalise the "Expert" tool to be able to answer questions on topics covered in our documentation (including metadata).

Some notes:
- I've created "categories" to narrow down the context needed for GPT to answer questions (i.e. system prompt is shorter, hence less expensive). I'm still able to provide GPT with the complete context. For instance, if you want to ask a question on metadata, you can select the category "Metadata". The system prompt used for this chat will be limited to metadata documentation.
- The Expert is now aware of the `variable.presentation.grapher_config` fields.

https://github.com/owid/etl/assets/18101289/bdd12c5a-516d-43ea-b024-9986f5127ac6


pabloarosado (approved):
Hey this is very nice! Thanks for doing this! I just tried it on my local grapher and worked well (although the answers felt a little "chatGPT-generic" instead of coming out of our actual documentation). A few comments/questions:
1. You remove some files from the docs, but where is that documentation?
2. Actually, PR servers survive for 3 days, not 24h. But that seems to be what we currently say in [the documentation](https://docs.owid.io/projects/etl/guides/staging-servers/)! So the answer is the expected one.

lucasrodes:
Thanks for reviewing, @pabloarosado!

On your points:

1. Files removed were not in use:
  - `docs/architecture/metadata/reference.md` is generated dynamically, so it should not be there.
  - `docs/getting-stated`: Were no longer referenced and used in the docs.

2. Good point. Should fix that section!

On generic GPT answers, that's a good point. Let's see if we can improve the system prompts in the future.

PR #2425, opened 2024-03-18
✨  snapshot: read file in archive

lucasrodes:
Sometimes, a snapshot comes as a ZIP/TAR file, which contains a relevant file inside. This PR makes it easier to directly read that relevant file.

**Example:** Suppose you have a file `archive.zip`, which when unzipped contains a file `data.csv`. To load `data.csv` as a table you can now do:

```python
snap = paths.load_snapshot("archive.zip")

# Load data from snapshot.
tb = snap.read_in_zip("data.csv")
```

To be able to reuse code, I've moved some logic out of `Snapshot` class.


Marigold (approved):
Nice!

PR #2428, opened 2024-03-19
🐛 generalise fillna table method

lucasrodes:
Pandas `fillna` _does not_ impose the usage of argument `value`. Therefore, `Table.fillna` should do the same.

Forcing it, makes calls to `.ffill` or `.fillna(how="ffill")` raise errors.

Example of error:

```python
# Load some table
tb = ...
tb.ffill()
```

leads to 

```
TypeError: Table.fillna() missing 1 required positional argument: 'value'
```


pabloarosado (approved):
Hi @lucasrodes thanks for fixing the `ffill` issue! Please consider adding some unit tests, given that this is a very important and frequently used method.

PR #2437, opened 2024-03-21
🎉  add Table.format()

lucasrodes:
Instead of 

```python
tb.underscore().set_index(["country", "year"], verify_integrity=True).sort_index()
```

use the now equivalent

```python
tb.format(["country", "year"])
```

or simply

```python
tb.format()
```


pabloarosado (approved):
Hey, nice, I've often thought about how often we repeat these commands. The code looks good.

I'm wondering how useful this is going to be as it is right now. For example, in meadow, I like to explicitly write `set_index(tb, verify_integrity=False)` when I know that the index is ambiguous. I think that's useful. Also, I often add `sort_index(axis=1)`. Or sometimes I may not want to sort at all.

So, I'm not sure if it's better to:
(a) close this PR without merging, and we just keep being explicit, using `underscore().set_index(...)...`, which usually is just one line of code, or
(b) add more arguments to `formatting()`, to cover for all those cases. The default arguments could be `index_columns=None, verify_integrity=True, underscore=True, sort_rows=True, sort_columns=True, **kwargs`.

What do you think?

lucasrodes:
I think we should go with b), I don't see why not. One can continue to explicitly use the current approach. So I don't think that going for a) makes sense.

I think that while there is a wide range of ways we use these three methods, there is one that is the most common one. Hence, having something for that will be useful and helpful.

lucasrodes:
@pabloarosado thanks for the proposal.

I would discard option 3.

Option 1 and 4 are essentially very similar. I think that the downside that you mention with option 1 (one could assume by default `formatting` sorts both columns and rows) is also in option 4.

We can try to go with option 2 and see. I'll ask the rest of the team.

lucasrodes:
@pabloarosado I avoided using format bc it is a keyword in python. I agree that 'formatting' is not the best. I'm flexible on this one though.

lucasrodes:
Based on the discussion on Slack, I am leaning towards having `sort_columns` set to `False` by default. I think that it is often the case that the order of the columns makes sense.

We could of course make an editorial decision and have columns _always_ sorted alphabetically. But I think it is safer for now to go with either solution 1 or 4 (from the ones you proposed).

PR #2441, opened 2024-03-22
Add documentation about tags limitation of the chart sync tool

pabloarosado:
As discussed [here](https://owid.slack.com/archives/C0193RW5E2J/p1711101052248769). Note that Mojmir had already added a comment in the instructions of the chart sync (but it was a bit hidden, so now that info is more evident).

### DIFF
--- a/apps/staging_sync/app.py
+++ b/apps/staging_sync/app.py
@@ -25,7 +25,14 @@ def main():
     directly to live. This tool allows you to first create charts on the staging server and then synchronize
     them to the live environment after merging.
 
-    ⚠️ Tags are synced only for **new charts** in the source server, any edits to tags in existing charts are ignored.
+    """
+    )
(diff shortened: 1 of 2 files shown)


apps/staging_sync/app.py
```diff
@@ -24,6 +24,8 @@ def main():
     before creating charts in the live environment. In some cases, it was even necessary to push your dataset
     directly to live. This tool allows you to first create charts on the staging server and then synchronize
     them to the live environment after merging.
+
+    ⚠️ Tags are synced only for **new charts** in the source server, any edits to tags in existing charts are ignored.
```
lucasrodes, line 28: We could have it as a `st.warning` or `st.info` as well.

lucasrodes (approved):
minor suggestions

PR #2443, opened 2024-03-23
📊 AI Papers with code - 2024 update

veronikasamborska1994:


etl/steps/data/garden/artificial_intelligence/2024-03-22/papers_with_code_coding.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Save outputs.
+    # Create a new garden dataset with the same metadata as the meadow dataset.
+    ds_garden = create_dataset(dest_dir, tables=[tb_combined], default_metadata=ds_meadow.metadata)
+
+    # Save changes in the new garden dataset.
+    ds_garden.save()
+
+
+def group_sort_filter(tb, column):
```
Marigold, line 40: You could put this into `shared.py` file and import it from all garden datasets to avoid code duplication.

etl/steps/data/grapher/artificial_intelligence/2024-03-22/papers_with_code_coding.py
```diff
@@ -0,0 +1,37 @@
+"""Load a garden dataset and create a grapher dataset."""
```
Marigold, line 1: Similarly here - you could have `shared.py` with method

```
def run(dest_dir: str, dataset_name: str):
  ...
```
and then call from every dataset
```
from . import shared
run = shared.run("papers_with_code_coding")
```

Marigold (approved):
Looks good!

etl/steps/data/grapher/artificial_intelligence/2024-03-22/papers_with_code_coding.py
veronikasamborska1994: yeaah sorry I was being lazy :D but done it now!

Issue #2238, opened 2024-01-31
Apply a strict schema to YAML metadata files


larsyencken:
## Problem

We would like to ensure that our YAML and `.meta.json` files have the metadata fields that we expect, and we would like to guard against accidental errors like [using the wrong type](https://github.com/owid/etl/issues/2064) for a field or having a typo in a field name.

## Proposed solution

Something along these lines would lock down our YAML in a helpful way:

1. Set up a schema for our current metadata standards
2. Allow YAML files to opt-in to that schema with a `$schema` parameter
3. Validate YAML files in CI against any schema they opt-in to
4. Add the `$schema` parameter and opt-in historical YAML files that are already compliant
5. Have CI check for compliance for datasets created after a certain date (i.e. force them to opt in)

alhridoy:
Hi @larsyencken could you advise which files you think would be good starting points to focus on for adding YAML schema validation?

Some areas I was considering are:

schemas/: To add the new JSON validation schema. I can start with the existing metadata standard.
etl/: Where we could integrate the validation during metadata parsing.
tests/: To add new test cases.

Thanks

larsyencken:
Mildly inclined to call it important to lock down our metadata like this, to prevent edge case errors.

lucasrodes:
Just as an FYI, and in case it affects other parts of the project: We currently have schemas in [`schemas/`](https://github.com/owid/etl/tree/master/schemas), which power:

- [Our documentation on Metadata](https://docs.owid.io/projects/etl/architecture/metadata/reference/). 
  - The documentation is rendered using [this script](https://github.com/owid/etl/blob/master/docs/ignore/generate_dynamic_docs.py).
- Our [Meta Expert](http://etl.owid.io/wizard/Meta%20Expert) tool to answer any questions on metadata.
  - It relies on the same functions as our documentation generation above. See [`apps/wizard/pages/meta_expert/prompts.py`](https://github.com/owid/etl/blob/master/apps/wizard/pages/meta_expert/prompts.py), which uses functions from [`etl.docs`](etl/docs.py).
- The hints and helper texts on Wizard when creating new steps (see screenshot below). These are rendered in the individual python scripts (e.g. [`apps/wizard/etl_steps/snapshot.py`](apps/wizard/etl_steps/snapshot.py) for snapshot). These also use functions from [`etl.docs`](etl/docs.py).
[...]

PR #2455, opened 2024-03-26
:tada: strict YAML metadata schema validation

Marigold:
Light implementation of https://github.com/owid/etl/issues/2238.

The original issue suggests to
> Add the $schema parameter and opt-in historical YAML files that are already compliant

In my opinion, adding a property to every YAML file would add too much boilerplate code. I'm not sure if we need such a granular approach. Similarly with versioning metadata schema. I don't think we have to check every metadata file at each point in time. If we always have a green CI, then bump the version in `VALIDATE_AFTER`, update schema and have green again, then it's sufficient. (Especially since we have linting enabled)

I fixed a couple of datasets, but added some to `EXCLUDE` list. I've also disabled checking for `description` as it is present in too many datasets (we still get linting error for it).

### DIFF
--- a/etl/steps/data/garden/covid/latest/variants.meta.yml
+++ b/etl/steps/data/garden/covid/latest/variants.meta.yml
@@ -1,6 +1,7 @@
-default_display: &default_display
-  yearIsDay: true
-  zeroDay: '2020-01-21'
+definitions:
+  default_display: &default_display
+    yearIsDay: true
+    zeroDay: '2020-01-21'
 
 dataset:
   title: COVID-19 - Variants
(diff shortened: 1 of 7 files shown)


Marigold:
> Eventually it would be good to check all .meta.json files, but I suppose that may be more complicated.

It would, though, maybe not be as complicated. The question is how valuable it is to do a full check. Since the data goes through objects like `Dataset` or `Variable`, there's already some validation (also you have type hints). There are some special cases like `description_key`, but it might be enough to have a special runtime validation for them (such as the one we added recently). I'd start with YAML validation and wait if it's enough or not.

Issue #2451, opened 2024-03-25
Chart sync tool requires staging-created-at argument


pabloarosado:
The chart sync tool in the wizard fails with an error that sounds as if the branch you are trying to sync did not exist. However, the real issue seems to be that the argument `staging-created-at` is required.

The workaround is to use the cli and pass that argument.

So I suppose that, either there is some bug in the backend (and the staging-created-at argument is not really necessary), or the wizard should ask the user to manually input that date before executing the chart sync.

This is the error I was getting now when trying to sync `update-faostat-data` to live:

```
2024-03-25 16:47:55,655 - [bugsnag] WARNING - No API key configured, couldn't notify
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in call
    return self.main(*args, **kwargs)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/rich_click/rich_command.py", line 126, in main
    rv = self.invoke(ctx)
    ^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/apps/staging_sync/cli.py", line 124, in cli
    staging_created_at = _get_staging_created_at(source, staging_created_at)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #2457, opened 2024-03-26
Chart sync CLI fails because of a `requests.exceptions.ConnectionError`


pabloarosado:
When running `etl chart-sync update-faostat-data .env.live --staging-created-at 2024-03-14` I get a `requests.exceptions.ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=80): Max retries exceeded with url: /admin/api/charts (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x16866b610>: Failed to establish a new connection: [Errno 61] Connection refused'))`.

NOTES:
1. I'm running this script from my local computer, with a SSH tunnel to production.
2. I temporarily replaced the `assert target_chart.config["isPublished"]` with a warning, because it was failing for chart 1031 (which is another issue: https://github.com/owid/etl/issues/2452 ).
3. When running it with `--dry-run` it works.

Full log:
```
(etl-py3.11) ➜  etl git:(master) ✗ etl chart-sync update-faostat-data .env.live --staging-created-at 2024-03-14
[...]

Issue #2452, opened 2024-03-25
Chart sync fails because of missing `isPublished`


pabloarosado:
When running `etl chart-sync update-faostat-data .env.live --staging-created-at 2024-03-14 --dry-run`, I get a `KeyError`.

<details><summary>Full log:</summary>
(etl-py3.11) ➜  etl git:(master) etl chart-sync update-faostat-data .env.live --staging-created-at 2024-03-20 --dry-run
2024-03-25 17:49:31 [info     ] staging_sync.start             chart_ids={1024, 1025, 1538, 1027, 1029, 1030, 1031, 7685, 5129, 1035, 525, 1037, 1038, 1039, 1040, 2576, 5138, 533, 534, 1046, 1559, 1562, 2086, 4652, 4653, 558, 1070, 1072, 1073, 571, 4156, 1088, 1089, 1090, 4675, 4676, 4677, 4678, 4679, 4680, 4681, 6216, 589, 590, 3676, 1127, 3175, 3176, 2673, 1149, 1154, 5624, 5116, 5625, 5626, 2713, 3738, 3236, 4774, 6550, 715, 2766, 4823, 2777, 2778, 2779, 2780, 2782, 2784, 1249, 1250, 2785, 2786, 2787, 2788, 743, 2789, 745, 746, 2790, 2791, 2792, 2793, 2794, 2795, 2796, 2797, 4848, 768, 787, 5348, 802, 823, 826, 5949, 4926, 841, 6989, 847, 848, 852, 853, 855, 856, 863, 864, 6500, 872, 4970, 4971, 4973, 878, 4974, 881, 886, 888, 3961, 3450, 891, 893, 5503, 4992, 3970, 3971, 4995, 5508, 3974, 5509, 3977, 3978, 3979, 3980, 3981, 3982, 3984, 1940, 1941, 1942, 1943, 1944, 3993, 1946, 3994, 1948, 1949, 3995, 3996, 3999, 4000, 4001, 4002, 4003, 4004, 4005, 4006, 4007, 4008, 4010, 939, 4011, 4012, 4013, 4014, 4015, 4016, 4017, 5550, 949, 951, 4539, 7108, 7110, 974, 975, 5584, 5075, 980, 3032, 3033, 4589, 4590, 1012, 1013, 1014, 1015, 1016, 1017, 1018, 1019, 1020, 1022, 1023} n=192
[...]

PR #2458, opened 2024-03-26
:bug: fix chart-sync issues

Marigold:
Fixes https://github.com/owid/etl/issues/2451, https://github.com/owid/etl/issues/2452 and possibly https://github.com/owid/etl/issues/2457.


pabloarosado (approved):
Thanks for fixing this!

PR #2461, opened 2024-03-27
:tada: automatic wildfires update

Marigold:
Adjustments needed to update wildfires automatically. The script will run `python snapshots/climate/latest/weekly_wildfires.py` and then run ETL to update wildfires dataset.

There are couple of minor differences to the current process:
- Data will be updated daily (nothing happens if there are no changes)
- Description short slightly changes its date format to `The 2024 data is incomplete and was last updated 2024-03-27.`

Is it safe to run it daily or does it have to be run once a week?

## TODO after merging

- [x] Enable wildfires in [Buildkite](https://buildkite.com/our-world-in-data/etl-automatic-dataset-updates-master/settings/steps)


veronikasamborska1994 (approved):
LGTM! I see no reason why you couldn't run it daily but I guess we'll find out :)

PR #2467, opened 2024-03-27
✨owidle: add column headers

tristannew:
I have been enjoying exploring the etl tool and the owidle game lately. I have been sitting on these changes locally and thought to push them in case they are useful 🙂 

They simply add headers to a users guesses to indicate what each column represents, as well as additional info in some cases about how things such as the score are calculated. Any feedback is really appreciated!

P.S. I am aware that the public cannot contribute datasets - I am not sure if this policy/preference extends to contributions like this. If so, feel free to close this PR and let me know!


lucasrodes:
hi @tristannew! 
Thanks for playing with `owidle`, and even more for proposing some improvements! I've looked at your code, and overall it looks good to me! Thanks for the suggestions 😃 

There is only one thing that makes me hesitant to merge it, and is the fact that the header titles are not aligned with the user inputs. I am not sure how to fix this at this moment. This comes from the fact that the header container has no border. 

One solution would be to put a border for the header row, but then I find it visually hard to distinguish it from the user's guesses. Maybe there is some way to tweak the CSS style of the header container passing some HTML to `st.markdow(..., unsafe_allow_html=True)`?

PR #2468, opened 2024-03-28
🐝 Update pandas to 2.2

Marigold:
Update pandas to 2.2.1. To be merged after https://github.com/owid/etl/pull/2468

I've extensively checked all datasets with
```
etl diff REMOTE data/ --include garden -e "faostat|papers_with_code_benchmarks|ihme_gbd|un_sdg|world_bank_pip|mortality_database|gho|sedlac|un_wpp" --verbose --workers 3
```
A few datasets have very minor changes. Otherwise, everything matches.

The new version introduces tons of new `Future Warnings`. I've tried to silence the most annoying ones (especially in core ETL code), but we still get tons of them. We should fix them on the fly.

Pandas 2.x introduces pyarrow data types (`pyarrow.float64`) in addition to standard numpy types (`np.float64`) and nullable types (`pd.Float64`). We're **not using them anywhere** yet. It'd be possible to load dataset from feather format in native pyarrow type and then work with it, but that's out of scope.

## Changes
- Increase epoch to rebuild all datasets
- `pivot_table` needs `observed=False`, otherwise raises a warning (it will be `observed=True` by default, I left it as `observed=False` for backward compatibility, even though using `observed=True` should be preferred for new code)
- `read_sql(q, engine)` started raising warnings, I created an utility function `read_sql` in `db.py`
- `DataFrame.append` has been replaced by `pd.concat
- `Series.fillna(method='ffill')` has been replaced by `Series.ffill`
- Fixed edge cases in our custom `.groupby`


pabloarosado (approved):
Massive! Thanks a lot for taking care of this. I have gone through all changes and I haven't spotted anything odd. The only two things that worry me a bit are:
1. The strange issue with the tables `merge` function. I haven't looked into it, but the comment scared me a bit (given how important that function is).
2. Your choices of `observed=False` and `True`. I can't think of any time where `observed=False` was useful. So I'm surprised that you used `False` most of the times. For example, in `etl/steps/data/explorers/wash/2024-01-06/who.py`, using `True` or `False` yields the exact same table, and there are no warnings (you can check that `tb.pivot_table(index=["country", "year"], columns="residence", values=cols, aggfunc="first", observed=True).equals(tb.pivot_table(index=["country", "year"], columns="residence", values=cols, aggfunc="first", observed=False))` is `True`), however, you set it to `False`. So, if you have checked that datadiff comes clean, then it's all good, I was just a bit confused about these choices.

Please let me know if there's anything else you'd like me to check here.

Marigold:
> I haven't looked into it, but the comment scared me a bit (given how important that function is).

Oh, I wrote it in a really scary tone :). Copying metadata fixes the problem, it's just that I haven't been able to write a failing unit test for it. (There's some low-level magic happening under the hood, luckily copying metadata is enough)

> I can't think of any time where `observed=False` was useful

I left it there because `observed=False` was the default in pandas 1.x and I wanted to stay as compatible as possible. On second thought, there's no reason to keep it like that. I'll remove it.

PR #2484, opened 2024-04-02
📊 Add data for organ donation and transplantation

pabloarosado:
Add snapshot, meadow, garden and grapher steps for data from the Global Observatory on Donation and Transplantation.


etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
```diff
@@ -0,0 +1,64 @@
...
+
+dataset:
+  update_period_days: 365
+
+
+tables:
+  organ_donation_and_transplantation:
+    variables:
+      n_organ_donors:
+        title: Total number of actual deceased organ donors
```
spoonerf, line 20: What does 'actual deceased' mean here? Maybe something for `description_short`

pabloarosado: Actual means donors from whom organs are taken, as opposed to potential donors.

pabloarosado:
Hi @spoonerf, thanks for the review and the suggestions. I just added the dataset so it's available for inspection. We can add more indicators (or regions) if needed later on.

PR #2485, opened 2024-04-02
📊 Add data on organ donors per million people

pabloarosado:
Add indicators per million people.

### DIFF
--- a/etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
+++ b/etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
@@ -133,4 +133,6 @@ tables:
       n_small_bowel_transplantation_per_million_people:
         title: Total number of small bowel transplantations per million people
         processing_level: major
-
+      population:
+        title: Population
+        processing_level: major
--- a/etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
+++ b/etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
@@ -64,3 +64,73 @@ tables:
         title: Total number of kidney-pancreas transplantations
       n_small_bowel_transplantation:
         title: Total number of small bowel transplantations
+      # Add indicators per million people.
+      n_organ_donors_per_million_people:
+        title: Total number of actual deceased organ donors per million people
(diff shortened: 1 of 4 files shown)


pabloarosado:
@spoonerf sorry for bothering you again with another PR, I decided to also quickly add indicators per million people, as you suggested :-).

etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
```diff
@@ -64,3 +64,75 @@ tables:
         title: Total number of kidney-pancreas transplantations
       n_small_bowel_transplantation:
         title: Total number of small bowel transplantations
+      # Add indicators per million people.
+      n_organ_donors_per_million_people:
```
spoonerf, line 68: Perhaps good to have the numDecimalPlaces for the per million indicators as 1

etl/steps/data/garden/health/2024-04-02/organ_donation_and_transplantation.meta.yml
```diff
@@ -64,3 +64,75 @@ tables:
...
+      n_pancreas_transplantation_per_million_people:
+        title: Total number of pancreas transplantations per million people
+        processing_level: major
+      n_kidney_pancreas_transplantation_per_million_people:
+        title: Total number of kidney-pancreas transplantations per million people
+        processing_level: major
+      n_small_bowel_transplantation_per_million_people:
+        title: Total number of small bowel transplantations per million people
+        processing_level: major
+      population:
```
spoonerf, line 136: Nit but you could probably drop this

spoonerf (approved):
Looks good! I am curious how complete this data is. Is it the case that these are reported transplants, and there could be many more non-reported ones? Or do you get the sense that most transplants that occur are captured in this dataset? It might be good to note this somewhere, if possible.

pabloarosado:
> Looks good! I am curious how complete this data is. Is it the case that these are reported transplants, and there could be many more non-reported ones? Or do you get the sense that most transplants that occur are captured in this dataset? It might be good to note this somewhere, if possible.

Thank you for the second review! I don't have a good intuition on how complete this dataset is, but a priori [the methodology](https://www.transplant-observatory.org/methodology/) made the dataset sound reasonably complete and trustworthy.

PR #2494, opened 2024-04-08
:sparkles: Improve datadiff

Marigold:
Improve datadiff based on my experience with migrating to pandas 2.x. I've simplified the output, improved performance and handled edge cases.

There's also some prep for pandas 2.x migration to make the subsequent review easier.

## Datadiff

You can now use `etl diff dataset_path_a dataset_path_b` which use particularly useful for comparing versions. For instance
```
etl diff data/garden/climate/2024-01-31/sea_ice_index data/garden/climate/2024-03-11/sea_ice_index --verbose
```
The output has also changed. Instead of showing confusing "changed index" messages, I just show new/removed/updated values with a random sample of 5 values. For instance
```
 = Table sea_ice_index
    ~ Dim location
        + New values: 4 / 1084 (0.37%)
                date            location
          2024-01-15 Northern Hemisphere
          2024-02-15 Northern Hemisphere
          2024-01-15 Southern Hemisphere
          2024-02-15 Southern Hemisphere
    ~ Dim date
        + New values: 4 / 1084 (0.37%)
                     location       date
          Northern Hemisphere 2024-01-15
          Northern Hemisphere 2024-02-15
          Southern Hemisphere 2024-01-15
          Southern Hemisphere 2024-02-15
    ~ Column sea_ice_extent (changed metadata, new data)
        -     date_accessed: '2024-01-31'
        ?                            --
        +     date_accessed: '2024-03-11'
        ?                           ++
        -     date_published: '2024-01-30'
        ?                             --
        +     date_published: '2024-03-10'
        ?                            ++

        + New values: 4 / 1084 (0.37%)
                     location       date  sea_ice_extent
          Northern Hemisphere 2024-01-15          13.916
          Northern Hemisphere 2024-02-15          14.607
          Southern Hemisphere 2024-01-15           3.961
          Southern Hemisphere 2024-02-15           2.142
```
(dimensions are unnecessarily repeated twice, but the code would be less elegant if I fixed this minor issue)

Datadiff is still a work in progress, though. We can improve the output based on our needs in the future. (Having specific use cases would be useful)

## Pandas 2.x preps

* Add utility function for `read_sql` and call it instead of `pd.read_sql` (it silences warnings and creates connection from engine)
* Fix edge cases for repack

### DIFF
--- a/apps/backport/backport.py
+++ b/apps/backport/backport.py
@@ -20,7 +20,7 @@ from apps.backport.datasync.datasync import upload_gzip_dict
 from etl import config, paths
 from etl import grapher_model as gm
 from etl.backport_helpers import GrapherConfig
-from etl.db import get_engine
+from etl.db import get_engine, read_sql
 from etl.files import checksum_str
 from etl.snapshot import Snapshot, SnapshotMeta
 
@@ -346,7 +346,7 @@ def _load_values(engine: Engine, variable_ids: list[int]) -> pd.DataFrame:
             "entityCode": "entity_code",
         }
     )
-    vf: pd.DataFrame = pd.read_sql(q, engine, params={"variable_ids": variable_ids})
+    vf = read_sql(q, engine, params={"variable_ids": variable_ids})
     df = df.merge(vf, on="variable_id")
 
     # try converting values to float if possible, this can make the data 50% smaller
--- a/apps/backport/bulk_backport.py
+++ b/apps/backport/bulk_backport.py
@@ -9,7 +9,7 @@ from owid.catalog.utils import underscore
 from sqlalchemy.engine import Engine
 
 from etl import config
-from etl.db import get_engine
+from etl.db import get_engine, read_sql
 from etl.snapshot import snapshot_catalog
 from etl.steps import load_dag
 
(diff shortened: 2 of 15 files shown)


pabloarosado:
Great work! Thanks for doing this. I think it would be good to update the documentation, since this new use case of comparing versions is not explained.

I tried the example you gave above and it worked well. However, I also tried `etl diff data/garden/faostat/2023-06-12/faostat_qcl data/garden/faostat/2024-03-14/faostat_qcl` and it printed at the end `No differences found`. But there clearly are differences between these two datasets. Maybe I'm misusing or misinterpreting something.

pabloarosado:
Oh, I think what's happening is that the first step `data/garden/faostat/2023-06-12/faostat_qcl` is not found. I suppose it would be good to make this case clearer: "Step not found", instead of printing "No differences found".

pabloarosado (approved):
The code looks good, although I haven't looked carefully enough to be able to follow all the details. I tried it on some examples and works as expected. The only issue I found is that, as I mentioned before, if you pass two steps, and only one is found, the output is "No differences found" which can be misleading. But feel free to merge and we can fix that in a future PR.

Marigold:
Good catch, fixed that. I'll add a section to docs when I finish all Datadiff changes that are planned.

Issue #2499, opened 2024-04-08
Snapshot: Getting an error when uploading a local file


spoonerf:
I'm getting this error when I try and upload a local csv to snapshot. I'll try it with a simple test csv to check it is not the csv that is problematic.

> .venv) fionaspooner@Fionas-MBP etl % python snapshots/who/2024-04-08/polio_afp.py
> Path to file: /Users/fionaspooner/Desktop/who_afp.csv
> 2024-04-08 16:09:23,799 - [bugsnag] WARNING - No API key configured, couldn't notify
> WARNING:bugsnag:No API key configured, couldn't notify
> Traceback (most recent call last):
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/snapshots/who/2024-04-08/polio_afp.py", line 25, in <module>
>     main()
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
>     return self.main(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
>     rv = self.invoke(ctx)
>          ^^^^^^^^^^^^^^^^
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
>     return __callback(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/snapshots/who/2024-04-08/polio_afp.py", line 21, in main
>     snap.create_snapshot(filename=path_to_file, upload=upload)
>   File "/Users/fionaspooner/Documents/OWID/repos/etl/etl/snapshot.py", line 175, in create_snapshot
>     add_snapshot(uri=self.uri, filename=filename, dataframe=data, upload=upload)
[...]

PR #2500, opened 2024-04-08
🔧  Attempting to fix ruamel issue

spoonerf:
Attempting to fix the ruamel issue (#2499) I was having, it appears to work (snapshot now runs) but I am not really sure why :)

### DIFF
--- a/etl/files.py
+++ b/etl/files.py
@@ -196,6 +196,7 @@ def ruamel_dump(d: Dict[str, Any]) -> str:
 
 def ruamel_load(f: io.TextIOWrapper) -> Dict[str, Any]:
     yaml = YAML(typ="rt")  # Create a YAML object with round-trip type
+    yaml.preserve_quotes = True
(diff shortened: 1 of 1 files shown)


etl/files.py
```diff
@@ -194,7 +195,8 @@ def ruamel_dump(d: Dict[str, Any]) -> str:
 
 
 def ruamel_load(f: io.TextIOWrapper) -> Dict[str, Any]:
-    return ruamel.yaml.load(f, Loader=ruamel.yaml.RoundTripLoader, preserve_quotes=True)
+    yaml = YAML(typ="rt")  # Create a YAML object with round-trip type
```
Marigold, line 198: Looks good! I'd just change it like this to keep it compatible
```
yaml = YAML(typ="rt")  # Create a YAML object with round-trip type
yaml.preserve_quotes = True
return yaml.load(f)  # Load the content using the new API
```

Marigold (approved):

PR #2506, opened 2024-04-10
🐛 Fix faostat animals data

pabloarosado:
## Summary
In FAOSTAT QCL dataset, the item "Meat, Total" is empty for the element "Producing Animals/Slaughtered". We therefore aggregate all items to generate that number. This PR makes that estimate more accurate.

## Details
The exact list of items that make up "Meat, Total" is not clear, and the names and descriptions of FAO items have changed over time. But I think the list we used to have was inaccurate (note that we included "Meat, lamb and mutton" as well as "Meat, sheep and goat"). It's not clear how we came up with this list, but it's probably the result of multiple changes in FAO item names, combined with our own renaming of items.

This PR uses another list that seems to be more accurate. It's taken from [FAOSTAT's definitions](https://www.fao.org/faostat/en/#definitions). I've done some offline checks to ensure there is no overlap between the items. I can't do that check with the number of slaughtered animals (since the data doesn't exist), but I checked that the aggregate of "Production" (in tonnes of meat) for this new list of items agrees with the production of "Meat, Total", within 1%, except for Bulgaria, Romania, Poland, Hungary, Czechia, Croatia, where the agreement is within 10% (except an outlier for Romania).

Overall, the resulting number of slaughtered animals for the World is almost unchanged (going from 83.86 billion to 83.30 billion animals in 2022). However, for some countries like New Zealand, the difference is significant. In this case, it may be due to double-counting sheep meat products. In other cases, like Bolivia, there was an increase in the number of slaughtered animals.
new_zealand_slaughtered_animals
bolivia_slaughtered_animals


etl/steps/data/garden/faostat/2024-03-14/faostat_qcl.py
```diff
@@ -31,18 +30,34 @@
 ITEM_CODE_MEAT_POULTRY = "00001808"
 # Item code for 'Meat, chicken'.
 ITEM_CODE_MEAT_CHICKEN = "00001058"
-# List item codes to sum as part of "Meat, total" (avoiding double-counting items).
+# Given that the number of slaughtered animals to produce all meat is not provided, we estimate it by aggregating the
+# number of slaughtered animals for each meat item.
+# List item codes to sum as part of "Meat, Total" (avoiding double-counting items).
+# This list can be found in https://www.fao.org/faostat/en/#definitions
+# by clicking on "Item Group" and selecting Item Code 1765 ("Meat, Total" with Domain Code "QCL").
```
spoonerf, line 37: Not a big deal but I'm not able to find the list following these instructions

spoonerf (approved):
Looks good to me!

PR #2511, opened 2024-04-10
📊 Update CO2 dataset

pabloarosado:
Update owid_co2 to use the latest Jones et al. (2024) data.


paarriagadap (approved):
Looks solid to me! I would only rebase to master to see if there is any trouble with the pandas update. I run it first and I had problems, but then I realized this PR is using an older version of master, so with poetry install I could run it

Issue #2497, opened 2024-04-08
Use only ingredients to compute checksums


Marigold:
ETL currently combines ingredients and **data** when computing checksums. Checksumming data could be inconsistent between systems (local vs server). When doing an `etl diff`, we would like to avoid having to download the data file; if we used ingredients-only checksums, we could rely on its metadata file instead.

```mermaid
graph LR

A.data --> B.data
B.py --> B.data
C.py --> C.data
B.data --> C.data
```

The dataset should be determined exclusively by its ingredients (data checksum is redundant). So a simpler alternative is only to include **ingredients** in checksum calculation.
```mermaid
graph LR

A.data --> B.py --> C.py --> D.py
```
Excluding data from checksum could cause problems in some edge cases, but we haven't found any real ones (perhaps editing data manually?).

Marigold:
Fixed by https://github.com/owid/etl/pull/2514

PR #2514, opened 2024-04-11
:sparkles: compute checksums from ingredients only

Marigold:
Implements https://github.com/owid/etl/issues/2497

The big drawback is that this will rebuild the entire ETL again.

## TODO:

- [x] Does it speed up `etl run --dry-run`?
  - Only a bit (we still calculate checksums of snapshots on disk to make sure they match checksum in their `*.dvc` file)
- [x] Does it help with datadiff checksum comparison?
  - Yes! This is a major blocker for datadiff right now. It'd speed it up by a lot.

### DIFF
--- a/apps/owidbot/etldiff.py
+++ b/apps/owidbot/etldiff.py
@@ -55,6 +55,10 @@ def cli(
 
     nbranch = _normalise_branch(branch) if branch else "dry-run"
 
+    # TODO: only include site-screenshots if the PR is from owid-grapher. Similarly, don't
+    # run etl diff if the PR is from etl repo.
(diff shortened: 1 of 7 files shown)


Marigold:
@pabloarosado I'm searching for a good moment to merge this as this will rebuild the entire ETL, again. Maybe we could merge it the weekend before our offsite? Or is it so annoying that we should park it until it's absolutely necessary?

pabloarosado:
> @pabloarosado I'm searching for a good moment to merge this as this will rebuild the entire ETL, again. Maybe we could merge it the weekend before our offsite? Or is it so annoying that we should park it until it's absolutely necessary?

I still don't fully understand the benefits of this change, but we could do it on that weekend, yes.

pabloarosado (approved):

PR #2525, opened 2024-04-15
✨  Adding WHO regions to regions.yml

spoonerf:
I'm not sure if the region_type should be continent?


etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -1847,3 +1847,221 @@
   end_year: 1902
   successors:
     - "ZAF"
+
+# WHO regios
```
pabloarosado, line 1851: Typo

etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -1847,3 +1847,221 @@
...
+- code: WHO_AMR
```
pabloarosado, line 1852: I'd add `defined_by: who` to all entries. That way, we have a field to distinguish between our and other institutions' definitions.

etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -1847,3 +1847,221 @@
...
+  name: "WHO Americas region"
+  region_type: "aggregate"
```
pabloarosado, line 1854: Yes, I think "aggregate" is probably a good choice.

etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -1847,3 +1847,221 @@
   end_year: 1902
   successors:
     - "ZAF"
+
+# WHO regios
+- code: WHO_AMR
+  name: "WHO Americas region"
```
pabloarosado, line 1853: I'm not sure about this name. This is the name that we will use everywhere, including charts. I'd suggest removing "region" from the names (since it's redundant). And I'm wondering whether the name should be, e.g. "Americas (WHO)", for consistency with our current naming conventions. What do you think?

pabloarosado (changes requested):
Thanks for including these region definitions! I haven't checked the members, but apart from that it looks good. I requested some changes, please let me know if you disagree.

pabloarosado:
@Marigold will this PR trigger a big ETL run, or is the `regions` dataset treated differently (I have a memory that this could be the case, because region aliases often change). Also, can you think of any other unexpected side effects from these changes? Thanks!

etl/steps/data/garden/regions/2023-01-01/regions.yml
```diff
@@ -1847,3 +1847,221 @@
...
+    - "IND"
+    - "IDN"
+    - "MDV"
+    - "MMR"
+    - "NPL"
+    - "THA"
+    - "TLS"
+    - "LKA"
+- code: WHO_WPAC
+- name: "WHO Western Pacific region"
```
pabloarosado, line 2038: Here there is an issue: I think you need to remove the extra "-" on the left of "name". Otherwise the previous item is stored as a region without a name, which causes other problems (I think that's why `nuclear_weapons_proliferation` step is failing [in buildkite](https://buildkite.com/our-world-in-data/etl-automated-staging-environment/builds/4049#018ee124-61e7-45c3-a6c7-9b859ff59369)).

pabloarosado (changes requested):

Marigold:
>  @Marigold will this PR trigger a big ETL run, or is the regions dataset treated differently (I have a memory that this could be the case, because region aliases often change). Also, can you think of any other unexpected side effects from these changes? Thanks!

Currently, any change to regions will trigger a big ETL run. There's an issue for ignoring aliases, but that wouldn't help anyway. I don't see any side effects.

Let me know when you merge this. I'll throw in [checksum updates](https://github.com/owid/etl/pull/2514) that would trigger ETL too.

pabloarosado:
I see from [datadiff](https://github.com/owid/etl/pull/2525#issuecomment-2056721655) that the changes in this PR are actually affecting other steps in unexpected ways. If it's just a few, we could just manually fix them. If it's many, we could create a new version of the `regions` dataset.

Marigold:
> I see from https://github.com/owid/etl/pull/2525#issuecomment-2056721655 that the changes in this PR are actually affecting other steps in unexpected ways. If it's just a few, we could just manually fix them. If it's many, we could create a new version of the regions dataset.

I'm looking at them. My hypothesis is that they happen because there have been some changes in master not merged into this branch. I'm working on some fix.

Marigold:
If we add `defined_by: who` to all entries, then datadiff would show no changes and we could merge this (after working hours).

etl/steps/data/garden/regions/2023-01-01/regions.yml
spoonerf: Good point, I think your suggestion to be consistent with our current naming conventions makes sense, I'll change it to that.

spoonerf:
Thanks @Marigold and @pabloarosado! I've made your suggested changes now - is it okay to merge after hours?

pabloarosado (approved):
Looks good!

pabloarosado:
@Marigold is there any way to show datadiff results again? @spoonerf if nothing else changes, then yes, it would be good to merge.

Marigold:
> @Marigold is there any way to show datadiff results again? @spoonerf if nothing else changes, then yes, it would be good to merge.

The owidbot message gets automatically updated (check the _Edited:_ at the bottom). It just might take more time with so many changes.

spoonerf:
Thanks! The datadiff looks okay to me, although there is an error at the bottom but I think that's unrelated. I'll merge after hours this evening.

PR #2526, opened 2024-04-16
:sparkles: add command sync.catalog for prefetching catalog from R2

Marigold:
Add command `sync.catalog` that fetches main channels from ETL catalog to local with `rclone`. The entire catalog has ~10gb, so this could be much faster than rebuilding it locally from scratch (~15 min for me). This is particularly useful when `ETL_EPOCH` is incremented or regions updated.

It assumes you already have `data/snapshots` locally (snapshots don't need to be updated after EPOCH increment). If not, it'd run the entire DAG anyway.

I plan to test it locally after today's regions update.


pabloarosado (commented):
Looks good!

PR #2527, opened 2024-04-16
Fix concat issue in new pandas version

pabloarosado:
Following [this discussion](https://owid.slack.com/archives/C02706JNFC6/p1713270597628399), I've patched the buggy `concat` behaviour of the latest pandas version. The issue and suggested workaround are explained in the code.


Marigold (approved):
Thanks! What a fancy warning; you really treat our future selves well.

PR #2534, opened 2024-04-17
:sparkles: add column owid_continent to grapher regions

Marigold:
Add column `owid_continent` with a name of the continent to `grapher/regions` for convenience (Bastian would find it useful, and I'd say it could be useful in the future, too). This won't rebuild ETL, just a few datasets that depend on `grapher/regions`.

### DIFF
--- a/etl/steps/data/grapher/regions/latest/regions.py
+++ b/etl/steps/data/grapher/regions/latest/regions.py
@@ -6,6 +6,7 @@ This dataset is not meant to be imported to MySQL and is excluded from automatic
 
 import ast
 import re
+from typing import Any, cast
 
 import pandas as pd
 from owid.catalog import Dataset, Table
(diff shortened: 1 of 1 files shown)


Marigold:
@pabloarosado I've created a new table `regions_by_owid` with owid-only defined countries and added `continent` column for Bastian.

pabloarosado:
Hi @Marigold sorry for the late review. This PR is in conflict with https://github.com/owid/etl/pull/2571
Let me take a look at both and decide a solution.

marcelgerber:
Interesting. I was initially under the assumption that `grapher/regions` is gonna be explicitly used by the Grapher codebase, and not as a dataset or as a source for other etl steps.
However, it seems that this now intended, right?

On the other hand, I can definitely see the appeal of this, and that this makes a (relatively common?) lookup a bit easier. I'm of course not an etl expert, but this could potentially also go into the main regions file directly?

pabloarosado:
I'd suggest closing this PR, and instead go with https://github.com/owid/etl/pull/2587 (we can discuss it later).

Marigold:
Closing in favor of https://github.com/owid/etl/pull/2587

Issue #2533, opened 2024-04-17
Chart-sync fails when syncing from local to staging


pabloarosado:
When trying to sync local grapher to a staging server (branch `update-climate-data-and-fix-co2-concentration-issue`), I get an issue, both when using the wizard and the cli. I haven't looked into it but the issue may be caused by unrelated work (about natural disasters, that is dealt with in another branch).

For now I give up on syncing from local to staging, and will simply redo chart revisions on staging. But I created the issue for the record.

```
(etl-py3.11) ➜  etl git:(update-climate-data-and-fix-co2-concentration-issue) etl chart-sync .env update-climate-data-and-fix-co2-concentration-issu --staging-created-at 2024-04-16 --dry-run
2024-04-17 12:57:25 [info     ] staging_sync.start             chart_ids={7174, 3207, 7175, 7568, 7569, 7570, 7571, 7572, 7573, 7574, 7575, 7576, 7577, 7578, 7579, 7580, 7581, 7582, 7583, 7584, 7585, 7586, 5421, 5427, 5428, 1850, 1851, 1853, 730, 731, 732, 2912, 1777, 1525} n=34
2024-04-17 12:57:25,872 - [bugsnag] WARNING - No API key configured, couldn't notify
Traceback (most recent call last):
  File "/Users/prosado/Documents/owid/repos/etl/etl/grapher_model.py", line 357, in migrate_to_db
    target_var = Variable.load_from_catalog_path(target_session, source_var.catalogPath)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/etl/grapher_model.py", line 1245, in load_from_catalog_path
    return session.exec(select(cls).where(cls.catalogPath == catalog_path)).one()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/sqlalchemy/engine/result.py", line 1522, in one
    return self._only_one_row(
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/sqlalchemy/engine/result.py", line 562, in _only_one_row
[...]

PR #2540, opened 2024-04-18
:sparkles: add --errors and --include/--exclude to chart-sync

Marigold:
Fix for https://github.com/owid/etl/issues/2533#issuecomment-2061210834

You can either exclude charts that have variables matching `--include`, e.g.
```
etl chart-sync .env update-natural-disasters-data --include natural --dry-run
```
or `--exclude`
```
etl chart-sync .env update-natural-disasters-data --exclude foo --dry-run
```
or simply ignore charts causing errors
```
etl chart-sync .env update-natural-disasters-data --errors warn --dry-run
```

@pabloarosado could you try merging this into your PR and testing it out? I've only tested it on a few simple cases.

### DIFF
--- a/apps/staging_sync/cli.py
+++ b/apps/staging_sync/cli.py
@@ -54,6 +54,12 @@ log = structlog.get_logger()
     updated in production. Default is branch creation date.""",
 )
 @click.option(
+    "--include",
+    default=None,
+    type=str,
+    help="""Include only charts with variables whose catalogPath matches the provided string.""",
+)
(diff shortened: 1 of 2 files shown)


Marigold:
On second thought, maybe the cleanest solution would be if Streamlit showed charts for sync and you'd select the ones you are interested in. It could be part of the new chart revision tool.

pabloarosado (approved):
This looks good and it's useful, thanks!

PR #2546, opened 2024-04-19
Improve add regions to table

pabloarosado:
I improved `add_regions_to_table` so that it can be used on tables with multiple dimensions (but by default it assumes the index is `["country", "year"]`.

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -267,6 +267,7 @@ def add_region_aggregates(
     df: TableOrDataFrame,
     region: str,
     countries_in_region: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


lucasrodes:
To be discussed: unsure if `country_col` can be considered a column we group by?

pabloarosado:
> To be discussed: unsure if `country_col` can be considered a column we group by?

I see your point: it's a bit redundant to need to specify `["country", "year", ...]` when those two columns are always assumed to be there. But maybe in the future we could generalize this function so that "year" is not required. In fact, currently it's only required to check for overlap between historical and current regions (otherwise it's not needed for the aggregations). So I think it makes sense to keep `index_columns` the way it is now.

PR #2558, opened 2024-04-24
:tada: add chart diff

Marigold:
Chart diff is a new streamlit-based tool that aspires to be a successor to the current chart approval tool. The idea behind it is to unify and simplify our tooling around charts:
- Remove chart approval
- Run `chart-sync` tool automatically after merging PR
- Kill the staging server right after `chart-sync` finishes (and not after 3 days)

Compared to the current workflow, the new one improves:
- Data managers don't have to manually `chart-sync` charts from staging after ETL gets deployed
- Not having to approve charts twice
- Replace chart approval with a more flexible tool

### Features

- Chart from staging & production shown side by side
- Approve / Reject buttons
- Smart sorting (either by "similarity score" or by `updatedAt`)
- owidbot would be posting to a PR numbers of approved / unapproved charts

### Detecting updated charts

We have to efficiently find charts updated on the staging server. We have several options:
- compare `updatedAt` between staging and prod
- compare `config.version`
- checksum `config` and compare it (would it be fast enough?)

It's an open question whether we want to include charts with **identical config, but different data**. If we wanted to, we could use `dataset.sourceChecksum` or somehow compute checksum for indicators.

### "Merge conflicts"

There's a period (potentially quite long for large datasets) between merging PR and running `chart-sync` where someone can edit charts in production. If that happens, we'd **overwrite the chart**, but notify us in Slack that a chart has been overwritten. It's up to us to check the previous version of the chart in Admin and decide what to do with it.

## TODO
- [x] functions to detect new / updated charts
- [x] Indicator upgrader shouldn't be creating chart revisions, but should be creating charts directly (through Admin API just like `chart-sync` does it).
  - https://github.com/owid/etl/pull/2575 
- [x] Get also a config diff view (in addition to the chart-pair view)
  - https://github.com/owid/etl/pull/2575 
- [x] Unapprove all charts. Imagine that you have detected a major problem, and don't want to bring all charts to live.
- [x] smart sorting with "similarity score" or `updatedAt`
  - For now only for `updatedAt`
- [x] Filtering options: Add option to "hide approved charts", "show only drafts", etc.
- [ ] run `chart diff` periodically against all open PRs in ETL and post results back with owidbot
- [ ] run `chart-sync` automatically after merging
  - Temporarily use `chart-sync` manually after merging
- [ ] deprecate chart approval tool in Admin & drop chart revisions table (it's one of our largest tables)
- [ ] Support for draft and deleted charts (maybe start with just signaling this)
- [ ] Edge case: we edit `chart A` on staging → we merge the PR from staging server into master → someone edits the `chart A` in production before ETL builds → collision! There are two versions for `chart A`, we should resolve this. 
- [ ] stress test it: what happens if we have 100 charts? 200? Do we need pagination?


lucasrodes:
I've created a separate issue collecting all the pending tasks: https://github.com/owid/etl/issues/2580.

I propose merging this PR into master soon, so we can start working with it and detecting more issues.

PR #2567, opened 2024-04-26
Update regions.yml

spoonerf:
Countries that were previously missing from the WHO regions, mostly small islands.


Marigold (approved):
Looks good, let's merge this before Monday.

PR #2570, opened 2024-04-27
🎉 Create tool to archive steps

pabloarosado:
Create CLI tool to archive steps, and add it to the ETL dashboard.

I also fixed the formatting of a few random dag files. By the way, @Marigold is there a way to ensure that dependencies in the dag file are always indented?

Fixes https://github.com/owid/etl/issues/1820


lucasrodes (approved):
LGTM

We might need some formatter for YML files to avoid committing unnecessary style changes to these.

Marigold:
> By the way, @Marigold is there a way to ensure that dependencies in the dag file are always indented?

Thanks for fixing the old ones. I believe that all **new** DAG files should be correctly intended. If I remember it correctly, we use either `ruamel_dump` or `yaml_dump` functions and both should be using the correct indentation. If that's not the case, please let me know!

PR #2571, opened 2024-04-27
📊 grapher/regions/latest: Add income group information

marcelgerber:
This is working towards https://github.com/owid/owid-grapher/issues/3517.

This is adding the 4 WB income groups as rows to the `regions.csv` file _which is only consumed by owid-grapher_.
These rows get the new entity type `income_group`, which we will then handle in grapher. We mostly care about the members of these income groups; and for these to be up to date, we will want to update this step to use the latest `wb/.../income_groups` classification once a new one becomes available.

This also assigns new entity codes to these four regions: `OWID_WB_LIC, OWID_WB_LMC, OWID_WB_UMC, OWID_WB_HIC`. The last part of these are the official 3-letter codes that the WB uses for these entities.

**Changing these entity codes in the future is going to be a major pain, so please let me know if you would want to use different ones instead.**

### DIFF
--- a/dag/main.yml
+++ b/dag/main.yml
@@ -229,6 +229,7 @@ steps:
   data://garden/regions/2023-01-01/regions:
   data://grapher/regions/latest/regions:
     - data://garden/regions/2023-01-01/regions
+    - data://garden/wb/2024-03-11/income_groups
 
   # Democracy and Human rights - V-Dem index
   data://meadow/democracy/2023-03-02/vdem:
@@ -666,7 +667,6 @@ steps:
     - data://garden/demography/2023-03-31/population
   data://grapher/missing_data/2024-03-26/who_md_suicides:
     - data://garden/missing_data/2024-03-26/who_md_suicides
-
 
   ######################################################################################################################
   # Older versions to be archived once they are not used by any other steps.
--- a/etl/steps/data/grapher/regions/latest/regions.py
+++ b/etl/steps/data/grapher/regions/latest/regions.py
@@ -9,11 +9,19 @@ import re
 
 import pandas as pd
 from owid.catalog import Dataset, Table
+from owid.catalog.processing import concat
 
 from etl.helpers import PathFinder, create_dataset
 
 # Get paths and naming conventions for current step.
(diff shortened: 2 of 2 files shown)


marcelgerber:
The output of this step can be found here: http://staging-site-grapher-regions-income-groups:8881/grapher/regions/latest/regions/regions.csv

pabloarosado:
Hi @marcelgerber, regarding codes for income groups, in other places of OWID I have seen "OWID_LIC", "OWID_LMC", "OWID_UMC", "OWID_HIC", for example [in FAOSTAT](https://github.com/owid/etl/blob/5c9c6bc5df5343c5823a2c9d62fb9ca74ed89e4c/etl/steps/data/garden/faostat/2024-03-14/shared.py#L139). I don't think those codes are relevant, but it makes me think that maybe these codes have been used also elsewhere, so maybe it's safer to stick to them.

This PR is in conflict with https://github.com/owid/etl/pull/2534. So let me think about both proposals and decide about a good solution.

marcelgerber:
Sure, I'll change the entity codes then to keep them consistent. Let me know how you want to proceed in regards to #2534.

pabloarosado:
> Sure, I'll change the entity codes then to keep them consistent. Let me know how you want to proceed in regards to #2534.

Don't worry about it. I've created another PR with a different logic: https://github.com/owid/etl/pull/2587
I'll discuss it tomorrow in the data architecture call and then decide which approach to take.

pabloarosado:
I'd suggest closing this PR, and instead go with https://github.com/owid/etl/pull/2587 (we can discuss it later).
Note that the resulting file you could use in grapher would be http://staging-site-create-external-channel:8881/external/owid_grapher/latest/regions/regions.csv

marcelgerber:
Nice. I've read through your proposal, and all of it makes a lot of sense to me. Thank you for thinking this through!

pabloarosado:
Hey @marcelgerber you can now access the regions data in https://catalog.ourworldindata.org/external/owid_grapher/latest/regions/regions.csv
Please let me know if there's any issue or you want to change anything there, thanks!

PR #2575, opened 2024-04-29
✨ wizard: indicator upgrade v2

lucasrodes:
### Summary
**indicator upgrade**
- It is an evolution of the former chart upgrade
- I've optimise performance
- Minor styling changes
- Code simplification
- It now updates charts directly using admin API

In **chart diff**:
- Add a chart config diff view

### TODO:
- Indicator mappings are no longer stored (before they were being somehow stored in the suggested chart table). We need to store this info, as discussed in some meetings. Maybe we need a table?

### Review
Go to http://staging-site-streamlit-chart-approval-indicator-upgrade/etl/wizard/ and test it


apps/wizard/pages/chart_diff/app.py
```diff
@@ -16,7 +18,7 @@
 
 CURRENT_DIR = Path(__file__).resolve().parent
 # TODO: unhardcode this
-SOURCE_ENV = "staging-site-streamlit-chart-approval"
+SOURCE_ENV = "staging-site-streamlit-chart-approval-indicator-upgrade"
```
Marigold, line 21: @lucasrodes you should change it after merging, right?

etl/chart_revision/v2/updaters/variable_update.py
```diff
@@ -465,57 +313,83 @@ def find_charts_from_variable_ids(variable_ids: Set[int]) -> List[gm.Chart]:
...
+    config_new = update_chart_config_map(config_new, indicator_mapping)
+    # Update dimensions
+    config_new = update_chart_config_dimensions(config_new, indicator_mapping)
+    # Update sorting
+    config_new = update_chart_config_sort(config_new, indicator_mapping)
+    # Update indicator references
+    return config_new
+
+
+def update_chart_config_map(
```
Marigold, line 341: Not for this PR, but perhaps having `ChartConfig` object with convenient methods would be helpful in multiple places. (It'd also help with autocomplete)

PR #2576, opened 2024-04-29
🐝 make the prompt for datainsight bot editable

danyx23:


danyx23:
The failing check is that I was connected via ssh and had been running wizard so it could not automatically be restarted.

Issue #2508, opened 2024-04-10
New ETL channel (or namespace) to track dependencies of external repositories


pabloarosado:
## Context & motivation

- Currently, we track dependencies of things that are used in Grapher charts; we will get an error if we delete a dependency
- But, we don't track all downstream uses:
  - Explorers
  - External repositories

We would love to get an error early, rather than breaking something.

## Proposal

Inspired by [this thread](https://owid.slack.com/archives/C03NA1B2P1U/p1712753409678569), I thought we could have a new ETL channel (or, as an easier alternative, a new namespace), e.g. `repository`, which hosts `latest` versions of steps that are loaded by external repositories, like `poverty-data`, `covid-19-data`, `co2-data`, `energy-data`, and also `owid-grapher`.

With this, if anyone accidentally deletes or archives a dependency of these repository steps, we will get an error by VersionTracker. We could also show these dependencies in the ETL Dashboard (and treat them as always active, even if they don't have charts).

Specifically, Sophia said it would be useful for her to have `latest` versions of `grapher/worldbank_wdi/2023-05-29/wdi/wdi#ny_gdp_pcap_pp_kd` and for `grapher/demography/2023-03-31/population/population#population_historical`. They could be easily created in the new channel or namespace.

## Technical notes

- Creating a new channel would be ideal, but it requires much more additional work. So, a much cheaper alternative would be to create a new namespace.

## Scope

- Decide on a special reserved namespace
- Make a new dataset for Sophia (for example), using a special namespace
- Modify the version tracker to always consider that namespace used

danyx23:
Another option for how to do this would be to add an alias system that we use to keep what are probably basically our MiMs in an easily accessible namespace.

I.e. the aliases would not be an ETL channel or something like this but a new kind of URI scheme that is mapped to full ETL paths. Maybe something to discuss in our offsite.

sophiamersmann:
Grapher has the following indicator IDs currently hardcoded:
- Continents: [123](https://owid.cloud/admin/variables/123) (probably ok)
- Population: [597929](https://owid.cloud/admin/variables/597929) (used in the admin as default size dimension for scatter plots)
- More population variables: [525709](https://owid.cloud/admin/variables/525709), [525711](https://owid.cloud/admin/variables/525711), [597929](https://owid.cloud/admin/variables/597929), [597930](https://owid.cloud/admin/variables/597930) (used to exclude the population indicator from being displayed in the sources modal in certain cases)

[...]

PR #2587, opened 2024-05-01
🎉 Create external channel

pabloarosado:
## Summary

This PR creates a new ETL channel called "external", that lets us add steps corresponding to external uses of our data, for example from other repositories (like [owid-grapher](https://github.com/owid/owid-grapher)). This channel can be used to produce data files in our catalog that are needed elsewhere, but not used by any grapher charts.

This PR fixes https://github.com/owid/etl/issues/2508

## Context

We are currently facing the following problems:
1. We have no way to track uses of ETL data by external repositories (like [energy-data](https://github.com/owid/energy-data) and [co2-data](https://github.com/owid/co2-data)) or even explorers (we have the "explorers" channel, but it doesn't fully solve the problem). Of course, we can't track all external uses of our catalog data (since we can't track how users ingest our data), but at least we should track all **OWID uses** of the data.
2. We need another data file with regions and income groups to be used by owid-grapher (see Marcel's PR: https://github.com/owid/etl/pull/2571 ). We need other data files from ETL (see Sophia's [request](https://owid.slack.com/archives/C03NA1B2P1U/p1712753409678569)).
3. We have a `grapher/regions/latest/regions` dataset, so that `owid-grapher` can ingest some data from ETL. But this is a very special use which doesn't fit with the logic of the `grapher` channel. In fact, [this step needs to be excluded](https://github.com/owid/etl/blob/0331aca383b71dcc1d04939e504c4772a14f841c/etl/command.py#L321) from ETL to avoid errors being raised (since this step should not end up in the grapher database). So it is a hack, rather than a solution.
4. We have 4 single-variable grapher datasets feeding the charts of our page on [world region definitions](https://ourworldindata.org/world-region-map-definitions). They were manually uploaded long time ago and are not maintained. There are other institutions with custom region definitions that we could include here.
5. We need a simple countries-regions file to be used elsewhere outside of ETL (see [discussion](https://owid.slack.com/archives/C6C0MCXU0/p1712907857167909) started by Bastian, and Mojmir's PR: https://github.com/owid/etl/pull/2534).

## Proposed solution

This PR attempts to solve all these problems with the following strategy:
1. We create an "external" ETL channel, where we can add all external uses of ETL steps. This can include direct loads of our ETL datasets like in the energy-data or co2-data repos, or less explicit uses of our data by owid-grapher. We could consider adding here explorers too.
  * For a separate PR: The current logic of `etl explorer-update` could be transferred to `etl d version-tracker`. The latter could check `owid-grapher`'s tsv files, and ensure that all dependencies are tracked by steps in the `external/owid-content` channel.
  * For a separate PR: The ETL dashboard could treat steps in the "external" channel differently, and assume they are always used (even if they don't have charts). This way we can always be informed if there are broken or outdated dependencies.
2. We create an `external/owid-grapher/latest/regions` dataset to be used by owid-grapher. This would replace Marcel's PR: https://github.com/owid/etl/pull/2571 . The resulting file can be downloaded at http://staging-site-create-external-channel:8881/external/owid_grapher/latest/regions/regions.csv
3. For now, we keep the `grapher/regions/latest/regions` file, since (I think) it's used by owid-grapher and possibly elsewhere. But we should eventually delete this hacky step (or at least stop misusing it). From now on we should use `external/owid-grapher/latest/regions` instead.
4. We create a `grapher/regions/2023-01-01/regions` step to feed the charts in our world regions page.
5. We add a new step to our `external` channel (which could depend on `grapher/regions/2023-01-01/regions`) to track the use of our countries-regions file. This would replace Mojmir's PR: https://github.com/owid/etl/pull/2534 . The resulting file can be downloaded at http://staging-site-create-external-channel:8881/external/notebooks/latest/regions/regions.csv

## Other considerations

We could consider (for consistency) adding the simplified countries-regions (point 5 above) as an additional table of the `garden/regions/2023-01-01/regions`, and possibly adding income groups there too. That would be more consistent than implementing that logic in grapher. However, for simplicity (and to avoid unintended side effects) we should avoid unnecessary changes to `garden/regions/2023-01-01/regions`. In the future, if we create a new version, we can consider adding more logic to it.

## TODO

If wee agree on this proposal:
* [x] Add external steps for Bastian's requested use of country-regions.
* [ ] Add external steps for Sophia's requested uses, and more generally, to all owid-grapher uses of ETL data that Sophia listed in [this comment](https://github.com/owid/etl/issues/2508#issuecomment-2047825217).
* [ ] Add World Bank, UN (and possibly UN institutions) to `garden/regions/2023-01-01/regions` and create charts for [our world map region definitions page](https://ourworldindata.org/world-region-map-definitions). Shall we show income groups in this page too?


lucasrodes (approved):
LGTM

lib/catalog/owid/catalog/datasets.py
```diff
@@ -41,7 +41,7 @@
 assert SUPPORTED_FORMATS[0] == PREFERRED_FORMAT
 
 # available channels in the catalog
-CHANNEL = Literal["garden", "meadow", "grapher", "backport", "open_numbers", "examples", "explorers"]
+CHANNEL = Literal["garden", "meadow", "grapher", "backport", "open_numbers", "examples", "explorers", "external"]
```
Marigold, line 44: Nit - We could add a constant listing all channels extracted from `Literal` type with
```
from typing import get_args
CHANNELS = get_args(CHANNEL)
```
which could be used to avoid enumerating all channels each time.

Marigold (approved):
Neat! This looks much better than I imagined when we discussed it.

lib/catalog/owid/catalog/datasets.py
pabloarosado: Good idea. I've noticed we use `CHANNEL.__args__` in many places. So maybe it's better to replace all those with CHANNELS (but better in a separate PR), to avoid confusion.

PR #2590, opened 2024-05-02
🐛 Adapt helper function that adds GDP to a table

pabloarosado:
Adapt helper function that adds GDP to a table, after the update of the Maddison Project Database.

The function now should work both when the old and when the new dataset is used.

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -730,7 +730,7 @@ def add_gdp_to_table(
     tb : Table
         Original table that contains a column of country names and years.
     ds_gdp : Dataset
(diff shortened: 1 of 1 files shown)


paarriagadap (approved):
Good! I saw it in the other issue already, so all fine.

I need to think in an expanded function that imports GDP in constant international dollars from the [World Bank](https://admin.owid.io/admin/variables/735657) or [Penn World Table](https://admin.owid.io/admin/variables/539736). Or even the other GDPs from the World Bank, like in [current US$](https://admin.owid.io/admin/variables/735650).

But that can wait.

PR #2591, opened 2024-05-02
✨ VersionTracker.steps_df: ~30% reduction in execution time (to be removed)

lucasrodes:
~31% execution time reduction in `VersionTracker().steps_df`. 

**In my machine, it went from 6.6 s ± 56.8 ms → 4.5s ± 75.3 ms.** (mean ± std. dev. of 7 runs, 1 loop each)

### `_add_columns_with_different_step_versions`
This has been achieved by slightly optimising the execution time of `_add_columns_with_different_step_versions`. In this function, there is a loop iterating over all "identifiers" of datasets. We only need to iterate over those identifiers that have more than one dataset associated, since those with only one dataset will have deterministic values for columns `"same_steps_forward"`, `"same_steps_backward`, `"same_steps_all"` and `"same_steps_latest"`. Note that the majority of the identifiers only have >` dataset (~1500 vs ~300 with more than one dataset). Therefore the loop is way faster!

### `_init_steps_df`
At the start of `_create_steps_df` there are various column assignments. I have placed all of these lines inside function`_init_steps_df` (no optimisation this far).

Optimisation has been done in assignments to `steps_df["direct_usages"]` and `steps_df["all_usages"]` which were taking the most chunk of the time in these assignments.

The new assignments are done with `_init_steps_df_ndim`, so that one can still use former implementation `_init_steps_df`.

### DIFF
--- a/etl/version_tracker.py
+++ b/etl/version_tracker.py
@@ -796,12 +796,10 @@ class VersionTracker:
         steps_df["all_active_usages"] = [
             self.get_all_step_usages(step=step, only_active=True) for step in self.all_steps
         ]
-        # TODO
         steps_df["all_usages"] = [self.get_all_step_usages(step=step) for step in self.all_steps]
         steps_df["state"] = ["active" if step in self.all_active_steps else "archive" for step in self.all_steps]
         steps_df["role"] = ["usage" if step in self.dag_all else "dependency" for step in self.all_steps]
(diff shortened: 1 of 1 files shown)


lucasrodes:
due to unknown issue in staging server, can't test ETL Dashboard. I've opted to create a new PR to launch a fresh staging server and test the optimised ETL dashboard there: https://github.com/owid/etl/pull/2593

PR #2608, opened 2024-05-06
:tada: periodic runs of chart-diff against all open PRs

Marigold:
Script that iterates over all open PRs by OWID members and runs `owidbot --services chart-diff` against them. Can be invoked with
```
etl d scan-chart-diff --dry-run
```

This script will be run from `etl-prod-2` as a CRON job every 10 minutes (could be lower if needed). I was considering running a Buildkite scheduled job, but that'd be probably too heavy.

In addition to the script, there are a couple of improvements:
- `owidbot` doesn't update `Last edited` timestamp when body hasn't changed
- If `.env.prod` exists, we use it to connect to prod (otherwise we keep using `staging-site-master`). By the way, this is getting a bit messy, maybe we could wrap staging & prod info in an object similar to `OWIDEnv`?

### DIFF
--- a/apps/wizard/pages/chart_diff/app.py
+++ b/apps/wizard/pages/chart_diff/app.py
@@ -3,8 +3,10 @@ from pathlib import Path
 
 import streamlit as st
 from sqlalchemy.engine.base import Engine
+from sqlalchemy.exc import OperationalError
 from sqlmodel import Session, SQLModel
 from st_pages import add_indentation
+from structlog import get_logger
 
 from apps.staging_sync.cli import _get_engine_for_env, _modified_chart_ids_by_admin, _validate_env
 from apps.wizard.pages.chart_diff.chart_diff import ChartDiffModified
@@ -13,6 +15,9 @@ from apps.wizard.utils import chart_html, set_states
 from apps.wizard.utils.env import OWID_ENV
 from etl import grapher_model as gm
 from etl.config import DB_HOST
+from etl.paths import ENV_FILE_PROD
+
+log = get_logger()
 
 # from apps.wizard import utils as wizard_utils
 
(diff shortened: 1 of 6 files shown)


apps/wizard/pages/chart_diff/app.py
```diff
@@ -22,8 +27,12 @@
 # TODO: simplify this
 SOURCE_ENV = DB_HOST  # "staging-site-streamlit-chart-approval"
 SOURCE_API = f"https://api-staging.owid.io/{SOURCE_ENV}/v1/indicators/"
-# TODO: switch to production once we are ready
-TARGET_ENV = "staging-site-master"
+
+if ENV_FILE_PROD.exists():
+    TARGET_ENV = ENV_FILE_PROD
+else:
+    log.warning(f"Production env file {ENV_FILE_PROD} not found, comparing against staging-site-master")
```
lucasrodes, line 34: I'd probably add a streamlit `st.warning` or something here, to signal this on the end user. It could have big implications if they don't know, right?

lucasrodes (approved):
LGTM!

Could you elaborate a bit more on:

> If .env.prod exists, we use it to connect to prod (otherwise we keep using staging-site-master). By the way, this is getting a bit messy, maybe we could wrap staging & prod info in an object similar to OWIDEnv?


Thanks!

apps/wizard/pages/chart_diff/app.py
Marigold: Added. It doesn't write anything to DB and `.env.prod` is read-only, so those implications are not as bad, but it's still good to know.

Marigold:
> Could you elaborate a bit more on:

What I meant was - if you have an idea how to refactor that crap, do it!

PR #2615, opened 2024-05-06
:tada: new chart-sync tool that supports chart-diff

Marigold:
Refactor `chart-sync` tool and integrate it with the new `chart-diff` tool. This **replaces** the existing `chart-sync` tool... it might be better to do `v2` of chart-sync or simply make everyone use the old way of creating charts in production for a while.

## Updates
- Replace status `rejected` by `unapproved`. In my head, "reject" is a stronger concept that might be introduced later. Rejected charts might not even be synced. Let me know if you disagree @lucasrodes
- Remove chart-sync app
- Leverage `ChartDiffModified` for working with two charts
- Simplified logic for creating / updating charts
- Ignore chart revisions
- Slack notification when syncing unapproved chart

### DIFF
--- a/apps/staging_sync/cli.py
+++ b/apps/staging_sync/cli.py
@@ -2,7 +2,7 @@ import copy
 import datetime as dt
 import re
 from pathlib import Path
-from typing import Any, Dict, Optional, Set, Union
+from typing import Any, Dict, List, Optional, Set, Union
 
 import click
 import pandas as pd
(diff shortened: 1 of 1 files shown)


lucasrodes:
> Replace status rejected by unapproved. In my head, "reject" is a stronger concept that might be introduced later. Rejected charts might not even be synced. Let me know if you disagree @lucasrodes

Sounds perfect!

PR #2620, opened 2024-05-07
✨ (owidbot) add svg tester commit links

sophiamersmann:
Adds commit links to the SVG tester summary. It's useful to have a direct link to the individual commit when there are many changes and GitHub freezes when opening the branch summary, which includes changes of all commits (it happens).

The script reads log files (`commit_{default,all}-views.log`) that contain commit IDs that have been added by the SVG tester Buildkite job.


Marigold (approved):
Welcome to Python land! Looks good, just run `make check` before merging please, it'll format modified files.

PR #2625, opened 2024-05-08
📊  Intact Forest Landscapes

spoonerf:
Hey @pabloarosado!

For this PR I am not sure how to handle the source. The author (Potapov) asked that we cite it as Potapov et al. (2017) and reference the paper that sets out the dataset. However, the dataset has been updated since, in 2021. Currently the source footer shows as `Potapov et al. (2017) (2021)`, which is not ideal. 

What do you think is the best solution here? `Potapov et al. (2017)`, which is what the author requested or `Potapov et al. (2021)` which is perhaps closer to what the dataset reference should be.


dag/forests.yml
```diff
@@ -0,0 +1,9 @@
+steps:
+  # Intact Forest Landscapes
+  data://meadow/forests/2024-05-08/ifl:
```
pabloarosado, line 3: I think it would be more useful to spell out this acronym, since there are many IFL things, and this one is not particularly well known (at least to me). But if it's too much hassle, you can leave it as-is.

snapshots/forests/2024-05-08/ifl.py
```diff
@@ -0,0 +1,25 @@
...
+
+from etl.snapshot import Snapshot
+
+# Version for current snapshot dataset.
+SNAPSHOT_VERSION = Path(__file__).parent.name
+
+
+@click.command()
+@click.option("--upload/--skip-upload", default=True, type=bool, help="Upload dataset to Snapshot")
+@click.option("--path-to-file", prompt=True, type=str, help="Path to local data file.")
```
pabloarosado, line 15: When creating a snapshot from a local file, it's convenient to explain where the file came from. I see that you mentioned it in the metadata file, but maybe it's more useful to do that in the docstring at the top of the script (up to you).

snapshots/forests/2024-05-08/ifl.xlsx.dvc
```diff
@@ -0,0 +1,35 @@
...
+      - Large natural forest areas are also important for maintaining ecological processes and supplying ecosystem services like water and air purification, nutrient cycling, carbon sequestration, erosion, and flood control.
+      - The conservation value of forest landscapes that are free from human disturbance is therefore high, although it varies among regions. At the same time the cost of conserving large unpopulated areas is often low. The same factors that have kept them from being developed, such as remoteness and low economic value, also help to reduce the cost of protecting them.
+
+      The concept of an Intact Forest Landscape (IFL) and its technical definition were developed to help create, implement, and monitor policies concerning the human impact on forest landscapes at the regional or country levels.
+    date_published: "2021-11-29"
+    # Citation
+    producer: Potapov et al. (2017)
```
pabloarosado, line 19: Producer should be simply "Potapov et al.". It should not include the year.

etl/steps/data/garden/forests/2024-05-08/ifl.meta.yml
```diff
@@ -0,0 +1,23 @@
...
+  update_period_days: 365
+
+tables:
+  ifl:
+    variables:
+      ifl_area:
+        title: Intact Forest Landscapes Area
+        unit: square kilometers
+        short_unit: km²
+        description_short: Area of [intact forest landscapes](#dod:intact-forest-landscapes). Intact forest landscapes are vast, continuous natural ecosystems without significant human disturbance or fragmentation.
```
pabloarosado, line 20: I find it a little odd that you have a DoD as well as an explicit explanation. Wouldn't it be enough to simply say:
(A) "Area of [intact forest landscapes](#dod:intact-forest-landscapes)."
or:
(B) "Area of [intact forest landscapes](#dod:intact-forest-landscapes), which are vast, continuous natural ecosystems without significant human disturbance or fragmentation."

etl/steps/data/garden/forests/2024-05-08/ifl.py
```diff
@@ -0,0 +1,46 @@
...
+    ds_meadow = paths.load_dataset("ifl")
+    # Load regions dataset.
+    ds_regions = paths.load_dataset("regions")
+    # Read table from meadow dataset.
+    tb = ds_meadow["ifl"].reset_index()
+
+    #
+    # Process data.
+    #
+    # Convert to square kilometers.
```
pabloarosado, line 25: Maybe, clarify in the comment that the original data is in 1000 square kilometers.

pabloarosado (approved):
It looks good! I just left very minor suggestions, feel free to ignore them.

pabloarosado:
Regarding the footer, the ideal solution for us and our users would be to use `Potapov et al. (2021)`, which already tells when the data was last updated. I suppose the author is not aware that we include a full citation in the "Learn more about this data" area, and they would be happy with a `Potapov et al. (2021)` in the footer + a proper citation to the 2017 paper. However, if the data provider (for some reason) strongly prefers citing the paper, then we should respect that, and have `Potapov et al. (2017)` in the footer.

etl/steps/data/garden/forests/2024-05-08/ifl.meta.yml
spoonerf: I am a little unsure about this, my understanding was that in cases where you can it is best to include a short explanation in the subtitle alongside a DoD. Mainly so that when users screenshot rather than download the chart, as is often the case, a brief explanation of the term is still on the chart. 

In this case, the DoD goes more in-depth than the information captured in the `description_short`.

PR #2632, opened 2024-05-09
📊 Add data on tsunamis, volcanoes and earthquakes

pabloarosado:
## Summary

Add data for earthquakes, volcanoes and tsunamis. This PR will replace four old dataset currently in grapher (on earthquakes, earthquake deaths, volcanoes, and volcano deaths), and add two charts on tsunamis and tsunami deaths.

All changes can be inspected from [the chart-diff page](http://staging-site-add-noaa-natural-disasters-data/etl/wizard/Chart%20Diff).

## Main changes

* Add snapshos/data steps for the Natural Hazards Data. This data consists of three files (one for each disaster type) that can be downloaded via an API.
* Minor improvements in the old EM-DAT natural disasters steps.
* Move steps related to natural disasters to its own dag file.
* Minor improvement in a helpers function.


spoonerf (approved):
Looks good but I had some more general comments about the charts:

* Should the 2024 data be included as it is incomplete? If you do include it, it might be worth setting the last year on the map to the last complete year.
* I don't think the tsunami chart should include the word 'significant', my understanding from the metadata is that only the earthquakes and volcanic eruptions have this 'significant' threshold. 
* Are you planning on adding decadal values for the NOAA NCEI data too? 
* Chart diff is very cool! 
* Is Chart diff automatically changing the title and subtitle of charts? That doesn't seem desirable, but maybe it is not reflecting what chart sync will do? 
Screenshot 2024-05-13 at 16 27 32

PR #2643, opened 2024-05-15
:bug: use proper Tailscale address for AdminAPI

Marigold:
Use Tailscale address http://owid-admin-prod.tail6e23.ts.net for AdminAPI instead of http://owid-admin-prod. The problem is that `owid-admin-prod` resolves to an address of the actual LXC container instead of going through the Tailscale. Using alias `owid-admin-prod.tail6e23.ts.net` fixes the problem.


pabloarosado (approved):
Thanks for fixing this, it was so confusing!

PR #2646, opened 2024-05-15
:tada: update github check in owidbot's chart-diff

Marigold:
Update the GitHub check in owidbot's chart diff in addition to posting it as the owidbot user (owidbot's updates are run every 5 minutes). The check is updated through the GitHub App. This also refactors existing GitHub utility functions.

The check itself is pretty bare-bones since the information about unapproved charts is already shown in owidbot's comment. We can improve the title/summary later.

Example of a check:
image

Example of owidbot comment:
image

### DIFF
--- a/apps/owidbot/chart_diff.py
+++ b/apps/owidbot/chart_diff.py
@@ -11,7 +11,7 @@ from . import github_utils as gh_utils
 log = get_logger()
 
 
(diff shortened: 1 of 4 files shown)


lucasrodes (approved):
thanks for working on this ー lgtm

PR #2655, opened 2024-05-17
:hammer: enable chart-diff to connect to production and clean up envs

Marigold:
## Motivation

Managing various envs (live, staging, local, etc.) has become more scattered and everyone is reinventing the wheel with their own helper functions. This PR unifies env management under existing `OWIDEnv` class.

## Implementation

Leverage `OWIDEnv` object for managing settings and engines for various environments. Instead of using default `etl.config`, you can now pass arbitrary config. There are also helper functions for creating `OWIDEnv` for staging servers etc.

I also removed `staging` from `OWIDEnvType` since we no longer use them (@lucasrodes can I remove `remote-staging` to `staging`?). Should I maybe also remove mentions of `live` from everywhere and replace them by `production`?

## Production env

Currently, we compare `chart-diff` against `staging-site-master` on staging servers. That's not optimal as we'd like to compare it against production. Hence, we add `ENV_FILE_PROD` env variable that points to production `.env` file (e.g. `.env.prod.read`). 

## TODO after merging
- [x] Add `ENV_FILE_PROD=.env.prod.read` variable and actual `.env.prod.read` file to all staging servers

### DIFF
--- a/apps/owidbot/chart_diff.py
+++ b/apps/owidbot/chart_diff.py
@@ -2,9 +2,9 @@ import pandas as pd
 from sqlmodel import Session
 from structlog import get_logger
 
-from apps.staging_sync.cli import _get_container_name, _get_engine_for_env, _modified_chart_ids_by_admin
+from apps.staging_sync.cli import _get_container_name, _modified_chart_ids_by_admin
 from apps.wizard.pages.chart_diff.chart_diff import ChartDiffModified
-from etl import config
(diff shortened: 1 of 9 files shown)


lucasrodes (approved):
> I also removed staging from OWIDEnvType since we no longer use them (@lucasrodes can I remove remote-staging to staging?). Should I maybe also remove mentions of live from everywhere and replace them by production?

Totally fine!

LGTM

PR #2660, opened 2024-05-20
✨ Let dashboard update and archive on staging servers

pabloarosado:
Currently, updating and archiving is only allowed when working locally. But this means that, if you are working on staging (especially with the new workflow) you can't archive from staging.

This PR lets the dashboard user update and archive steps from local or staging environments (but not from production).

NOTE: I haven't tested it, so I'm not sure if `ENV == "production"` is the right condition (but I suppose it should be).


Marigold (approved):
Looks good. Take a look at https://github.com/owid/etl/blob/master/apps/wizard/config/config.yml#L182 and set
```
disable:
  production: True
```
for Dashboard. That will hide it from wizard.

pabloarosado:
Thanks @Marigold, I want the dashboard to be visible in production, but not updating and archiving.

PR #2665, opened 2024-05-20
✨  Adding NTDs to schema tags

spoonerf:
I just added Neglected Tropical Diseases to the tags in the schema, as it's a new topic page.

Is this going to trigger a big etl rebuild or is it fairly safe to merge @Marigold ?


Marigold (approved):
Go for it, this doesn't trigger a rebuild.

PR #2674, opened 2024-05-21
:hammer: migrate to SQLAlchemy 2

Marigold:
We're currently using `SQLModel` as our ORM which is a wrapper around `SQLAlchemy`. Its main purpose is to provide better syntax and type hints. However, the new release of `SQLALchemy 2` implements pretty much the same features and is more mainstream and better maintained. I've been struggling with `SQLModel` lately and hope that this migration will be a good investment going forward.

Migration to SQLAlchemy 2 is a pretty big task that involves:
- Rewriting ORM classes
- Migrating `session.execute` commands
- Fixing type hints everywhere

What are the benefits?
- Unified SQLAlchemy queries
- More type checks (I removed a lot of `# type: ignore`)
- Improved performance and multithreaded access to DB (the original motivation was optimizing `chart-diff` performance)

Other changes:
- Remove unused classes from ORM (these can be added back as needed)
- Remove relationships between ORM classes (we don't use them now, but can add them in the future)

## How to review

Please skim through the PR to see if you spot anything unusual. You can also try it with your favorite wizard app.

## Testing

There's a high risk that something goes unnoticed. Our unit tests coverage of code executing ORM queries is unfortunately pretty low (e.g. wizard apps and CLIs). Here's a list of things to check to increase confidence in the migration:

- [x] Check all wizard apps (including Fast-track)
- [ ] Test dataset update workflow with indicator upgrader and chart-diff (I'll test it on WDI update)
- [x] Run ETL in staging and make sure data-diff doesn't find any differences

## TODO before merging
- [x] revert `ETL_EPOCH`


apps/backport/datasync/data_metadata.py
```diff
@@ -124,11 +121,11 @@ def _load_variable(session: Session, variable_id: int) -> Dict[str, Any]:
     """
 
     # Using the session to execute raw SQL and fetching one row as a result
-    result = session.execute(sql, {"variable_id": variable_id}).fetchone()  # type: ignore
+    result = session.execute(text(sql), {"variable_id": variable_id}).fetchone()
```
lucasrodes, line 124: i see how we can create a wrapper that applies `text(...)` to all entries. Maybe a simple extension of `Session`. Not urgent at all, can do in the future

apps/backport/backport.py
```diff
@@ -275,6 +275,7 @@ def _upload_data_metadata(lg: Any, backport_short_name: str, dry_run: bool) -> N
             display={},
             dimensions=None,
         )
+        db_var.id = db_variable_row["id"]
```
lucasrodes, line 278: is it not possible to pass `id` as an argument of `gm.Variable`?

apps/backport/datasync/data_metadata.py
Marigold: Yeah, I was thinking about that too and was actually surprised that SQLAlchemy doesn't already have such a helper function. (Feel free to add a helper function if you find yourself repeating it over and over again)

apps/backport/backport.py
Marigold: This is the tricky part with SQLAlchemy ORM and dataclasses. I had to define that field as
```
id: Mapped[int] = mapped_column(Integer, primary_key=True, init=False)
```
with `init=False`, otherwise it would always require `id` when creating the class (which is nonsense because `id` gets autoincremented after inserting to DB). This is a special case, in most cases you shouldn't be defining `id` manually, either you create a new object without `id` or you get one from DB which already has one.

That said, it's gonna take some time to get used to SQLAlchemy architecture.

PR #2675, opened 2024-05-21
:sparkles: GBD 2021 faster snapshot creation

Marigold:
Running `python snapshots/ihme_gbd/2024-05-20/gbd_cause.py` took quite long and "froze" at the end when creating the snapshot. The "froze" was caused by uploading a pretty large file as snapshot.

This PR tries to save memory wherever possible. It avoids unzipping file to disk and instead loads it directly to memory as dataframe. Then it optimizes it with `repack_frame`.

Finally, we concatenate all dataframes using our `concatenate` function that efficiently handles categorical variables.

(The previous approach might have been ok - it's just that uploading took so long. Zipped file should have similar size as dataframe with optimal types. One advantage of working with dataframes is that we don't have to unzip them in meadow step.)


spoonerf (approved):
Looks great!

PR #2685, opened 2024-05-23
Calculate historic daily calorie supply

pabloarosado:
I created a new dataset for the historical daily calorie supply, combining many different sources for different countries.
For more details, see https://github.com/owid/owid-issues/issues/1493

[Chart diff](http://staging-site-add-daily-calories-data/etl/wizard/Chart%20Diff) also shows additional changes on charts from other datasets, becauase also I did a few additional updates (on related datasets):
* Ensured we always refer to "daily calorie supply" instead of "consumption" or "intake" in various charts.
* Replaced [Meat consumption in the US (USDA, 2018)](http://staging-site-add-daily-calories-data/admin/datasets/3190) with data from FAOSTAT. The number from FAOSTAT are significantly higher (while the shape of the curves is similar). I don't know what [the original data](https://www.ers.usda.gov/data-products/food-availability-per-capita-data-system/) was exactly, so I can't tell the reason for this discrepancy. Maybe the original USDA data was the Loss-Adjusted Food Availability (which we have found some criticism about). So, for now, I think. it's best to keep using FAOSTAT. At some point we can move back to USDA (they actually have much more data, and probably FAOSTAT relies on them).
* Replaced [Daily protein supply - FAO (2017)](http://staging-site-add-daily-calories-data/admin/datasets/582) with data from FAOSTAT.


etl/steps/data/meadow/agriculture/2024-05-23/floud_et_al_2011.py
```diff
@@ -0,0 +1,45 @@
...
+
+    #
+    # Process data.
+    #
+    # Transform Europe data to have a year column.
+    tb_europe = tb_europe.melt(id_vars=["country"], var_name="year", value_name="daily_calories")
+
+    # Prepare US data.
+    tb_us = tb_us.rename(columns={"Year": "year", "Calories": "daily_calories"}, errors="raise").assign(
+        **{"country": "United States"}
```
spoonerf, line 31: I've not seen this 'assign(**{' before, it looks really useful!

etl/steps/data/garden/agriculture/2024-05-23/daily_calories_per_person.meta.yml
```diff
@@ -0,0 +1,39 @@
...
+        - Food Supply
+
+dataset:
+  update_period_days: 365
+
+tables:
+  daily_calories_per_person:
+    title: Daily calorie supply per person
+    variables:
+      daily_calories:
```
spoonerf, line 15: Perhaps setting the decimal places to 1 or even 0 could be useful here.

snapshots/agriculture/2024-05-23/fogel_2004.csv.dvc
```diff
@@ -0,0 +1,30 @@
+# Learn more at:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+meta:
+  origin:
+    # Data product / Snapshot
+    title: The Escape from hunger and Premature Death
```
spoonerf, line 6: hunger -> Hunger

spoonerf (approved):
Looks great, very methodical, and incredibly well documented!

PR #2693, opened 2024-05-24
📊 health: ghe age-std and m-f ratio

lucasrodes:
[Tracking issue](https://github.com/owid/owid-issues/issues/1547)

We want to migrate our charts from using GHO to GHE estimates.

To this end, we need to estimate new indicators in our GHE dataset:

- Age-standardized death rates from cause self-harm. I've estimated these rates for all causes.
- Male-to-female ratio of death-rate from self-harm. I've created a new table not to affect the main table.


spoonerf (approved):
Looks good, thanks for the general tidy up too!

PR #2694, opened 2024-05-27
:sparkles: rename staging_sync to chart_sync

Marigold:
We all make mistakes. Especially when naming things...

- Rename `staging_sync` to `chart_sync`
- Rename `chart_diff` to `chart-diff` in etlwiz CLI so that we can call `etlwiz chart-sync` (this is consistent with `indicator-upgrade`)


lucasrodes (approved):
LGTM.

I've added some more renamings that I think should be there too. Feel free to change if not needed.

Marigold:
Thanks, @lucasrodes! I totally missed those 🤦

PR #2696, opened 2024-05-27
:bar_chart: switch WDI from sources to origins

Marigold:
Remove misc function `add_origins_to_wdi` and update dependency `garden/worldbank_wdi/2023-05-29/wdi` to `garden/worldbank_wdi/2024-05-20/wdi` in all steps. I used ChatGPT-generated `wdi.sources.json` to get a producer (previously `source.name`) for a source from WDI metadata. Initially, I wanted to avoid ChatGPT and just dump the full source name as producer, but it was too messy and longer than 255 characters. If we wanted to polish it even further, we could use ChatGPT to create a full origin from given source name (not sure if it's worth it though).

Here is an example of an [old chart](https://admin.owid.io/admin/charts/317/edit) and a [new chart](http://staging-site-wdi-sources-to-origins/admin/charts/317/edit). It's almost identical except for indicator's description, which has been moved to `description_from_producer`.

### DIFF
--- a/etl/steps/data/garden/worldbank_wdi/2024-05-20/wdi.py
+++ b/etl/steps/data/garden/worldbank_wdi/2024-05-20/wdi.py
@@ -1,13 +1,12 @@
 import json
 import re
 import zipfile
-from copy import deepcopy
 from pathlib import Path
 from typing import Any, Dict, Optional
 
 import pandas as pd
 import structlog
-from owid.catalog import Source, Table, VariableMeta
+from owid.catalog import Table, VariableMeta
 from owid.catalog.utils import underscore
 
 from etl.data_helpers import geo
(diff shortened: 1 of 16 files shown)


etl/steps/data/garden/worldbank_wdi/2024-05-20/wdi.py
```diff
@@ -84,18 +82,12 @@ def mk_omms(table: Table) -> Table:
...
-        owid_data_url=source.owid_data_url,
-        date_accessed=source.date_accessed,
-        publication_date=source.publication_date,
-        publication_year=source.publication_year,
-        published_by=f"Our World in Data based on {source.published_by}",
-    )
+    omm_origin = table["sp_pop_totl"].metadata.origins[0].copy()
+
+    # rewrite "World Development Indicators" to "Our World in Data based on World Bank"
+    omm_origin.title = "Our World in Data based on World Bank"
```
pabloarosado, line 88: Why this? The title of the data product should be "World Development Indicators". Also, we shouldn't use "Our World in Data based on..." anymore.

etl/steps/data/garden/worldbank_wdi/2024-05-20/wdi.py
```diff
@@ -110,7 +102,7 @@ def mk_omms(table: Table) -> Table:
         title="Urban population living in slums",
         description=(
             "Total urban population living in slums. This variable is calculated"
-            f" by Our World in Data based on the following variables from {source.published_by}:"
+            f" by Our World in Data based on the following variables from {published_by}:"
             f' "{table[urb_pop_code].metadata.title}"'
             " and"
             f' "{table[slums_pct_code].metadata.title}"'
```
pabloarosado, line 108: Sometimes (e.g. omm_urb_pop_slum) the added description is "None". It would be better if this text is only added if the description is not None (but I'm no sure at this point if this can affect other things down the line).

etl/steps/data/garden/worldbank_wdi/2024-05-20/wdi.py
pabloarosado: And now that we are modifying the origin of each individual indicator, why not editing the full citation too, to be consistent? We could simply prepend the `producer` to the current full citation.

pabloarosado (approved):
I suggested some small changes, apart from that, it looks good.

etl/steps/data/garden/worldbank_wdi/2024-05-20/wdi.py
Marigold: Good idea, thanks!

PR #2697, opened 2024-05-27
:hammer: refactor metadata classes

Marigold:
Move all utility methods into a `Base` class.


lucasrodes (approved):
So clean, thanks! 😍

Maybe can rename this to `MetaBase` instead? Not really urgent tho...

PR #2706, opened 2024-05-29
✨ data workflow: various enhancements

lucasrodes:
### AI-tools maintenance
- Combine "Expert" and "Datasette Oracle"
- Add GPT4o
- Update cost estimates
- Remove "Meta Expert"

### Wizard theme
- Align theme of Wizard with OWID's site

### Chart diff
- Add reject option

### Indicator upgrader
- Fix error distribution tab
- Improve ergonomics
- Show score only if datapoints changed
- Show confusion matrix if indicator has few values or is categorical

### owid bot
- Support `reject` option

### Others
- Bump streamlit to 1.35 version


lucasrodes:
@Marigold I've added a reject option to chart-diff. I think we should now go and add support on the chart-sync side for this.

As we've discussed in other threads, I think that the behaviour of chart-sync should be:

- check if _all charts_ have been reviewed (either approved or rejected)
- if some is pending (i.e. `approval_status='pending'`), owidbot should raise an error and CI/CD should fail (as now).
- once the PR is merged, only those charts with `approval_status='approved'` should be merged (i.e. overwrite master counterparts if existent). The remaining ('rejected' or 'pending' if user ignored the CI/CD error) should be dropped.

Marigold (approved):
Looks good! I've added rejection logic to owidbot and chart-sync and tested it on this PR. Could you take a look at my commits and check if they're ok? If yes, we can test this by rejecting all charts, getting ✅ and merging.

PR #2707, opened 2024-05-29
:bug: get proper container name when using weird branch names

Marigold:
Fix `STAGING=fix/wizard-3 etlr ...` and `STAGING=1` with branch names containing `/` and other unusual characters.


lucasrodes (approved):
Thanks! LGTM!

PR #2711, opened 2024-05-29
:hammer: set override=False when loading env

Marigold:
Don't override system environment variables by those from `.env` file. We've [enabled it](https://github.com/owid/etl/pull/2272) some time ago for reasons no one remembers. It's unintuitive behaviour. Imagine I use `STAGING=mojmir` in my `.env` file and I run `STAGING=other etlr ...`. I'm expecting that the env variable in front of the command has higher priority, but it gets overwritten.

(I want to have `STAGING=1` in my `.env` and optionally set `STAGING=mojmir` to override it.)


lucasrodes (approved):
lgtm!

PR #2712, opened 2024-05-30
:bug: wrap raw queries in text()

Marigold:
I saw some errors in BugSnag about raw queries not being wrapped in `text(...)` which is needed by SQLAlchemy. This PR fixes that and add proper type hints to get a warning in the future.


lucasrodes (approved):
thanks, Mojmir!

PR #2721, opened 2024-05-31
📊  cigarette data

antea04:
Branch to fix issue: owid/owid-issues#1490 - cigarette sales data has to be completely replaced, since data is wrong for many countries and not in etl yet.

Tasks to import data into etl:
- [x] create snapshot
- [x] import data from excel files (meadow)
  - [x] import 'standard' cases (own tab)
  - [x] import 'special' cases (multiple countries in one tab'
- [x] explore/ test data from meadow step
- [x] clean up dataset to use in grapher
  - [x] fix running years (e.g. 1951-55)
  - [x] fix footnotes etc. in data
 - [x] push dataset to grapher
 - [x] Update graph ([here](https://ourworldindata.org/grapher/sales-of-cigarettes-per-adult-per-day))

Graph linked above is updated (changed from using 'manufactured and hand-rolled cigarettes' to only 'manufactured cigarettes'.
Good to know: countries shown per default have only minor changes in the data but other countries (e.g. Netherlands, Australia) have now correct underlying data & visualization


snapshots/smoking/2024-05-30/cigarette_sales.zip.dvc
```diff
@@ -0,0 +1,28 @@
+# Learn more at:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+meta:
+  origin:
+    # Data product / Snapshot
+    title: International Smoking Statistics
+    date_published: "2016-12-19"
```
pabloarosado, line 7: Oh wow! I didn't know that this dataset has not been updated since 2016! In such cases, it's good to consider whether putting much effort into the dataset is worth it. Have you checked if there is an alternative, more recent dataset on this topic elsewhere?

snapshots/smoking/2024-05-30/cigarette_sales.zip.dvc
```diff
@@ -0,0 +1,28 @@
...
+    title_snapshot: International Smoking Statistics - Cigarette Sales
```
pabloarosado, line 8: The field `title_snapshot` is only necessary if there are multiple snapshots for the same data product.
For example, [the Global Carbon Budget](https://github.com/owid/etl/tree/master/snapshots/gcp/2023-12-12) has multiple files, so the title of all snapshot files is `Global Carbon Budget`, but each file has a `title_snapshot`, e.g. `Global Carbon Budget - Fossil CO2 emissions`.
In this case, as far as I can tell, there is only one (and I suppose we are not expecting to create more). So `title_snapshot` is not required. Alternatively, you could create a file for each country, but that would be too many files, which is annoying, so a zip file is a much better choice (as you did).

snapshots/smoking/2024-05-30/cigarette_sales.zip.dvc
```diff
@@ -0,0 +1,28 @@
...
+
+    # Citation
+    producer: Barbara Forey, Jan Hamling, John Hamling, Alison Thornton, Peter Lee
```
pabloarosado, line 11: The field `producer` should be the name of the institution. If it was a paper, it would be something like `Forey et al.`.
In this case, it's neither an institution nor a paper. So I think it should be "P N Lee Statistics and Computing Ltd" (which is not a great name, in my opinion, but that's what their main page seems to show!). Alternatively, `Forey et al.` would be a fine choice (I leave it up to you).

snapshots/smoking/2024-05-30/cigarette_sales.zip.dvc
```diff
@@ -0,0 +1,28 @@
...
+    attribution: International Smoking Statistics (2016)
+    citation_full: |-
+      International Smoking Statistics WEB Edition
+      © Barbara Forey, Jan Hamling, John Hamling, Alison Thornton, Peter Lee, 2006-2016
```
pabloarosado, line 15: Do they explicitly ask to use this citation? If yes, then it's fine (we should follow the data providers' recommendations). If not, I think the copyright does not belong here. We have a field for license below (although it's currently not surfaced in data pages, but it should be at some point).
In this case, the full citation should mention the producer (as well as the authors) and the title of the data product.
So, I'd suggest:
P N Lee Statistics and Computing Ltd - International Smoking Statistics WEB Edition.
Barbara Forey, Jan Hamling, John Hamling, Alison Thornton, Peter Lee (2006-2016).

snapshots/smoking/2024-05-30/cigarette_sales.zip.dvc
```diff
@@ -0,0 +1,28 @@
...
+
+    # Files
+    url_main: http://www.pnlee.co.uk/ISS.htm
+    date_accessed: 2024-05-30
+
+    # License
+    license:
+      name: © Barbara Forey, Jan Hamling, John Hamling, Alison Thornton, Peter Lee, 2006-2016
```
pabloarosado, line 23: The license url should be the place where you find the copyright. In this case: http://pnlee.co.uk/ISS.htm

pabloarosado (changes requested):
Hi @antea04! I haven't finished the review (I will later). So far it looks very good! But for now I have already some suggested changes, so you can start working on them.
I have also noticed [in chart diff](http://staging-site-cigarette-data/etl/wizard/chart-diff) that the USSR extends unil 2016, which seems to be a mistake (the data should probably end in 1991).

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+                "Year",
+                "Manufactured cigarettes - Millions",
+                "Manufactured cigarettes - Number/adult/day",
+            ]
+            for col in addl_cols:
+                tb_from_excel[col] = "NaN"
+            tb_from_excel = tb_from_excel.head(format["num_rows"])
+
+        # drop empty columns
+        tb_from_excel.drop(labels=["Unnamed: 3", "Unnamed: 6", "Unnamed: 9", "Unnamed: 12"], axis=1, inplace=True)
```
pabloarosado, line 229: I think it's clearer to use `.drop(columns=...)`. Also, it's safer to add `errors="raise"` to ensure that the code will fail if the columns are not found (which could mean that the data is not as expected).

pabloarosado: Also, it's usually better practice to do `tb_from_excel = tb_from_excel.some_operation(inplace=True)` rather than `tb_from_excel.some_operation(inplace=True)`. It's clearer that you are replacing `tb_from_excel` by something different.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+
+        # change header and remove sub-header columns
+        tb_from_excel.drop(labels=[0, 1], axis=0, inplace=True)
+        tb_from_excel.columns = NEW_COLUMNS
```
pabloarosado, line 233: In general, it's safer to rename columns rather than assigning column names this way. Imagine if, on a new dataset update, the order of columns change. But in this case (especially, given that the dataset may never be updated again), it's fine.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+
+        tb_from_excel["country"] = cty
+        country_tables.append(tb_from_excel)
```
pabloarosado, line 236: I think `append` is going to be deprecated. You can use `concat` instead.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+    "Unnamed: 8",
+    "Unnamed: 9",
+    "All tobacco products",
+    "Unnamed: 11",
+    "Unnamed: 12",
+]
+
+
+def is_not_string(value):
+    return not isinstance(value, str)
```
pabloarosado, line 142: I don't think this function is necessary. You are creating more code, and simply having `not isinstance(value, str)` is clear enough.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+        tb_from_excel.columns = NEW_COLUMNS
+
+        tb_from_excel["country"] = cty
+        country_tables.append(tb_from_excel)
+
+    # concatenate tables and delete rows without data
+    smoking_data_tb = pr.concat(country_tables)
+
+    # remove repeating headers
+    smoking_data_tb = smoking_data_tb[smoking_data_tb["year"].notna()]
```
pabloarosado, line 242: Why not `smoking_data_tb = smoking_data_tb.dropna(subset="year")`?

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+        "Unnamed: 6",
+        "Unnamed: 7",
+        "Unnamed: 8",
+        "Unnamed: 9",
+        "Unnamed: 10",
+        "Unnamed: 11",
+        "Unnamed: 12",
+    ]
+
+    tb_from_excel = Table()
```
pabloarosado, line 175: I think this definition is not necessary, since you define tb_from_excel below on the first iteration.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+
+        elif cty == "USSR and fSU":
+            tb_from_excel = pr.read_excel(
+                zf.open("{}{}".format(folder_name, cty_sheet["file_name"])),
+                sheet_name=cty_sheet["sheet_name"],
+                metadata=snap_meta,
+                origin=snap.metadata.origin,
+                header=9,
+            )
+            tb_from_excel["Unnamed: 4"] = "NaN"
```
pabloarosado, line 206: Why not using `pd.NA` or `np.nan` instead?

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+        country_tables.append(tb_from_excel)
+
+    # concatenate tables and delete rows without data
+    smoking_data_tb = pr.concat(country_tables)
+
+    # remove repeating headers
+    smoking_data_tb = smoking_data_tb[smoking_data_tb["year"].notna()]
+    smoking_data_tb = smoking_data_tb[smoking_data_tb["manufactured_cigarettes_millions"].apply(is_not_string)]
+
+    # change data types to string
```
pabloarosado, line 245: It would be good to explain why year needs to be a string (because the most logical choice would be int). Give some examples of the content of "year" in the comment, so that it's clear why you use str.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+
+        # change header and remove sub-header columns
+        tb_from_excel.drop(labels=[0, 1], axis=0, inplace=True)
+        tb_from_excel.columns = NEW_COLUMNS
+
+        tb_from_excel["country"] = cty
+        country_tables.append(tb_from_excel)
+
+    # concatenate tables and delete rows without data
+    smoking_data_tb = pr.concat(country_tables)
```
pabloarosado, line 239: Minor suggestion (up to you) my preference would be `tb_smoking`, instead of `smoking_data_tb`. In general, having "data" as part of the name of a table is redundant.

pabloarosado (changes requested):
A few more suggestions (feel free to disregard them if you disagree).

pabloarosado (changes requested):
The review is finished, great work! I have been very picky and made many suggestions. Some of them are generic, just to let you know common conventions we follow. Feel free to disregard minor suggestions, and please let me know if you disagree or have further questions.

snapshots/smoking/2024-05-30/cigarette_sales.zip.dvc
edomt: Hi @pabloarosado — there's a lot of hidden context behind this update, all explained [here](https://github.com/owid/owid-issues/issues/1490). In a nutshell, @spoonerf and I specifically selected this dataset because we really need to fix it _and_ ti's good practice for @antea04. So no worries here!

antea04: Like Ed said (and mentioned via Slack) - as an exercise and also since this data set is referenced in a few different places around the site it was useful to fix it - but I do agree that more recent data would also be interesting!

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,168 @@
...
+                year_int = int(year.split("/")[0])
+            if year_int > 0:
+                dict_from_row["year"] = year_int
+                new_df.append(dict_from_row)
+                continue
+            elif "-" in year:  # timeframe given in excel
+                timeframe = year.split("-")
+                start_year = int(timeframe[0])
+                end_year = int(timeframe[1])
+                if end_year < 100:
```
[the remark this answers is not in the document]
antea04: Yes, I agree! I added comments - I hope it is clearer now!

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,168 @@
...
+    # reorder columns
+    tb = tb[["year", "country"] + COLS_WITH_DATA]
+
+    # drop rows with no data (nan values)
+    tb = tb.dropna(how="all", subset=COLS_WITH_DATA)
+
+    # drop rows with zeros and NaNs
+    tb_temp = tb.fillna(0)
+    rows_to_drop = tb_temp[(tb_temp[COLS_WITH_DATA] == 0).all(axis=1)].index
+    tb = tb.drop(rows_to_drop)
```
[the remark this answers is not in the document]
antea04: yes, the data set is a bit messy - some rows are completely empty (only NaN values) and some have zeros in some columns and NaN in the rest, so I am removing those different types one by one. Let me know if it is clearer now (I adjusted the comments)!

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,168 @@
...
+    # load population data
+    ds_population = paths.load_dataset("population")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow["cigarette_sales"].reset_index()
+
+    # Process data.
+
+    # Fix years column (change dtype to integer and expand timeframes)
+    df_years_fixed = Table(standardise_years(tb), metadata=tb.metadata)
```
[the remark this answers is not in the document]
antea04: I am assuming this is re: line 118 and I totally agree :D Let me know if you meant something else.

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,168 @@
...
+    tb_temp = tb.fillna(0)
+    rows_to_drop = tb_temp[(tb_temp[COLS_WITH_DATA] == 0).all(axis=1)].index
+    tb = tb.drop(rows_to_drop)
+
+    # cast columns to float
+    for col in COLS_WITH_DATA:
+        tb = cast_to_float(tb, col)
+
+    # harmonize countries (also exclude east and west germany)
+    tb = geo.harmonize_countries(
```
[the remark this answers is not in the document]
antea04: Yeah, that makes sense! I excluded them from the table after adding the combined column - I was try to sort of "preserve" them in case somebody is interested specifically in data from East/ West Germany but I agree that the trade-off there is probably not worth it.

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,168 @@
...
+                    dict_from_row = row.to_dict()
+                    dict_from_row.update({"year": year_in_timeframe})
+                    new_df.append(dict_from_row)
+    return pd.DataFrame(new_df)
+
+
+def include_split_germany(tb, ds_population):
+    """Include data for Germany 1945-1990 in the table by taking weighted average of East and West Germany data"""
+    germany_tb = tb[tb["country"].isin(["West Germany", "East Germany"])]
+    germany_tb = geo.add_population_to_table(germany_tb, ds_population, interpolate_missing_population=True)
```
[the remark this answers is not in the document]
antea04: I actually thought about it when doing it this way - I decided against it for two reasons:
- They are using WHO estimates for their population figures which I don't think correspond 1:1 to our population data (feel free to correct me if I am wrong here!) and I didn't know how big the effect from that change would be
- This way I am not using any absolute populations but only the share/ ratio between East/ West Germany which - even as the dataset is sparse - does not change much over that timeframe (from like 74:26 to 79:21), so I assume the effect of the interpolation to be a small error overall. 

Let me know if I have a mistake in my thinking - otherwise I would leave the function as is

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
antea04: I totally agree in principle - this dataset was just so inconsistent (column names were using upper/ lowercase letters/ had typos/ had footnotes) that catching all the "edge cases" would have made the code even messier. So I agree that it is not the best practice, but I would keep it here as is (as you also said)

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
antea04: country_tables is a list, so I think this is the way to do it (I append tb_from_excel as new element at the end of the list) - but this comment reminds me that I should give the variable a better name so that is immediately clear

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
antea04: I do agree - I think the linter (Ruff?) complained when I tried to commit/ push it like that (I think since I assign tb_from_excel within the for loop that is not strictly guaranteed to be run, even if it always will be). I can figure out what exactly was the problem here in case we want to adjust the global linter config - but for now I would leave the definition here.

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
```diff
@@ -0,0 +1,265 @@
...
+        "Unnamed: 12",
+    ]
+
+    tb_from_excel = Table()
+
+    for cty, cty_sheet in COUNTRY_MAP.items():
+        if cty not in SPECIAL_CASES:
+            # open excel file
+            tb_from_excel = pr.read_excel(
+                zf.open("{}{}".format(folder_name, cty_sheet["file_name"])),
```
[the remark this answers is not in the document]
antea04: oh amazing - that's so cool! 😎

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.meta.yml
```diff
@@ -0,0 +1,115 @@
...
+  update_period_days: 365
+
+
+tables:
+  cigarette_sales:
+    variables:
+      manufactured_cigarettes:
+         title: Annual sales of manufactured cigarettes
+         unit: cigarettes
+         short_unit: cigarettes
```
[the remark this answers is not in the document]
antea04: this also affects the way it is presented in the chart (on the axis), right? Because that is the reason I kept it as a short unit - I think it could be confusing if the axis did not have that label as I would maybe assume volume of cigarettes sold to be in packs/ per weight than actual cigarettes. Or does it then default to "unit" in the chart?

(This is all theoretical as we are not using this indicator in a graph as of now but also for the future.)

etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
antea04: So if i comment this line out and run "make test" in the console it fails with the following error:
/Users/tunaacisu/Code/etl/etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py
  /Users/tunaacisu/Code/etl/etl/steps/data/meadow/smoking/2024-05-30/cigarette_sales.py:226:25 - error: "tb_from_excel" is possibly unbound (reportUnboundVariable)
1 error, 0 warnings, 0 informations 
Completed in 9.855sec
make: *** [check-typing] Error 1

If this is unwanted behavior, maybe this can be adjusted - but for now I found it easiest to just define tb_from_excel before the for-loop starts

etl/steps/data/garden/smoking/2024-05-30/cigarette_sales.meta.yml
```diff
@@ -0,0 +1,115 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    processing_level: minor
```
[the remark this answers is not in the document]
pabloarosado: That's an open discussion. I had a strict criterion on this, but we couldn't reach an agreement. In many cases, the data is identical to the original, so that should be mino processing. In this case, the data is not identical, for example due to the processing of historical regions like Germany. So that to me is enough to call it major. But again, there are no clear rules for now on this choice.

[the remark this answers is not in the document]
antea04: okay, that makes sense to me! Will follow that logic in the future!

pabloarosado (approved):
Great work! Thanks for the changes. The only thing left I noticed is that the garden step still throws a warning. This happens because East and West Germany still are in the excluded countries file. Simply remove that file (and then don't mention it when harmonizing country names). Apart from that, this PR is ready to go!

PR #2727, opened 2024-06-04
:bar_chart: migrate GHE from sources to origins

Marigold:
Fix for https://github.com/owid/owid-issues/issues/1564. Adding origins should enable Data pages for GHE.

I didn't spend much time improving the metadata and data pages, so they look pretty terrible. Check out [this one](http://staging-site-migrate-ghe-to-origins/admin/datapage-preview/527002) for example. There's a long title, no default countries, etc. It's not worse than what it used to be, though. Should we invest time in improving it, or do we just merge it as it is and fix it in the next update?

## Changes
- use snapshot instead of walden
- refactor code
- move metadata to origin at snapshot level


spoonerf:
Ah, thanks for doing this! 

Hmm, yeah, the data pages don't look great, do they? I'm off the rest of the week, but next week I can have a bash at adding some Jinja templating, which wouldn't be much work (just a couple of hours, I guess?) and probably go quite a long way in terms of improving the basic metadata.

Marigold:
That'd be great! This is definitely not urgent.

spoonerf (approved):

spoonerf:
@Marigold I've had a stab at improving the metadata, I think it's improved enough for this time round, but happy to make some more changes if you think it's worth it.

Issue #2735, opened 2024-06-05
🐛 Running `etl --help` crashes because of metagpt


pabloarosado:
When running `etl --help` on master, I get the following error:
```
(etl-py3.11) ➜  etl git:(fix-etl-help-crash) ✗ etl --help

 Usage: etl [OPTIONS] COMMAND [ARGS]...

 Run OWID's ETL client.
 Create ETL step templates, compare different datasets, generate dependency visualisations, synchronise charts across different servers,
 import datasets from non-ETL OWID sources, improve your metadata, etc.

 Note: For a UI experience, refer to CLI etlwiz.

╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help      Show this message and exit.                                                                                               │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Run ETL steps ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ run  Generate datasets by running their corresponding ETL steps.                                                                      │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Data ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ harmonize        Generate a dictionary with the mapping of country names to OWID's canonical names.                                   │
│ diff             Compare all datasets from two catalogs and print out a summary of their differences.                                 │
│ graphviz         Generate a Graphviz DOT file to see all dependencies.                                                                │
│ compare          Compare two dataframes/tables/datasets in terms of their structure, values and metadata.                             │
│ update           Update one or more steps to their new version, if possible.                                                          │
│ archive          Archive one or more steps.                                                                                           │
│ explorer-update  Update one or more explorer (tsv) files.                                                                             │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
2024-06-05 10:05:48,755 - [bugsnag] WARNING - No API key configured, couldn't notify
Traceback (most recent call last):
  File "/Users/prosado/Documents/owid/repos/etl/.venv/bin/etl", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/prosado/Documents/owid/repos/etl/.venv/lib/python3.11/site-packages/rich_click/rich_command.py", line 367, in __call__
    return super().__call__(*args, **kwargs)
[...]

PR #2740, opened 2024-06-05
🔨 gpt utils

lucasrodes:
- Remove duplicate GPT utils code
- Fix https://github.com/owid/etl/issues/2735: Remove `@dataclass` from class definition. This is causing errors likely because it is incompatible for the class to be a `dataclass` and inherit from `ChatCompletion` at the same time.


pabloarosado (approved):

PR #2743, opened 2024-06-05
:tada: save checksums for indicators in MySQL

Marigold:
PR with MySQL migration https://github.com/owid/owid-grapher/pull/3690

Right before uploading data & metadata to R2 as JSON files, calculate checksums for data and metadata. If checksums are matching ones in MySQL, it means that the data is identical and there's no need to upload the same files.

Checksums will be added as we refresh datasets. We could also bump `ETL_EPOCH` and populate all datasets, but perhaps only after making sure this concept works.

For metadata, we exclude fields `updatedAt`, `dataChecksum`, `metadataChecksum` (otherwise we'd always get a different checksum).

## Performance

Tested on two staging servers with command
```
time poetry run etlr grapher://grapher/ihme_gbd/2024-05-20/impairments --force --private --grapher --only
```

Calculating checksums doesn't take much time (I ran a few benchmarks and it turned out that computing checksum from JSON we're about to upload to R2 is the fastest. Still, it's a fraction of the entire process). If checksums match and we don't reupload files, it can speed up grapher:// step by ~50% (this could be even more pronounced if you have poor internet connection).

## Links
My previous attempt for improving grapher upsert performance could be relevant. https://github.com/owid/etl/pull/1564

### DIFF
--- a/apps/backport/datasync/data_metadata.py
+++ b/apps/backport/datasync/data_metadata.py
@@ -98,7 +98,7 @@ def add_entity_code_and_name(session: Session, df: pd.DataFrame) -> pd.DataFrame
         missing_entities = set(unique_entities) - set(entities.entityId)
         raise ValueError(f"Missing entities in the database: {missing_entities}")
 
-    return pd.merge(df, entities, on="entityId")
+    return pd.merge(df, entities.astype({"entityName": "category", "entityCode": "category"}), on="entityId")
 
 
 def variable_data(data_df: pd.DataFrame) -> Dict[str, Any]:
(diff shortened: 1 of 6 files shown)


pabloarosado (approved):
Awesome! The code looks good (just a small thing, in the grapher migration, you defined the checksum columns as VARCHAR(60), and here as 64).
Have you tested it on a big dataset, like FAOSTAT? If not, let me know, I can help.

Marigold:
Yeah, I tested it on GBD (which is large) and now on FAO as per your advice and it runs as fast as before.

PR #2746, opened 2024-06-06
🎉 Ensure steps affecting `external` channel are not archivable

pabloarosado:
* Ensure that steps that affect any step in the `external` channel are not archivable.
* Move the logic of the update status of steps from StepUpdater to VersionTracker, and ensure the archiving logic of StepUpdater and VersionTracker are aligned.
* Include the update statuses "Archived" (for steps that have already been archived) and "Not yet used" (for steps that may be in their latest version, but not used in any charts or by any external channel).
* Move a few things around for consistency, and fix broken test.

NOTE: This will make VersionTracker a bit slower, but hopefully not too much. If so, we can figure out ways to speed some things up.

### DIFF
--- a/apps/step_update/cli.py
+++ b/apps/step_update/cli.py
@@ -1,7 +1,6 @@
 import difflib
 import sys
 from datetime import datetime
-from enum import Enum
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


Marigold (approved):
Code looks good!

PR #2750, opened 2024-06-06
🐝 Downgrade ipython

pabloarosado:
For some reason, ipython 8.18.0 was not working (it seems to be incompatible with prompt_toolkit).


Marigold (approved):
Thanks for fixing this!

PR #2752, opened 2024-06-06
:tada: show chart diffs with changed checksums

Marigold:
## Motivation

Chart-diff currently only shows charts with changed config (due to updating variable ids by indicator upgrader). We'd like to have an option to show all charts with changes - either coming from change to the underlying data or metadata.

## Summary

Use data and metadata checksums to optionally show chart differences, in addition to just changed chart configurations. This information is displayed as tags next to a chart slug. Currently, only charts with changed configurations are shown by default, and approving them is enough for owidbot to give you a ✅. However, if desired, you can enable the option to show data and metadata changes in the filters.

The new function `modified_charts_by_admin` assumes that all charts or datasets were created by an Admin user with ID=1. This is a temporary solution until we determine how to display the real name of the creator.

I think we should carefully test this before making it the default. I am concerned that updating datasets, such as regions or population, would trigger a review for almost all charts.

image

image

## How to test it

Go to chart-diff on the staging server and select "metadata" in filters. You should see a cherry blossom chart with updated metadata.

@lucasrodes I noticed that warnings have been turned off. Is that expected? I want to show a warning if `GRAPHER_USER_ID` isn't Admin.

## TODO before merging:
- [x] undo modifications to cherry blossom

### DIFF
--- a/apps/wizard/app_pages/chart_diff/app.py
+++ b/apps/wizard/app_pages/chart_diff/app.py
@@ -9,7 +9,7 @@ from sqlalchemy.orm import Session
 # from st_copy_to_clipboard import st_copy_to_clipboard
 from structlog import get_logger
 
-from apps.chart_sync.cli import _modified_chart_ids_by_admin
+from apps.chart_sync.cli import modified_chart_ids
 from apps.wizard.app_pages.chart_diff.chart_diff import ChartDiff
 from apps.wizard.app_pages.chart_diff.show import st_show
 from apps.wizard.app_pages.chart_diff.utils import WARN_MSG, get_engines
@@ -81,7 +81,8 @@ def get_chart_diffs_from_grapher(max_workers: int = 10) -> dict[int, ChartDiff]:
     """
     # Get IDs from modified charts
     with Session(SOURCE_ENGINE) as source_session:
-        chart_ids = _modified_chart_ids_by_admin(source_session)
+        with Session(TARGET_ENGINE) as target_session:
+            chart_ids = modified_chart_ids(source_session, target_session)
 
     # Get all chart diffs in parallel
     with ThreadPoolExecutor(max_workers=max_workers) as executor:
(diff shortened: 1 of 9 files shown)


apps/wizard/app_pages/chart_diff/chart_diff.py
```diff
@@ -78,7 +87,7 @@ def slug(self) -> str:
         """
         if self.target_chart:
             assert self.source_chart.config["slug"] == self.target_chart.config["slug"], "Slug mismatch!"
-        return cast(str, self.source_chart.config["slug"])
+        return self.source_chart.config.get("slug", "no-slug")
```
lucasrodes, line 90: when is the case that a chart won't have a slug?

etl/steps/data/garden/biodiversity/2024-01-25/cherry_blossom.meta.yml
```diff
@@ -1,5 +1,5 @@
 dataset:
-  update_period_days: 365
+  update_period_days: 368
```
lucasrodes, line 2: is this value change expected? or was it only to test things out?

Marigold: Testing things, I'll remove it before merging.

apps/wizard/app_pages/chart_diff/app.py
```diff
@@ -47,8 +48,15 @@
...
-# st.warning("- " + "\n\n- ".join(warn_msg))
+# WARN_MSG += ["This tool is being developed! Please report any issues you encounter in `#proj-new-data-workflow`"]
+
+if config.GRAPHER_USER_ID != 1:
+    WARN_MSG.append(
+        "Your `GRAPHER_USER_ID` is not set to 1 (Admin). Please start using STAGING or change it in your env file."
+    )
+
+if WARN_MSG:
+    st.warning("- " + "\n\n- ".join(WARN_MSG))
```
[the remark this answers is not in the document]
Marigold: Added more info.

apps/wizard/app_pages/chart_diff/chart_diff.py
Marigold: I created a new draft chart for testing and it didn't have slug for some reason. It's an edge case which won't happen in real life, but I had to fix it for development.

PR #2754, opened 2024-06-06
🎉 Create etl cli tool to start a new pr server from scratch

pabloarosado:
(NOTE: I updated the logic of the script in my last commit)
With this script, you can just run "etl d start-server --new-branch some-branch-name" and it will create a new draft PR (with its corresponding staging server). What do you think?


apps/utils/start_server.py
```diff
@@ -0,0 +1,121 @@
...
+    type=str,
+    default=None,
+    help="Name of new branch to create (if not given, current branch will be used).",
+)
+@click.option(
+    "--base-branch", type=str, default=None, help="Name of base branch (if not given, 'master' will be used)."
+)
+def cli(new_branch: Optional[str] = None, base_branch: Optional[str] = None) -> None:
+    if not GITHUB_TOKEN:
+        log.error("Define GITHUB_TOKEN in your ETL .env file. Then run this tool again.")
```
Marigold, line 48: Can we either show full instructions how to create one or at least inform to run `etl d start-server --help` to get them?

apps/utils/start_server.py
```diff
@@ -0,0 +1,121 @@
...
+Usage:
+- If you use this script from the master branch on your local ETL repo, specify a --new-branch argument.
+  This will create a new local branch, which will be pushed to remote.
+- If you have already created a new local branch, run this script without specifying any argument.
+  The 'master' branch will be assumed as the base branch of the new draft pull request.
+- If you want the base branch to be different from 'master', specify it with the --base-branch argument.
+
+The resulting draft pull request will be new_branch -> base_branch.
+
+To create a GitHub token, go to GitHub settings. On the left, at the bottom, choose Developer settings.
```
Marigold, line 12: You can just use a link here https://github.com/settings/tokens

apps/utils/start_server.py
```diff
@@ -0,0 +1,121 @@
...
+    log.info("Creating an empty commit.")
+    repo.git.commit("--allow-empty", "-m", "Start a new staging server")
+
+    log.info("Pushing the new branch to remote.")
+    repo.git.push("origin", new_branch)
+
+    log.info("Creating a draft pull request.")
+    headers = {"Authorization": f"token {GITHUB_TOKEN}", "Accept": "application/vnd.github.v3+json"}
+    data = {
+        "title": f":construction: Draft PR for branch {new_branch}",
```
Marigold, line 111: I'd appreciate having an option to specify PR title and message of the first commit (they tend to be identical quite often). For instance
```
etl d start-server --new-branch my-branch -c ":bar_chart: My dataset update"
```

apps/cli/__init__.py
```diff
@@ -73,6 +73,7 @@ def _lazy_load(self, cmd_name):
             "run-python-step": "etl.run_python_step.main",
             "map-datasets": "apps.utils.map_datasets.cli",
             "scan-chart-diff": "apps.utils.scan_chart_diff.cli",
+            "start-server": "apps.utils.start_server.cli",
```
Marigold, line 76: We should think twice about naming it right. I don't have a strong preference and will leave it up to you (`start-server` is good, don't get me wrong!). Here are my 2c:
- I can imagine having other server-specific commands, for instance `refresh-server` or `destroy-server` (these would be just wrappers around `owid-lxc` commands with better instructions). Maybe we could have a group of commands `server start`, `server refresh`, etc.?
- It seems like the important thing here is to create a PR, starting a server is a side effect, so perhaps `create-pr` name would be more intuitive? (this goes against my previous point, feel free to pick one...)

This is a great idea! The only risk is that people will start creating servers like crazy and I'll have to start destroying them sooner than after 7 days (that's gonna change anyway with the chart-diff workflow).

Marigold (approved):
This is a great idea! The only risk is that people will start creating servers like crazy and I'll have to start destroying them sooner than after 7 days :) (that's gonna change anyway with the chart-diff workflow).

apps/utils/start_server.py
pabloarosado: Hmm, I thought it's a good idea to have a generic title for the draft PR. This way it's easier to spot if some Draft PR was forgotten.
Then, when making it a real PR, ready for review, the user will have to assign a proper title to it. What do you think? Assigning the title from the terminal is also a bit inconvenient. But if you think we should have the option, I'm happy to do that.

apps/cli/__init__.py
pabloarosado: Given that, for now, we don't have any other server operations, I've renamed it `etl d draft-pr`. We can merge it for now like this and then discuss it with Lucas on Monday, and possibly rename it.

apps/utils/start_server.py
Marigold: Changing the title in PR is fine, but changing the commit message (to the same title) is annoying. Having an option to set both in CLI would be ultimate.

> This way it's easier to spot if some Draft PR was forgotten.

I don't think that's a big deal. We don't have that many PRs anyway.

apps/cli/__init__.py
Marigold: Sounds good!

PR #2760, opened 2024-06-07
:hammer: improve memory usage of GBD dataset

Marigold:
I found that GBD dataset converts some of categorical columns into objects. This exploded memory and was a likely cause for recent outages of `foundation-1`. It went deeper and the conversion to objects happened in `data_helpers` modules. This PR is more of a hotfix. It reduces memory usage from tens of GB to <10GB and speed it up 2x. I plan to temporarily increment ETL_EPOCH to trigger full ETL rebuild on this staging server to make sure other datasets are not affected.

Pandas is unpredictable when it comes to categoricals. Native operations such as `concat` or `merge` sometimes convert categorical variables to objects. Since we already have `pr.concat` and `Table.merge`, it might be helpful to fix handling of categorical variables there (it could be as simple as converting it back to category after doing the operation). What do you think @pabloarosado?

Another alternative would be to start using `string` type that should be more efficient. However, this would mean changing a lot of datasets or completely rewriting helper functions.

## TODO before merging
- [x] revert ETL_EPOCH to original value


pabloarosado (approved):
Thanks for fixing this, @Marigold!
The code looks good, but I'm a bit worried about touching the geo functions. They are used everywhere, and they are very tricky (and handling categoricals very often leads to complicated issues). Would it be possible to just fix things for `ihme_gbd`, for now?
Then, once we have the indicator checksums implemented, we can improve `pr.concat` properly, to handle categoricals, and increase `ETL_EPOCH`, with better visibility of what is changing. What do you think?

PR #2767, opened 2024-06-07
✨ Improve selection of old and new datasets in indicator upgrader

pabloarosado:
This PR improves the experience of choosing datasets for indicator upgrader, to avoid unexpected changes that could lead to wrong dataset mappings.

The following conditions must be fulfiled:
* When (re)starting the indicator upgrader page, the new datasets dropdown shows by default a new dataset that has not yet been mapped (if any).
* Switching "Show step names" on and off should never change the choices of the dropdowns.
* Switching "Show all datasets" on does not change the choices of the dropdowns.
  * However, switching it off will make the new dataset dropdown show a new dataset, and the old dataset dropdown will show its corresponding old dataset.
* Changing the choice of old dataset should never change the choice of the new dataset.
   * However, changing the new dataset (when "Show all datasets" is off) should switch the old dataset to the corresponding one.


apps/wizard/app_pages/indicator_upgrade/dataset_selection.py
```diff
@@ -145,10 +196,9 @@ def build_dataset_form(df: pd.DataFrame, similarity_names: Dict[str, Any]) -> "S
         )
         reset_indicator_form()
 
-    # Get IDs of datasets
-    dataset_old_id = display_name_to_id_mapping[dataset_old]
-    dataset_new_id = display_name_to_id_mapping[dataset_new]
-    set_states({"migration_new_id": dataset_new_id})
+    # Get IDs of the chosen old and new datasets.
+    dataset_old_id = display_name_to_id_mapping.get(dataset_old) or display_step_to_id_mapping.get(dataset_old)
+    dataset_new_id = display_name_to_id_mapping.get(dataset_new) or display_step_to_id_mapping.get(dataset_new)
```
lucasrodes, line 201: instead, you could use:

```py
dataset_old_id = display_name_to_id_mapping.get(dataset_old, display_step_to_id_mapping.get(dataset_old))
dataset_new_id = display_name_to_id_mapping.get(dataset_new, display_step_to_id_mapping.get(dataset_new))

lucasrodes (approved):
lgtm ー let's try it on future updates and see how it behaves.

apps/wizard/app_pages/indicator_upgrade/dataset_selection.py
pabloarosado: Thanks, but I think the current option is clearer.

PR #2770, opened 2024-06-09
📊 world happiness report

antea04:
See also: owid/owid-issues#1524
Update Happiness data with 2024 world happiness report and add aggregates over continent/ age group

Goals:
- [x]  Check that the methodology of the WHR is consistent across years, and it is okay for us to combine data across multiple reports, as we currently do
  - [x] If that is okay combine the new data with the current data in the etl
- [x] Calculate population-weighted continent and income group averages and potentially a global average of the Cantril Ladder Scores
- (data for this currently not available - won't do) Import the Cantril Ladder Scores for the different age-groups and draft a chart showing the differences across age-groups

Low-level tasks
- [x] Snapshot/ import WHR 2024 data
- [x] Read methodology and whether the data is comparable over years
- [x] Figure out whether data for age group is available (reached out - no answer yet)
- [x] Continent/ income group averages calculation (garden step)
- [x] Update existing charts (check charts by indicator)
  - (data for this not available) New chart (comp over age groups)


etl/steps/data/garden/happiness/2024-06-09/happiness.meta.yml
```diff
@@ -0,0 +1,81 @@
...
+
+
+
+
+
+
+
+
+
+      # testing_variable:
```
spoonerf, line 47: Maybe this section of hashed out config can be deleted

etl/steps/data/garden/happiness/2024-06-09/happiness.py
```diff
@@ -0,0 +1,87 @@
...
+import owid.catalog.processing as pr
+
+from etl.data_helpers import geo
+from etl.helpers import PathFinder, create_dataset
+
+# Get paths and naming conventions for current step.
+paths = PathFinder(__file__)
+
+
+REGIONS = {reg: reg_dict for reg, reg_dict in geo.REGIONS.items() if reg != "European Union (27)"}
```
spoonerf, line 12: Fancy!

etl/steps/data/garden/happiness/2024-06-09/happiness.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Harmonize country names
+    tb = geo.harmonize_countries(
+        df=tb, countries_file=paths.country_mapping_path, excluded_countries_file=paths.excluded_countries_path
+    )
+
+    # Process data (add population weighted averages for continents & income groups)
+
+    # save data of Northern Cyrpus and Somaliland to concat later (they do not have population in population dataset)
+    tb_countries_wo_population = tb[tb["country"].isin(["Northern Cyprus", "Somaliland"])]
```
spoonerf, line 48: You could consider excluding these countries, but I'm also happy if you want to keep them in.

spoonerf: Also perhaps a slightly neater way to achieve the same thing would be:

```
msk = tb["country"].isin(["Northern Cyprus", "Somaliland"])
tb_countries_wo_population = tb[msk]
tb = tb[~msk]
```

spoonerf (approved):
Looks good to me! Just some minor comments on my side, but I would wait for Pablo R's review before merging. 

I'm not sure why the Marimekko chart isn't showing as expected on chart-diff, perhaps you could report this in #proj-new-data-workflow. 
Screenshot 2024-06-14 at 11 16 01

etl/steps/data/garden/happiness/2024-06-09/happiness.meta.yml
```diff
@@ -0,0 +1,81 @@
...
+        - Happiness & Life Satisfaction
+
+
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 365
+
+
+tables:
```
spoonerf, line 16: If possible, it would be good to retain the links to the previous reports ([like here](https://github.com/owid/etl/blob/41f9b1d7f40fdf19fd8270cb79d03237d5e6708a/etl/steps/data/garden/happiness/2023-03-20/happiness.meta.yml#L18)), as this dataset is a combination of many annual. I'm not sure what would be the best place to do this, though.

This information is stored in each of the Snapshots of the source files, but these do not use the Origin format. It could be a bit of a headache to fix, I'd be curious to hear what @pabloarosado thinks about it.

etl/steps/data/garden/happiness/2024-06-09/happiness.py
antea04: I would keep them in - they only show up if they get selected in the line chart and there I think it might be nice to have them available.

snapshots/happiness/2024-06-09/happiness.xls.dvc
```diff
@@ -0,0 +1,30 @@
...
+      Helliwell, J. F., Layard, R., Sachs, J. D., De Neve, J.-E., Aknin, L. B., & Wang, S. (Eds.). (2024). World Happiness Report 2024. University of Oxford: Wellbeing Research Centre.
+
+    # Files
+    url_main: https://worldhappiness.report/ed/2024/
+    url_download: https://happiness-report.s3.amazonaws.com/2024/DataForFigure2.1+with+sub+bars+2024.xls
+    date_accessed: 2024-06-09
+
+    # License
+    license:
+      name: ""
```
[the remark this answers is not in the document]
antea04: I did struggle with that because on the website there is no license mentioned or explicit instructions for how to (re-)publish the data. THe closest thing would be the citation and isbn on the report page (https://worldhappiness.report/ed/2024/). I added this now as license url - let me know if that works also!

snapshots/happiness/2024-06-09/happiness.xls.dvc
```diff
@@ -0,0 +1,31 @@
...
+    # Data product / Snapshot
+    title: World Happiness Report
+    description: |-
+      The World Happiness Report is a partnership of Gallup, the Oxford Wellbeing Research Centre, the UN Sustainable Development Solutions Network, and the WHR’s Editorial Board.
+      It reviews the state of happiness in the world today and shows how the science of happiness explains personal and national variations in happiness.
+    date_published: "2024-03-08"
+    version_producer: 2024
+
+    # Citation
+    producer: "University of Oxford: Wellbeing Research Centre"
```
pabloarosado, line 14: This could be simply "Wellbeing Research Centre". That's how they call themselves in multiple public places. And you mention Oxford in the full citation, so that should be fine.

etl/steps/data/meadow/happiness/2024-06-09/happiness.py
```diff
@@ -0,0 +1,37 @@
...
+    # Retrieve snapshot.
+    snap = paths.load_snapshot("happiness.xls")
+
+    # Load data from snapshot.
+    tb = snap.read()
+
+    # add year and country columns
+    tb["year"] = 2023  # 2024 report -> 2023 data
+    tb["country"] = tb["Country name"]
+    tb = tb.drop(columns=["Country name"])
```
pabloarosado, line 22: Instead you could simply do `tb = tb.rename(columns={"Country name": "country"}, errors="raise")`.

etl/steps/data/grapher/happiness/2024-06-09/happiness.py
```diff
@@ -0,0 +1,32 @@
...
+    #
+    # Load garden dataset.
+    ds_garden = paths.load_dataset("happiness")
+
+    # Read table from garden dataset.
+    tb = ds_garden["happiness"]
+
+    #
+    # Process data.
+    #
```
pabloarosado, line 21: If there is no processing, you can remove the "Process data" section.

etl/steps/data/garden/happiness/2024-06-09/happiness.py
```diff
@@ -0,0 +1,86 @@
...
+        ds_regions=ds_regions,
+        ds_income_groups=ds_income_groups,
+        min_num_values_per_year=1,
+    )
+
+    # Divide the sum of the cantril ladder score times population by the total population
+    tb["cantril_ladder_score"] = tb["cantril_times_pop"] / tb["population"]
+
+    # drop unneeded columns
+    tb = tb.drop(columns=["cantril_times_pop", "population"])
```
pabloarosado, line 71: Given that population was crucial for this calculation, you could leave it as part of the data. Imagine that in some months from now you discover some issue in one of the regions. Having population there would help to find out what happened (yes, we also have the population dataset, so it's not totally necessary).

etl/steps/data/garden/happiness/2024-06-09/happiness.meta.yml
pabloarosado: Are we expecting to keep using previous versions of the report every year? In that case, maybe it's worth moving the old step to the new metadata format. If the snapshot doesn't exist anymore, we could just change the snapshot dvc file, and the garden meta yml file (and possibly some minor things in the code).

pabloarosado (approved):
Looks good! I suggested that maybe it's worth adapting the step of the previous year to the new metadata format, if we are planning on re-using it in the future.

etl/steps/data/garden/happiness/2024-06-09/happiness.meta.yml
antea04: Thanks for the review @pabloarosado - indeed I think it would be good to move the old data set to origins as well. I will follow the steps you laid out - but what do you mean by "if the snapshot doesn't exist anymore"? the old snapshots from 2012-2023 are still under etl/snapshots/happiness

pabloarosado: Good idea! What I meant is that the data provider's data may not be available anymore. If that's the case, you will not be able to execute the snapshot .py file. However, even then, you can simply modify the content of the .dvc file, and ETL will read the new metadata as expected. There's no need to re-execute the corresponding .py file. Does that make sense?

pabloarosado:
> Looks good to me! Just some minor comments on my side, but I would wait for Pablo R's review before merging.
> 
> I'm not sure why the Marimekko chart isn't showing as expected on chart-diff, perhaps you could report this in #proj-new-data-workflow. Screenshot 2024-06-14 at 11 16 01

[That chart seems to be a draft chart in production](https://admin.owid.io/admin/variables/577009), and it's not working in production either, so I would simply ignore it (or even consider removing it).

spoonerf:
> [That chart seems to be a draft chart in production](https://admin.owid.io/admin/variables/577009), and it's not working in production either, so I would simply ignore it (or even consider removing it).

Good point! I just deleted it.

PR #2771, opened 2024-06-09
🎉 chart-diff: conflict resolver

lucasrodes:
TL;DR: Tool to resolve conflicts if chart in production has been modified after modifying chart in staging.

## Summary
It may happen that someone edits a chart in production after we've started working on changes to the chart in a staging server.

Whenever this happens, we signal this in chart diff with some "Conflict" messages.

Now, in addition to signaling this, we will have the 'conflict resolver', which helps the end user resolve this conflicts. To do so, it compares the chart configs from production and staging and asks the user to pick the value from one or the other; and, optionally, edit it further. When happy, the user can click on 'resolve conflicts' and the chart is updated in staging.

To actually see the updated chart, the user needs to click on 'refresh' after going through the resolver.

image

### Other minor changes
- Change emojis to material-icons.

## Review
There is one chart to review, which I've changed a bit in staging and then just bumped the version in production. Maybe play with it. Edit it again in staging, bump version again, etc.


apps/wizard/app_pages/chart_diff/conflict_resolver.py
```diff
@@ -0,0 +1,172 @@
...
+            # Consolidate changes
+            config = deepcopy(self.diff.source_chart.config)
+            for field_key, field_resolution in self.value_resolved.items():
+                if self.value_resolved[field_key] is None:
+                    config.pop(field_key, None)
+                else:
+                    config[field_key] = as_valid_json(field_resolution)
+
+            # Push to staging
+            api = AdminAPI(SOURCE.engine, grapher_user_id=1)
```
Marigold, line 121: (In my other PR I'm assuming that everything on the staging server is done by Admin user.)

Marigold (approved):
Nice! I played with it a bit and it worked. How about having prod / staging values expanded by default so that you can compare them right away?

Conflicts most often come up when I'm updating variables of a large dataset and someone edits a typo in the title in production. Would it work well for that case?

apps/wizard/app_pages/chart_diff/conflict_resolver.py
lucasrodes: should i change to use the defaults? I am not sure what happens if the user is not using STAGING, and then the chart gets modified by X user id. This would prevent the tool to work properly.

But at the same time these tools are not currently working for local dev.

Marigold: It's fine. I'll let everyone know to change it.

PR #2777, opened 2024-06-10
✨ Improve `etl d draft-pr` tool

pabloarosado:
Following [this discussion](https://owid.slack.com/archives/C014SGT6W9X/p1718024551640549), now you don't need to write `--new-branch` and you can simply pass the name of the new branch as an argument.

I haven't implemented the `--ready-to-review` idea, because we don't want to create a PR that is ready for review with a draft title. If we do that, then we also need an argument for the title. At this point, I think it's just easier and more transparent to do it from github. It feels safer if the tool just drafts things. What do you think?

### DIFF
--- a/apps/utils/draft_pull_request.py
+++ b/apps/utils/draft_pull_request.py
@@ -3,7 +3,7 @@
 Usage:
 - If you have already created a new local branch, run this script without specifying any argument.
   The 'master' branch will be assumed as the base branch of the new draft pull request.
-- If you are in the master branch on your local ETL repo, specify a --new-branch argument.
+- If you are in the master branch on your local ETL repo, pass the name of the new branch as an argument.
   This will create a new local branch, which will be pushed to remote to create a new draft pull request.
 - If you want the base branch to be different from 'master', specify it with the --base-branch argument.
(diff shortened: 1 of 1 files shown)


lucasrodes:
Makes sense to not have the `--ready-to-review` for now. hat is your opinion on replacing `draft-pr` to `pr`, though? No strong opinion on my end, just looking for short commands :p

pabloarosado:
> Couldn't find the behaviour for when no argument for `new-branch` is passed. Do you need to set the variable to the current branch?

That is already implemented, when `new_branch` is None, in line 81. Isn't that what you meant?

lucasrodes:
> I think etl pr may be "too short" (it could be typed by accident). Let's leave it at etl draft-pr for now.

Alright! (not that it matters, but I meant `etl d pr`)

lucasrodes:
> That is already implemented, when new_branch is None, in line 81. Isn't that what you meant?

Got it ー  didn't see it in the last changes and was confused. Makes sense, thanks!

lucasrodes (approved):
thanks, pablo!

PR #2790, opened 2024-06-11
:tada: Create a map bracketer tool

pabloarosado:
## Summary

This PR creates a wizard page to ease the task of creating map brackets, although **it still has several important limitations**.

## Description

For now, the map bracketer does the following:
* Loads a specific explorer file.
* Selects a specific indicator of this explorer (among those that appear in an explorer view with a map tab).
* Chooses an optimal configuration of map brackets that agree with our guidelines.
* Lets the user tweak those brackets.
* Writes the brackets in the explorer file.

While creating it, a few possibly useful things were also created:
* Some helper functions to round numbers to significant figures and powers of ten.
* An `Explorer` object that let's you easily read, modify, and write an explorer file. For now it only works with indicator-based explorers.

## For reviewers

* @paarriagadap would you like to be the "beta tester"? To try the tool, run `etlwiz` on your local machine and play around with it. I won't have much more time in the coming weeks to improve it, but maybe there are some small changes to make it useful.
  * NOTE: When you press "Save brackets in explorer file", it will edit your local explorer .tsv file (inside `owid-content`, which is expected to be located in the same folder as `etl`). You can do as many changes as you want just to play around with it, and then do `git restore . ` in `owid-content` to revert changes.
* @lucasrodes I don't think a thorough review of the code is necessary, but feel free to have a look at `apps/wizard/app_pages/map_brackets.py`, in case you see some low hanging fruit to make it better. There are lots of TODOs (I'll handle them when/if they become important). I think we could easily figure out a better scoring system for the brackets, but I didn't have much time to play with it (check `_get_bracket_score`). Maybe we could use a part of a data meeting to brainstorm this score.
* @Marigold could you please have a look at `etl/grapher_model.py` and the function `load_mappable_regions_and_ids` in `apps/wizard/app_pages/map_brackets.py`? I needed to get the mapping of entity ids to country names, but I couldn't find an easier way. Feel free to also inspect `etl/explorer_helpers.py`, in case you find the idea of `Explorer` useful.

Thanks!


etl/grapher_model.py
```diff
@@ -103,6 +103,25 @@ class Entity(Base):
     code: Mapped[Optional[str]] = mapped_column(VARCHAR(255))
     updatedAt: Mapped[Optional[datetime]] = mapped_column(DateTime, init=False)
 
+    @classmethod
+    def load_entity_mapping(cls, session: Session, entity_ids: List[int]) -> Dict[int, str]:
```
Marigold, line 107: This looks good. (If we find it too slow, we can try `pd.read_sql` instead, but for hundreds of records this should be fast enough).

Marigold (approved):
Both grapher_helpers.py and Explorer class look good!

apps/wizard/app_pages/map_brackets.py
```diff
@@ -0,0 +1,923 @@
...
+    # Add a dropdown for color scheme.
+    # TODO: Add full list of color schemes.
+    mb.color_scheme = st.selectbox(  # type: ignore
+        label="Color scheme (list not yet complete!)",
+        options=["BuGn", "BinaryMapPaletteA", "YlOrBr"],
+        help="Color scheme for the map.",
+    )
+
+    # Add toggle to control whether the brackets are based only on the data for the latest year, or all years.
+    mb.latest_year = st.toggle("Consider only latest year", mb.latest_year)
```
lucasrodes, line 605: maybe add a help text here to explain the user how this is relevant?

apps/wizard/app_pages/map_brackets.py
```diff
@@ -0,0 +1,923 @@
...
+        reload_all_brackets=True,
+        reload_rank=True,
+        reload_optimal_brackets=True,
+    )
+    mb.percentile_min, mb.percentile_max = st.slider(
+        label="Percentile range covered by brackets",
+        min_value=0,
+        max_value=100,
+        value=(MIN_PERCENTILE, MAX_PERCENTILE),
+        help="Currently, this only affects linear brackets",
```
lucasrodes, line 646: maybe elaborate a bit more on what 'percentile range' means?

apps/wizard/app_pages/map_brackets.py
```diff
@@ -0,0 +1,923 @@
...
+            + list(BRACKET_LABELS["custom"].values())
+        },
+    }
+    # Create a bracket type selector with a select box.
+    bracket_type = st.selectbox(
+        "Select linear or log-like",
+        options=bracket_type_labels,
+        # NOTE: By default, the first option will be selected, which is expected to be the optimal one.
+        index=0,
+        # horizontal=True,
```
lucasrodes, line 678: help message explaining what this is? e.g. 'at owid we use different strategies for the map bracket bins. These are mostly divided into two categories: log and linear. ...'

apps/wizard/app_pages/map_brackets.py
```diff
@@ -0,0 +1,923 @@
...
+            st.error("No brackets possible.")
+            st.stop()
+        min_selected, max_selected = st.select_slider(  # type: ignore
+            "Select a range of values:",
+            options=mb.brackets_positive,
+            value=(
+                abs(mb.brackets_optimal[mb.bracket_type]["lower"]),
+                abs(mb.brackets_optimal[mb.bracket_type]["upper"]),
+            ),  # default range
+        )
```
lucasrodes, line 708: help message?

apps/wizard/app_pages/map_brackets.py
```diff
@@ -0,0 +1,923 @@
...
+            st.stop()
+        mb.update_custom_brackets(brackets_manual)
+    elif mb.bracket_type in list(BRACKET_LABELS["log"].values()):
+        # Create a slider for positive values.
+        # TODO: Maybe have a toggle to be able to manually select negative and positive values separately.
+        if len(mb.brackets_positive) <= 1:
+            st.error("No brackets possible.")
+            st.stop()
+        min_selected, max_selected = st.select_slider(  # type: ignore
+            "Select a range of values:",
```
lucasrodes, line 702: Can use the same structure of the title of the other slider: `Value range`

It is implicit that the user needs to select (no need to say it). No need to add the `:`

lucasrodes (commented):
This is looking fantastic, Pablo! Thanks for doing this work. I think this will become very useful in the time to come.

I've added some comments, mostly suggestions on adding help messages to make it clearer what the different options are.

Also, do you think you could some accompanying docs for the tool? Would be nice to have it in the same PR, but it's fine if not. You can add a section in https://docs.owid.io/projects/etl/guides/. I think that can be a nice way to introduce users to the tool, and motivate its usage. Also so we can start making references to it from other parts of the docs.

Finally, I am not sure what is the best way to display the indicators in the dropdown. Currently you are using variable IDs, which makes a lot of sense. I wonder if catalogPaths is something we could have as an alternative also.

pabloarosado:
@paarriagadap I'll merge for now, so we can all start playing around with it. But feel free to give it a try. You can add bugs and suggestions in this issue: https://github.com/owid/etl/issues/2915

PR #2791, opened 2024-06-11
✨ chart-diff: add metadata-diff view, bug fixes

lucasrodes:
- [x] **New dataset**: Plastic surgery dataset.
- [x] **Bug fixed**: Charts changed due to config not showing.
- [x] **Feature**: Compare metadata changes with some diff tool. ✨ Include GPT-summary, which only works locally (could we add OpenAI creds in staging servers?).
- [x] ~**Feature**: Compare data changes with some explorer-like tool.~ For a future PR.
- [x] Merge filtering options 'Chart modified / new' and 'Chart changes type' into one.
- [x] Change reject/approval options for metadata/data-triggered changes.
- [x] Change owidbot message.

## BUG: Charts without no change type tag are not being shown. 
- [x] DONE

<details>
  <summary>Bug explained!</summary>
  These charts are: new charts and charts whose config has changed for special fields. I fixed the bug for 'new charts'.

For charts with only changes in fields specified by [configs_are_equal](https://github.com/owid/etl/blob/data/plastic-surgery/apps/wizard/app_pages/chart_diff/chart_diff.py#L240-L246) are not considered as different by this function. Hence, these are not tagged with the label `change_type='config'`. Consequently these charts are not shown unless we enable 'show all'.
This, however, is not consistent with what [modified_charts_by_admin](https://github.com/owid/etl/blob/data/plastic-surgery/apps/chart_sync/cli.py#L492) from chart_sync is doing. It gets charts that have suffered any change in the config. And this is important, because the charts obtained by `modified_charts_by_admin` are those listed by owidbot.

Consequently, in this PR we observe that there are **3 charts** listed by @owidbot, yet only two charts are shown by default in chart-diff.

We should align these behaviours. Some ideas:
1. Further filter the charts obtained by `modified_charts_by_admin` by checking their configs and ignoring them based on the same criteria as `configs_are_equal`.
2. Change the logic from `configs_are_equal` and just tag a chart with _any_ change with `change_type='CONFIG'`.
3. Leave as-is, but make @owidbot ignore charts with changes in fields mentioned in `configs_are_equal`. We would have charts with such minor config changes in chart-diff, but wouldn't be shown.

My personal preference is either 1 or 2. I find 3 confusing.
</details>

## BUG: There are leaks and charts from other workflows are shown
likely fixed by https://github.com/owid/etl/pull/2797


lucasrodes:
@Marigold FYI, I found couple of minor bugs. Tried to explain them in the PR description. 

I estimate that you will start working like 2 hours before me, so I thought I'd ping you in case you want to look into this before I start working. Otherwise I'll work on it tomorrow.

Marigold:
Thank you @lucasrodes. Just a heads up that I hotfixed two of those bugs on master & this branch, but there are other things to fix (or at least to discuss them). I'll write it up as soon as I finish firefighting.

lucasrodes:
@Marigold I'd say you can play around with charts that originate from 'metadata' changes. If you run this locally, you will get gpt summaries (bc there are no credentials on staging to access openai gpt).

on this matter, do you think we could have credentials added to staging servers? or is it risky?

lucasrodes:
Also, not really the need to review the new dataset pipeline.

apps/wizard/app_pages/chart_diff/show.py
```diff
@@ -160,6 +210,97 @@ def _show_chart_diff_controls(self):
...
+        Note that to access the metadata, one needs to retrieve the JSON metadata files from the S3 bucket.
+        """
+        if st.button(
+            "🔎 Metadata differences",
+            f"btn-meta-diff-{self.diff.chart_id}",
+        ):
+            self._show_metadata_diff_modal()
+
+    @st.experimental_dialog("Metadata differences", width="large")  # type: ignore
+    def _show_metadata_diff_modal(self) -> None:
```
Marigold, line 227: This looks really nice! Just one nit - I think we could ignore all `id` fields and `updatedAt` when comparing metadata (they're not going to be part of metadata checksum anyway).

apps/wizard/app_pages/chart_diff/show.py
```diff
@@ -319,6 +433,42 @@ def show(self):
         else:
             self._show()
 
+        self.clean_cache()
+
+
+def compare_dictionaries(dix_1: Dict[str, Any], dix_2: Dict[str, Any]):
```
Marigold, line 439: We have tons of useful "diff"-like functions scattered across our repo (for dataframes, dicts, yamls, etc.). One day we can unify it into a single library (if it doesn't turn out to be too bespoke).

apps/wizard/utils/env.py
```diff
@@ -37,7 +37,10 @@ class UnknownOWIDEnv(Exception):
 
 
 class OWIDEnv:
-    """OWID environment."""
+    """OWID environment.
+
+    TODO: maybe worth moving to etl.config
```
Marigold, line 42: Good idea, this has become useful in other parts of our codebase, too.

Marigold (approved):
This looks great Lucas, thanks!

apps/wizard/app_pages/chart_diff/show.py
lucasrodes: totally;

apps/wizard/app_pages/chart_diff/show.py
lucasrodes: That makes sense; I'll create the list of columns to ignore somewhere, and use it here and in the checksum estimation. (after you merge https://github.com/owid/etl/pull/2797)

lucasrodes: Nah, actually I'll abstract your function logic of popping irrelevant fields and will use it here and there

PR #2795, opened 2024-06-12
AI Epoch dataset - small edits 💄 

veronikasamborska1994:
small changes to metadata + cleaning up domains with less than 20 systems to be "Other"


Marigold (approved):
Looks good!

PR #2797, opened 2024-06-12
:bug: fix metadata checksum

Marigold:
Metadata checksum should be as consistent as possible. There are problematic fields like `updatedAt` that can change even though it has no effect on metadata and should be excluded from checksum calculation.

This PR adds a couple more fields that cause problems
 - `dataChecksum`, `metadataChecksum` - including these would lead to a feedback loop of ever-changing checksums
 - `dimensions` - these are not causing problems, but are already captured in `dataChecksum`
 - `IDs` - staging server and production can build ETL separately and end up with different ids (but same metadata)

## TODO after merging:
- [ ] delete values in `variables.metadataChecksum` in production DB (we're not going to backfill them all, this will be done in the future when incrementing ETL epoch)

### DIFF
--- a/apps/backport/datasync/data_metadata.py
+++ b/apps/backport/datasync/data_metadata.py
@@ -15,7 +15,7 @@ from tenacity.retry import retry_if_exception_type
 from tenacity.stop import stop_after_attempt
 from tenacity.wait import wait_fixed
 
-from etl import config
+from etl import config, files
 from etl.db import read_sql
 
 log = get_logger()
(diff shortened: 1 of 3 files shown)


lucasrodes:
> `dimensions` - these are not causing problems, but are already captured in dataChecksum

Would they? For instance, if we use **indicator upgrader** and change the use of an indicator for another, an ID in `dimensions` will be changed. This should captured as a 'config change'. I just wonder, would such changes be captured by checksum and config?

Marigold:
Sorry, I don't think I got it, but let me try anyway :).

> For instance, if we use indicator upgrader and change the use of an indicator for another, an ID in dimensions will be changed.

If you change an **indicator** (or did you want to say entity?) then grapher config will change (which is the most important). `dimensions` are computed directly from data. The thing is that if you change an entity name, with dimensions it'd show up as both **data change** and **metadata change**.

lucasrodes:
ah, nvm, I got confused. I was thinking of `chart.config.dimensions`! All clear sir!

lucasrodes (approved):
LGTM

PR #2799, opened 2024-06-12
🔨 git: get changed files

lucasrodes:
Pablo implemented a function to detect changes in files between branches. It is currently placed in [apps/utils/map_datasets](https://github.com/owid/etl/blob/master/apps/utils/map_datasets.py#L29). This PR moves it to `etl.git` instead.


pabloarosado (approved):
Looks good!

PR #2801, opened 2024-06-12
:bar_chart: Update IMF expenditure data in the long run

paarriagadap:
Upload an updated version of IMF's dataset to correct a problem in the calculation of the data made 7 years ago.

This is a very straightforward code, mostly from Wizard.

Main issue https://github.com/owid/owid-issues/issues/1567

---
- Create object `ChartDiffLoader` to assist the loading and retrieval of ChartDiff objects.
- Optimisation of chart-diff loading: run SQL queries for a batch of charts. Results in 85% improvement in computation time.


lucasrodes:
I'm taking over this branch for some days to try to fix some issues with chart leakage into chart-diff

PR #2819, opened 2024-06-16
📊 UNESCO Institute for Statistics (UIS) - Education (OPRI)

veronikasamborska1994:


etl/steps/data/garden/unesco/2024-06-16/education_opri.py
```diff
@@ -0,0 +1,112 @@
...
+            update_metadata(meta, display_decimals=0, unit="boys", short_unit="")
+        elif "female" in name_lower:
+            update_metadata(meta, display_decimals=0, unit="girls", short_unit="")
+        else:
+            # Default metadata update when no other conditions are met.
+            update_metadata(meta, 0, " ", " ")
+    return tb
+
+
+def update_metadata(meta, display_decimals, unit, short_unit):
```
spoonerf, line 97: Remember to add the 'types' to each of your function arguments e.g. `unit:str`

etl/steps/data/garden/unesco/2024-06-16/education_opri.py
```diff
@@ -0,0 +1,112 @@
...
+        meta.display = {}
+
+        # Convert the 'name' variable to lowercase to make text matching easier
+        name_lower = tb[column].title.lower()
+
+        #
+        # Update metadata units, short_units and number of decimal places to display depending on what keywords the variable name contains
+        #
+        # Fill out units and decimal places
+        if "%" in name_lower:
```
spoonerf, line 77: I don't think you should change this, but you could check out Jinja templates as an alternative (an example [here](https://github.com/owid/etl/blob/master/etl/steps/data/garden/ihme_gbd/2024-05-20/gbd_cause.meta.yml)), I find them a bit more readable than adding metadata in the script, but up to you!

spoonerf (approved):
Looks mostly good, I think you could exclude a lot more of the odd regions and just have one entity for World, I am pretty sure they are the same across the different organization's versions.

There's a couple of entities showing as numbers e.g in [this variable](http://staging-site-unesco-edstats/admin/datapage-preview/935577), they are showing as: 369931 & 369932, not sure what is going on there.

etl/steps/data/garden/unesco/2024-06-16/education_opri.countries.json
```diff
@@ -0,0 +1,324 @@
...
+  "WB: Post-demographic dividend": "Post-demographic dividend (WB)",
+  "WB: Pre-demographic dividend": "Pre-demographic dividend (WB)",
+  "WB: Small states": "Small states (WB)",
+  "WB: South Asia": "South Asia (WB)",
+  "WB: South Asia (IDA & IBRD)": "South Asia (IDA & IBRD) (WB)",
+  "WB: Sub-Saharan Africa": "Sub-Saharan Africa (WB)",
+  "WB: Sub-Saharan Africa (excluding high income)": "Sub-Saharan Africa (excluding high income) (WB)",
+  "WB: Sub-Saharan Africa (IDA & IBRD)": "Sub-Saharan Africa (IDA & IBRD) (WB)",
+  "WB: Upper middle income (July 2023)": "Upper middle income July 2023 (WB)",
+  "WB: World": "World (WB)",
```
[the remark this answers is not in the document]
veronikasamborska1994: almost exactly the same though there are some small differences in some indicators. I've used the SDG one as it has the best coverage!

etl/steps/data/garden/unesco/2024-06-16/education_opri.py
veronikasamborska1994: thank you! I've tried it but I think it will be tricky to add the long descriptions to the metadata using Jinja templates. looks like everything that we might want to use to define metadata in Jinja needs to be a part of the index that then gets pivoted in the grapher stage? maybe I am missing something though..

PR #2821, opened 2024-06-17
🐛 investigate bug in fast-track

lucasrodes:


lucasrodes:
### DRAFT: Joe – Difference in Gini, 1980-2018, PIP vs WID data
(`draft_joe_gini_diff_1980_2018`)

The error comes from adding rows with year 2020, but without an entity.

image

pabloarosado (approved):
Looks good. Given that one or more of these steps are breaking the ETL, let's keep them out of the dag for now and see if the problem is fixed.

PR #2844, opened 2024-06-19
📊 Epoch AI datasets on large-scale AI models

veronikasamborska1994:


Marigold (approved):
Looks good! This is a new dataset, not an update, right?

veronikasamborska1994:
> Looks good! This is a new dataset, not an update, right?

It's a new dataset but with very similar code to another dataset that I've decided not to use (hence the question on slack). No related published charts yet!

PR #2854, opened 2024-06-20
:honeybee: catalog: typing

lucasrodes:


Marigold:
@lucasrodes I spent some time trying to make it work, but with no success (even the simplest overriding of `__getitem__` didn't work). I'd close this and use `.loc[]` rather than square brackets. It's possible that pandas might fix that in their core code in the future.

lucasrodes:
hey @Marigold, I think that makes sense for the `[]` issue. We could still add overloads for the other methods though.

Marigold:
> I think we could still add overloads for the other methods though.

For sure. The question is whether to add them all at once or do it on the fly (which is what I'm doing). Some methods with a lot of arguments could end up having tons of `@overload` decorators.

lucasrodes:
Alright, yeah, will be a long run.

Should we then try to push these changes or discard them for the time being?

Marigold:
Sure, why not.

Marigold (approved):
Looks good.

lucasrodes:
@Marigold still need to fix the tests not passing CI/CD. I'm looking at it.

Marigold (approved):
Looks good.

PR #2859, opened 2024-06-20
✨ indicator upgrader: pagination, performance, ui/ux

lucasrodes:
For large datasets, indicator upgrader is not working as good as it could.

This PR attempts to incorporate pagination to the app.

Code
- Create objects `IndicatorUpgrade`, `IndicatorUpgradeShow`. Very much as we did for chart-diff
- Added usage of Pagination
- Update to Streamlit v1.36
- Improve performance: Removed unnecessary widgets (disabled `selectbox` for old indicators)
- Place old and new indicator on top of each other (instead of left-right). This way, long names will be readable. So far, they were hidden with `...`.
- Pagination: Add `on_click` option, so we can act (externally) on user's interaction.
- Harmonizer: Signal better when user adds custom input by disabling selectbox.

UX/UI
- Make experience closer to Chart-diff: 
- Improve performance by avoiding complete page re-runs.

TODO:
- [x] Pass tests
- [x] Further code cleaning
- [x] Improve header, options
- [x] Pagination with selectbox +/-?
- [x] Make sure explorer-mode works
- [x] Make app work with auto. Also combination auto + manual


lucasrodes:
@Marigold I'm not sure why the baking of the staging server is failing. It looks like the bake is trying to remove some variables for which there are charts... But I am not sure why this occurs here and not on master.

Marigold:
@lucasrodes that was because of [this issue](https://github.com/owid/etl/issues/2871). I ran `cd owid-grapher && make refresh` on your staging server to get a working copy of prod MySQL.

Marigold (approved):
Looks good! I finally got a chance to try out the Explore mode. I might reuse it in terminal for data-diff or even get data-diff its own streamlit app.

PR #2866, opened 2024-06-21
 :bar_chart: Reduce memory footprint of surface temperature

Marigold:
Use smaller type float32 to reduce memory footprint. Check out data-diff below for differences (it's on the sixth decimal place). If this doesn't work, I guess we can just move the entire `_load_data_array` to snapshot.


Marigold:
@veronikasamborska1994 could you have a look when you have time, please? Is the reduced precision a problem, or is it ok?

veronikasamborska1994 (approved):
sorry I somehow missed this! I think it's definitely fine! and thanks!

PR #2870, opened 2024-06-22
🎉 harmonize: interactive shell

lucasrodes:
Based on the form recently shared with the team, it was suggested that the Harmonizer was easily accessible also from the Interactive Shell in VS Code.

This PR adds this, along with its corresponding documentation. Essentially, one can run the following from the interactive shell now:

```py
from etl.harmonize import harmonize_ipython
harmonize_ipython(
  tb,
  "country",
  paths.country_mapping_path
)
```

A form will be displayed and the user will be able to generate the harmonized country mappings.

Additionally, I've re-factored the code so that all our endpoints (Wizard, terminal, and now interactive shell) use the same code.


lucasrodes:
cc @pabloarosado

etl/harmonize.py
```diff
@@ -70,20 +75,45 @@ def harmonize(
...
+
+    # Export
+    harmonizer.export_mapping()
 
-    mapping = interactive_harmonize(geo_column, mapping, institution=institution, num_suggestions=num_suggestions)
-    with open(output_file, "w") as ostream:
-        json.dump(mapping, ostream, indent=2)
+
+def harmonize_ipython(
+    tb: Table, column: str, output_file: str, num_suggestions: int = 100, institution: Optional[str] = None
```
pabloarosado, line 99: Argument "column" should not be mandatory. Its default value can be defined as a global variable. Alternatively, you can make it None by default (an also None by default in Harmonizer, and if None, then its value is set to "country").

pabloarosado: Ideally, argument `output_file` should also be optional. At some point in the future, the path to the excluded countries should also become an argument. Maybe `paths` could be given as an optional argument, instead. If that's too much hassle, you can leave it this way.

pabloarosado (approved):
Hey, thanks for doing this! I was expecting something much less fancy. I made minor suggestions. The only issue I've noticed is that currently, when using the tool on a long list of ambiguous entities, the window goes down to the bottom after every interaction. That's a bit annoying, but maybe there's a workaround.

https://github.com/owid/etl/assets/12246978/8d1507e3-e16b-4687-bce0-5bb2de2c1261

etl/harmonize.py
lucasrodes: you need to provide a value to the export file. I don't see how you can avoid that.

It's either as-is now, or provide paths and access the country mapping file from within the class.

I think the current state is more general and flexible, since it can be used regardless of whether you have paths or not.

What I can do, is adding an additional argument to accept paths alternatively

lucasrodes: > Argument "column" should not be mandatory. Its default value can be defined as a global variable. Alternatively, you can make it None by default (an also None by default in Harmonizer, and if None, then its value is set to "country").

do you mean to basically have the default 'country' value?

I think that's a good option.

pabloarosado: Yes, that's what I mean. I thought that maybe you didn't do it that way because the Harmonizer already has "country" defined as a default argument. For that reason I suggested defining a global variable or setting it to None. But it's definitely not necessary (in most cases, indeed, the column name is going to be "country" anyway).

Issue #2871, opened 2024-06-23
`chart-sync` fails when user renames variables


Marigold:
If someone renames **short name** of a variable, e.g. from `dummy_a` to `dummy_b` and modifies a chart which uses that variable, then ETL deploy will fail after merging the staging server.
The way ETL works is that it adds variable `dummy_b` and then tries to remove unused variable `dummy_a`. The problem is that the chart hasn't been synced yet and still depends on `dummy_a`, hence cannot be deleted and ETL fails with the error message 
```
ValueError: Variables used in charts will not be deleted automatically:
```

I don't have a good solution, but here are some options:
- Change that error to a warning. ETL would finish and `chart-sync` would swap the variable. The only problem is that "ghost" variable would still be in Admin until next ETL run of that dataset
- Make ETL deploy fail, but still run `chart-sync`
- Separate "cleanup_ghost_variables" into a step after both ETL and chart-sync

Marigold:
Fixed by https://github.com/owid/etl/pull/2890

PR #2890, opened 2024-06-25
 :bug: Cleanup variables after chart-sync

Marigold:
Fix for https://github.com/owid/etl/issues/2871, see the description there.

I fixed it by changing the error to a warning and setting a different checksum so that those variables get cleaned on next ETL run. It means that the dataset will be rerun twice, but it's the easiest solution and this should happen rarely.


lucasrodes (approved):
lgtm!

PR #2895, opened 2024-06-25
✨ chart-diff: performance, ux/ui

lucasrodes:
- Increase iframe height: footer buttons were hidden in some display settings. Changed from 500px to 600px.
- Changed pagination mode. Before arrows were used to change page; now a number input.
- remove `etlci` command alias, since it is not used. Instead, we use `etl` which is equivalent.
- improve presentation of icon in expander from chart-diff
- `chart_html`: Reduce the HTML code for the chart. Try to keep just the bare minimum to render the chart.
- use `st.experimental_fragment` for the chartdiff blocks, so that if the user interacts with one box, only code within the box is re-executed. I.e. avoids long re-runs of complete page.


Marigold (approved):
Sweet! Can't wait for the speed-up from fragment.

PR #2897, opened 2024-06-26
 :sparkles: Add roundingMode to variable.display in YAML

Marigold:


sophiamersmann (approved):
Looks good to me, thanks a lot, Mojmir!

ps: What does the pattern '<%' mean? (Just wondering if I forgot to implement something)

Marigold:
> ps: What does the pattern <% mean? (Just wondering if I forgot to implement something)

That's ETL thing. Sometimes we use dynamic Jinja templates in metadata with `<%` tags so we allow it in JSONschema too.

PR #2900, opened 2024-06-26
🐛 indicator-upgrader: team feedback

lucasrodes:
Errors found

- [x] `ValueError`: caused by not finding a key in `st.session_state.indicator_upgrades_ignore`. (first image). It worked fine after refresh. Refreshing is fine for a small dataset, but not great if user has done substantial work already. After selecting some indicators from dropdowns, the user went to select another one and this error was raised.
- [ ] A user reported weird charts appearing in chart-diff. After looking into them, they seem to originate from updating the referenced indicator IDs in the chart config via the indicator upgrader. Unclear if it was due to a human mistake or some logic in indicator-upgrader failed.  **UNCLEAR FOR NOW ー TO EXPLORE IF IT IS REPORTED AGAIN**

image


lucasrodes:
I was able to reproduce the first error. It seems to occur when you click the first 'next' button for the second time.

Seems that the session_state might not be properly configured.

PR #2903, opened 2024-06-26
:bar_chart: Create poverty projections dataset from World Bank

paarriagadap:
Discussed here https://github.com/owid/owid-issues/issues/1519


etl/steps/data/garden/wb/2024-06-26/poverty_projections.meta.yml
```diff
@@ -0,0 +1,101 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    processing_level: minor
+    presentation:
+      topic_tags:
+        - Poverty
+  description_key_estimated_extreme_poverty: Extreme poverty here is defined as living below the International Poverty Line of $2.15 per day.
```
antea04, line 8: Is there a benefit to making all the key points separate? The first four points always show up together in the descriptions below, if possible, I would combine them in one variable.

etl/steps/data/meadow/wb/2024-06-26/poverty_projections.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # Load data from snapshot.
+    tb_number = snap_number.read()
+    tb_share = snap_share.read()
+
+    #
+    # Process data.
+    #
+    # Rename columns
+    tb_number = tb_number.rename(columns={"million_poor": "headcount_215"})
```
antea04, line 27: Nitpicky: I would either multiply with a million here or rename this column only when you multiply

antea04 (approved):

antea04 (approved):

antea04 (approved):
Looks good from my side - I only left teeny tiny comments. Is there any reason you did not approve the charts in chart-diff? They look good from my end, so I went ahead and approved them, but you might want to take another look before you merge!

etl/steps/data/garden/wb/2024-06-26/poverty_projections.meta.yml
paarriagadap: No, there's no benefit. I thought any option would break the parameter, but I applied anchors and they worked. Thanks!

paarriagadap:
Thank you @antea04! I actually approved them before, so I don't know if something was broken or not. I will report if it happens next time

PR #2906, opened 2024-06-27
📊 update of OECD dataset on road fatalities

antea04:
(linked to issue owid/owid-issues#1252)

**Included in PR:**

- [x] updated OECD dataset
- [x] Added OECD data on passenger kilometers by road and rail (separate snapshot)
- [x] updated and added road safety charts
  - [x] total reported road fatalities
  - [x] road fatalities per 1 billion passenger-kilometers
  - [x] passenger kilometers for road & rail transport
- [x] figured out whether [OECD data has an error](https://github.com/owid/owid-issues/issues/1252#issuecomment-1771364275) -> yes, reached out to OECD about it

**Things to note:**

- Some countries (e.g. Russia) have really high fatalities/B passenger kilometers. I double-checked the data, but that is consistent with the OECD data and it seems to be the actual reported numbers. Feel free to double-check, but this seems to be an actual data point/ reporting issue, rather than a data artifact.


etl/steps/data/meadow/oecd/2024-07-01/passenger_travel.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # drop columns not needed
+    tb = tb[COLS_TO_KEEP]
+
+    # Rename columns
+    tb = tb.rename(columns={"Reference area": "country", "TIME_PERIOD": "year"})
+
+    # fix dtypes and scale up data
+    tb["year"] = tb["year"].astype("Int64")
+    tb["OBS_VALUE"] = tb["OBS_VALUE"].astype("Int64") * 1_000_000
```
paarriagadap, line 29: It's not clear to me why you are scaling up? It's related to me not knowing about passenger-kilometers

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+    # drop old columns
+    tb = tb.drop(columns=["fatalities", "injured", "crashes", "deaths", "injuries", "accidents_involving_casualties"])
+
+    # remove passenger travel by road if it is identical to bus or car column
+    tb["passenger_kms_road"] = tb.apply(lambda x: check_road_passenger_travel(x), axis=1)
+
+    # add death per million inhabitants
+    tb = geo.add_population_to_table(tb, ds_population)
+    tb["deaths_per_million_population"] = (tb["accident_deaths"] / tb["population"]) * 1_000_000
+    # drop population as well as old death per million and per vehicle (these numbers are wrong)
```
paarriagadap, line 121: Why are they wrong?

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+            "Rail": "passenger_kms_rail",
+            "Road": "passenger_kms_road",
+            "Passenger cars": "passenger_kms_car",
+            "Buses": "passenger_kms_bus",
+        }
+    )
+
+    # harmonize country names
+    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path, warn_on_unused_countries=False)
+    tb_old = geo.harmonize_countries(
```
paarriagadap, line 92: Is it not harmonized already?

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+    "passenger_kms_road",
+    "passenger_kms_car",
+    "passenger_kms_bus",
+    "deaths_per_billion_kms",
+]
+
+
+def add_origins(tb: Table, cols: list, origins: Origin) -> Table:
+    for col in cols:
+        tb[col].origins = origins
```
paarriagadap, line 28: you can also use copy_metadata as
```
for col in cols:
    tb[col] = tb[col].copy_metadata(tb["country"])
```

Or any other variable that had not lost the origins.

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+    return tb
+
+
+def check_road_passenger_travel(tb_row: pd.Series):
+    """Check if road passenger travel is identical to bus or car passenger travel. If so, remove road passenger travel."""
+    road_passengers = tb_row["passenger_kms_road"]
+    # if road passenger travel is NA or both bus and car passenger travel are NA, there is nothing to compare
+    if pd.isna(road_passengers) or (pd.isna(tb_row["passenger_kms_bus"]) & pd.isna(tb_row["passenger_kms_car"])):
```
paarriagadap, line 36: Shouldn't you use and instead of & here?

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+        return road_passengers
+    # check whether road passenger travel is identical to bus travel
+    elif not pd.isna(tb_row["passenger_kms_bus"]) and tb_row["passenger_kms_bus"] == road_passengers:
+        return None
+    # check whether road passenger travel is identical to car travel
+    elif not pd.isna(tb_row["passenger_kms_car"]) and tb_row["passenger_kms_car"] == road_passengers:
+        tb_row["passenger_kms_road"] = None
```
paarriagadap, line 43: Why do you prefer to define passenger_kms_road here instead of just leaving `return None` as the elif above?

etl/steps/data/garden/oecd/2024-07-01/road_accidents.meta.yml
```diff
@@ -0,0 +1,120 @@
...
+
+tables:
+  road_accidents:
+    variables:
+      accident_deaths:
+        title: Road accident deaths
+        unit: deaths
+        short_unit: deaths
+        description_short: Number of reported deaths due to road accidents.
+        description_processing: |-
```
paarriagadap, line 33: Does this only apply to accident_deaths? If yes, you maybe need to add description_processing for the other indicators (since they are all tagged as with major processing)

paarriagadap (changes requested):
Looks good! Though it's weird how the older (2013) steps were built. What I would do, would be to reconstruct the historical data in separate steps and then include it to the new version of the data. That way origins would be clearly seen and the data will show proper citation.

It's confusing that (as far as I see) the file for those steps is a combination of the data from the snapshot.

I will request changes for now only because I see that the new charts only show data for 9 countries. Is that OK?

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
antea04: They come from the OECD and don't make sense - death per million inhabitants is off by a factor of 10 and death per million vehicles is somehow 10x smaller than per inhabitant, which does not add up with the vehicle to people ratio.

etl/steps/data/garden/oecd/2024-07-01/road_accidents.meta.yml
antea04: Will add processing description on indicator level!

etl/steps/data/meadow/oecd/2024-07-01/passenger_travel.py
antea04: Oh, the data is just given in millions - I added it to the comment to make it clearer

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def choose_latest_data(tb_row: pd.Series, old_col: str, new_col: str):
+    if pd.isna(tb_row[new_col]):
+        return tb_row[old_col]
+    else:
+        return tb_row[new_col]
+
+
+def run(dest_dir: str) -> None:
```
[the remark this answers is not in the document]
antea04: I like having it on top personally (maybe a leftover habit from the jupyter notebook days) but its certainly no hill to die on for me, so I'll change it :D I think I have seen both in old etl steps, so I don't know if there is a general guidance - maybe it would be good to have a uniform way to do it

antea04:
> Looks good! Though it's weird how the older (2013) steps were built. What I would do, would be to reconstruct the historical data in separate steps and then include it to the new version of the data. That way origins would be clearly seen and the data will show proper citation.
> 
> It's confusing that (as far as I see) the file for those steps is a combination of the data from the snapshot.
> 
> I will request changes for now only because I see that the new charts only show data for 9 countries. Is that OK?

Thanks for the review! I will take another look at the historic data - when I last checked the data in the linked websites was not easily exportable and I felt like it was not worth the added effort of building e.g. fast-track steps for every piece of that dataset. But I will take another look and see if I can think of something.

And yeah - only 9 countries showing up is definitely wrong. I will take a look at what happened there. Weirdly, it did not reset the chart-diff apparently - I think it should probably, right?

etl/steps/data/garden/oecd/2024-07-01/road_accidents.py
```diff
@@ -0,0 +1,149 @@
...
+    #
+    # Load inputs.
+    #
+    # Load meadow dataset.
+    ds_meadow = paths.load_dataset("road_accidents", channel="meadow", version="2024-07-01")
```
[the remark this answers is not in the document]
antea04: I personally think it looks confusing if I am importing another garden step thats named the same - I prefer the clarity of the version, even if it is more of a hassle to update. But I will look into seperating the historic data in individual steps, so maybe this will not even be necessary in the end.

paarriagadap (approved):
I can now see the rest of countries! Thank you! What was the problem? Anyway, I approve now

PR #2914, opened 2024-07-01
✨ mab bracketer: ux/ui

lucasrodes:
## UX/UI improvements

The main motivation for these improvements is to have the chart shown more prominently so that the user doesn't have to constantly scroll down.

Actions taken:
- Move description and limitations to a `popover` (labeled as "Learn about it") so that it doesn't take much vertical space.
- Don't show the different 'use cases' radio buttons, since only one is supported. Signal this in the `Learn about it` popover.
- Put dataset and indicator selection in a container with border so it is easily found.
- Sidebar is shown by default. It was set to `collapsed`; I thought there we don't necessarily need to hide it this time since the page doesn't need much width. 
- Move configuration parameters to the side bar.
- Move export config button to the side bar, so that it is always visible.

## Future work
There are some minor things that we could tackle in https://github.com/owid/etl/pull/2790.

I'll leave comments in there once we merge this.

Basically, some improvements I can think of are to add more help messages to streamlit widgets.

### DIFF
--- a/apps/wizard/app_pages/map_brackets.py
+++ b/apps/wizard/app_pages/map_brackets.py
@@ -659,7 +659,9 @@ def map_bracketer_interactive(mb: MapBracketer) -> None:
 
     # Select bracket type.
     bracket_type_labels = {
+        # Add optimal choice
         f"Optimal ({mb.brackets_optimal['optimal']})": mb.brackets_optimal["optimal"],
+        # Rest of the options
         **{
             value: value
             for value in list(BRACKET_LABELS["log"].values())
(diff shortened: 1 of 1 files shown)


apps/wizard/app_pages/map_brackets.py
```diff
@@ -658,20 +659,23 @@ def map_bracketer_interactive(mb: MapBracketer) -> None:
...
+            for value in list(BRACKET_LABELS["log"].values())
+            + list(BRACKET_LABELS["linear"].values())
+            + list(BRACKET_LABELS["custom"].values())
+        },
     }
     # Add optimal choice.
-    bracket_type_labels[f"Optimal ({mb.brackets_optimal['optimal']})"] = mb.brackets_optimal["optimal"]
+    # bracket_type_labels[f"Optimal ({mb.brackets_optimal['optimal']})"] =
     # Create a bracket type selector with radio buttons.
     # NOTE: By default, the last option will be selected, which is expected to be the optimal one (the last element in bracket_type_labels, defined above).
```
pabloarosado, line 673: The previous comment would need to be changed to make sense, "last" -> "first".

pabloarosado (approved):
Nice, thanks for the improvements!

PR #2931, opened 2024-07-03
:sparkles: wb: Add median income in constant US$ in PIP data

paarriagadap:
Main discussion here https://github.com/owid/owid-issues/issues/1598


paarriagadap:
This will not be implemented, because we decided to make the calculation and not publish a new chart. The code in this PR was useful for the calculation, though.

PR #2938, opened 2024-07-05
🐝 rename is_indexable -> isIndexable

sophiamersmann:
We want to rename the `charts.is_indexable` column to `isIndexable` in Grapher (see https://github.com/owid/owid-grapher/pull/3773)


larsyencken (approved):
Good to go :shipit:

PR #2943, opened 2024-07-08
:bar_chart: maternal mortality: update long run dataset

antea04:
see also owid/owid-issues#1432

### What I have done in this pull request:

- [x] add the [gapminder data set](https://www.gapminder.org/data/documentation/gd010/) on maternal mortality to the etl
- [x] add the latest [UN MMEIG data](https://www.who.int/publications/i/item/9789240068759) to the etl 
- [x] combine UN MMEIG, Gapminder and WHO Mortality database data sets to [long run maternal mortality](http://staging-site-maternal-mortality/admin/datasets/6612) indicators
  - [x] GM data pre 1950, 1950-1985 calculated indicators from the WHO Mortality database & UN WPP and post 1985 UN MMEIG data
  - [x] includes maternal deaths, live births, maternal mortality ratio and maternal mortality rate
  - [x] calculated regional aggregates for areas with >90% population coverage
- [x] Updated all graphs which were using older versions of those indicators (s. note)
- [x] imported rest of the indicators of [UN MMEIG data set](http://staging-site-maternal-mortality/admin/datasets/6610) to grapher

### Notes

- (**Sources/ Origins**) I added the sources from the old WHO mortality database data set as origins to our new data set, since they would not show up in the grapher admin otherwise. Please let me know whether this is an appropriate way to do it or if there is an easier way to properly propagate both sources & origins in one data set

- (**Data artifacts from combining different sources**) Since the data source changes in 1985 from reported/ calculated figures to the UN estimates, in some countries there is a little jump in the data - usually small but occasionally big enough to be very noticable (e.g. Brazil 4x, Russia, Kyrgyzstan, Ecuador 2x). 
I have added a footnote to the chart, but let me know if you think I should remove these data points pre-1985)

- (**Replaced data sets**) I have replaced both the indicator in the old [long run dataset](https://admin.owid.io/admin/datasets/522) as well as the maternal mortality ratio and deaths in the [World Development Indicators by the World Bank](https://admin.owid.io/admin/datasets/6543). All these charts are now using the new long run indicators


etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.py
```diff
@@ -0,0 +1,225 @@
...
+    # drop all columns that are not 1) long run or 2) not related to maternal mortality
+    cols_to_keep = ["country", "year", "maternal_deaths", "mmr", "live_births", "mm_rate"]
+    tb = tb[cols_to_keep]
+
+    # drop rows where there is no data for maternal mortality indicators (introduced by aggregating regions)
+    tb = tb.dropna(subset=["maternal_deaths", "mmr", "mm_rate"], how="all")
+
+    # fix dtypes (coerce errors since NAs are not accepted otherwise)
+    tb["maternal_deaths"] = (
+        pd.to_numeric(tb["maternal_deaths"], errors="coerce")
```
spoonerf, line 143: Is `pr.to_numeric()` not possible? I'm not sure if it's included or not.

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.py
```diff
@@ -0,0 +1,225 @@
...
+    tb_region = geo.add_population_to_table(tb_region, ds_population, population_col="total_population")
+    tb_region["share_population"] = tb_region["population"] / tb_region["total_population"]
+    tb_region = tb_region[tb_region["share_population"] >= threshold]
+
+    tb_region = tb_region.drop(columns=["total_population", "share_population"])
+    tb = pr.concat([tb_region, tb_no_regions])
+    return tb
+
+
+def sources_to_origins_who(ds: Dataset) -> list[Origin]:
```
spoonerf, line 200: Nicely done!

etl/steps/data/grapher/un/2024-07-08/maternal_mortality.py
```diff
@@ -0,0 +1,40 @@
...
+    #
+    # Load inputs.
+    #
+    # Load garden dataset.
+    ds_garden = paths.load_dataset("maternal_mortality")
+
+    # Read table from garden dataset.
+    tb = ds_garden["maternal_mortality"]
+
+    # only include columns not in long run dataset data://garden/maternal_mortality/2024-07-08/maternal_mortality.py
```
spoonerf, line 19: I think you should keep all the columns here and preferentially use the UN MMEIG data for most of the maternal mortality charts, excluding the top two in [this dataset](https://admin.owid.io/admin/datasets/522).

The long-run variable is really good and important to have, but I think because we make some assumptions and combine the data in a slightly complicated way, it's better to keep the UN MMEIG-only variable as the primary one used in most charts. 

Could you also change the bottom two charts in [this dataset](https://admin.owid.io/admin/datasets/522) to the UN MMEIG-only variable?

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.meta.yml
```diff
@@ -0,0 +1,63 @@
...
+
+tables:
+  maternal_mortality:
+    title: Maternal mortality (Gapminder/UN MMEIG, long run)
+    variables:
+      mmr:
+        title: Maternal mortality ratio
+        description_short: |-
+          The estimated number of women who die from [maternal conditions](#dod:maternal-mortality) per 100,000 live births, based on data from death certificates, large-scale surveys, and statistical modeling.
+        unit: deaths per 100k live births
```
spoonerf, line 32: Could you have 100,000 instead of 100k

spoonerf (changes requested):
Hey @antea04! 

This is looking really good, you've done a great job, especially with the metadata. There's a few things I think should be changed: 

* I think because of the jumps in the data e.g. for Brazil and many other countries, you should only use the WHO/UN data for countries with data in the GapMinder dataset, i.e. countries in this file - etl/steps/data/garden/gapminder/2024-07-08/maternal_mortality.countries.json. I don't think adding WHO/UN data prior to the UN MMEIG makes sense for other countries as the jump in the data is pretty considerable, I guess because the cause-of-death reporting rates are lower in many of these countries? I think it makes sense to use it as a bridge between Gapminder and UN MMEIG, the jumps are pretty minimal for these countries and we can probably assume for these countries that cause of death recording is better. 

* I would include all the MMEIG variables in grapher and the mmr variable here should be the primary one used in most charts, apart from the top two in [this dataset.](https://admin.owid.io/admin/datasets/522) Having long-run data is important and we should definitely have some charts showing the very long-view. But I think for most of our charts we should be using the most straightforward data, with great metadata, where we haven't had to make many adjustments or assumptions, and in this case that's the MMEIG-only data.

antea04:
Hey @spoonerf! Thank you for the quick and thorough review!

I completely agree with your second point and will update the UN MMEIG data set and the corresponding charts. 

Regarding the point about the jumps in the data:
>   I think because of the jumps in the data e.g. for Brazil and many other countries, you should only use the WHO/UN data for countries with data in the GapMinder dataset, i.e. countries in this file - etl/steps/data/garden/gapminder/2024-07-08/maternal_mortality.countries.json. I don't think adding WHO/UN data prior to the UN MMEIG makes sense for other countries as the jump in the data is pretty considerable, I guess because the cause-of-death reporting rates are lower in many of these countries? I think it makes sense to use it as a bridge between Gapminder and UN MMEIG, the jumps are pretty minimal for these countries and we can probably assume for these countries that cause of death recording is better.

I have added the WHO/UN data specifically for the countries that do not show up in the Gapminder data. Saloni noticed when reviewing the data, that countries such as France only have data from 1985 onwards, even though (relatively high-quality) mortality data is available much earlier.  I think for many European countries (e.g. France, Spain, Austria etc.) the WHO/UN data is very good and there is no or only a minimal jump to the UN MMEIG estimates. 

On the other hand the WHO/ UN estimates add very little to the Gapminder figures, since they are very similar/ often identical. 

I see three options on how to proceed with the WHO/ UN data:
- I can take it out completely, which solves the issue with the jumps and simplifies the data set overall. Then the data for some countries from 1950-1985 would be lost, but that's the trade-off. This would roughly correspond to your suggestion, since the Gapminder data and WHO/ UN is very similar.
- I can leave it in as-is and make the reason for the jump clearer in the FASTT/ processing notes of the long-run data
- I can decide for which countries to include the WHO/ UN data with a decision rule (e.g. if the jump is within 50/30% of the 1984 reported MMR). This would include the countries with good reporting and eliminate big jumps, but it's an (a bit arbitrary) edit to the existing data and not very transparent at first glance for our readers. 

Let me know what you would prefer.

spoonerf:
Ah, apologies for my misunderstanding there. That is a tricky situation! 

I am leaning towards either your first or third option. Potentially [this variable from the WHO MDB](https://admin.owid.io/admin/variables/789511) could be used as a heuristic for data quality; that is how it is described in the WHO MDB:
Screenshot 2024-07-16 at 15 36 16

It could be that when this value is below 10% then the WHO/UN data is used, otherwise it is not used, although I think we would still have some jumps e.g. for Chile. 

@antea04 Do you have a sense for why there is a jump in the data, is it purely that WHO MDB misses some maternal deaths? 

@saloni-nd, would you be able to give your opinion on which approach might be better here?

antea04:
> Ah, apologies for my misunderstanding there. That is a tricky situation!

Ah, no problem at all! I agree, I don't have a strong sense of what would be the best solution. I am curious to hear Saloni's opinion as well!
> 
> @antea04 Do you have a sense for why there is a jump in the data, is it purely that WHO MDB misses some maternal deaths?

Yes, I think that is the main reason. Looking at [maternal deaths](http://staging-site-maternal-mortality/admin/variables/946982) from the long-run data set (WHO pre-1985, UN after) you can see the same jumps in 1985. 

It might be that some of the increase also comes from the way maternal deaths are counted: while the UN considers all deaths that stem from causes that are related or aggravated by pregnancies, it could be that some of those deaths are attributed to the underlying factor in the WHO mortality DB. (e.g. a death caused by HIV that is aggravated by the pregnancy would be classified as "infectious disease" rather that "maternal causes"). Since the WHO mortality DB gets its data from the country's death certifications, it might be that some countries are systematically mis-coding maternal deaths for any reason.

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.meta.yml
```diff
@@ -0,0 +1,63 @@
+definitions:
+  common:
+    processing_level: major
+    description_processing: |-
+      - We combined data from multiple sources (Gapminder, WHO mortality database, UN world population prospects, UN MMEIG) to display the historic development of maternal mortality.
```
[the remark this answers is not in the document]
antea04: Thank you!

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.py
antea04: I tried it and it didn't work :( I also looked for equivalent functions (something like tb.to_numeric()), but that also doesn't exist. Maybe we should add it?

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.py
antea04: Thank you!

snapshots/un/2024-07-08/maternal_mortality.zip.dvc
```diff
@@ -0,0 +1,34 @@
+# Learn more at:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+meta:
+  origin:
+    # Data product / Snapshot
+    title: Trends in maternal mortality
+    description: |-
+      The United Nations Maternal Mortality Estimation Inter-Agency Group (MMEIG) – comprising WHO, the United Nations Children’s Fund (UNICEF), the United Nations Population Fund (UNFPA), the World Bank Group and the United Nations Department of Economic and Social Affairs, Population Division (UNDESA/Population Division) has collaborated with external technical experts on a new round of estimates covering 2000 to 2020. The estimates represent the most up to date, internationally-comparable MMEIG estimates of maternal mortality, using refined input data and methods from previous rounds.
```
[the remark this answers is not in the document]
antea04: True, thank you for spotting that!

saloni-nd:
Hey @antea04 @spoonerf, sorry for the late reply as I was away!

If I'm understanding correctly, there seems to be a jump at 1985 when the UN MMEIG data starts (which tries to adjust for underreporting) compared to both Gapminder and the WHO MDB/UN WPP data before that.

For example, in Brazil's data in the [indicator](http://staging-site-maternal-mortality/admin/variables/946982).

Using the other indicators for a threshold would be tricky — (such as [share of ill-defined deaths](https://ourworldindata.org/grapher/deaths-registered-with-an-ill-defined-cause-of-death) or [share of deaths registered](https://ourworldindata.org/grapher/share-of-deaths-registered)) — because these indicators are only about data quality in the most recent years.

I think another decision rule like the degree of jump would also be somewhat arbitrary, and hard for us to justify in text.

Since it seems to affect our existing historical dataset with Gapminder data, I personally think the best option would be to clarify which datasets are used for each time period, perhaps in the footnote. What do you think?

antea04:
Hi @saloni-nd! No problem at all, thanks for getting back to us so quickly after your time off. 

> If I'm understanding correctly, there seems to be a jump at 1985 when the UN MMEIG data starts (which tries to adjust for underreporting) compared to both Gapminder and the WHO MDB/UN WPP data before that.
> 
> For example, in Brazil's data in the [indicator](http://staging-site-maternal-mortality/admin/variables/946982).

Yes, that is exactly right! I think also Brazil is the "worst offender" in this case as it jumps to almost 4x its 1984 value in 1985.

> Using the other indicators for a threshold would be tricky — (such as [share of ill-defined deaths](https://ourworldindata.org/grapher/deaths-registered-with-an-ill-defined-cause-of-death) or [share of deaths registered](https://ourworldindata.org/grapher/share-of-deaths-registered)) — because these indicators are only about data quality in the most recent years.

Oh, okay I was not aware of that - that would also explain why e.g. Chile still has such a high jump even though they do not record many ill-defined or unregistered deaths.

> I think another decision rule like the degree of jump would also be somewhat arbitrary, and hard for us to justify in text.

I agree - we are just taking out data that jumps just because it jumps, which is not based on anything. 

> Since it seems to affect our existing historical dataset with Gapminder data, I personally think the best option would be to clarify which datasets are used for each time period, perhaps in the footnote. What do you think?

I think that is valuable and the best option if we don't want to get rid of all the WHO/ UN data. Currently, I have a footnote on the charts that reads: "_Note: Deaths prior to 1985 are reported figures by the nations and might be undercounted. Sudden jumps in mortality rate in 1985 are a consequence of combining data sources and do not necessarily reflect a real increase in mortality_."

In the processing notes I then have a more detailed explanation: _The years before 1985 are based on the WHO mortality database and the UN WPP where available, and the Gapminder data set otherwise. The years after 1985 are based on the UN MMEIG data set._ 

I could combine them in the footnote maybe? Along the lines of _"Data before 1985 is based on reported figures from the WHO mortality database where available, and Gapminder otherwise. Data points after 1985 are estimated counts from the UN MMEIG. Sudden jumps in mortality rate in 1985 are a consequence of combining data sources."_

What do you think? Would that work? 

Thanks so much for taking the time and your insights, Saloni, we really appreciate it! 🥰

saloni-nd:
Thank you very much @antea04 !

For the footnote, how about:

> Note: Prior to 1985, only reported data are available, which are likely to underestimate the true maternal mortality rate. From 1985, estimates are shown, which aim to adjust for underreporting and misclassification. Sudden changes do not necessarily reflect an underlying increase.

And for the processing notes:

> The dataset combines three sources: WHO Mortality Database (before 1985), Gapminder (before 1985, if WHO Mortality Database data are unavailable), UN MMEIG (1985 onwards). The WHO Mortality Database contains reported figures from countries, and is likely to underestimate the true maternal mortality rate. The UN MMEIG aims to estimates the true rate, by adjusting for underreporting and misclassification. Sudden jumps in mortality rate in 1985 are a consequence of combining data sources.

spoonerf:
I'm not 100% certain we _should_ use it, but the ['Share of ill-defined deaths' ](https://admin.owid.io/admin/charts/7945/edit) is a time series*, so it _could_ be used as a proxy for data quality over time if I'm not mistaken? For example, in 1984, 22.5% of Brazil's deaths were recorded as ill-defined, which seems like a lot to me! 
 
Screenshot 2024-07-22 at 11 16 34

*The [Grapher chart](https://ourworldindata.org/grapher/deaths-registered-with-an-ill-defined-cause-of-death) we have is shown as a bar chart of the latest year only.

saloni-nd:
Oh, I didn't realise that indicator went back that far! Thank you @spoonerf.

I agree we could use it as a proxy of data quality, but I'm not sure if we should set a threshold based on it. 

Some of the trendlines also look quite strange, for example in Ukraine, Belarus and Russia, where the % of ill-defined deaths used to be very low. 

The Soviet Union [used to classify](https://onlinelibrary.wiley.com/doi/epdf/10.1111/j.1600-0447.1998.tb10763.x) some causes of death as 'secret', so I guess their data could be quite strange. I'm not sure if there are similar problems elsewhere.

antea04:
Thank you both @spoonerf and @saloni-nd ☺️ 

I will adjust the notes like you suggested, Saloni! Thanks for you feedback!

> I'm not 100% certain we should use it, but the ['Share of ill-defined deaths' ](https://admin.owid.io/admin/charts/7945/edit)is a time series*, so it could be used as a proxy for data quality over time if I'm not mistaken? For example, in 1984, 22.5% of Brazil's deaths were recorded as ill-defined, which seems like a lot to me!

Thank you Fiona! I did not notice that 😅 I also lean towards not using it for reasons Saloni mentioned and also because some data points are missing. We would need to decide how we handle those years, which might make it (even) harder to understand and retrace how we decided which data points to include. 

E.g. in 1984: 
Screenshot 2024-07-23 at 14 09 39

Let me know if you think there would be a better option - otherwise I will continue to include all the data. Thank you!

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.py
```diff
@@ -0,0 +1,225 @@
...
+
+def run(dest_dir: str) -> None:
+    """Creates long running data set on maternal mortality combining following sources (in brackets - timeframes available for each source):
+    - Gapminder maternal mortality data (1751 - 2008)
+    - WHO mortality database (1950 - 2020) & UN WPP (1950 - 2020)
+    - UN MMEIG maternal mortality data (1985 - 2020)
+    We combine them following the hierarchy below, where the most recent data is used when available:
+    - UN MMEIG >> WHO/UN >> Gapminder
+
+    For the timeframe ~1950-1984 we calculate maternal mortality ratio and rate out of WHO mortality database and UN WPP by using:
```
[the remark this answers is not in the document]
antea04: see discussion on github PR - I think it makes sense to include the WHO/UN data nonetheless. I added a big disclaimer in the footnote and hope this makes it clearer for readers browsing through the data

spoonerf:
Hey @antea04, that's a good point. I don't think there is an easy way to separate out which countries we feel have a nice smooth trend between WHO MDB and UN MMEIG and those where there is a bit of a jump. 

I do agree with both you and @saloni-nd that it is valuable to have this additional years in the dataset from the WHO MDB, particularly for countries [like France](http://staging-site-maternal-mortality/grapher/maternal-mortality?tab=chart&country=~FRA) where we can be fairly confident that the data quality is high and the splice in 1985 between the two datasets is relatively smooth.  

However, I am just not quite sure that the footnote is enough for cases [like Brazil](http://staging-site-maternal-mortality/grapher/maternal-mortality?tab=chart&country=~BRA), where the datasets just seem like they really aren’t comparable, and it seems at first glance like there is a mistake in the data, particularly in Brazil’s case as we don’t really gain much from the WHO data, just a few additional years.

But, as you mentioned, it’s hard to make this distinction between countries where the trend is smooth and the ones where it is jumpy in a way that isn’t arbitrary. @edomt it would be good to get your opinion on this.

edomt:
> However, I am just not quite sure that the footnote is enough for cases [like Brazil](http://staging-site-maternal-mortality/grapher/maternal-mortality?tab=chart&country=~BRA), where the datasets just seem like they really aren’t comparable, and it seems at first glance like there is a mistake in the data, particularly in Brazil’s case as we don’t really gain much from the WHO data, just a few additional years.

I completely agree: it is a big risk to show such a big discontinuity in cases where the OMM only gives us a few more years, and while the footnote does a good job of explaining the issue, we should assume that most people won't see it.

If you think there isn't a robust way to differentiate "good countries" from "bad countries" automatically, I can suggest two ways forward:

1. you could default to not including the pre-1985 data and only include it for a select number of countries for which you've reviewed the data and approved it (e.g., France).

2. you could include pre-1985 data for all countries but add a clear warning as an annotation for all countries where you see a big discontinuity. By "annotation", I mean the gray text that can be displayed below the country name (for example, "tests performed" and "people tested" [here](https://ourworldindata.org/explorers/coronavirus-data-explorer?facet=none&Metric=Tests&Interval=Cumulative&Relative+to+Population=true&Color+by+test+positivity=false&country=USA~BRA~JPN~DEU))

(One open question is whether annotations can be set through the ETL; I'm not sure they can. I meant to create a GitHub issue for @spoonerf to look at this — I'll do this now.)

etl/steps/data/garden/maternal_mortality/2024-07-08/maternal_mortality.py
spoonerf: Oh sorry I missed this! It looks like it is figured out now?

etl/steps/data/garden/un/2024-07-08/maternal_mortality.meta.yml
```diff
@@ -0,0 +1,148 @@
...
+dataset:
+  update_period_days: 365
+
+
+tables:
+  maternal_mortality:
+    variables:
+      births:
+        title: Estimated live births
+        description_short: The number of live births in a given year.
```
spoonerf, line 48: Perhaps worth doubling down on the 'estimated' with "The estimated number of live births in a given year.'

spoonerf (approved):
It looks good! Just a couple of minor comments: It's good to emphasize which variables are estimated in their `description_short`. 

Great work on this! It wasn't an easy one, but you led it really well and got a good result!

etl/steps/data/garden/un/2024-07-08/maternal_mortality.meta.yml
antea04: true! I'll include it!

PR #2950, opened 2024-07-09
📊 minerals: Add BGS minerals data

pabloarosado:
## Summary

In this PR:
* I created snapshot and data steps to fetch and process [the BGS' World Mineral Statistics dataset](https://www2.bgs.ac.uk/mineralsuk/statistics/wms.cfc?method=searchWMS).
* I have not created any grapher steps yet, since we need to decide what data and how we will present it (we may probably create an explorer with this and other datasets). I have not worked on the garden metadata yet, but I will come back to it depending on what indicators we end up using.
  * I have not even added a unit, because I'm not sure if they are consistent across all rows (individual data points have footnotes, so I wouldn't be surprised if some other units are also used, like short tons). Most likely, it's all metric tonnes, but I'll have to double-check.
* I have also added a very simple `read_from_df` table method to `owid.catalog`, to be able to easily create a table with metadata (and indicators metadata).
* I added some region aliases.

## Review

* [x] @spoonerf could you please review the `bgs` snapshot and data steps? Some of the regions in the dataset were a bit unusual, it would be good to double-check if the mapping makes sense.
* [x] @spoonerf could you please check if the aliases in the `regions.yml` file are correct?
* [x] @lucasrodes could you please have a look at the changes in `owid.catalog`?

Thank you both!


spoonerf (approved):
Looks all good to me, you're a data wizard making a nice clean dataset out of this!!

lucasrodes (approved):
LGTM!

PR #2951, opened 2024-07-09
:hammer: add read_table method to read table without setting primary_key

Marigold:
Reading tables from disk is fast (<1s). What is slow is setting a primary key with `set_index` (>6s). Some steps using population through helper functions could be speeded up by avoiding setting & resetting the index.

This PR adds `read_table` method as an alternative to `ds[table_name]` that lets you optimize the reading.

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -600,7 +600,7 @@ def add_population_to_dataframe(
 
     # Load population data.
     if ds_population is not None:
-        population = ds_population["population"].reset_index()
+        population = ds_population.read_table("population", primary_key=[])
     else:
         population = _load_population()
     population = population.rename(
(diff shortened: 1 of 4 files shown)


pabloarosado (approved):
Thanks for this! This implementation makes sense and would be very useful, but I'm hesitating because of two things:
1. When would we set `primary_key` different from either `None` or `[]`. I suppose `read_table` would probably fail if you pass anything that is not the correct primary key, right?
2. By default, the most convenient would be if `read_table()` returned the table with already reset index (since we almost always do `ds[table_name].reset_index()`).

So I'm wondering if, instead of `primary_keys`, the argument should be boolean, like `reset_index` (which would be True by default). Does that make sense?

Marigold:
You're right, `reset_index` is a better argument. I changed the default to True (to be different from `ds[table_name]`).

PR #2966, opened 2024-07-10
📊 ai: Epoch July update

veronikasamborska1994:
hi @lucasrodes! nothing has changed between the last time you looked at the code about a month ago! just updating the dataset with the newest version


lucasrodes (approved):
thanks for the update, Veronika! LGTM!

PR #2973, opened 2024-07-12
📊 population: update historical dataset

lucasrodes:
[Tracking issue](https://github.com/owid/owid-issues/issues/1561)

- Migrate Grapher dataset to ETL
  - Import historical WPP data via Fast-Track, using the data we had manually transcribed in the past.
    - I've added more data from revisions we hadn't transcribed. These are available from WPP's site: https://population.un.org/wpp/Download/Archive/Standard/. We now have data for all revisions since 1990.
  - Add data from recent reviews 2022 and 2024 via ETL (from Garden datasets)
- Update the chart with a gradient color scheme, and add two samples for 2002 and 2012 (to avoid large time gaps).

### DIFF
--- a/dag/demography.yml
+++ b/dag/demography.yml
@@ -44,6 +44,16 @@ steps:
   data://grapher/un/2024-07-12/un_wpp_full:
     - data://garden/un/2024-07-12/un_wpp
 
+  # WPP Historical comparison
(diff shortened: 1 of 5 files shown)


lucasrodes:
Left: current chart
Right: improved chart

image

PR #2974, opened 2024-07-12
📊 Add USGS minerals data

pabloarosado:
## Main changes

Sorry, this PR has snowballed a bit! Here's a summary of the main changes:

### Data changes
* Added USGS data on minerals.
* Fixed various issues on BGS data on minerals.
* Created steps to combine both, and build an explorer.
* Moved steps about minerals from `energy` to a new `minerals` namespace (with its own dag yaml file).

### ETL changes
* Added an `--explorer` flag to `etl run`, that creates an explorer and writes a `tsv` file to `owid-content`. We [discussed](https://www.notion.so/owid/2024-Data-architecture-60c39d558abf4eb2973cc52447d066fb?pvs=4#fff55afb361242719c6dc4dfaaa54869) this option in the data-architecture call and agreed that it's a move in the right direction. The implementation is not optimal, though:
  * Currently, `--explorer` sets an environment variable to 1. It would be better to handle this differently.
  * In theory, the new explorers step depends on the `grapher://grapher` step, so that's what should be added to the dag. However, this fails because `GrapherStep` does not have a `checksum_output()` method implemented, which makes the ETL break. For now, I used the `data://grapher` step as a dependency instead.
* Improved the `Explorer` object, so that it can handle catalog-path-based explorers. It can also easily translate from variable-id to catalog-paths (so we could use it for all existing indicator-based explorers).
* Added an `EXPLORERS_DIR` variable to `.env.example`. This variable is necessary if you want to create explorers from ETL, and your local `owid-content` is not in the same folder as `etl` (unlikely).

## For reviewers
This PR is not yet ready to be merged, but I expect that more and more small issues related to minerals data are going to come up. So it would be better to merge this PR soon, and then have additional PRs to fix issues and improve metadata.
* @Marigold could you please review the ETL changes (`etl/command.py`, `etl/config.py`, `etl/explorer_helpers.py`, `etl/helpers.py`)?
* @lucasrodes I'm going to keep fixing things in the coming days. But if you don't mind, please have a general look at the rest of the data-related changes (as you did with some of the files already) and let me know if you see any major issues. No need for a thorough review (and I still need to improve metadata significantly). If you want to have a look at the resulting explorer, you can [preview it](http://staging-site-add-usgs-minerals-data/admin/explorers/preview/minerals) (and untick the "Display locally edited explorer" option on the top-right). But again, it's not yet fully ready, I'll talk to Hannah about various issues.


etl/steps/data/garden/usgs/2024-07-15/mineral_commodity_summaries.py
```diff
@@ -0,0 +1,582 @@
...
+        d = d.rename(columns={"Reserves_kt_2023": "Reserves_kt"}, errors="raise")
+    if mineral == "POTASH":
+        # Reserves are given as "Reserves_ore_kt" and "Reserves_ore_K2O_equiv_kt".
+        # For consistency with other minerals, keep only the former.
+        # NOTE: This happens for 2023 and 2024, not 2022.
+        d = d.drop(columns=["Reserves_ore_K2O_equiv_kt"], errors="ignore")
+    if (year == 2023) & (mineral == "PEAT"):
+        # There are spurious notes in "Reserves_kt", and no column for "Reserves_notes".
+        # Fix this.
+        index_issue = d[d["Reserves_kt"] == "Included with “Other countries.”"].index
```
lucasrodes, line 151: Use `loc`

etl/steps/data/garden/usgs/2024-07-15/mineral_commodity_summaries.py
```diff
@@ -0,0 +1,582 @@
...
+    # There is also at least one case (2023 Nickel) of ">100000000". Remove the ">".
+    series = series.str.replace(">", "", regex=False)
+    # There is also at least one case (2022 Helium) of "<1". Remove the "<".
+    series = series.str.replace("<", "", regex=False)
+    # There is also at least one case (2022 Vermiculite) of 'Less than ½ unit.'. Replace it by "0.5".
+    series = series.str.replace("Less than ½ unit.", "0.5", regex=False)
+    # There is also at least one case (2024 Sand and gravel (industrial)) of numbers starting with "e", e.g. "4200".
+    # I suppose that probably means "estimated" (and the corresponding notes mention the word "estimated").
+    # Remove those "e".
+    series = series.str.replace("e", "", regex=False)
```
lucasrodes, line 208: I'd suggest passing a dictionary to `replace`, can make your could more readable.

```suggestion
    series = series.str.replace(
        {
            # Remove spurious commas from numbers, like "7,200,000".
            ",": "",
            # There is also at least one case (2023 Nickel) of ">100000000". Remove the ">".
            ">": "",
            # There is also at least one case (2022 Helium) of "<1". Remove the "<".
            "<": "",
            # There is also at least one case (2022 Vermiculite) of 'Less than ½ unit.'. Replace it by "0.5".
            "Less than ½ unit.": "0.5",
            # There is also at least one case (2024 Sand and gravel (industrial)) of numbers starting with "e", e.g. "4200".
            # I suppose that probably means "estimated" (and the corresponding notes mention the word "estimated").
            # Remove those "e".
            "e": "",
        },
        regex=False
    )    
```

lucasrodes (commented):
as I was checking this great contribution, I thought I'd leave some general suggestions :)

etl/steps/data/explorers/minerals/latest/minerals.py
```diff
@@ -0,0 +1,52 @@
...
+    metric_dropdown = []
+    commodity_dropdown = []
+    for column in tb.drop(columns=["country", "year"]).columns:
+        if tb[tb[column].notnull()].empty:
+            continue
+        metric = column.split("_")[0].capitalize()
+        commodity = " ".join(column.split("_")[1:]).capitalize()
+        variable_ids.append([f"{ds.metadata.uri}/{tb.metadata.short_name}#{column}"])
+        metric_dropdown.append(metric)
+        commodity_dropdown.append(commodity)
```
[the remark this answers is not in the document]
pabloarosado: Hey, thanks! Your suggestion is actually very useful, not just for clarity, but it speeds up the code significantly. I didn't realise that, by doing `tb[tb[column].notnull()].empty`, a new dataframe is created on every iteration. When removing it, the code is ~10 times faster ❤️

etl/steps/data/meadow/usgs/2024-07-15/historical_statistics_for_mineral_and_material_commodities.py
```diff
@@ -0,0 +1,174 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_path = Path(temp_dir)
+
+            # Extract supply_demand files.
+            supply_demand_dir = temp_path / "supply_demand"
+            zipf.extractall(path=temp_path)
+
+            # Iterate through the extracted files and apply process_file().
+            for file_path in supply_demand_dir.glob("*.xlsx"):
+                supply_demand_data[file_path.stem] = pr.ExcelFile(file_path)
```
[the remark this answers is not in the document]
pabloarosado: Good idea. Ideally, we should assert that a list of expected files are in the zip, but also (optionally) warn if new files are added (because, if new minerals are included, I'd like to fetch them too). Feel free to implement it yourself or create an issue (I don't have the appetite for this PR).

etl/steps/data/meadow/usgs/2024-07-15/historical_statistics_for_mineral_and_material_commodities.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Remove spurious "Unnamed: X" columns.
+    df = df.drop(columns=[column for column in df.columns if column.startswith("Unnamed")])
+
+    # Add commodity and unit columns.
+    df["commodity"] = sheet_name
+    df["unit"] = unit
+
+    # Extract notes written below the table in the first column.
+    notes = list(df[~df["Year"].astype(str).str.isdigit()]["Year"])
```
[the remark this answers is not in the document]
pabloarosado: Good point, I'm used to str, but I suppose "string" is superior.

etl/steps/data/meadow/usgs/2024-07-15/historical_statistics_for_mineral_and_material_commodities.py
```diff
@@ -0,0 +1,174 @@
...
+    #
+    # Process and combine data for all commodities.
+    combined = combine_data_for_all_commodities(supply_demand_data=supply_demand_data)
+
+    # Create a table with metadata.
+    tb = pr.read_from_df(data=combined, metadata=snap.to_table_metadata(), origin=snap.metadata.origin, underscore=True)
+
+    # Columns contain a mix of numbers and strings (I suppose corresponding to footnotes).
+    # For now, save all data as strings.
+    tb = tb.astype({column: str for column in tb.columns})
```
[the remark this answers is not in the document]
pabloarosado: Hmmm, if I store the "year" column as a `str`, ETL (or, I guess, the repack library?) will pack it as an integer. But if I store it as `"string"`, then it's stored as a real string (and the following step fails). So, shall we adapt repack to also be smart about `"string"` columns? @Marigold (not urgent)
For now, I'll leave the code with `str`, to avoid the issues.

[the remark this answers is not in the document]
Marigold: Good point. `repack` indeed leaves `"string"` type alone and doesn't convert it to int / categorical. I think we should be consistent and repack it in the same way as we repack `str` / `object`

pabloarosado:
Hmmm, I guess this was closed automatically because of Mojmir's fix on a comment from this PR.

etl/steps/data/garden/usgs/2024-07-15/mineral_commodity_summaries.py
pabloarosado: This doesn't work, but I've changed it for better performance in the following way:
```
    series = (
        # Remove spurious commas from numbers, like "7,200,000".
        series.str.replace(",", "", regex=False)
        # There is also at least one case (2023 Nickel) of ">100000000". Remove the ">".
        .str.replace(">", "", regex=False)
        # There is also at least one case (2022 Helium) of "<1". Remove the "<".
        .str.replace("<", "", regex=False)
        # There is also at least one case (2022 Vermiculite) of 'Less than ½ unit.'. Replace it by "0.5".
        .str.replace("Less than ½ unit.", "0.5", regex=False)
        # There is also at least one case (2024 Sand and gravel (industrial)) of numbers starting with "e", e.g. "4200".
        # I suppose that probably means "estimated" (and the corresponding notes mention the word "estimated").
        # Remove those "e".
        .str.replace("e", "", regex=False)
    )
```

Marigold:
@pabloarosado fixed the issue with unit tests on master.

pabloarosado:
Thanks @lucasrodes for the suggestions!

pabloarosado:
> @pabloarosado fixed the issue with unit tests on master.

Thanks @Marigold for the fix!

etl/helpers.py
```diff
@@ -1048,3 +1058,51 @@ def create_dag_archive_file(dag_file_archive: Path) -> None:
...
+
+class DatasetAndExplorer:
+    def __init__(self, dataset, explorer):
+        self.dataset = dataset
+        self.explorer = explorer
+
+    def save(self):
+        # Save ETL dataset to disk.
+        self.dataset.save()
+        if EXPLORER:
```
Marigold, line 1071: This didn't work for me. You're changing the env dynamically in `command.py`, but this value is from `config.py` when modules were loaded. You should use `os.environ.get("EXPLORER")` instead (that fixed it for me).

dag/minerals.yml
```diff
@@ -0,0 +1,61 @@
...
+  #
+  # Minerals - Minerals.
+  #
+  data://grapher/minerals/2024-07-15/minerals:
+    - data://garden/minerals/2024-07-15/minerals
+  #
+  # Minerals - Minerals explorer.
+  #
+  data://explorers/minerals/latest/minerals:
+    # TODO: The following would be the right dependency, but it fails because GrapherStep.checksum_output() is not implemented.
```
Marigold, line 59: Why is `data://grapher` step not sufficient?

Marigold (approved):
Looks good! You've created a very good workaround. I'll try to refactor it into a special step in a different PR, but it shouldn't block you from merging it.

etl/helpers.py
pabloarosado: That's odd, the current implementation works for me.
Screenshot 2024-07-29 at 10 22 03
But your suggestion works too, so I'll implement your suggestion (and remove EXPLORER from config). At some point we can refactor to handle this properly.

dag/minerals.yml
pabloarosado: Hmmm. Just to be able to execute the steps, the `data://grapher` dependency will work. But, if you don't execute the `grapher://grapher` step, then you won't see the changes in the admin.
However, given that the `--explorer` flag also forces `--grapher`, then maybe the dependency on the `data://grapher` step is enough. I'll remove the TODO.

PR #2981, opened 2024-07-16
📊 ai: CSET investments and articles

veronikasamborska1994:
hi @spoonerf! I need to get in touch with the data provider to find out whether we should be using the data for the more recent years as it's currently in complete but it'd be good to have this code reviewed so I can merge when I know the answer :) thanks very much!


etl/steps/data/garden/artificial_intelligence/2024-07-16/cset.meta.yml
```diff
@@ -0,0 +1,232 @@
...
+dataset:
+  update_period_days: 365
+
+
+tables:
+  cset:
+    variables:
+      disclosed_investment:
+        title: Disclosed investment
+        unit: US dollars
```
spoonerf, line 52: It sounds like the adjusting for inflation situation is pretty complicated, but not sure if it makes sense to have something here in line with the [harmonizing document?](https://docs.google.com/document/d/1tcknTryJAxOR0yv8x38s3jkONq-TY3T4HNCEcZQ_c-s/edit?pli=1#heading=h.qd6mxo3fpwgz) e.g. constant 2021 US$

etl/steps/data/garden/artificial_intelligence/2024-07-16/cset.py
```diff
@@ -0,0 +1,118 @@
...
+    # List of members representing different regions CSET and use these for World aggregation
+    members = [
+        "North America",
+        "Europe",
+        "Asia Pacific",
+        "Africa",
+        "Latin America and the Caribbean",
+        "Oceania",
+        "OECD",
+        "Five Eyes (Australia, Canada, New Zealand, UK, and the US)",
```
spoonerf, line 26: Maybe there is some double-counting here? E.g. Australia is in there twice, US is in a few different groups too.

spoonerf (approved):
Looks good! I just had a couple of small queries but nothing big.

etl/steps/data/garden/artificial_intelligence/2024-07-16/cset.py
veronikasamborska1994: ah sorry I am actually excluding these for the World aggregation despite the comment above :) I've changed that now!

PR #2985, opened 2024-07-17
📊 demography: population density

lucasrodes:
- Update population density to use latest population data available.
- Instead of creating an additional dataset, add indicator to Population main dataset.


etl/steps/data/garden/demography/2024-07-15/population.meta.yml
```diff
@@ -65,6 +65,23 @@ tables:
...
+  population_density:
+    common:
+      presentation:
+        attribution: |-
+          HYDE (2023); Gapminder (2022); UN WPP (2024); UN FAO (2024)
+    variables:
+      population_density:
+        title: Population density
+        description_short: |
+          Population per unit of area.
```
spoonerf, line 78: Nit pick but perhaps just worth stating what the unit of area is.

spoonerf (approved):
LGTM!

PR #2991, opened 2024-07-18
📊 Country latitude centroids

spoonerf:
Hey @veronikasamborska1994! 

I thought it might be useful to add a latitude variable to Grapher; I thought it shows the hump-shape of temperature anomalies quite nicely around the equator, e.g. [here](http://staging-site-centroids/admin/charts/7942/edit). It could be useful for other environmental variables too e.g., there is a latitudinal diversity gradient with species numbers being much higher around the equator.

### DIFF
--- a/etl/steps/data/grapher/geography/2024-07-18/latitude.py
+++ b/etl/steps/data/grapher/geography/2024-07-18/latitude.py
@@ -1,5 +1,8 @@
 """Load a garden dataset and create a grapher dataset."""
 
+from owid.catalog import Table
+from owid.catalog import processing as pr
+
 from etl.helpers import PathFinder, create_dataset
 
 # Get paths and naming conventions for current step.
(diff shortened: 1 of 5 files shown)


etl/steps/data/grapher/geography/2024-07-18/latitude.py
```diff
@@ -0,0 +1,57 @@
...
+    #
+    # Load inputs.
+    #
+    # Load garden dataset.
+    ds_garden = paths.load_dataset("nat_earth_110")
+
+    # Read table from garden dataset.
+    tb = ds_garden["nat_earth_110"].reset_index()
+    tb = tb[["name", "latitude_centroid", "absolute_latitude"]]
+    tb["year"] = 2024
```
veronikasamborska1994, line 19: I think it might be useful to have these not just for 2024 but maybe dating back for every year back to 1900s?

veronikasamborska1994 (approved):
hi @spoonerf ! this is a great addition!!! thanks so much for adding these :) I only have one suggestion but I think it'd be good to have these for not just year 2024 (also as we wouldn't be able to use it since the anomalies data isn't complete until the year ends).

etl/steps/data/grapher/geography/2024-07-18/latitude.py
spoonerf: Great point! I'll add it for 1900-2100.

PR #2995, opened 2024-07-22
:sparkles: exclude certain datasets from metadata changes

Marigold:
Fix for https://github.com/owid/etl/pull/2995

This PR reduces false positives from data & metadata changes (all are described in code in detail)
- Exclude certain datasets like COVID and AI Epoch from `METADATA CHANGE` in chart-diff (fixing the root cause turned out to be too challenging)
- Use `catalogPath` instead of `variableId` when comparing data & metadata checksums
- Remove data & metadata checksums change when variable is updated by indicator upgrader

If you want to test it, you can check that running it against `cset-ai` shows only relevant config changes
```
STAGING=cset-ai streamlit run apps/wizard/app_pages/chart_diff/app.py
```

### DIFF
--- a/apps/wizard/app_pages/chart_diff/chart_diff.py
+++ b/apps/wizard/app_pages/chart_diff/chart_diff.py
@@ -15,6 +15,16 @@ ADMIN_GRAPHER_USER_ID = 1
 log = get_logger()
 
 
+# Some datasets like COVID or certain AI datasets use {TODAY} in their metadata, making the metadata dependent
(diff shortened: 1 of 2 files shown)


lucasrodes:
Sweet, I see there are three charts appearing in chart-diff if I enable 'show all' option ー is that expected?

Marigold:
Those charts use different `descriptionKey`. I have no idea why, but I'll look into it in a different PR.

PR #3007, opened 2024-07-23
📊 ai: papers with code ImageNet

veronikasamborska1994:


Marigold (approved):
Code looks good, but the [new chart](http://staging-site-papers-with-code-imagenet/etl/wizard/chart-diff) doesn't show Top-5 accuracy. Is that expected?

veronikasamborska1994:
> Code looks good, but the [new chart](http://staging-site-papers-with-code-imagenet/etl/wizard/chart-diff) doesn't show Top-5 accuracy. Is that expected?

thank! yep, they stopped updating it!

PR #3012, opened 2024-07-25
:sparkles: Refactor repack_series function to handle string dtype

Marigold:
Fix https://github.com/owid/etl/pull/2974#discussion_r1691003507

`"string"` type should be handled in the same way as `object` (`str`) by `repack`.


pabloarosado (approved):
Thanks, that was quick! Mojmir workspace is still faster than copilot.

PR #3018, opened 2024-07-25
📊 ai: adoption by organizations (AI Index)

veronikasamborska1994:
hi @antea04 ! I am adding another dataset from the AI index report that I previously asked you to review. It's very simple and pretty short. no rush to review, just when you get a chance to take a look! thanks very much!

### DIFF
--- a/etl/steps/data/garden/artificial_intelligence/2024-06-28/ai_adoption.meta.yml
+++ b/etl/steps/data/garden/artificial_intelligence/2024-06-28/ai_adoption.meta.yml
@@ -20,9 +20,10 @@ tables:
     variables:
        pct_of_respondents:
         title: Organizations reporting having adopted AI
-        description_short: Share of companies using AI technology (e.g., machine learning, computer vision, or natural language processing) in at least one business function, based on an average of 1,763 companies that responded to a global survey over those years.
+        description_short: Share of companies using AI technology (e.g., machine learning, computer vision, or natural language processing) in at least one business function.
         description_key:
           - Share of respondents who reported that their organizations have adopted AI in at least one function, with a breakdown by region. The regional groupings, as defined by McKinsey & Company, classify countries into specific categories. The "Developing Markets" grouping includes countries such as India, Latin America, and the Middle East and North Africa (MENA). The "Greater China" grouping includes Hong Kong and Taiwan.
+          - The number of companies that respond to the survey varies between years and ranged between 1365 and 1843 between 2021 and 2024.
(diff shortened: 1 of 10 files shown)


etl/steps/data/garden/artificial_intelligence/2024-06-28/ai_adoption.meta.yml
```diff
@@ -0,0 +1,29 @@
...
+dataset:
+  update_period_days: 365
+
+
+tables:
+  ai_adoption:
+    variables:
+       pct_of_respondents:
+        title: Organizations reporting having adopted AI
+        description_short: Share of companies using AI technology (e.g., machine learning, computer vision, or natural language processing) in at least one business function, based on an average of 1,763 companies that responded to a global survey over those years.
```
antea04, line 23: I don't fully understand where the 1763 comes from, but since the n for this survey seems to change each year (according to McK n=(1843, 1492, 1684, 1365) for 2021-24) I would argue giving the average here is error-prone and not super insightful - I would probably say something like (more than XXX companies or between XXX and YYY companies), but ultimately up to you.

antea04 (approved):
Hi Veronika!

The PR looks good to me! I had two little comments regarding the metadata on the number of respondents, but feel free to handle that in the way that seems best to you!

One tiny point about the chart - currently the AI Index is listed twice as a source with two different years. You could see if you want to include only the latest issue (2024) or list the source only once with both years (2023, 2024), or if you prefer to leave it as is ☺️ 

Thanks for your patience and from my side it is good to go!

etl/steps/data/garden/artificial_intelligence/2024-06-28/ai_adoption.meta.yml
veronikasamborska1994: good point! I've removed it from the subtitle and added a note on this in description_key now! thanks!

veronikasamborska1994:
> Hi Veronika!
> 
> The PR looks good to me! I had two little comments regarding the metadata on the number of respondents, but feel free to handle that in the way that seems best to you!
> 
> One tiny point about the chart - currently the AI Index is listed twice as a source with two different years. You could see if you want to include only the latest issue (2024) or list the source only once with both years (2023, 2024), or if you prefer to leave it as is ☺️
> 
> Thanks for your patience and from my side it is good to go!

thanks a lot for the review @antea04 ! the sources were automatically propagated via the metadata but I agree it's much better to shorten it :)

PR #3019, opened 2024-07-26
 :bug: Make widget keys unique in chart-diff

Marigold:
I saw errors from chart-diff on this [staging server](http://staging-site-yougob/etl/wizard/chart-diff?change_type=new&change_type=config) about `DuplicateWidgetID`. Adding suffix `chart_id` to all those widgets fixed them, but I'm not sure if there could be any side effects.


lucasrodes (approved):
shouldn't cause any issue ー thanks, Mojmir!

PR #3020, opened 2024-07-26
📊 WHO Mortality DB 2024 Update

spoonerf:
Hey @antea04! 

This is an update of the WHO Mortality DB. I've duplicated the steps used in a previous update on the `who-mdb` branch and then made further changes on the `who-mdb-update` branch, so you should just be seeing the differences between those two branches (apart from Snapshot, which I forgot to do this for 🙈). 

Let me know if you have any questions!


etl/steps/data/garden/who/2024-07-26/mortality_database.py
```diff
@@ -17,7 +14,7 @@ def run(dest_dir: str) -> None:
     # Load inputs.
     #
     # Load meadow dataset.
-    ds_meadow = cast(Dataset, paths.load_dependency("mortality_database"))
+    ds_meadow = paths.load_dataset("mortality_database")
 
     # Read table from meadow dataset.
     tb = ds_meadow["mortality_database"]
```
antea04, line 20: I have learned from Pablo that you can now combine getting the table and reset index by using tb = ds_meadow.read_table("mortality_database") :)

snapshots/who/2024-07-26/mortality_database.csv.dvc
```diff
@@ -0,0 +1,35 @@
...
+
+          Where they are collected, death registration data are the best source of information on key health indicators, such as life expectancy, and death registration data with cause-of-death information are the best source of information on mortality by cause, such as maternal mortality and suicide mortality.
+
+          WHO requests from all countries annual data by age, sex, and complete ICD code (e.g., 4-digit code if the 10th revision of ICD was used). Countries have reported deaths by cause of death, year, sex, and age for inclusion in the WHO Mortality Database since 1950.
+
+          Data are included only for countries reporting data properly coded according to the International Classification of Diseases (ICD). Today the database is maintained by the WHO Division of Data, Analytics and Delivery for Impact (DDI) and contains data from over 120 countries and areas. Data reported by member states and selected areas are displayed in this portal’s interactive visualizations if the data are reported to the WHO mortality database in the requested format and at least 65% of deaths were recorded in each country and year.
```
antea04, line 14: Maybe we could write something like "The WHO only includes data, which is/ are properly coded according to ..." - It would be easier to parse and easier to understand who makes this choice to include/ not include data.

snapshots/who/2024-07-26/mortality_database.csv.dvc
```diff
@@ -0,0 +1,35 @@
...
+    date_published: "2024-02-21"
+
+    # Citation
+    producer: WHO Mortality Database
+    citation_full: |-
+      WHO Mortality Database, Data, Analytics & Delivery for impact (DDI), World Health Organization (2024)
```
antea04, line 20: tiniest nitpick, but they refer to themselves as "WHO Division of Data, Analytics and Delivery for Impact (DDI)", so maybe I would stick to that (except if the citation is given somewhere like that, but I couldn't find it)

snapshots/who/2024-07-26/mortality_database.py
```diff
@@ -0,0 +1,128 @@
+"""Script to create a snapshot of dataset 'WHO Mortality Database (2022)'.
+The data download for this step is quite manual.
```
antea04, line 2: [Not a comment] I thought this said "minimal" and already wanted to protest 😂

etl/steps/data/grapher/who/2024-07-26/mortality_database.py
```diff
@@ -16,7 +14,7 @@ def run(dest_dir: str) -> None:
     # Load inputs.
     #
     # Load garden dataset.
-    ds_garden = cast(Dataset, paths.load_dependency("mortality_database"))
+    ds_garden = paths.load_dataset("mortality_database")
 
     # Read table names from garden dataset.
     table_names = sorted(ds_garden.table_names)
```
antea04, line 20: Maybe there is a good reason not to do it, but I would pull all this in a list comprehension like:
tables = [ds_garden[tab] for tab in sorted(ds_garden.table_names)]

and then ds_grapher = create_dataset(dest_dir, tb=tables ...)

etl/steps/data/garden/who/2024-07-26/mortality_database.py
```diff
@@ -50,7 +50,8 @@ def tidy_sex_dimension(tb: Table) -> Table:
     Improve the labelling of the sex column
     """
     sex_dict = {"All": "Both sexes", "Female": "Females", "Male": "Males", "Unknown": "Unknown sex"}
-    tb["sex"] = tb["sex"].replace(sex_dict, regex=False)
+    # tb["sex"] = tb["sex"].replace(sex_dict, regex=False)
```
antea04, line 53: Do you anticipate that we will need to revert these changes again at some point? Otherwise I would take these comments out (they are a bit confusing imo)

etl/steps/data/garden/who/2024-07-26/mortality_database.py
```diff
@@ -28,20 +25,23 @@ def run(dest_dir: str) -> None:
...
 
 def tidy_causes_dimension(tb: Table) -> Table:
     """
-    To clarify blood disorders are also included in this group.
+    Improve the labelling of the cause column to make it clear that blood disorders are included in this category.
     """
     cause_dict = {"Diabetes mellitus and endocrine disorders": "Diabetes mellitus, blood and endocrine disorders"}
-    tb["cause"] = tb["cause"].replace(cause_dict, regex=False)
+    tb["cause"] = tb["cause"].cat.rename_categories(lambda x: cause_dict.get(x, x))
+    # tb["cause"] = tb["cause"].replace(cause_dict, regex=False)
```
antea04, line 44: s. other comment

etl/steps/data/garden/who/2024-07-26/mortality_database.py
```diff
@@ -79,127 +80,6 @@ def tidy_age_dimension(tb: Table) -> Table:
         "[All]": "all ages",
     }
 
-    tb["age_group"] = tb["age_group"].replace(age_dict, regex=False)
-
+    # tb["age_group"] = tb["age_group"].replace(age_dict, regex=False)
```
antea04, line 83: s. other comment

etl/steps/data/garden/who/2024-07-26/mortality_database.meta.yml
```diff
@@ -4,7 +4,10 @@ definitions:
     presentation:
       topic_tags:
         - Causes of Death
-
+    processing_level: minor
+    description_key: ["The ICD 10 codes that define this cause of death are << icd10_codes  >>."]
+  footnote: |-
+    <% if age == "Age-standardized" %>To allow for comparisons between countries and over time, this metric is [age-standardized](#dod:age_standardized).<%- endif -%>
```
antea04, line 10: [Not a comment] This is a general question I personally have but - this does not show up in the grapher automatically, right? I checked this for the age-standardized indicators in the staging server and the footer is not prepopulated. Or is this a "WIP"-feature?

spoonerf: Ah no, you are totally right, I applied this wrong 🤦🏼‍♀️, good catch!

antea04: Oh really? Because this happened to me before as well and since the footer box in grapher doesn't have the little clickable link I always assumed this just doesn't work (yet 😅) Could you tell me how you fixed it?

spoonerf: It should be there now 🙏🏼

antea04 (approved):
Hi Fiona! 

I just approved the PR ☺️  As you can see I only left a few small comments regarding code suggestions and metadata nitpicks. Feel free to disregard them, if you prefer keeping the code this way.

I also have some thoughts on the charts:

- In the meta.yml file you added into the subtitle that indicators are age-standardized indicators, but many of the charts are still only mentioning it in the footnote. Maybe it makes sense to add it to the subtitle if it does not make it too long (with a DoD)

- For some of the charts showing death rates (e.g. [maternal mortality](http://staging-site-who-mdb-update/grapher/death-rate-from-maternal-disorders), [digestive diseases](http://staging-site-who-mdb-update/grapher/death-rate-from-digestive-diseases-who-mdb) and [diabetes](http://staging-site-who-mdb-update/grapher/death-rate-from-diabetes-who-mdb?time=1996)) it would be good to add somewhere (e.g. the footnote) that not all deaths are registered with a cause of death

Thank you for trusting me with the review! Let me know if you have any questions or you disagree on any points!

etl/steps/data/garden/who/2024-07-26/mortality_database.py
spoonerf: Ah no way, I didn't know that, thanks!

snapshots/who/2024-07-26/mortality_database.csv.dvc
spoonerf: Good point, I will update this!

snapshots/who/2024-07-26/mortality_database.csv.dvc
spoonerf: Yeh that's a good point, I'll use "WHO Division of Data, Analytics and Delivery for Impact (DDI)"  instead

etl/steps/data/grapher/who/2024-07-26/mortality_database.py
spoonerf: Ah very good point, shall do this

etl/steps/data/garden/who/2024-07-26/mortality_database.meta.yml
spoonerf: Ah in this case I was using the wrong index name to make the Jinja work. I think it depends a little on whether there is something in the footnote already, if there is it won't be overwritten. 

However, it does seem a little bit unpredictable when it works and when it doesn't. I'll take a look into it. Do you have an example to hand of when it didn't work for you?

antea04: Ah, that makes sense! I think in all the cases I tried it, there was a footnote already on the chart and I could not get the one from the etl to show up 😅 I will keep that in mind though - if I run into another issue with it I'll let you know!

PR #3021, opened 2024-07-26
🎉 Add new export:// step for creating TSV files and committing to Github repos

Marigold:
## Motivation:

We often want to do something after we create the dataset - create TSV explorer file for commit to `owid-content` or pushing data as csv to Github repository. These side effects or "actions" currently require manual effort.

## Solution:

Add new `ActionStep` named `action://actions/namespace/version/short_name`. This step can call arbitrary functions (e.g. `create_explorer` or `commit_file_to_github`) and will be executed only once (and again if dependencies change). It is triggered with `--action` flag (similar to `--grapher` flag).

This is an extension to Pablo's explorers solution (see the [discussion in Notion](https://www.notion.so/owid/2024-Data-architecture-60c39d558abf4eb2973cc52447d066fb?pvs=4#2eda2f8fc43b4cefb2ee6aa3f82a128e)).

## TSV files for explorers

In addition to creating a local file in `owid-content`, we'll also upload it to the staging server.

## Committing to Github repositories

Call `commit_file_to_github` to commit the file. It runs in dry mode unless `ETL_ACTIONS_COMMIT` is set in `.env` (supposed to be true only in production env).

I tried migrating `co2_data` step from `external` channel. It's just a proof of concept, we might not want to update it whenever there's a minor change in data (we are not creating new commits if files are identical).

## TODO before merging
- [ ] Make sure all files like minerals.py for creating explorer have the most recent changes from master


Marigold:
@pabloarosado I renamed it to `ExportStep` and used directory structure like `export://github/co2_data/latest/owid_co2` or `export://explorers/minerals/latest/minerals`. Let me know what you think.

etl/files.py
```diff
@@ -282,3 +282,30 @@ def watch_folder(path: Path) -> Generator[Path, None, None]:
...
+        raise ValueError(f"The target '{target}' is not properly formatted. Expected format: 'user@host:/remote/path'.")
+
+    try:
+        # Construct the scp command
+        scp_command = ["scp", str(local_file_path), target]
+
+        # Execute the command
+        subprocess.run(scp_command, check=True, text=True, capture_output=True)
+        print(f"File {local_file_path} successfully uploaded to {target}")
+    except subprocess.CalledProcessError as e:
```
pabloarosado, line 310: Is there a way to ensure this cannot happen by accident in production? For example, if I run an explorer step with `--export` and I forget that I'm currently in the `master` branch.

etl/steps/export/explorers/minerals/latest/minerals.py
```diff
@@ -180,4 +180,4 @@ def run(dest_dir: str) -> None:
     #
     # Create a new explorers dataset and tsv file.
     ds_explorer = create_explorer(dest_dir=dest_dir, config=config, df_graphers=df_graphers)
-    ds_explorer.save()
+    ds_explorer.write()
```
pabloarosado, line 183: Maybe, instead, it's better to rename `Explorer.write` -> `Explorer.save` for consistency. Up to you!

etl/steps/export/github/co2_data/latest/owid_co2.py
```diff
@@ -0,0 +1,363 @@
...
+
+    # Prepare outputs.
+    combined = prepare_outputs(combined=combined, ds_regions=ds_regions)
+
+    gh.commit_file_to_github(
+        combined.to_csv(),
+        repo_name="co2-data",
+        file_path="owid-co2-data.csv",
+        commit_message=":bar_chart: Automated update",
+        branch="master",
```
pabloarosado, line 362: I'm not sure about this. Would it be possible to create a PR instead (maybe some of what we do for `draft-pr`)? If that's too complicated, no problem.
Usually, when updating CO2 or energy data, I also manually update the README a bit, to explain the changes and update the versions. Ideally, those README changes could also happen here. But then we would need to commit multiple files.

pabloarosado: One option would be to have an `etl.helper` function to handle whether you want to push directly or create a PR, and which files you want to commit.

pabloarosado (approved):
This is great! All your choices make sense to me, and this makes the workflow to write to external repos and explorers much easier and intuitive.
I left some suggestions and things I'm not sure about, but feel free to disregard them, we can improve them later on.
One additional thought/question is: If you want to run the etl to write to an explorer step, do you have to type `etlr [step] --grapher --export`? If so, it's not a big deal, but maybe we could also have an `--explorer` flag that does both (as we have now). If you think that's an overkill, no problem.

lucasrodes (commented):
This looks fantastic, thanks for working on this, Mojmir!

Just two comments:
- It would be nice if we could add a section to our docs. Maybe [here](https://docs.owid.io/projects/etl/architecture/workflow/) or [here](https://docs.owid.io/projects/etl/architecture/workflow/other-steps/). This section could list the various 'export' steps that we have, explaining their purposes and perhaps with a small code-snippet or links to example steps. I find it useful if the user knows right away that they need to use `create_explorer` instead of `create_dataset`, or `gh.commit...`, etc.
- I've [migrated our COVID-19 explorer to ETL](https://github.com/owid/etl/tree/master/etl/steps/data/explorers/covid/latest), in case there is any tweak needed

etl/files.py
Marigold: Good question! This is handled by `if config.STAGING:` condition in a function that's calling this one and should never push to `master` (even if it does, it'd push to `staging-site-master` which is not synced back to production).

etl/steps/export/explorers/minerals/latest/minerals.py
Marigold: Makes sense. I was overly creative.

etl/steps/export/github/co2_data/latest/owid_co2.py
Marigold: Oh right, I didn't realize that CO2 and monkeypox would have different update frequency and automatic updates might not be feasible. Creating a PR would definitely be an option, but it'd be better to do it during the next CO2 update to get a better sense of what it entails.

How about this workflow for now:
1. You create a PR with `etl d pr my-update ...`
2. In the export step, we only commit to a branch if an env variable is set. E.g.
```
if os.environ["CO2_BRANCH"]:
   gh.commit_file_to_github(..., branch=os.environ["CO2_BRANCH"])
```
3. You call it with `CO2_BRANCH=my-update etlr ... --export`

You can iterate on this, make sure your PR looks good, and then merge.

For monkeypox, it'd be committing right to master if `ETL_ACTIONS_COMMIT` env is set (in production).

pabloarosado: Thanks. It's not ideal to have an env variable for each repos (one for co2, one for energy, one eventually for poverty, maybe others). But for now if it works, I'm fine with it.

PR #3028, opened 2024-07-29
 :bug: Helper function `geo.add_population_to_dataframe` uses old version of population dataset

Marigold:
We had an error in nightly build about `garden/health/2023-08-16/deaths_karlinsky` not having `garden/owid/latest/key_indicators` as dependency, despite having `data://garden/demography/2023-03-31/population` as dependency.

It turned out that it even though we were expecting version `2023-03-31/population` in dependencies, function `geo.add_population_to_dataframe` for adding population is using `key_indicators` by default (instead of population dataset) and **key_indicators use outdated `2022-12-08/population`**.

There are a couple of options how to fix it:
- Replace deprecated `geo.add_population_to_dataframe` function by `geo.add_population_to_table` everywhere
- Make `key_indicators` use the latest `2023-03-31/population` population dataset

Perhaps we could do both?

This PR fixes only a single case so far, but I can add more once we agree what the fix should be.

### DIFF
--- a/etl/data_helpers/geo.py
+++ b/etl/data_helpers/geo.py
@@ -547,7 +547,7 @@ def harmonize_countries(
     return df_harmonized  # type: ignore
 
 
-def add_population_to_dataframe(
+def _add_population_to_dataframe(
     df: TableOrDataFrame,
     ds_population: Optional[Dataset] = None,
     country_col: str = "country",
@@ -733,7 +733,7 @@ def add_population_to_table(
 
     """
     # Create a dataframe with an additional population column.
-    df_with_population = add_population_to_dataframe(
+    df_with_population = _add_population_to_dataframe(
         df=tb,
         ds_population=ds_population,
         country_col=country_col,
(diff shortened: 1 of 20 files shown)


pabloarosado:
Hi @Marigold the best solution is replacing `add_population_to_dataframe` by `add_population_to_table` everywhere. That way we not only fix the issue for now, but also for the future. But if for some of the steps it gets tricky (maybe FAOSTAT?), then the other solution is also fine.

etl/data_helpers/geo.py
```diff
@@ -76,6 +76,7 @@ def _load_population() -> Table:
     # WARNING: This function is deprecated. All datasets should be loaded using PathFinder.
     ####################################################################################################################
     log.warning(f"Dataset {DATASET_POPULATION} is silently being loaded.")
+    __import__("ipdb").set_trace()
```
pabloarosado, line 79: Remove?

pabloarosado (approved):
Looks good.

PR #3038, opened 2024-07-29
📊 Update World Bank income groups

paarriagadap:


pabloarosado (approved):
Nice! It all looks good. And I also like your changes in the color of the map chart.
Regarding what to do in an update, it's better to not update all usages: Instead, we update them little by little (as you decided to do in the end). That's usually the way we handle updates of datasets that have many usages. So, this PR is ready to merge!

paarriagadap:
Thank you, @pabloarosado!

PR #3041, opened 2024-07-30
📊  Updating CVD chart to latest WHO MDB data

spoonerf:
HI @veronikasamborska1994, I just updated the WHO Mortality DB so thought I would update this too. Some bigger code changes as the format of the data changed a bit, but the [resulting chart looks all good](http://staging-site-who-mdb-cardio/grapher/share-of-deaths-from-underlying-causes).


veronikasamborska1994 (approved):
that's great! thanks a lot @spoonerf :) code looks good!

PR #3044, opened 2024-07-30
 :bar_chart: Update ghe

Marigold:
[View diff without step copy](https://github.com/owid/etl/pull/3044/files/53981a880c548a33f590b8c6060eea8206dedc52..HEAD)

Code hasn't changed much since the last update (see the link above). I just renamed a few columns and added caching to the snapshot to help with frequent timeouts of their API.

The data have changed a lot in some cases, but the general trends still hold (see discussion below).

@spoonerf should I rephrase, why has the data changed from the last update in 2019 (i.e. mention reliance on GBD 2021 and counting of acute overdose deaths)?

## TODO after merging
- [ ] Remap indicators for [chart 6837](https://admin.owid.io/admin/charts/6837/edit). I messed up its config on staging, and it'll be easier to do it separately (chart diff is rejected)
- [ ] Rename slug from 2019 to 2021 in [chart 6438](https://admin.owid.io/admin/charts/6438/edit)


Marigold:
@spoonerf I'm comparing data from our current GHE version 2022-09-30 to the new version 2024-07-30 and there are some striking differences. This really surprised me since the data is downloaded from the same API. It looks like they must have revised their data at some point.

Here are two examples - one is ischaemic heart disease from Bolivia that shows different trends
image

and the second one is cocaine use disorders from the USA
image

It's not just these two observations, most of the data has changed. Do you have an idea what could be going on, or what could I be doing wrong?

Code for creating those plots
```python
tb_new = Dataset(paths.DATA_DIR / "meadow/who/2024-07-30/ghe").read_table('ghe')
tb_old = Dataset(paths.DATA_DIR / "meadow/who/2022-09-30/ghe").read_table('ghe')

def compare(tb_new, tb_old, country, cause):
    tb_new = tb_new.query(f'country == "{country}" and cause == "{cause}" and sex == "TOTAL"')
    s_new = tb_new.query('age_group == "TOTAL"').set_index('year')

    tb_old = tb_old.query(f'country == "{country}" and cause == "{cause}" and sex == "BTSX"')
    s_old = tb_old.query('age_group == "ALLAges"').set_index('year')

    return pd.DataFrame({
        "old_ghe": s_old.death_count,
        "new_ghe": s_new.val_dths_count_numeric,
    }).round(1).plot(title=f'Deaths from {cause} disease in {country}')

compare(tb_new, tb_old, 'BOL', 'Ischaemic heart disease')
compare(tb_new, tb_old, 'USA', 'Cocaine use disorders')
```

spoonerf:
Hey @Marigold!

I _think_ it's likely that they changed their method somewhat between the two versions. That's not super unusual for this kind of dataset. This dataset has the GBD data as an input, and that also changed quite a lot for some variables in the recent update. 

For example, for drug dependence deaths, [in the new documentation](https://cdn.who.int/media/docs/default-source/gho-documents/global-health-estimates/ghe2021_cod_methods.pdf?sfvrsn=dca346b7_1), they say: _"GHE drug dependence deaths have been updated for all countries drawing on GBD2021 cause fractions. Additionally, estimates for the USA have been updated using US death registration data together with published analyses of drug-specific overdose deaths in the USA."_ which probably explains the second chart. 

Overall, it looks like the general trends and patterns in the charts are pretty similar, so I think they are all good. It would probably be useful and interesting to get @saloni-nd's opinion on this ([ here's the chart-diff](http://staging-site-ghe-update/etl/wizard/chart-diff)). I think of particular interest are the suicide charts, as they look like they have gotten a lot tighter with the other sources (for the US anyway!). Do you have any concerns at all about the differences @saloni-nd?

saloni-nd:
Hey @spoonerf, thank you for tagging me! 

I agree, it looks like they're mostly fine:

- Since there's new data available, their most-recent years match the death registry data closer (for suicides).

- For drug overdoses, the GHE methods says that they now also count 'acute' (short term) drug overdoses, which is probably the reason for the large increase.

> The injury codes for accidental poisoning by alcohol and by opioids are now used to code acute intoxication
deaths from alcohol and acute overdose deaths by opioids. These deaths have been remapped to alcohol
use disorders and drug use disorders respectively (see Annex Table A).

- I'm not sure about Bolivia's data. The WHO GHE says that their underlying data was 'very low' quality, so they didn't use Bolivia's data in their model, meaning their trends are entirely estimated based on other countries' data. Therefore a change in overall methodology might have led to a very large change for Bolivia.

spoonerf (approved):
Looks great just a couple of tiny comments. The new conflict resolver tool looks brilliant too!

There was a few charts with manually overwritten source footers, I updated them so now they are taking the ETL metadata directly.

PR #3049, opened 2024-07-31
:sparkles: Post grapherConfigETL update to Admin API

Marigold:
ETL part of https://github.com/owid/owid-grapher/pull/3793.

`grapherConfig` can't be directly saved to MySQL, because there are side effects to it (it updates all relevant chart configs). Instead of duplicating the side effect logic in ETL, we'll rather call Admin endpoints that handle all that:
```
PUT /variables/:variableId/grapherConfigETL
DELETE /variables/:variableId/grapherConfigETL
```

It works fairly well. I haven't optimized it much, since only a handful of indicators use `grapher_config`.

## Changes
- `grapherConfigETL` is no longer in `*.metadata.json`, is that ok?
- This will change checksum of a lot of indicators, we might have to trigger a full ETL rebuild with new epoch

### DIFF
--- a/apps/backport/datasync/data_metadata.py
+++ b/apps/backport/datasync/data_metadata.py
@@ -253,23 +253,17 @@ def _variable_metadata(
 
     schemaVersion = row.pop("schemaVersion")
     processingLevel = row.pop("processingLevel")
-    grapherConfigETLJson = row.pop("grapherConfigETL")
-    grapherConfigAdminJson = row.pop("grapherConfigAdmin")
     licenseJson = row.pop("license")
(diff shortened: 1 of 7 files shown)


sophiamersmann (approved):
The Grapher model and API calls look good to me! The rest I can't really say :)

We're now using a plain string UUID as the primary key for the `chart_configs` table, so the type is now `char(36)` instead of `binary(16)`. This is probably also relevant for your previous PR that adds the `chart_configs` table.

Removing `grapherConfigETL` from metadata json files is a good call! (I also checked with Daniel and he agrees). I'll make sure that Grapher codebase doesn't rely on it.

PR #3070, opened 2024-08-05
 :sparkles: Add --step-update to draft-pr command

Marigold:
Automate PR creation after updating step in ETL dashboard (see our docs on [Updating data](https://docs.owid.io/projects/etl/guides/data-work/update-data/)).

## Motivation

We update a step with ETL dashboard, but then have to manually create a PR and potentially reference branch or a link to Github diff that excludes copying steps. This PR automates it a bit, while keeping it flexible enough.

## Solution

After you update your step, StepUpdater CLI will show you a command:
```
Create PR automatically with:
etl d draft-pr update-{short_name} --title ":bar_chart: Update {short_name}" --step-update
```

An extra flag `--step-update` will commit the updated steps and post a link to view diff without copied steps (we used to use reference branch for this). See https://github.com/owid/etl/pull/3044 for example.

If this proves to be useful, we can add a modal window to ETL dashboard that would run the command for you.

### DIFF
--- a/apps/step_update/cli.py
+++ b/apps/step_update/cli.py
@@ -360,6 +360,11 @@ class StepUpdater:
                     log.error(f"Stopped because of a failure on step {step}.")
                     break
 
+            # Tell user how to automatically create PR
+            short_name = steps[-1].split("/")[-1].split(".")[0]
+            cmd = f'etl d draft-pr update-{short_name} --title ":bar_chart: Update {short_name}" --step-update'
+            log.info(f"Create PR automatically with:\n  {cmd}")
+
     def _archive_step(self, step: str) -> None:
         # Move a certain step from its active dag to its corresponding archive dag.
 
--- a/apps/utils/draft_pull_request.py
+++ b/apps/utils/draft_pull_request.py
@@ -107,12 +107,18 @@ description = "- " + "\n- ".join(
     "-s",
     help="Scope of the PR (only relevant if --title is given). This text will be preprended to the PR title. \n\n\n**Examples**: 'demography' for data work on this field, 'etl.db' if working on specific modules, 'wizard', etc.",
 )
+@click.option(
+    "--step-update",
+    is_flag=True,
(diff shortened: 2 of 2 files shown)


pabloarosado (commented):
Thanks Mojmir! This is a good idea. But wouldn't you be working on master when running the `etl d draft-pr --step-update` command? If so, then you would be accessing `production` database, and the ETL dashboard will not work properly. You'd need to first create a staging server to be able to access a non-production database to start with. Maybe I'm missing something. How do you envision the workflow?

Marigold:
Oh right, I didn't even know that ETL Dashboard connects to MySQL. Does it only grab info with function `_add_info_from_db` or is it used for anything else? I didn't notice when I was updating GHE because it was connecting to my local MySQL.

There are several alternatives:
- Just start from `master` branch and ignore the additional info from MySQL. It's fine for step update. It'll function properly after you create the branch.
- Instead of connecting to local / staging, make ETL Dashboard connect to read-only production (if `ENV_FILE_PROD` is set) or to `staging-site-master`. This is what we do in Chart diff by default.
- Start by creating staging server with raw `etl d draft-pr`, then use ETL Dashboard to update the step and let it commit files & update PR description

What do you think makes most sense?

lucasrodes:
nice that this is being added!

my two cents to this command:

- you can choose the category of the PR with `-c` flag (it adds emoji automatically)
- i'd appreciate if we replaced `etl d draft-pr` to simply `etl pr`. I personally find it is starting to be quite long.

lucasrodes:
@Marigold does the current error in CI/CD look familiar?

@pabloarosado is there anything left to review?

Marigold:
We haven't resolved the [connection to production](https://github.com/owid/etl/pull/3070#issuecomment-2272910960) MySQL yet. Maybe we can briefly discuss it on Thursday? I'd go with the first proposal and revisit it if this function becomes more widely used.

PR #3074, opened 2024-08-05
📊 ai: Epoch August update

veronikasamborska1994:
hey @Marigold! these are monthly updates so no substantial changes to the code from the previous version though the underlying datasets have changed a bit so some charts look quite different!


Marigold (approved):
Charts look good. I skimmed through code and found no issues.

PR #3075, opened 2024-08-06
 :tada: Fetch datasets from catalog instead of rebuilding them with PREFER_DOWNLOAD

Marigold:
## Motivation

Rebuilding datasets takes a while, especially when someone edits regions or increments `ETL_EPOCH`. Rather than rebuilding those datasets, it could be faster to fetch them from the catalog. We currently have the command `make sync.catalog` which syncs the whole catalog and is useful when you remember to run it, but no one is using it and it's not very flexible.

## Solution

If the environment variable `PREFER_DOWNLOAD` is true, instead of running a step, first check if the dataset's checksum matches the one in the catalog. If it does, download all files instead of running the step. This is especially useful when combined with `--workers N`, which parallelizes the process. This significantly speeds up ETL builds and workflow (no more "ugh, someone updated population again"). It could also speed up staging servers and reduce their CPU usage.

## Issues

It downloads files from S3 at the moment, because we don't have table names in the dataset's `index.json` file. If we had them, we could fetch all files over HTTPS, which could be even faster.


larsyencken:
I haven't had a chance to look at this before leaving on break, but the idea is solid and something I was thinking about for a related project 👌 

It should help with a range of situations and cut resource use down.

lucasrodes (approved):
lgtm!

question: why don't use `owid.catalog` library here?

Marigold:
> question: why don't use owid.catalog library here?

Good question! The main reason is performance - I wanted more control of how the files are downloaded to make sure it's as snappy as possible. (That said, it should be possible to do it through the catalog, too)

PR #3079, opened 2024-08-06
✨ Avoid unexpected behaviour from `Explorer` when writing to tsv files

pabloarosado:
Raise error if not all items in the `yVariableIds` column of the graphers dataframe are lists.


antea04 (approved):
Just tested it - works fab 👍  Thank you!

PR #3083, opened 2024-08-07
 :bar_chart: Migrate monkeypox to ETL

Marigold:
Migrate monkeypox from https://github.com/owid/monkeypox to ETL. There are minimal changes to the data (I've used our own population estimates which changed some metrics on the 5-th decimal place).

The repository is still updated by script in CRON. The next step will be implementing new `ExternalStep` that would be committing CSV to the repo.

The PR also adds [detection](https://github.com/owid/etl/issues/3104) when `date` is used instead of `year` and modifies the dataset by adding `yearIsDay` to all columns before inserting it to MySQL.

Explorer in [staging](http://staging-site-monkeypox/admin/explorers/monkeypox).

## TODO after merging
- [ ] Update CSV link in explorer to https://catalog.ourworldindata.org/explorers/who/latest/monkeypox/monkeypox.csv


etl/steps/data/garden/who/2024-08-07/monkeypox.py
```diff
@@ -0,0 +1,196 @@
...
+
+    # Concatenate with tb
+    return pr.concat([tb, regions])
+
+
+def add_population_and_countries(tb: Table) -> Table:
+    ds_population = paths.load_dataset("population")
+
+    # Use population from 2022. Using population per year results in charts with
+    # cumulative cases per million showing jumps around new year.
```
lucasrodes, line 153: heads up that for COVID, I had the same 'problem'. I decided to use 'daily population', which I estimated by linearly-interpolating the yearly values.

I implemented this function:

https://github.com/owid/etl/blob/5c62946e9451e3f62b3b308aee31644e4af25759/etl/steps/data/garden/covid/latest/shared.py#L94-L113

which then I call like:

https://github.com/owid/etl/blob/5c62946e9451e3f62b3b308aee31644e4af25759/etl/steps/data/garden/covid/latest/sequence.py#L398-L403

etl/steps/data/garden/who/2024-08-07/monkeypox.excluded_countries.json
```diff
@@ -0,0 +1,2 @@
+[
```
lucasrodes, line 1: can remove?

etl/helpers.py
```diff
@@ -205,44 +237,33 @@ def create_dataset(
...
-    assert match, f"Could not parse path {str(dest_dir)}"
+        if ds.metadata.channel == "grapher":
+            # if a table contains `date` instead of `year`, adapt it for grapher
+            if "date" in table.all_columns and "year" not in table.all_columns:
+                if "date" in table.index.names:
+                    table = table.reset_index("date")
+                    table = gh.adapt_table_with_dates_to_grapher(table)
+                    table = table.set_index("year", append=True)
+                else:
+                    table = gh.adapt_table_with_dates_to_grapher(table)
```
lucasrodes, line 258: Very nice to have out-of-the-box support! Thanks

In COVID, I've been manually adapting the columns to have `year` instead of `date`, where `year` would be 'days since zero_day`:

https://github.com/owid/etl/blob/5c62946e9451e3f62b3b308aee31644e4af25759/etl/steps/data/grapher/covid/latest/cases_deaths.py#L24

using function from

https://github.com/owid/etl/blob/5c62946e9451e3f62b3b308aee31644e4af25759/etl/steps/data/grapher/covid/latest/shared.py#L33-L47

I suppose this is no longer needed? Do you know if your changes could break this?

lucasrodes (approved):
lgtm ー  minor comments

lucasrodes:
btw, don't fully understand why I had to commit anything in https://github.com/owid/etl/pull/3083/commits/c8f1b5e0f440aeb464aedc6c83932482f8419d81. I got a CONFLICT for some reason.

etl/helpers.py
Marigold: I also didn't know we already have a function `adapt_table_with_dates_to_grapher` for this. It might make sense to just get rid of all `yearIsDay` in our codebase and let it all happen automatically.

> Do you know if your changes could break this?

I hope not. If it causes problems, let me know and I'll look into it.

PR #3088, opened 2024-08-08
✨ map bracketer - work with paths instead of ids 

antea04:


pabloarosado (approved):
Thanks a lot for this! It looks good, we can debug and improve it as we go.

PR #3106, opened 2024-08-12
🐛  log.info() only takes a single string argument

spoonerf:
Fix for snapshot not working


Marigold (approved):
Nice catch, thank you!

PR #3132, opened 2024-08-16
✨ Improve minerals explorer

pabloarosado:


pabloarosado:
Discarded: We decided to combine BGS and USGS in a different way (see `improve-minerals-explorer-3`).

Issue #3125, opened 2024-08-15
wizard: zeroDay new behaviour causing unexpected behaviour in Explore Mode


lucasrodes:
### One-liner

`zeroDay` set automatically is causing unexpected behaviour in _Explore mode_ in **Indicator Upgrader**.

### Context & details
Recently, there were some ergonomic improvements for data pipelines that use `date` instead of year (I think [here](https://github.com/owid/etl/pull/3083), probably?).

I was comparing indicator [grapher/covid/latest/vaccinations/vaccinations#total_vaccinations](http://staging-site-covid-vax/admin/variables/738594) (old) with [grapher/covid/latest/vaccinations_global/vaccinations_global#total_vaccinations](http://staging-site-covid-vax/admin/variables/984968) (new), which should be rather similar. However, I was getting unexpected error values.

<img width="732" alt="image" src="https://github.com/user-attachments/assets/5f9e8014-2551-4164-83c9-bd9beace3ea9">

This is causing the underlying hardcoded 'year' column to miss-match between indicators.

**Explore mode** is very helpful to make sure that the timeseries of two indicators match approximately. We should make sure that it is also reliable when using `date` columns. We probably need to convert '`year`' to an actual date using the `zeroDay` value of both indicator. Only then we can align the time-series.

Marigold:
Fixed by https://github.com/owid/etl/pull/3137. Please reopen if this appears again.

PR #3137, opened 2024-08-16
:bug: Fix explore mode for dates

Marigold:
If `yearIsDay`, convert integer year to date in Explore Mode. It could help with https://github.com/owid/etl/issues/3125. I tested it and it looked fine, although I haven't been able to replicate the original issue.

This PR also updates some AI datasets to use `date` instead of manually setting `yearIsDay`.


etl/grapher_helpers.py
```diff
@@ -782,7 +782,7 @@ def adapt_table_with_dates_to_grapher(
         tb = tb.drop(columns=["year"], errors="raise")
 
     # Ensure date column is in datetime format.
-    tb[date_column] = pd.to_datetime(tb[date_column])
+    tb[date_column] = pd.to_datetime(tb[date_column].astype(object))
```
lucasrodes, line 785: maybe can use `astype("string")`?

lucasrodes (approved):
lgtm! thanks for doing this.

FYI, My recent additions of COVID datasets include some usages of `yearIsDay`.

etl/grapher_helpers.py
Marigold: I don't recall why, but it wasn't working for some reason.

Issue #2905, opened 2024-06-27
Upgrade pyright to a newer version


larsyencken:
## Problem

VSCode uses a newer version of Pyright than what we do in linting. We are frozen on a earlier version because of a huge number of updates we would need to do to fix linting errors.

## Impact

We are seeing warnings in the editor that we do not see in CI, which is annoying, and gets people used to ignoring warnings in the editor rather than fixing them.

## See earlier attempt

- https://github.com/owid/etl/issues/1113

larsyencken:
We know this is large and annoying to fix. Eventually we will want to, but we can also live with the status quo for a while longer.

Marigold:
I tried upgrading pyright to 1.1.373 which is the version used by Pylance in VSCode. There are ~2700 errors that need to be deat with. In summary, newer pyright version can detect more typing issues, but most of them are not very useful, e.g.
- `Argument of type "FrameOrSeries" cannot be assigned to parameter "df" of type "DataFrame"`
- `Argument of type "DataFrame | None" cannot be assigned to parameter "df" of type "DataFrame"`
- `Import "shared" could not be resolved`
- `Argument of type "str | None" cannot be assigned to parameter "url"`

They are often pandas-related typing issues that are mostly cosmetic and don't improve quality of code.

## Upgrade plan

Parse errors from `pyright --outputjson` and use `# type: ignore` to silence the following types of errors:
- All recipes in `etl/steps/`
- All tests
- Errors in core etl code that cannot be easily fixed

Marigold:
Done. Note that we've decided to relax type checks for recipes, and CI now passes even with failing type check. Users still see issues underlined red in VSCode which should be enough.

PR #3148, opened 2024-08-19
:honeybee: Upgrade pyright to match VSCode errors

Marigold:
Implementation of https://github.com/owid/etl/issues/2905.

I tried to manually fix as many errors as possible in our core codebase. For the rest, I ran a script to ignore lines for snapshots / steps.

We've decided to make type checks for recipes in `etl/steps` **optional**. CI won't fail any more if your code is not type hint proof. Issues will be still shown in VSCode as **problems**, so it's up to you what do you decide to do with them.

We made it optional because users were getting tons of useless type errors like:
- Missing implementation of `Table.method` causing the type to be DataFrame, while in runtime it is Table as expected
- Pandas functions returning `FrameOrSeries` type
- Not using `.loc[]` returning `DataFrame` type instead of `Table`
- etc.

Those kinds of errors were typically solved by adding `# type: ignore` at the end of the line and were just extra work with no value.

This is going to rebuild a large part of ETL.

## How to review

Don't review all files! These are just files with added `# type: ignore`. Instead, checkout this PR, run `poetry install` and make sure type errors show up in VSCode as "Problems". Also try running `make test` to make sure there aren't any leftover type errors.


lucasrodes (approved):
lgtm!

PR #3165, opened 2024-08-21
 :sparkles: engineering: refactor Step paths

Marigold:
Refactor Step paths like channel / namespace / version / name. Move the logic from bespoke functions to steps themselves and use `Step` properties to access them.

Add `CatalogPath` to centralize its parsing.

An idea - we could have `DatasetPath`, `VariablePath` and `StepPath` to distinguish between `data://...`, `data://...#variable`. These would behave similarly to `pathlib.Path` and support operations such as `/` or `+` / `@` (as a replacement of `#`)

## TODO
- [ ] Refactor function `get_all_changed_catalog_paths`
- [ ] Should we use dataset / name?

### DIFF
--- a/apps/metadata_migrate/cli.py
+++ b/apps/metadata_migrate/cli.py
@@ -15,7 +15,7 @@ from sqlalchemy.orm import Session
 from etl import config
 from etl import grapher_model as gm
 from etl.command import main as etl_main
-from etl.db import get_engine, read_sql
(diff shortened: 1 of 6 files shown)


Marigold:
Not yet owidbot, I'd really like to finish this one at one point.

Marigold:
Closing, it didn't get anywhere after a year. It turned out that using strings is easier than checking the reference for `CatalogPath` each time. Maybe one day...

PR #3175, opened 2024-08-22
📊 wildfires: visualising average and standard deviations

veronikasamborska1994:


spoonerf (commented):
Nice! I think the average makes sense. However, for the max and min I think it might be better to do it slightly differently. At the moment they don't really represent a 'real' year, just all the cumulative max and min values for each country across the dataset, right? 

I think it might be more meaningful to show the actual years that had the min and max values in the 52nd week, if that makes sense. I guess that will be different for each country and maybe a bit fiddlier to calculate!

veronikasamborska1994:
> chart-diff

ah I think I see what you mean - for each country select the year that had a total maximum/minimum at the end of the year and use the cumulative values of these?

veronikasamborska1994:
> Nice! I think the average makes sense. However, for the max and min I think it might be better to do it slightly differently. At the moment they don't really represent a 'real' year, just all the cumulative max and min values for each country across the dataset, right?
> 
> I think it might be more meaningful to show the actual years that had the min and max values in the 52nd week, if that makes sense. I guess that will be different for each country and maybe a bit fiddlier to calculate!

ah I think I see what you mean - for each country select the year that had a total maximum/minimum at the end of the year and use the cumulative values of these?

spoonerf:
> ah I think I see what you mean - for each country select the year that had a total maximum/minimum at the end of the year and use the cumulative values of these?

Yep exactly! I'm assuming that's what they have done on the [source charts](https://gwis.jrc.ec.europa.eu/apps/gwis.statistics/seasonaltrend) too, but hard to tell. Could be nice to set the max & min to be projections too, so they are dashed and slightly less prominent, but up to you!

veronikasamborska1994:
> > ah I think I see what you mean - for each country select the year that had a total maximum/minimum at the end of the year and use the cumulative values of these?
> 
> Yep exactly! I'm assuming that's what they have done on the [source charts](https://gwis.jrc.ec.europa.eu/apps/gwis.statistics/seasonaltrend) too, but hard to tell. Could be nice to set the max & min to be projections too, so they are dashed and slightly less prominent, but up to you!

yes, this is definitely a good idea! It was a bit fiddly but I've now done it! :)

spoonerf:
> > > ah I think I see what you mean - for each country select the year that had a total maximum/minimum at the end of the year and use the cumulative values of these?
> > 
> > 
> > Yep exactly! I'm assuming that's what they have done on the [source charts](https://gwis.jrc.ec.europa.eu/apps/gwis.statistics/seasonaltrend) too, but hard to tell. Could be nice to set the max & min to be projections too, so they are dashed and slightly less prominent, but up to you!
> 
> yes, this is definitely a good idea! It was a bit fiddly but I've now done it! :)

Amazing, it looks ace!

PR #3178, opened 2024-08-22
📊 tourism: UNWTO supplementary datasets

veronikasamborska1994:


dag/tourism.yml
```diff
@@ -28,3 +28,15 @@ steps:
     - data://garden/regions/2023-01-01/regions
   data://grapher/oecd/2024-08-21/co2_air_transport:
     - data://garden/oecd/2024-08-21/co2_air_transport
+  data://meadow/tourism/2024-08-17/unwto_environment:
```
antea04, line 31: I would separate these blocks by data set and add a comment, so you can find them quickly when e.g. updating these data sets

snapshots/tourism/2024-08-17/unwto_environment.xlsx.dvc
```diff
@@ -0,0 +1,30 @@
...
+      https://www.unwto.org/tourism-statistics/economic-contribution-SDG [17-08-2024].
+
+    # Files
+    url_main: https://www.unwto.org/tourism-statistics/economic-contribution-SDG
+    url_download: http://pre-webunwto.s3.eu-west-1.amazonaws.com/s3fs-public/2024-04/Indicator-12_b_1-2022-UN_Toiurism_april2024_update.xlsx
+    date_accessed: 2024-08-17
+
+    # License
+    license:
+      name: © UNWTO, 92844/14/23
```
antea04, line 26: You could link their copyright page (https://www.unwto.org/copyright) but I don't know if that is strictly necessary

etl/steps/data/garden/tourism/2024-08-17/unwto_gdp.py
```diff
@@ -0,0 +1,36 @@
...
+
+def run(dest_dir: str) -> None:
+    #
+    # Load inputs.
+    #
+    # Load meadow dataset.
+    ds_meadow = paths.load_dataset("unwto_gdp")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow["unwto_gdp"].reset_index()
```
antea04, line 18: if you want, you can replace this with tb = ds_meadow.read_table("unwto_gdp"), which automatically resets the index

etl/steps/data/garden/tourism/2024-08-17/unwto_gdp.meta.yml
```diff
@@ -0,0 +1,25 @@
...
+
+tables:
+  unwto_gdp:
+    variables:
+      tourism_share_gdp:
+        title: GDP from tourism as a share of total GDP
+        unit: '%'
+        short_unit: '%'
+        description_from_producer: Tourism Direct GDP (TDGDP) is defined as the sum of the part of gross value added (at basic prices) generated by all industries in response to internal tourism consumption plus the amount of net taxes on products and imports included within the value of this expenditure at purchasers’ prices.
+        description_short: GDP from tourism includes tourism industries and other industries that directly serve visitors.
```
antea04, line 25: you could add the gdp dod here (#dod:gdp), if you want/ think it's helpful

etl/steps/data/garden/tourism/2024-08-17/unwto_environment.meta.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      tsa_tables:
+        title: Number of tables in the Tourism Satellite Account
+        unit: tables
+        description_short: Number of tables in the TSA that track key economic aspects of tourism, including expenditure, production, and employment.
+
+      total_tables:
+       title: Total number of tables
+       unit: tables
+       description_short: Monitoring is expressed as the number of implemented tables across the Tourism Satellite Account and the System of Environmental and Economic Account. These tables track key economic and environmental aspects of tourism, including expenditure, production, employment, and resource flows.
```
antea04, line 32: Hm, so looking at this description and the chart of this indicator, I (as a layperson) do not really understand what this is supposed to mean. 

I think this is mostly due to this metric being a bit abstract and not 1-to-1 corresponding with easily understandable things, but maybe you could elaborate a bit more on these tables (in the description_key or a dod). E.g. what is the TSA/ SEEA (an organisation? a method?), who is implementing/ tracking these, what are they used for? What does it really mean to have 1 table vs 8 tables here? 

Of course, if you don't have this information or there is no way to make it easily accessible to a wider audience, disregard this.

antea04 (approved):
Hi Veronika!

The PR looks good from my side! I left some very minor comments, feel free to decide what you implement.

The only thing is that I find the indicator 12.b.1 a bit hard to understand, especially what the values mean (what is really better about having 8 tables vs 1 table? I assume you are somehow tracking more metrics, but is that really the case?)

If you find a way to make that a bit clearer for a general audience, that would be great, but maybe this is just a weird metric/ already as clear as possible, then just disregard it.

Let me know if anything else is unclear, otherwise feel free to merge 👍

snapshots/tourism/2024-08-17/unwto_environment.xlsx.dvc
veronikasamborska1994: ah thanks! we got a permission from them to use the data last year because under their actual terms we wouldn't be allowed to which is why I left it out but I'll add it anyway! thanks!

etl/steps/data/garden/tourism/2024-08-17/unwto_gdp.py
veronikasamborska1994: nice! pretty sure these just get prefilled by wizard but good to know we have this function!

etl/steps/data/garden/tourism/2024-08-17/unwto_environment.meta.yml
veronikasamborska1994: thank you! I've added a more comprehensive description_key and slightly edited the description_short to make it a bit clearer!

PR #3194, opened 2024-08-26
:tada: Multi-dimensional indicators

Marigold:
Define multi-dimension indicators in export step in ETL. These steps load the accompanying YAML metadata file and dynamically add **views**. Export step creates a multi-dim config and upserts it to a table `multi_dim_data_pages`.

Export steps must be called with the `--export` flag, e.g.
```
etlr export://multidim/energy/latest/energy --export
```

### [Causes of death](http://staging-site-multi-dim-db/admin/grapher/mdd-causes-of-death)

Nice example of a dataset with dimensions that can be easily mapped to the YAML file. I had to distinguish between "All causes" that are the sum and side-by-side comparison of all causes.

### [Energy](http://staging-site-multi-dim-db/admin/grapher/mdd-energy?source=all&metric=total)

This example involves custom low-level code for creating views, since the underlying dataset doesn't have dimensions and use specific naming. It'd be cleaner if the dataset had dimensions, but I don't think it's worth creating a new transformed grapher dataset just for this case.


etl/multidim.py
```diff
@@ -0,0 +1,149 @@
...
+    multidim config file. If the dimension is missing from groupby, it will be set to "all".
+
+    :params config: multidim config file
+    :params combinations: dictionary with dimension names as keys and values as dimension values, use * for all values
+        e.g. {"metric": "*", "age": "*", "cause": "All causes"}
+    :params table: catalog path of the grapher table
+    :params engine: SQLAlchemy engine
+    """
+
+    # Get all allowed valuse from choices
```
danyx23, line 85:
```suggestion
    # Get all allowed values from choices
```

danyx23 (approved):
This looks very cool! It looks good from my side - the only thing that we might want to tweak is the schema (thanks for adding that btw!). I'll ask Marcel if he can have a look at that and tell us if we should extend it.

PR #3197, opened 2024-08-26
🐛 Fix `filter_to_subgraph` when passing the `only` argument

pabloarosado:
I noticed that `to_dependency_order(dag=dag, includes=STEPS, only=True)` was returning a list of steps that was larger than STEPS. But if I understand correctly, `only` should avoid that, right?
This PR may fix it (but I'm not sure if that behaviour was as expected, and I'm misinterpreting the meaning of `only`).

As an example:
```
from etl.steps import load_dag, to_dependency_order
STEPS = ['data://grapher/war/2023-09-21/ucdp', 'data://garden/war/2023-09-21/ucdp']
steps = to_dependency_order(
    dag=load_dag(),
    includes=STEPS,
    excludes=[],
    downstream=False,
    only=True,
)
```
as far as I understand, should return `['data://garden/war/2023-09-21/ucdp', 'data://grapher/war/2023-09-21/ucdp']`, but instead it was returning `['data://garden/war/2023-09-21/ucdp', 'data://grapher/war/2023-09-21/ucdp',  'data://garden/war/2023-09-21/ucdp_prio']`.

(this issue [affected](https://owid.slack.com/archives/C0193RW5E2J/p1724671526845349) @lucasrodes while using StepUpdater)

### DIFF
--- a/apps/step_update/cli.py
+++ b/apps/step_update/cli.py
@@ -342,6 +342,7 @@ class StepUpdater:
                 excludes=[],
                 downstream=False,
                 only=True,
+                exact_match=True,
             )
 
             message = "The following steps will be updated:"
--- a/etl/steps/__init__.py
+++ b/etl/steps/__init__.py
@@ -76,13 +76,18 @@ def to_dependency_order(
     excludes: List[str],
     downstream: bool = False,
     only: bool = False,
+    exact_match: bool = False,
 ) -> List[str]:
     """
     Organize the steps in dependency order with a topological sort. In other words,
     the resulting list of steps is a valid ordering of steps such that no step is run
     before the steps it depends on. Note: this ordering is not necessarily unique.
     """
-    subgraph = filter_to_subgraph(dag, includes, downstream=downstream, only=only) if includes else dag
+    subgraph = (
+        filter_to_subgraph(dag, includes, downstream=downstream, only=only, exact_match=exact_match)
+        if includes
+        else dag
+    )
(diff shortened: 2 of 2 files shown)


Marigold:
It actually works correctly. It runs all steps that **re.findall-match** strings from `includes`. `data://garden/war/2023-09-21/ucdp_prio` is matched because it contains `data://garden/war/2023-09-21/ucdp` as a substring.

To only get the matching steps, you have to add `$` as the suffix (to match the end of the string). Running
```
etlr data://grapher/war/2023-09-21/ucdp data://garden/war/2023-09-21/ucdp$ --dry-run --only
```
works as expected.

pabloarosado:
Thanks @Marigold! I have added an `exact_match` argument, to use when we want to "include" only an exact list of steps. What do you think?

Marigold (approved):
Isn't is easier to add `$` than to use `--exact_match`? I'd personally never use it since I always type just a part of the step and almost always get what I want, but if you find it useful, feel free to merge it.

pabloarosado:
> Isn't is easier to add `$` than to use `--exact_match`? I'd personally never use it since I always type just a part of the step and almost always get what I want, but if you find it useful, feel free to merge it.

Using `$` is indeed easier, if you are always aware that you are passing a regular expression pattern instead of a step. I think `exact_match` is much more understandable, and should be included in `etlr` as well. You can keep using `$` if you find that more convenient.

PR #3204, opened 2024-08-27
Add exact-match to etlr options

pabloarosado:
Following the discussion in https://github.com/owid/etl/pull/3197 I think we should have an `exact-match` option in `etlr`, to run steps that happen to be substrings of other steps (e.g. `data://grapher/war/2023-09-21/ucdp`,  substring of `data://garden/war/2023-09-21/ucdp_prio`).

Currently `etlr data://garden/war/2023-09-21/ucdp --only --dry-run --force`:
```
--- Would run 2 steps:
1. data://garden/war/2023-09-21/ucdp
2. data://garden/war/2023-09-21/ucdp_prio
```
And if you want to run only the desired step you need to do `etlr data://garden/war/2023-09-21/ucdp$ --only --dry-run --force`, which is a bit of a hack. To avoid such unexpected executions, users not only need to remember this hack, but also know that there are other steps that will be matched to the given argument.

Instead, this PR lets you do `etlr data://garden/war/2023-09-21/ucdp --only --exact-match --dry-run --force`:
```
--- Would run 1 steps:
1. data://garden/war/2023-09-21/ucdp
```

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -93,6 +93,12 @@ log = structlog.get_logger()
     help="Only run the selected step (no upstream or downstream dependencies). Overrides `--downstream` option.",
 )
 @click.option(
+    "--exact-match",
+    "-x",
+    is_flag=True,
+    help="Steps should exactly match the arguments (if so, pass the steps with their full name, e.g. 'data://garden/.../step_name').",
+)
+@click.option(
     "--exclude",
     "-e",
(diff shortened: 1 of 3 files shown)


Marigold (approved):
The PR looks good, but I wouldn't say that using $ is a hack :).

PR #3205, opened 2024-08-27
 :hammer: Remove --step-update and make committing explicit

Marigold:
Revert `etl pr ... --step-update` and update ETL docs. User has to manually commit generated files now and insert the diff link in the description.


apps/utils/draft_pull_request.py
```diff
@@ -224,19 +213,6 @@ def cli(
         log.error(f"Failed to create draft pull request:\n{response.json()}")
         return
 
-    # modify the PR body to include the diff link
-    if step_update:
-        diff_link = f"{js['html_url']}/files/{repo.head.commit.hexsha}..HEAD"
```
pabloarosado, line 229: I think this was quite useful, I'm sad to see this go!

apps/utils/draft_pull_request.py
```diff
@@ -199,11 +193,6 @@ def cli(
     log.info("Creating an empty commit.")
     repo.git.commit("--allow-empty", "-m", title or f"Start a new staging server for branch '{new_branch}'")
 
-    if step_update:
-        log.info("Committing all files.")
```
pabloarosado, line 203: What if we simply print the list of files to be commited, and ask for confirmation, before adding and committing?

pabloarosado (commented):
Thanks for the improvements @Marigold, I think that some of the stuff you did here were very useful, but the main issue is automatically adding and committing everything. Maybe we could simply ask for confirmation (instead of reverting everything). What do you think?

apps/utils/draft_pull_request.py
pabloarosado: If there are snapshot files in the list, the printed message should mention that snapshot scripts would need to be executed prior to committing (or otherwise expect CI to fail, as [suggested](https://github.com/owid/etl/pull/3205#discussion_r1732534404) by @lucasrodes).

pabloarosado (approved):
Feel free to try my suggestions, or to simply go ahead and merge this PR as it is (with Lucas' suggestion about snapshots), if you don't feel like spending more time on these details. Either option is ok but not ideal in any case. Sorry for the back and forth.

docs/guides/data-work/update-data.md
```diff
@@ -58,19 +58,23 @@ This guide assumes you have already a [working installation of `etl`](../../../g
...
     - You can close the Wizard (kill it with ++ctrl+c++).
 
-- **Run `etl pr .. --step-update` to commit changes and create a PR**
+- **Run `etl pr ..` to create a PR**
 
     Run the command you copied from the ETL Dashboard:
 
     ```bash
-    etl pr update-{short_name} --title ":bar_chart: Update {short_name}" --step-update
+    etl pr update-{short_name} --title ":bar_chart: Update {short_name}"
```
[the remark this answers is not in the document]
Marigold: Good point. It's displayed in the terminal which is a bit hidden. I removed the mention of ETL dashboard and only left the command.

docs/guides/data-work/update-data.md
```diff
@@ -58,19 +58,23 @@ This guide assumes you have already a [working installation of `etl`](../../../g
...
     ```
 
-    This will create a new git branch in your local repository, which will be pushed.
+    This will create a new git branch in your local repository with empty commit, which will be pushed.
     It will also create a draft pull request in github, and a staging server.
+
+    Then, commit generated files and push them to the new branch. This commit will contain only copied steps, subsequent changes
+    will be added as separate commits.
+
```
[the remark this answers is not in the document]
Marigold: Updated the comment with a warning. You should get a decent error now telling you that you should probably run the snapshot.

apps/utils/draft_pull_request.py
Marigold: I thought that too, but now I think it's more intuitive to just copy the link from the docs or select correct commits if you're a reviewer. I'm sure we'll find a better way to display this in the future.

Marigold:
> Maybe we could simply ask for confirmation (instead of reverting everything). What do you think?

I've actually changed my mind after our discussion. It's nicer to have ETL dashboard and git operations separated. It's just two extra commands that the user has to run, and it's way more transparent. If we decide to integrate them together, we'd better do it all in ETL dashboard without having to run CLI.

PR #3214, opened 2024-08-29
🔧  Fixing the GBD region percentage aggregate error

spoonerf:
....but skipping it for risk factor datasets. Also attempting to make the aggregating function a little bit more modular and less horrible.


spoonerf:
Hey @veronikasamborska1994 - I _think_ this is an okay solution to the regional aggregates issue in the GBD code. Curious to hear what you think, but no rush!

veronikasamborska1994:
> Admin

hey @spoonerf! It looks like this code would only work for gbd_cause dataset. Maybe a cleaner way would be just to add an argument that would be false by default and only run the add_regions_to_percent if you set it to true in the gbd_cause script? I think adding the weighted averages functionality would take quite a bit of thought as we'd probably just want to build it into the geo module so probably best if we go with your solution for this one.

veronikasamborska1994 (approved):
this works as a solution - if you wanted to make it a bit cleaner you could do something along these lines in your `add_regional_aggregates` but feel free to merge either way so we can fix this bug: 
```
    tb_out = pr.concat([tb_rate, tb_rate_regions], ignore_index=True)  # type: ignore

    if run_percent:
        tb_percent_regions = add_regions_to_percent(tb_number, regions, index_cols)
        # Check there aren't any values above 100
        tb_percent = pr.concat([tb_percent, tb_percent_regions], ignore_index=True)
        assert tb_percent["value"].max() <= 100 or tb_percent.shape[0] == 0
        assert tb_percent["value"].min() >= 0 or tb_percent.shape[0] == 0

        # Combine all the metrics back together
        tb_out = pr.concat([tb_out, tb_percent], ignore_index=True)
```

p.s. I am also happy to merge this for you later tonight as not to block the etl!

spoonerf:
@veronikasamborska1994 - That is a very good idea! I've added that to the code. If it passes CI okay I'll merge shortly 🤞🏼

PR #3223, opened 2024-08-30
📊 cancer: GCO Cancers Attributable to Alcohol

veronikasamborska1994:
hey @Marigold ! do you know what we usually do with cases where the data includes (below, above a certain threshold)? here is an example [chart](http://staging-site-cancer-infections/admin/charts/8021/edit). I think that's why the chart-diff is failing


Marigold:
> I think that's why the chart-diff is failing

Sorry, missed that post. You were right, there was an issue with chart-diff. Those charts should show up now.

Marigold (approved):
The code looks good, though I don't understand what "All cancers" means in the bar chart. Shouldn't it be rather "Other cancers"?

image

veronikasamborska1994:
> chart-diff

it's just an average of all cancers I think! including those for which alcohol wouldn't be a significant risk factor but I'll keep it in mind and maybe we shouldn't include it in that particular chart!

PR #3225, opened 2024-09-02
✨ Raise error in etl pr if branch already exists locally or remotely

pabloarosado:
Currently, if you use `etl pr` with a branch name that does not exist locally, but does exist remotely, an error is raised, but after the local branch has been created.
With this PR, an error is raised and no local branch is created.

I have also changed `origin.fetch()` to `origin.fetch(prune=True)`, because often the script was raising an error, claiming that the proposed new branch already existed in remote, when it was actually removed there. If you think this can have unwanted side effects, we can do `origin.fetch(prune=prune)` and add `---prune` as an optional boolean flag (but I think there is no danger setting it to True).

### DIFF
--- a/apps/utils/draft_pull_request.py
+++ b/apps/utils/draft_pull_request.py
@@ -149,7 +149,10 @@ def cli(
 
     # Update the list of remote branches in the local repository.
     origin = repo.remote(name="origin")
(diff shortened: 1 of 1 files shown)


Marigold (approved):
Nice!

Issue #3182, opened 2024-08-23
Use `groupby(..., observed=True)` by default


Marigold:
When working with categorical variables, not using `observed=True` creates groups with no records. This is unintuitive and causing tons of headaches. We set `observed=True` manually pretty much everywhere, so it'd be nice if this was the default.

Pandas also throws a warning saying that `observed=True` will be default in the future.

PR #3226, opened 2024-09-02
 :hammer: Use `groupby(..., observed=True)` by default

Marigold:
Fix for https://github.com/owid/etl/issues/3182

In this PR, I incremented ETL_EPOCH and ran datadiff on all our datasets. I fixed those that weren't using `observed` argument and were showing differences (these are typically harmless missing values that don't make it to grapher anyway).

## TODO before merging
- [x] Undo `ETL_EPOCH` change

### DIFF
--- a/etl/config.py
+++ b/etl/config.py
@@ -179,7 +179,7 @@ SUBSET = env.get("SUBSET", None)
 MAX_VIRTUAL_MEMORY_LINUX = 32 * 2**30  # 32 GB
 
 # increment this to force a full rebuild of all datasets
-ETL_EPOCH = 6
+ETL_EPOCH = 5
 
 # any garden or grapher dataset after this date will have strict mode enabled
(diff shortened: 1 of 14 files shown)


pabloarosado (approved):
Thanks! It's a bit annoying to need to specify `observable=True` on every groupby, and it's almost always the desired behaviour.
Maybe for a separate PR, we could consider if `as_index=False` should also be the default (following the same logic by which `read_table` by default returns tables wihout index). People usually are forced to `reset_index()` after grouping operations.

Marigold:
> Maybe for a separate PR, we could consider if as_index=False should also be the default (following the same logic by which read_table by default returns tables wihout index). People usually are forced to reset_index() after grouping operations.

I wouldn't rush into it. `observed=True` will be default in future version of pandas, but `as_index=False` won't (and I'm personally used to having groups in the index, others might be too).

PR #3236, opened 2024-09-03
 :hammer: engineering: refactor upserts to MySQL

Marigold:
Refactor upserts to MySQL from `grapher://` step. This allows us to compare checksums of data & metadata by indicator and skip upserts to `variables` table in MySQL if metadata doesn't change (previously we only skipped uploading to R2). This should speed up cases where you only work on a single indicator.

It's not going to be any faster than our current solution if checksums differ. I'll try tackling that in the next PR where I switch from threads to asyncio.

## Changes
- Data & metadata checksums are now computed from dataframe and `VariableMeta` dataclass instead of the final JSON that we upload to R2.
- Use `hash_any` to calculate checksums instead of `md5` that we usually use.
- Minimize `set_index` and `reset_index` operations.
- Start using `catalogPath` more.
- Deprecate old code.


lib/catalog/owid/catalog/utils.py
```diff
@@ -200,3 +202,28 @@ def dynamic_yaml_to_dict(yd: Any) -> dict:
...
+
+def hash_any(x: Any) -> int:
+    """Return unique deterministic hash for an arbitrary object. This is useful if you can't make your dataclasses
+    frozen but still want to use operations such as `set` or `unique`.
+
+    The problem with using `hash` function directly is that Python uses random hash seed (PYTHONHASHSEED) for
+    security reasons. So the hash of the same object can be different in different runs for strings. This is
+    why we use md5 hash for strings.
+    """
+    if is_dataclass(x):
```
pabloarosado, line 215: Wow, this function is quite obscure, it would be nice to explain a bit more what's going on.

pabloarosado (approved):
Looks good (although I can't quite follow all the logic, I'd need more context to understand the process). Thanks for doing this!
Can I simply fetch this branch and execute a few grapher steps to see if it works as expected?

lib/catalog/owid/catalog/utils.py
Marigold: Added some comments.

Marigold:
> Can I simply fetch this branch and execute a few grapher steps to see if it works as expected?

That'd be very helpful 🙏

PR #3255, opened 2024-09-09
:hammer: engineering: switch from threads to asyncio in grapher upserts

Marigold:
Switch from threads to asyncio when upserting to MySQL. This simplifies the code a bit and might make it slightly faster, but at the cost of introducing asyncio to our codebase. Whether this is worth it depends on how faster and convenient grapher step becomes.

@pabloarosado if you run into a situation where grapher upserts are annoyingly long, can you ping me and I can revisit the performance on that concrete example? (By the way, we've recently merged an improvement that upserts only indicators that changed, not the entire dataset. This could already speed things up)


Marigold:
With a heavy heart, I'm closing this one. Recent optimizations, like upserting variables only when checksum changes, have made it less painful and asyncio approach is not much faster than threads. It's more efficient and "elegant", but that's not worth mixing async and sync code.

One day, we might open it again...

PR #3256, opened 2024-09-09
📊  Add flu testing data to ETL

spoonerf:
Currently this dataset goes directly to the grapher.


Marigold (approved):
Looks good! I added a couple of commits with fixes and also removed `yearIsDay` hack. It's now possible to use `date` instead of `year` and let ETL handle the conversion automatically.

PR #3258, opened 2024-09-09
🔨 engineering: add field definition for new generated md5 column on chart_configs

danyx23:
This catches up the ETL side data model with a recently added column on the chart_configs table


Marigold (approved):
Sweet, who doesn't love md5s?

PR #3262, opened 2024-09-10
 :bug: gbd-risk: Add missing percentages

Marigold:
Percentages from `gbd_risk` dataset such as [this one](https://admin.owid.io/admin/variables/937511) got removed by [this commit](https://github.com/owid/etl/commit/93e92ce428abd88a98bd08b819064cbf335fec88). Because of that, ETL tries to rebuild `gbd_risk` every time it runs and then throws a warning.

This PR adds `run_percent=True` argument to regional aggregates to put them back. It also raises an error on staging servers when there are variables that cannot be deleted automatically.

### DIFF
--- a/etl/grapher_import.py
+++ b/etl/grapher_import.py
@@ -458,13 +458,17 @@ def cleanup_ghost_variables(engine: Engine, dataset_id: int, upserted_variable_i
         if rows:
             rows = pd.DataFrame(rows, columns=["chartId", "variableId"])
 
-            # show a warning
-            log.warning(
-                "Variables used in charts will not be deleted automatically",
-                rows=rows,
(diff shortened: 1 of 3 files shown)


spoonerf (approved):
Ah thanks for finding this! So we don't want regional aggregates for percentages for gbd_risk as there's some weird stuff going on their (some 'risks' reduce your chance of dying), so run_percent has to stay False. 

But I think there was a broader issue in shared.py, meaning that all aggregate percentages were getting dropped unless run_percent = True, which isn't what we wanted. 

Have changed it now, I think that also fixes the original issue?

Marigold:
> Have changed it now, I think that also fixes the original issue?

Yep, much better!

PR #3271, opened 2024-09-11
✨  flu: use only positive and negative tests in denominator

spoonerf:


lucasrodes:
@spoonerf
what's the status of this PR?

PR #3282, opened 2024-09-13
🔧  mpox: Using xmart API endpoint

spoonerf:


Marigold (approved):
The endpoint works. Unfortunately the latest data is from 2024-09-04, so there's no new data compared to the existing snapshot.

spoonerf:
@Marigold Shall we merge for now and then continue looking for where the newer data can be sourced?

Marigold:
> Shall we merge for now and then continue looking for where the newer data can be sourced?

Up to you. I fear that using xmart endpoint might be hiding the real problem. I'd wait till Monday to see if running `python snapshots/who/latest/monkeypox.py` updates any data.

spoonerf:
@Marigold Okay sounds good, let's hold fire. Is it currently preventing the [second snapshot](https://github.com/owid/etl/blob/0330030d5246c5fe3f78b9fa6f57b0e018ed6b4e/scripts/update-monkeypox.sh#L10) in the automated script running?

Marigold:
> @Marigold Okay sounds good, let's hold fire. Is it currently preventing the [second snapshot](https://github.com/owid/etl/blob/0330030d5246c5fe3f78b9fa6f57b0e018ed6b4e/scripts/update-monkeypox.sh#L10) in the automated script running?

Oh, you're right, it fails and never executes the second one. Then I think we should merge this one, even if it doesn't add any data (I just checked).

PR #3290, opened 2024-09-13
📊 cancer: NHS stats on diagnosis routes and survival rates

veronikasamborska1994:


etl/steps/data/grapher/cancer/2024-09-13/diagnosis_routes_survival.py
```diff
@@ -0,0 +1,33 @@
...
+# Get paths and naming conventions for current step.
+paths = PathFinder(__file__)
+
+
+def run(dest_dir: str) -> None:
+    #
+    # Load inputs.
+    #
+    # Load garden dataset.
+    ds_garden = paths.load_dataset("diagnosis_routes_survival")
```
spoonerf, line 14: It looks like some of the survival rates are above 100%, which surprised me. Perhaps it would be a nice addition to description_key to make a note of why that could be?

etl/steps/data/garden/cancer/2024-09-13/diagnosis_routes_by_route.meta.yml
```diff
@@ -0,0 +1,44 @@
...
+    <% endif %>
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 365
+
+
+
+tables:
+  diagnosis_routes_by_route:
```
spoonerf, line 32: Worth adding display.numDecimalPlaces: 0 for integers and maybe 1 for the %

spoonerf (approved):
Looks great, some minor comments!

etl/steps/data/grapher/cancer/2024-09-13/diagnosis_routes_survival.py
veronikasamborska1994: nicely spotted! I also noticed that and just thought it could be because of the age-standardization but can't see anything specific on this on their website atm! I'll try to dig into this a bit more a bit later!

PR #3292, opened 2024-09-16
:bug: Improve chart-sync

Marigold:
Couple of bugfixes for `chart-sync`

- Don't use default grapher schema `004`, but always load the schema specified in `config["$schema"]` (otherwise we get errors in chart-diff)
- Remap variableId of `sortColumnSlug`, this was previously left out (also try mapping new unknown fields)

### DIFF
--- a/apps/wizard/app_pages/chart_diff/conflict_resolver.py
+++ b/apps/wizard/app_pages/chart_diff/conflict_resolver.py
@@ -2,7 +2,6 @@ import json
 from copy import deepcopy
 from typing import cast
 
-import requests
 import streamlit as st
 import structlog
 from requests.exceptions import HTTPError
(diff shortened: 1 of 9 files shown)


apps/wizard/app_pages/chart_diff/conflict_resolver.py
```diff
@@ -153,7 +152,7 @@ def resolve_conflicts(self, rerun: bool = False):
                 config.pop(field, None)
 
             # Verify config
-            config_new = validate_chart_config_and_set_defaults(config, schema=get_schema("004"))
+            config_new = validate_chart_config_and_set_defaults(config, schema=get_schema_from_url(config["$schema"]))
```
lucasrodes, line 155: could it be the case that this raised a `KeyError` exception bc `$schema` is not a property in `config`?

lucasrodes (approved):
lgtm, thanks!

apps/wizard/app_pages/indicator_upgrade/charts_update.py
```diff
@@ -105,7 +105,7 @@ def push_new_charts(charts: List[gm.Chart], schema_chart_config: Dict[str, Any])
             config_new = update_chart_config(
                 chart.config,
                 st.session_state.indicator_mapping,
-                schema_chart_config,
+                get_schema_from_url(chart.config["$schema"]),
```
lucasrodes, line 108: also, possible `KeyError` raised sometimes?

apps/wizard/app_pages/chart_diff/conflict_resolver.py
Marigold: Good question. Fortunately, `$schema` is never null.

PR #3308, opened 2024-09-18
📊 mpox: get data from the WHO Shiny App

spoonerf:
Hey @Marigold - it looks like the data for suspected cases is base64 encoded in the [WHO Shiny App](https://worldhealthorg.shinyapps.io/mpx_global/#26_Case_definitions). I've had a go at extracting it here, but not sure how solid the snapshot method is. 

This could be a new source of suspected cases and deaths from the WHO, but annoyingly they don't include data for either of these prior to 2024, when I think there were suspected cases from Sept 2023.


Marigold (approved):
Sweet! I've never seen a data being provided in HTML element. Very nice find :)

lucasrodes:
@spoonerf
what's the status of this PR?

PR #3311, opened 2024-09-18
:tada: Add data explaining capability to Insighter

larsyencken:
The insighter has one mode at the moment which uses a chart view.

This PR adds a second, where it attempts to capture all the trends in a raw dataset.


larsyencken:
Merging without review as part of prep for Stockholm.

larsyencken:
We should simplify that code when we have a new CSV API.

PR #3320, opened 2024-09-20
 :bug: Deprecate charts.list_charts

Marigold:
With the introduction of data catalog, we no longer have a way to get a list of slugs for all charts. This was used in `charts.list_charts()` function in data catalog.

I don't know how often was this function used. If we want to keep it, we'd have to add an endpoint that would return all charts slugs or call public Datasette.


Marigold:
Merging to make ETL CI pass again.

PR #3335, opened 2024-09-30
✨ wizard: option to show archived datasets

lucasrodes:
Pablo Arriagada reported that he could not view archived datasets in Indicator Upgrader (as part of his work in https://github.com/owid/etl/pull/3343).

To overcome this, he unarchived a dataset.

This PR solves this by allowing the user to show archived datasets if they click the 'show all datasets' option.

image


paarriagadap (approved):
It works for me! Thanks!

I also found another chart with the exact same issue:
https://admin.owid.io/admin/variables/819727

PR #3341, opened 2024-09-30
 :sparkles: Use real users on staging servers

Marigold:
Goes together with https://github.com/owid/owid-grapher/pull/4083 that automatically logs Tailscale users to Admin. This means that going to http://staging-site-feature won't ask you to login as `admin` user and won't show an annoying popup about compromised password.

On ETL side, charts are created by logged user instead of ETL and synced correctly to production by chart-diff.

Datasets will still be created by ETL user, because they're created in production and there's no easy way how to know who created it. That is, unless we add `owner` to dataset metadata and start using it.

## TODO before merging
- [x] Merge PR in owid-grapher https://github.com/owid/owid-grapher/pull/4083

### DIFF
--- a/apps/chart_sync/admin_api.py
+++ b/apps/chart_sync/admin_api.py
@@ -25,7 +25,7 @@ def is_502_error(exception):
 
 
 class AdminAPI(object):
-    def __init__(self, owid_env: OWIDEnv, grapher_user_id: Optional[int] = None):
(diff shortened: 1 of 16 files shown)


pabloarosado (approved):
Thanks, this is very useful! The code looks good. I'm not sure how I can test it. Please let me know if you want me to test it on my current branch or elsewhere.

PR #3346, opened 2024-10-01
 :bug: Add back old version of epoch dataset

Marigold:
@veronikasamborska1994 some datasets still use `data://meadow/artificial_intelligence/2024-09-09/epoch`, so I had put it back to fix ETL.


veronikasamborska1994:
oh shit sorry! looks like I've accidentally pushed to master before finishing - thanks for catching it!

Issue #3347, opened 2024-10-01
`uv` auto-installation is funky and unusual


larsyencken:
## Context

We recently switched from `poetry` to `uv`, a faster package manager with a bunch of other benefits.

## Problem

In the switch, we:

- Hardcoded `~/.cargo/bin/uv` as the path for `uv` in `defaults.mk` and elsewhere
- Install `uv` into `~/.cargo/bin/uv`, but only if you do not have it installed already somewhere

This both puts `uv` in a strange place that should be managed by `cargo`, and only if you have a Rust toolchain stalled, and it's not comfortably compatible with managing `uv` through a different system such as Homebrew.

## Expected behaviour

The `Makefile` should ask people to install `uv` if it detects they do not have it (e.g. `brew install uv`, `pip install uv`).

In CI, we should simply install it ourselves to a known good path (e.g. `/usr/local/bin/uv`), or otherwise put it in `~/.local/bin/uv` and make sure that is in the `PATH`, as we do for other packages we install.

PR #3348, opened 2024-10-01
:bug: Allow `uv` to be managed outside of `.local/cargo/bin`

larsyencken:
Fixes https://github.com/owid/etl/issues/3347

- Lets the user decide where and how to install `uv`, and makes them responsible for putting it in their `PATH`
- Gives an error if `uv` is not installed, suggesting `brew install uv` as a resolution for the user


Marigold:
@larsyencken how about this? `.cargo/env` adds `uv` to the path which works for CI, and it doesn't do anything for users who installed it via brew.

(Alternatively, I can install uv explicitly in all Buildkite steps & CI)

```
install-uv-default:
	@if ! command -v uv >/dev/null 2>&1; then \
		echo '==> UV not found. Installing...'; \
		curl -LsSf https://astral.sh/uv/install.sh | sh; \
	fi

.venv-default: install-uv .sanity-check
	@echo '==> Installing packages'
	@if [ -n "$(PYTHON_VERSION)" ]; then \
		echo '==> Using Python version $(PYTHON_VERSION)'; \
		export UV_PYTHON=$(PYTHON_VERSION); \
	fi
	[ -f $$HOME/.cargo/env ] && . $$HOME/.cargo/env || true && uv sync --all-extras
```

larsyencken:
@Marigold Overall it looks good, especially if there's nowhere else we need `uv`, i.e. if all other parts of the recipe just use `.venv/bin/<blarg>` to do their thing.

PR #3351, opened 2024-10-01
📊 AI: AI epoch adding regressions code

veronikasamborska1994:
hey @Marigold this is the remainder of the code that I accidentally pushed to master that caused the bug! it's a monthly update so most of these have been reviewed several times before except this new bit of [code](https://github.com/owid/etl/pull/3351/files#diff-cf05e0cd3896171000fab8e3b62f740a68e97bc61fa784794d8c4be59150aca7) with regressions


veronikasamborska1994:
thanks @Marigold! nicely spotted - I guess I was changing some things and population sizing is the default which I didn't remove!

PR #3352, opened 2024-10-02
🧹 Attempt to run all snapshots to time them and count failures

pabloarosado:
I created a script to time and execute all snapshots, and store the result in an accompanying json file (we can move things elsewhere afterwards).

The script is running now.

@Marigold I've renamed the `.config/rclone` folder in the staging server to ensure no snapshots are actually uploaded (which could mess things up). Please let me know if you think there can be any other unexpected side effects of running all snapshot scripts.


Marigold:
Nice! I'd try running it with `--upload` on the staging server and then committing it all to the PR. It's safe, there are no side-effects, every snapshot is uploaded to a new file and doesn't overwrite the previous one.

Marigold (approved):

pabloarosado:
The resulting file has attempted to execute 440 snapshots, of which:
* 20% failed
* 4% took longer than 100s (and where therefore interrupted)
* 86% were executed successfully.

Results file: [snapshot_execution_times.json](https://github.com/user-attachments/files/17231534/snapshot_execution_times.json)

@Marigold I don't understand why we would run them again with `upload=True`. I don't like the idea of creating dozens or a few hundred new snapshots that are totally disconnected from the code. I'm not sure if there's an easy way to prune them afterwards, but it sounds a little bit risky to me, and I don't see much benefit. What would be the motivation?

PR #3355, opened 2024-10-02
 :hammer: Make schema version in grapher_config: compulsory

Marigold:
If grapher_config.$schema is missing from the metadata, set it to the default value (which is the latest version, hardcoded in the config). The schema is now required by the Admin, and we also want to ensure that our configs are always using the latest schema whenever possible.

I've added schema validation on the ETL side, and, to my surprise, all of our existing grapher configs were valid.

## TODO
- [x] Validate grapher configs in grapher step
- [x] Undo ETL_EPOCH after testing that all configs make it through
- [x] Make sure all configs are valid with the latest schema

### DIFF
--- a/apps/chart_sync/admin_api.py
+++ b/apps/chart_sync/admin_api.py
@@ -11,7 +11,6 @@ import structlog
 from requests.adapters import HTTPAdapter, Retry
 from requests.exceptions import HTTPError
 from sqlalchemy import text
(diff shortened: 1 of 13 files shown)


sophiamersmann (approved):
Thanks, Mojmir! 🙏🏻

I can't say anything about the particulars, but overall looks good to me!

PR #3361, opened 2024-10-02
 :broom: Rerun all snapshots

Marigold:
Based on https://github.com/owid/etl/pull/3352.

Try running all snapshots, upload them to S3 and check how many of them have updated data.

## Notes
- This runs even outdated snapshots that might already have a newer version available. We should [prune them](https://github.com/owid/etl/issues/3364) and then run this again


Marigold:
Keep it open for a bit longer.

PR #3365, opened 2024-10-03
🎉 Add vscode extension to find ETL step files more easily

pabloarosado:
The idea was to create a VSCode extension that makes browsing in ETL easier.

The goal is to have a shortcut to open a search bar (similar to `cmd+p`), type a file name, and files should appear sorted in descending order of date (given in the paths). Even more ideal would be to show first garden, then meadow, then grapher, then the rest.

I created two versions of `*.vsix` files. The first version sorts better (although it doesn't seem to be perfectly right). The second version has a better response (it shows relative paths without showing file names twice), however it clearly doesn't sort well. I tried to combine both solutions, but didn't manage to make it work.

To install a version:
* `cmd+shift+p` -> `Extensions: Install from VSIX`. Choose `find-latest-etl-step-0.0.1.vsix`.
* Restart VSCode. For some reason, you may need to open a random file before the extension works.
* Type `ctrl+shift+L`, and type a file name.

Ideally, you should first see the latest version of a step (usually a garden step). It would be nice to be able to type "meadow step_name" to find the latest meadow file for that step, but I didn't manage to achieve that.


lucasrodes:
Haven't tested, but I just read the name of this issue and fell in love ー thanks Pablo, great idea!

Marigold:
Nice! I spent some time (actually, too much time...) trying to get searches like "meadow cherry" to work, but apparently, QuickPick doesn’t support [fuzzy matching](https://github.com/microsoft/vscode/issues/34088), and I couldn’t find a way to hack it. I think it should be possible to sort the results not just by date, but also by channel.

(It took me a while to figure out how to run the extension in development mode. If anyone wants to develop this further and hear about my experience, feel free to reach out.)

pabloarosado:
Thanks both! I also spent some time just setting things up.
Creating the extension was quite easy, I think all you need is:
```
npm install -g yo generator-code 
yo code
```
This will create all the boilerplate code.

Then, to develop and debug, I think I've figured out the easiest way to set it all up:
* Go to the root folder of the extension, `find-latest-etl-step`.
* Run `npm run install`.
* Open a VSCode on that folder. On the terminal, run `npm run watch` (and keep this open).
* Then go to the debug pane, and hit the play button to Run Extension. This will open a new VSCode window.
* On the new window, hit `cmd+shift+p` and select `Find Latest ETL Step` to test the extension.
* You can now make changes in the code, then go to the new window, and hit `cmd+r` to refresh. Then rerun `cmd+shift+p` and select `Find latest ETL step`.

Once you are happy with the result, bump up the version in `package.json` (although for now version 0.0.2 was not properly working, so I'll try to get that to work), go to the terminal, ensure the extension compiles, and package it.
```
npm run compile
vsce package
```
This creates a new `.vsix` file for the new version.

@Marigold is that more or less what you found?

Marigold:
Yeah, that's what I learned the hard way :). Could you add your instructions README?

pabloarosado:
OK, I'll start using it in the coming days, and, if it turns out to work well, I'll add the instructions to the ETL documentation and announce it to the rest of the team.

PR #3366, opened 2024-10-03
🚧 famines: world peace dataset

veronikasamborska1994:
hi @Marigold! not sure why the staging server is failing here and looks like it's missing [variables](http://staging-site-famines-wpf/admin/charts/8106/edit) that are unrelated to this PR?


Marigold (approved):
Code and charts look good. ETL is failing because it was created a long time ago (and at a wrong time when master was failing with "Variables used in charts will not be deleted automatically"). It should be good once you merge it.

PR #3376, opened 2024-10-07
:bar_chart: Update World Bank PIP to Sept 2024 release (fix)

paarriagadap:
This one adds a fix from https://github.com/owid/etl/pull/3372
There was a problem with a repeated dag for the PIP explorer file.


Marigold (approved):
Looks good!

PR #3385, opened 2024-10-08
✨ anomalist: improve utils

lucasrodes:
This PR re-organizes several grapher/db/s3 utils in our library.

We had some tools spread in multiple modules, within `etl/` and in `apps/`. The idea is to compile these in centralised modules, so that it is easier to improve them, their maintenance, and enhancement.

Changes include:
- `etl.db`: This module already existed but included a mixture of database utils and methods to query specific bits of data from our database. From now on, it should only include database utils.
- `etl.grapher_io`: This is a new module, which should contain all our tools to access data/metadata (and possibly other bits of content) from our Grapher database/api (mysql/s3). The important bit here, is that it accesses non-etl content. Its content comes from various places: new content, from `etl.db`, from, `apps.backport.datasync.data_metadata`, etc.
- `etl.grapher_model`: Added some functions to some ORM objects. For instance, class `Variable` has now a `get_data` method to access it's data (i.e. it accesses s3 data).


etl/grapher_io.py
```diff
@@ -0,0 +1,796 @@
...
+    with db_conn.cursor() as cursor:
+        cursor.execute(query)
+        result = cursor.fetchall()
+
+    assert len(result) == 1, f"Ambiguous or unknown dataset name '{dataset_name}'"
+    dataset_id = result[0][0]
+    return dataset_id
+
+
+def get_variables_in_dataset(
```
Marigold, line 410: Eventually, I think we should move this as a method to `grapher_model.Dataset` so that all dataset-related methods are in one place. (Not now though!)

etl/grapher_io.py
```diff
@@ -0,0 +1,796 @@
...
+    dataset_id: int, only_used_in_charts: bool = False, db_conn: Optional[pymysql.Connection] = None
```
Marigold, line 411: `db_conn` should eventually also get deprecated. We should only use `engine` or `session` and `read_sql` function from `db.py`.

lucasrodes: this makes sense. I've added a @deprecated tag

PR #3396, opened 2024-10-10
🎉 anomalist: Detect new datasets automatically

pabloarosado:
* Fix bug in the mapping generated by indicator upgrader.
* Let anomalist show the full list of grapher datasets, and by default select new grapher datasets.
* Some refactoring was necessary, and some additional changes in utils functions.
* I created some temporary energy_mix and electricity_mix datasets, and changed the dag accordingly.
  THIS CHANGES SHOULD NOT BE MERGED TO MASTER! We can consider pushing it to `wizard-anomalist`, but I'm not sure if the syncing will do the right thing. So, after inspection, I'm happy to revert the changes in the data, for safety.

TODO
* For now, anomalist does nothing with the selected datasets. Currently, it loads indicators from a single dataset based on its URI. But I suppose we need to (1) load all indicators from all datasets, (2) based on dataset ids (rather than URI, since we will load grapher data, as far as I understand). I can look into this.


lucasrodes (approved):
LGTM, thanks Pablo! Also for re-arranging things.

I've consolidated some code from indicator_upgrade.utils.get_datasets and anomalist.utils.get_datasets_and_mapping_inputs into generic utils.io.get_steps_df

PR #3397, opened 2024-10-10
Fix missing archived grapher datasets in version tracker

pabloarosado:
Some archived grapher datasets were missing in `VersionTracker.steps_df`. The reason was that, when extracting the relevant info from the database, steps were assign either `data://` or `data-private://` prefixes based on `isPrivate`. But there were grapher datasets that were manually archived (by setting `isArchived` and `isPrivate` to 1) which came from non-private ETL steps.
This PR fixes the issue, and those archived grapher datasets now appear in `steps_df`.

### DIFF
--- a/etl/grapher_io.py
+++ b/etl/grapher_io.py
@@ -358,10 +358,11 @@ def _fetch_metadata_from_s3(variable_id: int, env: OWIDEnv | None = None) -> Dic
         return {}
 
 
-##############################################################################################
+#######################################################################################################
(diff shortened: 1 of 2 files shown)


lucasrodes (approved):
LGTM, thanks Pablo!

I've added some brief docs.

Also a reminder, that we should probably slowly stop using db_conn in favour of OWIDEnv objects. This basically means refactoring some functions in etl.grapher_io and their usage (for the future).

PR #3398, opened 2024-10-11
 :hammer: Cache indicator data from S3

Marigold:
Cache indicator data from S3 to disk along with its ETag. When fetching it again, first compare its checksum to the cached version, and if they match, load it from disk instead. This should improve data loading speed in Anomalist.


lucasrodes (approved):
LGTM, thanks for implementing this!

PR #3405, opened 2024-10-15
🎉 anomalist: Improve Anomalist backend

pabloarosado:
Various improvements on anomalist backend.

### DIFF
--- a/apps/anomalist/anomalist_api.py
+++ b/apps/anomalist/anomalist_api.py
@@ -26,7 +26,7 @@ log = structlog.get_logger()
 INDEX_COLUMNS = ["entity_id", "year"]
 
 # Define anomaly types.
-ANOMALY_TYPE = Literal["version_change", "time_change", "lost"]
+ANOMALY_TYPE = Literal["upgrade_change", "time_change", "upgrade_missing"]
 
 
 def load_latest_population():
@@ -118,7 +118,7 @@ class AnomalyDetector:
             new = self.df[self.df["entity_id"] == row["entity_id"]][["entity_id", "year", variable_id]]
             new["country"] = map_series(new["entity_id"], self.entity_id_to_name)
             new = new.drop(columns=["entity_id"]).rename(columns={row["variable_id"]: variable_name}, errors="raise")
-            if score_name == "version_change":
+            if score_name == "upgrade_change":
                 variable_id_old = row["variable_id_old"]
                 old = self.df[self.df["entity_id"] == row["entity_id"]][["entity_id", "year", variable_id_old]]
                 old["country"] = map_series(old["entity_id"], self.entity_id_to_name)
@@ -148,10 +148,10 @@ class AnomalyDetector:
                 ).show()
 
 
-class AnomalyLostData(AnomalyDetector):
+class AnomalyUpgradeMissing(AnomalyDetector):
     """New data misses entity-years that used to exist in old version."""
 
-    anomaly_type = "lost"
+    anomaly_type = "upgrade_missing"
 
     def get_score_df(self) -> pd.DataFrame:
         # Create a dataframe of zeros.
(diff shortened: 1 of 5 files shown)


apps/anomalist/anomalist_api.py
```diff
@@ -0,0 +1,406 @@
...
+}
+
+########################################################################################################################
+
+# AGGREGATE ANOMALIES:
+
+
+def load_latest_population():
+    # NOTE: The "channels" parameter of the find function is not working well.
+    candidates = find("population", channels=("grapher",), dataset="population", namespace="demography").sort_values(
```
Marigold, line 44: We could also load it from the local catalog with
```
population = Dataset(DATA_DIR / "grapher/owid/latest/key_indicators")["population"]
```
Might be better than `find`.

apps/anomalist/detectors.py
```diff
@@ -0,0 +1,193 @@
...
+        variable_mapping: Dict[int, int],
+        entity_id_to_name: Dict[int, str],
+    ) -> None:
+        self.variable_ids = variable_ids
+        self.variable_mapping = variable_mapping
+
+        # Initialize data objects.
+        self.df = df
+        self.metadata = metadata
+        self.entity_id_to_name = entity_id_to_name
```
Marigold, line 66: Do we need to convert ids to names in every detector? We could just convert entity ids to names in `anomaly_detection` and all detectors would work with them. Or is there a benefit in passing them as ids?

apps/anomalist/detectors.py
```diff
@@ -0,0 +1,193 @@
...
+
+    return df_score_long
+
+
+class AnomalyDetector:
+    anomaly_type: str
+
+    def __init__(
+        self,
+        variable_ids: List[int],
```
Marigold, line 54: Passing data to `__init__` and accessing them via `self.` can get tricky. An alternative is to keep it stateless and pass only hyperparameters and config to `__init__` and actual data to `get_score_df` method. This would make it look like this
```
def get_score_df(
        self, df: pd.DataFrame, variables: list[gm.Variable], variable_mapping: dict[int, int]
    ) -> pd.DataFrame:
```

It makes it clear what data is it using and prevents unexpected memory leaks (when you don't delete the object). Scikit-learn uses that pattern too in `estimator.fit(X, y)` and `estimator.predict(X)`.

apps/anomalist/detectors.py
pabloarosado: I suppose we can just use entity names everywhere. I thought sticking to entity ids made sense, just because we are using indicator ids everywhere too. But I can change that to use only entity names.

Marigold (approved):
Looks good! I had a couple of suggestions, feel free to go what your instincts tell you.

(By the way, this is the code I had before going on holiday)

```py
class AnomalyDetector:
    anomaly_type: str

    def get_text(self, entity: str, year: int) -> str:
        return f"Anomaly happened in {entity} in {year}!"

    def get_score_df(
        self, df: pd.DataFrame, variables: list[gm.Variable], variable_mapping: dict[int, int]
    ) -> pd.DataFrame:
        raise NotImplementedError


class LostAnomalyDetector(AnomalyDetector):
    # New data misses entity-years that used to exist in old version.

    anomaly_type = "lost"

    def get_score_df(self, df: pd.DataFrame, variables: list[gm.Variable], variable_mapping) -> pd.DataFrame:
        # Create a dataframe of zeros.
        df_lost = df.copy()
        # Make 1 all cells that used to have data in the old version, but are missing in the new version.
        for variable_id_old, variable_id_new in variable_mapping.items():
            affected_rows = df[(df[variable_id_old].notnull()) & (df[variable_id_new].isnull())].index
            df_lost.loc[affected_rows, variable_id_new] = 1

        return df_lost


class GPAnomalyDetector(AnomalyDetector):
    anomaly_type = "gp"

    def get_text(self, entity: str, year: int) -> str:
        return f"There is an outlier {entity} in {year}!"

    def get_score_df(
        self, df: pd.DataFrame, variables: list[gm.Variable], variable_mapping: dict[int, int]
    ) -> pd.DataFrame:
        score_df = df * np.nan
        ...
```

PR #3411, opened 2024-10-15
:sparkles: Integrate GP outlier detector

Marigold:
Add `AnomalyGaussianProcessOutlier` detector that calculates z-score as a deviation from fitted gaussian process. It's not yet optimized, so I'm only calculating anomalies for a couple of countries as a proof of concept.

Additional changes:
- Remove `dataset_ids` from `anomaly_detection` and only keep it in CLI to load variables from given datasets
- When creating long format, drop NaN values instead of filling them with zeros. Is this ok or are you somehow relying on that @pabloarosado?
- Convert year and entity to categorical in order to reduce size in MySQL


lucasrodes:
I think I experienced something related to the second point: this might be leading to some errors in the UI atm.

I haven't checked but i think that something like this is happening:

- if variable_mapping is empty or None (i.e. no upgrade) upgrade-type anomalies are still generated and filled w zeroes.
- This anomalies are added in anonalies DB table, with zero-filled dataframes.
- In the UI, we then get anomaly objects of upgrade-type, even if it does not apply.


I think that we should not add rows with anomalies if none are found or we are not asked to.

Therefore, Id say: (i) fill with NaNs if it is not applicable (instead of zero-filling) and (ii) add row w anomaly only if applicable.

apps/anomalist/detectors.py
```diff
@@ -109,14 +133,15 @@ class AnomalyUpgradeMissing(AnomalyDetector):
...
-        df_lost = pd.DataFrame(np.zeros_like(df), columns=df.columns)[INDEX_COLUMNS + variable_ids]
-        df_lost[INDEX_COLUMNS] = df[INDEX_COLUMNS].copy()
+        df_lost = self.get_zeros_df(df, variable_ids)
 
         # Make 1 all cells that used to have data in the old version, but are missing in the new version.
         for variable_id_old, variable_id_new in variable_mapping.items():
             affected_rows = df[(df[variable_id_old].notnull()) & (df[variable_id_new].isnull())].index
             df_lost.loc[affected_rows, variable_id_new] = 1
 
+        __import__("ipdb").set_trace()
```
pabloarosado, line 143: Delete.

apps/anomalist/gp_detector.py
```diff
@@ -13,52 +13,54 @@
...
 
-    def get_score_df(self, df: pd.DataFrame, variables: list[gm.Variable]) -> pd.DataFrame:
-        score_df = cast(pd.DataFrame, df.sample(1))
-        score_df.iloc[0, :] = [random.normalvariate(0, 1) for _ in range(df.shape[1])]
-        return score_df
+class AnomalyGaussianProcessOutlier(AnomalyDetector):
+    anomaly_type = "gp_outlier"
 
+    def get_score_df(self, df: pd.DataFrame, variable_ids: List[int], variable_mapping: Dict[int, int]) -> pd.DataFrame:
+        # Create a dataframe of zeros.
```
pabloarosado, line 25: The comment needs to be adjusted. But also, you could indeed just use a dataframe of zeros, and then define the z-score in absolute value. Or is there any benefit in keeping the sign of the deviations?

pabloarosado (approved):
Thanks, it looks good! I agree with dropping nans and zeros in the long format scores dataframe.
I suggested making the z-score strictly positive, but it's up to you. Feel free to merge, I'm happy to look into how to speed up and optimize the GP regressor.

apps/anomalist/detectors.py
Marigold: Whoops, thanks for noticing this.

apps/anomalist/gp_detector.py
Marigold: Z-score is converted to absolute value at the very end. You're right that it doesn't matter whether we use zeros or nans. I found nans more useful for visualization purposes when prototyping, so I'm going to keep them, but feel free to change it if you want.

PR #3420, opened 2024-10-16
🎉 anomalist: Experiment with different anomaly detection methods

pabloarosado:
* Improved the script to play around with anomaly detection for a specific dataset, for a given detector (it has a CLI).
* Some small refactoring and improvements in Anomalist.
* I was trying different simple anomaly detection strategies, but they were all slow and didn't add much value compared to current simple methods. They could be optimized (as you did with GP detector) but I don't think they yield as good result as GP, so it may not be worth it.


Marigold (approved):
It's good that you included all your attempts with other anomaly detectors. If we start detecting anomalies in the background on staging, their runtime might not be such a problem.

PR #3422, opened 2024-10-16
 :sparkles: Add max_time and n_jobs to GP

Marigold:
- Stop detecting anomalies after `max_time` in GP. They are processed in random order, but with probability proportional to country population (entities without population are given mean probability). It's **not being used yet**, but will be useful when we start detecting anomalies in production
- Add `n_jobs` for multiprocessing. This works poorly on my shitty laptop, but might provide speed up on our beefy server
- Return long format from GP detector and handle it in anomalist_api

### DIFF
--- a/apps/anomalist/gp_detector.py
+++ b/apps/anomalist/gp_detector.py
@@ -1,71 +1,165 @@
 import random
 import time
 import warnings
+from multiprocessing import Pool
 from typing import Dict, List, Optional
 
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import structlog
+from joblib import Memory
 from sklearn.exceptions import ConvergenceWarning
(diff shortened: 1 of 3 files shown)


apps/anomalist/gp_detector.py
```diff
@@ -1,72 +1,163 @@
...
+        items = _processing_queue(
+            list(df_wide.index.unique()),
+        )
 
-        return df_score
+        start_time = time.time()
+
+        results = []
+
+        items = items[:1000]
```
pabloarosado, line 87: Is this necessary?

pabloarosado (approved):
Looks good. In a way, it's good that you have an old laptop, since it forces you to optimize as much as possible :-P
We can consider also that the user may initialize the anomaly detection process and go to have a cup of coffee while it runs. It will depend on how good the results are.

PR #3425, opened 2024-10-17
📊 Add regions to the UNDP HDR data set

antea04:
For a DI @simonvanteutem would like to show regions (as in continents and income groups) in the [HDR data set](https://admin.owid.io/admin/datasets/6475), esp. for [this indicator.](https://admin.owid.io/admin/variables/899547) 

I've added population-weighted averages for regions where more than 80% of countries have data.


simonvanteutem:
Hey @antea04! I still can't find Europe (only Europe and Central Asia) or Africa (only sub-Saharan Africa). Is this the result of the 80% requirement, or something else? Let me know if I can help with anything!

Thank you!

Simon

antea04:
@simonvanteutem - Tou need to use the [data set on the staging server ](http://staging-site-regions-schooling/admin/datasets/6475) - there you will find all the Regions :)

paarriagadap (approved):
All good! I added a minor comment.

I also corrected the code to have population in people instead of million people

PR #3427, opened 2024-10-17
 :bug: Fix anomalist bugs

Marigold:
- Normalize gp_outlier to (0, 1)
- Let detectors own anomaly text instead of having it in the app
- Fix a couple of bugs I found during testing


pabloarosado (approved):
Thanks, it looks good!

PR #3429, opened 2024-10-18
✨ anomalist: Improve automatic detection of new datasets

pabloarosado:
* Simplify the way we detect new datasets, and with this speed up the start of Anomalist.
* Infer a variable mapping when indicator upgrader has not been yet executed.


Marigold (approved):
Code looks good! I'm a bit confused by the way we get new datasets, but I'll ask on slack.

PR #3434, opened 2024-10-18
🐛 anomalist: Fix bug with unknown indicators and long loading time

pabloarosado:
* Stop storing `dfScore`, since the load of anomalies from DB takes much longer than when only storing `dfReduced`, and we are currently not using `dfScore` anywhere.
* GP detector was searching for anomalies in all indicators in the incoming `df`. However, by construction, `df` includes data for both new and old indicators (and detecting anomalies in the old indicators is unnecessary). This was causing anomalist to raise an error. Now the detection takes half of the time for datasets for which there is an old version.


Marigold (approved):
Nice fixes.

PR #3437, opened 2024-10-18
✨ anomalist: Small improvement in Anomalist filters

pabloarosado:
* In the filter section, select detectors to show, instead of detectors to hide (to be consistent with other filters).


Marigold (approved):
Thanks, this is better!

PR #3441, opened 2024-10-21
:sparkles: Persist filter values in the URL for Anomalist

Marigold:
- Persist selected datasets and filter values in the URL
- Display anomalies from DB at startup (don't wait for a click on "Detect anomalies")
- A couple of small performance optimizations


pabloarosado (approved):
This is nice! On one of your comments: Yes, I think it doesn't make sense anymore to detect new datasets based on file changes. Since anomaly detection now runs automatically, we can simply list new datasets as any dataset for which there are anomalies detected.
That said, it would be good if the user could know that anomaly detection is currently running (especially if the results they are seeing are incomplete).

PR #3446, opened 2024-10-21
:bar_chart: Add UN SD census data 

antea04:
Adding the dates of the last population census from the [UN SD Population Census Programme](https://unstats.un.org/unsd/demographic-social/census/censusdates/).  This covers years 1985-2023. 

Indicators created:
- Year of last census
- Year since last census

Notes:
- I'm importing this from a local file which I created semi-manually by copy/ pasting the table on the website and then writing a small script to clean it. I don't know how to best document the process of creating the file from the website, so I'm happy to learn about best practices here. 
- There are currently two charts on "population census recently completed" ([World Bank](https://ourworldindata.org/grapher/population-census-world-bank)/ [Brambor et al](https://ourworldindata.org/grapher/population-census-brambor)). The Brambor data set is long run, but I guess the WB one could be replaced by this indicator? I've created a [draft chart](http://staging-site-unsd-census/admin/charts/8199/edit) to see how that would look like - I'd appreciate input here.


etl/steps/data/meadow/un/2024-10-21/census_dates.py
```diff
@@ -0,0 +1,33 @@
...
+    # Retrieve snapshot.
+    snap = paths.load_snapshot("census_dates.csv")
+
+    # Load data from snapshot.
+    tb = snap.read()
+
+    #
+    # Process data.
+    tb = tb.drop_duplicates()
+    tb = tb[tb["Date"].apply(len) > 3]
```
veronikasamborska1994, line 22: could be nice to explain why you are doing this filtering here. I am sure there is a good reason so maybe add a comment so it's easier to remember why you've done this next time you update the data?

etl/steps/data/grapher/un/2024-10-21/census_dates.meta.yml
```diff
@@ -0,0 +1,28 @@
...
+
+
+tables:
+  years_since_last_census:
+    variables:
+      years_since_last_census:
+        title: Years since last census
+        description_short: The number of years since the last census of this country
+        unit: years
+        short_unit: "y"
```
veronikasamborska1994, line 22: is this something we do for the year short units? it doesn't look super intuitive to me when I look at the chart especially since you are coding the color mapp based on whether it happened in the last 10 years

etl/steps/data/grapher/un/2024-10-21/census_dates.meta.yml
```diff
@@ -0,0 +1,28 @@
...
+      last_census:
```
veronikasamborska1994, line 23: unlikely you'd use this so I'd probably drop this column and exclude it and instead create an indicator with [0,1] for whether there was a census in the last 10 years

etl/steps/data/garden/un/2024-10-21/census_dates.excluded_countries.json
```diff
@@ -0,0 +1,2 @@
+[
```
veronikasamborska1994, line 1: you don't need this file if you aren't excluding any countries so you can remove it here and the link to it in the garden step

etl/steps/data/garden/un/2024-10-21/census_dates.py
```diff
@@ -0,0 +1,194 @@
...
+    tb["date"] = tb["date"].replace(MONTHS_DICT)
+    tb["country"] = tb["country"].apply(clean_country)
+
+    # convert date to datetime
+    tb["date_as_year"] = tb["date"].apply(date_as_year)
+
+    #
+    # Process data.
+    #
+    tb = geo.harmonize_countries(
```
veronikasamborska1994, line 79: once you remove the empty excluded_countries file you just remove `excluded_countries_file=paths.excluded_countries_path` here

etl/steps/data/garden/un/2024-10-21/census_dates.py
```diff
@@ -0,0 +1,194 @@
...
+    ds_meadow = paths.load_dataset("census_dates")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow["census_dates"].reset_index()
+
+    # remove timeframes from date
+    tb["date"] = tb["date"].apply(lambda x: x.split("-")[1] if "-" in x else x)
+
+    # add comments
+    tb["comment"] = tb.apply(lambda x: get_comment(x["date"], x["country"]), axis=1)
```
veronikasamborska1994, line 66: are you actually using this bit for anything? I guess it could be a useful bit of info for the metadata but atm I don't know if this is necessary

etl/steps/data/garden/un/2024-10-21/census_dates.py
```diff
@@ -0,0 +1,194 @@
...
+    tb_census.m.short_name = "years_since_last_census"
+
+    for col in tb_census.columns:
+        tb_census[col].metadata = tb["date"].m
+        tb_census[col].metadata.origins = tb["date"].m.origins
+    return tb_census
+
+
+def add_uk(tb_census):
+    years = [int(x) for x in range(1985, 2024)]
```
veronikasamborska1994, line 177: you could define the years at the top so it's easier to edit on the next update. or just do max and min if appropriate

veronikasamborska1994 (approved):
LGTM! I'd possibly add a separate indicator with 1 and 0 for whether there was a census in the last 10 years and use that for the map as it's a bit confusing atm once you hover over the different countries and see a range of values but that's up to you of course!

etl/steps/data/garden/un/2024-10-21/census_dates.excluded_countries.json
antea04: Ah good point, thanks!

etl/steps/data/garden/un/2024-10-21/census_dates.py
antea04: Currently not, but I also did not want to lose this information, in case we get questions or comments about this data. Should I better remove it?

etl/steps/data/garden/un/2024-10-21/census_dates.py
antea04: Good point, will do!

etl/steps/data/grapher/un/2024-10-21/census_dates.meta.yml
antea04: I thought I had seen this before, but now that I checked it doesn't seem to be the case! I removed it :)

etl/steps/data/grapher/un/2024-10-21/census_dates.meta.yml
antea04: I'll add an indicator for that!

etl/steps/data/garden/un/2024-10-21/census_dates.py
veronikasamborska1994: up to you! keep it if you think it can be useful. I just wondered if you were planning to use it for something else and then forgot to remove it.

veronikasamborska1994:
> @veronikasamborska1994 Thanks for the review, Veronika!
> 
> Re local files: so there are **no files** on the website - there is only [this table](https://unstats.un.org/unsd/demographic-social/census/censusdates/) which I linked in the snapshot.
> 
> I've gotten my local file (which I am uploading to the snapshot) by manually copy-pasting this table into a text file and writing a small script to convert that text file into a csv. This is probably not the best practice, so I'm open for other ideas on how to get the data and also how to document this!
> 
> I've also replaced the [chart using the World Bank data](https://ourworldindata.org/grapher/population-census-world-bank) with this data (as it covers more years and more countries). Feel free to have a look at the [chart diff](http://staging-site-unsd-census/etl/wizard/chart-diff) and see if this looks good!
> 
> Thanks again! 🥰

the chart looks good @antea04 ! I think the best practice would probably be to read the HTML and extract the data directly from the website but tbh given this is one chart I wouldn't go through all of this work just for this unless you are very experienced in this sort of thing. I think others on the team might have also suggested trying to get in touch with the source and ask for this data in a spreadsheet format but again I don't think it's worth waiting for them to respond (though of course you could still do this now too).  

my solution would be to describe what you've done to get the data at the top of the snapshot script. if you have any python code to process your text file you could actually include it in the snapshot script too and do that conversion within the script. this solution shouldn't take you long and also will mean every bit of code that you've written for this is on the etl.

antea04:
Thanks, Veronika! That's very helpful!

- I've now added my manual process/ cleaning script to the snapshot step - I think that's a really good solution to this.
- I had reached out to the UN SD before starting work on this to see if they have a csv/xls version and noone got back to me. I figured following up will take more time than manually importing it, but for the sake of nice code and reproducability I can follow up again 😄 

Thanks for your review and feedback - I'll merge this now and will amend it if I hear back from the UN SD.

PR #3447, opened 2024-10-21
:hammer: Anomalist performance

Marigold:
- Add function `variable_data_df_from_catalog` that loads data from catalog instead of R2
  - This is orders of magnitude faster (especially for huge datasets), but I haven't tested it much
  - There's no fallback to R2
- Add envs `ANOMALIST_MAX_TIME` and `ANOMALIST_N_JOBS` which have higher values on the staging server
- When using CLI, use new datasets by default (e.g. `etl anomalist --anomaly-types gp_outlier --force`)


pabloarosado (approved):
It looks good! But I think it makes sense to fallback to R2 if the dataset does not exist locally. Think of the case that Anomalist is used on a separate grapher dataset (not one of the new ones the user is working on). Those variables may not exist locally.

Marigold:
> Think of the case that Anomalist is used on a separate grapher dataset (not one of the new ones the user is working on). Those variables may not exist locally.

I'm afraid that falling back to getting variables from R2 would be confusing. Loading a large dataset from R2 takes minutes as opposed to seconds when loading it from local catalog. It'd be confusing to see UI frozen without knowing what's going on. We'd have to signal that we're using fallback in the UI. Staging servers have a complete catalog, so this would only happen locally - in which case it's better to rebuild the dataset in my opinion.

Anyway, I'd keep it as it is now, but if we run into this problem, I'm happy to add the fallback.

PR #3450, opened 2024-10-22
 :sparkles: Add errors to chart-sync

Marigold:
Add field `error` to `ChartDiff` object. It's used for errors that cannot be automatically fixed. We just had a case when someone replaced a chart by a different one in production (with different chart id) and kept the same slug. chart-sync tried to create a new chart, but Admin API rejected it because the slug was already taken.

This PRs shows error in chart-sync UI and in owidbot summary if chart-sync would create a new chart, but the slug is already taken.

Other changes:
- Show error messages from AdminAPI
- Enable bugsnag for chart-sync
- Skip charts that were already synced

This is how it looks
image

### DIFF
--- a/apps/wizard/app_pages/chart_diff/chart_diff.py
+++ b/apps/wizard/app_pages/chart_diff/chart_diff.py
@@ -248,7 +248,7 @@ class ChartDiff:
         checksums_diff = cls._get_checksums(source_session, target_session, chart_ids)
 
         # Get all slugs from target
-        slugs_in_target = set(read_sql("SELECT slug FROM chart_slug_redirects", target_session)["slug"]) | {"blablabla"}
+        slugs_in_target = set(read_sql("SELECT slug FROM chart_slug_redirects", target_session)["slug"])
 
         # Build chart diffs
         chart_diffs = []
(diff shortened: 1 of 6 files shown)


apps/wizard/app_pages/chart_diff/chart_diff.py
```diff
@@ -241,6 +247,9 @@ def from_charts_df(
         # Get checksums
         checksums_diff = cls._get_checksums(source_session, target_session, chart_ids)
 
+        # Get all slugs from target
+        slugs_in_target = set(read_sql("SELECT slug FROM chart_slug_redirects", target_session)["slug"])
```
pabloarosado, line 251: Does `chart_slug_redirects` contain all chart slugs? I guess `chart_configs` does.

pabloarosado (commented):
Thanks for fixing this. I'm not sure I'm following the logic. I left a comment. If you have checked this works, feel free to merge. Otherwise let me know if I can help checking.

apps/wizard/app_pages/chart_diff/chart_diff.py
Marigold: Good catch! It should be actually both as it'd fail on redirected slugs, too.

PR #3451, opened 2024-10-23
✨ anomalist: Improve scores and test anomaly detectors

pabloarosado:


Marigold (approved):
Thanks for improving GP detector, using scale looks quite elegant. The new score table is much better than before.

PR #3452, opened 2024-10-23
:sparkles: Prettify Anomalist scores table

Marigold:
- Convert z-score to p-value adjusted for multiple hypothesis testing instead of ad-hoc `min(z, 3) / 3`
  - This significantly lowers GP scores
  - Whether to adjust p-value is a question... time series with a lot of points (e.g. cherry blossom) give more reasonable scores, but it could be too low overall
- Prettify table with scores and add columns for all anomalies
   image

### DIFF
--- a/apps/anomalist/gp_detector.py
+++ b/apps/anomalist/gp_detector.py
@@ -80,7 +80,7 @@ class AnomalyGaussianProcessOutlier(AnomalyDetector):
 
     @staticmethod
     def get_text(entity: str, year: int) -> str:
-        return f"There are some outliers for {entity}! These were detected using Gaussian processes. There might be other data points affected."
+        return f"There are some outliers for {entity} in year {year}! These were detected using Gaussian processes. There might be other data points affected."
 
     def get_score_df(self, df: pd.DataFrame, variable_ids: List[int], variable_mapping: Dict[int, int]) -> pd.DataFrame:
         # Convert to long format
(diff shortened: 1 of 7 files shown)


apps/anomalist/gp_detector.py
```diff
@@ -149,17 +151,28 @@ def get_score_df(self, df: pd.DataFrame, variable_ids: List[int], variable_mappi
...
+        # df_score_long["anomaly_score"] = np.minimum(df_score_long["z"] / 3, 1)
 
-        return df_score_long
+        # Convert z-score into p-value
+        df_score_long["p_value"] = 2 * (1 - norm.cdf(np.abs(df_score_long["z"])))
+
+        # Adjust p-values for multiple testing
+        df_score_long["adj_p_value"] = df_score_long.groupby(["entity_name", "variable_id"]).p_value.transform(
+            lambda p: multipletests(p, method="fdr_bh")[1]
+        )
```
pabloarosado, line 162: Nice! I also implemented a more natural normalization on my branch, instead of clipping above 3. Using the CDF works well, but in practice it doesn't make much of a different for big anomalies. Everything above ~3 sigmas gets a score of 100%.
I didn't know about multipletests! Have you checked if it makes a bit difference?

apps/wizard/app_pages/anomalist/app.py
```diff
@@ -456,6 +462,35 @@ def _change_chart_selection(df, key_table, key_selection):
...
+def _score_table(df: pd.DataFrame, df_all: pd.DataFrame, indicator_id: int) -> pd.DataFrame:
+    """Return a table of scores and other useful columns for a given indicator. Return
+    styled dataframe."""
+    # Get scores of anomalies for the same indicator
+    gf = df_all[df_all.indicator_id == indicator_id]
+    gf = gf.pivot(index=["entity_name", "year"], columns=["type"], values=["score"])["score"].add_prefix("score_")
+    gf = gf.groupby(["entity_name"]).max().reset_index()
+
+    # Select columns to display and add scores of all detectors
+    cols = ["entity_name", "score_weighted", "views"]
```
pabloarosado, line 474: This is very nice! Why not showing the population as well (maybe the current population, instead of the population of the anomaly year). Up to you.

pabloarosado (approved):
Thanks, this is very nice!

apps/wizard/app_pages/anomalist/app.py
Marigold: Yep, good idea. Added that.

apps/anomalist/gp_detector.py
Marigold: > Have you checked if it makes a bit difference?

Yeah, it does make a big difference, although I'm not really sure if using multiple hypothesis adjustments makes sense for residuals. It's not motivated by any theory, just by the need to "make the number smaller for larger samples".

PR #3198, opened 2024-08-26
:sparkles: Improve Jinja templating experience

Marigold:

PR #3456, opened 2024-10-23
:lipstick: oecd: add metadata for private grants

paarriagadap:
Following comment https://github.com/owid/owid-issues/issues/1620#issuecomment-2431805867

### DIFF
--- a/etl/steps/data/garden/oecd/2024-08-21/official_development_assistance.meta.yml
+++ b/etl/steps/data/garden/oecd/2024-08-21/official_development_assistance.meta.yml
@@ -180,6 +180,19 @@ definitions:
     minor
     <%- endif -%>
 
+  description_short_oda_net_disbursements: &description_short_oda_net_disbursements |-
+    "{definitions.short_oda} {definitions.short_net_disbursements} {definitions.short_constant_usd}"
+
+  description_short_private_grants: &description_short_private_grants |-
+    "{definitions.short_net_private_grants} {definitions.short_net_disbursements} {definitions.short_constant_usd}"
+
+  description_short_by_donor: &description_short_by_donor |-
+    <% if donor == "Private donors" %>
+    *description_short_private_grants
+    <% else %>
+    *description_short_oda_net_disbursements
+    <%- endif -%>
+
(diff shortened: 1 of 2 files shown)


etl/steps/data/garden/oecd/2024-08-21/official_development_assistance.meta.yml
```diff
@@ -178,6 +180,38 @@ definitions:
...
+    - "{definitions.key_constant_usd}"
+    - "{definitions.key_oda_list_of_countries}"
+
+  description_key_private_grants: &description_key_private_grants
+    - "{definitions.key_net_private_grants}"
+    - "{definitions.key_private_reporting}"
+    - "{definitions.key_net_disbursements}"
+    - "{definitions.key_constant_usd}"
+
+  description_key_by_donor: &description_key_by_donor
```
paarriagadap, line 209: Is it possible to avoid this error, @Marigold?

Marigold: This was a tough one. The thing is that you cannot use YAML's anchors and references in combination with Jinja templating. This is further complicated by the fact that `description_key` is a list 🤦.

I committed a hacky solution which seems to be working, but is not great. I'll add this example to a [tracking issue](https://github.com/owid/etl/pull/3198) for improving templating experience.

paarriagadap: The only issue I see is that both `description_short` and `description_key` are shown in quotes. Is there a way to solve that?
image

Marigold: I removed double quotes from contents of `description_short_oda_net_disbursements`. If it is a multi-line string (`|-`) then these quotes are unnecessary.

paarriagadap: It works! Thank you again! I am not sure why this also fixed description_key, but I am happy anyway.

PR #3467, opened 2024-10-28
📊 energy: Add data on battery prices

pabloarosado:


pabloarosado:
I will handle this in the PR about energy prices: https://github.com/owid/etl/pull/3499

Issue #3277, opened 2024-09-12
:sparkles: enhance `dataset.read_table(...)` method with type normalisation


Marigold:
## Motivation

Historically, we've been using function `dataset["my_table"]` to access table from a dataset. Recently, a new helper method `dataset.read_table(reset_index: bool = False)` has been added that lets us read the table with reset index which is significantly faster for large dimensional datasets.

## Concept

We could add more functionality to `read_table` and make it de facto standard to read tables. These could be:
- Retype all columns to "standard" types (e.g. `uint8 -> int64`, `Float16 -> float64`) and categorical to `string` type
- ~~underscore column names etc. (see `.format` method)~~
  - Should already be in snake case by the time we're talking tables

larsyencken:
We think type standardising would remove some common footguns for data folk:

- `categorical` → `string`
- `Float*` -> `Float64`
- `Int*` -> `Int64`

PR #3474, opened 2024-10-31
 :sparkles: Convert dtypes to safe types Float64, Int64 and string

Marigold:
Implements https://github.com/owid/etl/issues/3277

Adds new parameter `ds.read_table("my_table", safe_types=True)` that converts all types to "safe" types `Float64`, `Int64` or `string[python]`. It's true by default.

(We've also considered the name `unpack` in the past instead of `safe_types`. I don't have a preference.)

## int64 vs Int64

Should we use numpy's `int64` type or nullable pandas type `Int64` (or even pyarrow types)? This PR goes with nullable types, but there are some hidden risks to it:
- Mixing `np.nan` and `pd.NA` creates a nightmare of compatibility issues. We should always use `pd.NA`
- Sometimes `np.nan` sneaks in, for instance division 0/0 returns `np.nan` even if we work with nullable types
```python
a = pd.Series([0]).astype("Int64")
b = pd.Series([0]).astype("Int64")
# the first item is np.nan, not pd.NA!
c = a / b
# this is actually False, `pd.isnull` doesn't detect np.nan when dtype is Int64
pd.isnull(c)
# this doesn't work either...
c.fillna(pd.NA)
# only this fixes it
c.mask(np.isnan(c), pd.NA)
```

## TODO before merging:
- [x] Increment ETL_EPOCH and make sure old datasets don't change
- [x] Undo ETL_EPOCH increment

### DIFF
--- a/lib/repack/tests/test_repack.py
+++ b/lib/repack/tests/test_repack.py
@@ -16,8 +16,8 @@ def test_repack_non_object_columns():
     df2 = df.copy()
     df2 = repack.repack_frame(df2, {})
 
-    assert df2.myint.dtype.name == "uint8"
-    assert df2.myfloat.dtype.name == "float32"
+    assert df2.myint.dtype.name == "UInt8"
+    assert df2.myfloat.dtype.name == "Float32"
     assert_frame_equal(df, df2, check_dtype=False)
 
 
@@ -27,6 +27,9 @@ def test_repack_object_columns():
             "myint": [1, 2, None, 3],
             "myfloat": [1.2, 2.0, 3.0, None],
             "mycat": ["a", None, "b", "c"],
+            "myintstr": [1, 2, 3, "4"],
+            "nans1": [1, 2, 3, pd.NA],
+            "nans2": [1, 2.1, 3, np.nan],
         },
         dtype="object",
     )
@@ -35,8 +38,11 @@ def test_repack_object_columns():
 
     df_repack = repack.repack_frame(df_repack)
     assert df_repack.myint.dtype.name == "UInt8"
-    assert df_repack.myfloat.dtype.name == "float32"
+    assert df_repack.myfloat.dtype.name == "Float32"
     assert df_repack.mycat.dtype.name == "category"
+    assert df_repack.myintstr.dtype.name == "UInt8"
+    assert df_repack.nans1.dtype.name == "UInt8"
+    assert df_repack.nans2.dtype.name == "Float32"
 
 
 def test_repack_frame_with_index():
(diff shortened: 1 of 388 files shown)


lib/repack/tests/test_repack.py
```diff
@@ -253,3 +253,73 @@ def test_repack_string_type():
...
+    assert df_safe.index.levels[1].dtype == "string[python]"  # type: ignore
+    # Check column dtype
+    assert df_safe["float_col"].dtype == "Float64"
+
+
+def test_to_safe_types_with_nan():
+    # Create a DataFrame with NaN values
+    df = pd.DataFrame(
+        {
+            "int_col": [1, 2, 3],
```
pabloarosado, line 308: It seems that is_integer_dtype returns False if it's a series of integers with nans. So, it's testing for `int`, not for `Int64`. I suppose that, ideally, it should return True, and we would load all integer columns (with possible nans) as `Int64`.
Should we include that extra logic? I'm not sure about it.

pabloarosado (approved):
Thanks for doing this! Changing types after repacking is a common headache.
I'm a bit unsure of the preferred dtypes. Maybe we should discuss them on the data architecture call (I agree with your choices, but I'm not sure if there may be some negative side effects).

Marigold:
I've updated it to use nullable types & string[pyarrow]. The first commit contains changes to steps that are not worth checking (datadiff confirmed they match), so please only look at [changes from the second commit](https://github.com/owid/etl/pull/3474/files/6cb9bfdb2dbe8305df5919d9ca1c2f9b6021a9ce..HEAD).

## Changes
- Repack library converts everything to **nullable types and categoricals** rather than a mix of nullable types, numpy types and categoricals
- New function `to_safe_types` that converts to `Int64`, `Float64` and `string[pyarrow]` types
- New method `ds.read(...)` (it used to be called `read_table`) that reads a table and by default resets index and converts to safe types
- Reading data from snapshot with `snap.read(...)` also uses safe types by default
- Update pandas to 2.2.3 (fixes problems with pyarrow)
- Division 0/0 returns `pd.NA` when working with Tables (it doesn't cover edge cases, but if they happen, you'd get an error when saving a dataset)

Marigold:
@pabloarosado This is finally ready for review. Check the comment above for a summary of changes.

I wanted to go one step further and enable `pd.options.future.infer_string = True`, which makes the new `string[pyarrow]` the default. I ran into some issues, though, so I'm leaving it for either a future PR or for the actual upgrade to pandas 3.

pabloarosado (approved):
Huge, thanks for this massive refactor! I have quickly scanned through all files and it all looks good. And you already checked for data changes. So I think this is ready to go!

Issue #3423, opened 2024-10-17
Add `catalogPath` to datasets table in MySQL


Marigold:
## Motivation

During Anomalist development, we ran into a situation where it would have been helpful to get the `catalogPath` of datasets in MySQL.

Right now we have to construct it ourselves as `grapher/{namespace}/{version}/{shortName}`. It'd be nice to have it there as a column, similarly to `datasets.variables`.

## Implementation

- Fill `catalogPath` when upserting datasets to MySQL. Use the path to check if a dataset should be updated or inserted.
- Add an extra column `tables` with a list of tables in dataset (as JSON)
- Backfill for existing datasets using the `catalogPath` of their indicators

PR #3475, opened 2024-10-31
 :sparkles: Add catalogPath column to datasets table

Marigold:
Implements https://github.com/owid/etl/issues/3423

Add columns `datasets.catalogPath` and `datasets.tables` and populate them when upserting datasets to MySQL. Only `catalogPath` is backfilled, `tables` are left as null (will be populated at next ETL_EPOCH increment).


pabloarosado (approved):
Looks good!

PR #3477, opened 2024-10-31
📊 Temporary dataset to simulate a data anomaly

pabloarosado:


pabloarosado:
This data was generated to produce [the DI on data anomalies](https://ourworldindata.org/data-insights/spotting-and-fixing-data-issues-how-we-help-improve-data-quality-on-and-off-our-publication), we don't need to keep it in ETL, so I'll close it without merging.

PR #3478, opened 2024-11-01
 :sparkles: Upsert mdims via API

Marigold:
Upsert mdims via API instead of upserting them directly to MySQL.


rakyi (approved):
LGTM!

BTW, there is one more change we might like to do already, even though it's optional currently. We added a new key to the top level mdim config, `grapherConfigSchema`, which is the schema to apply to every view grapher config, if present. The latest schema is currently `"https://files.ourworldindata.org/schemas/grapher-schema.005.json"`.

Obviously it means that if we do provide a grapher config schema override for a view, it should use that schema.

Could you add it too, if it's easy?

PR #3480, opened 2024-11-01
📊 homicides: UNODC 2024

spoonerf:


snapshots/homicide/2024-10-30/unodc.xlsx.dvc
```diff
@@ -0,0 +1,45 @@
...
+    attribution_short: UNODC
+
+    # Files
+    url_main: https://dataunodc.un.org/dp-intentional-homicide-victims
+    url_download: https://dataunodc.un.org/sites/dataunodc.un.org/files/data_cts_intentional_homicide.xlsx
+    date_accessed: 2024-10-30
+
+    # License
+    license:
+      name: CC BY 4.0
```
antea04, line 39: I'm leaving this comment because I am always wondering about this: 

The UN ODC itself does not explicitly release their materials under a CC BY 4.0 license (at least not under their terms-of-use link), but the Terms of Use do seem very similar to the conditions of this license. But at the same time the unodc.org and the dataunodc.un.org website have a copyright notice in the footer which links to this blanket UN copyright notice (https://www.un.org/en/about-us/copyright). 

With UN data sets where they don't explicitly release it under a license I always put "custom license" - © United Nations and then link the terms of use, where they allow the distribution. But I don't know what's the best/ correct way to do it, so I'd like your input!

snapshots/homicide/2024-10-30/unodc.xlsx.dvc
```diff
@@ -0,0 +1,45 @@
...
...lected from national authorities with the annual United Nations Survey of Crime Trends and Operations of Criminal Justice Systems (UN-CTS). National focal points working in national agencies responsible for statistics on crime and the criminal justice system and nominated by the Permanent Mission to UNODC are responsible for compiling the data from the other relevant agencies before transmitting the UN-CTS to UNODC. Following the submission, UNODC checks for consistency and coherence with other data sources. Data on homicide from public health sources were primarily obtained from the WHO Mortality Database.10 This dataset is a comprehensive collection of mortality data by cause of death, sex, and age group conducted yearly by the WHO with Member States. Deaths coded with Internatioanl Classification of Disease (ICD10) codes X85-Y09 (injuries inflicted by another person with intent to injure or kill), and ICD10 code Y87.1 (sequelae of assault), generally correspond to the definition of intentional homicide. The population data used to calculate homicide rates is sourced from the World Population Prospect, Population Division, United Nations Department of Economic and Social Affairs.
```
antea04, line 8: nitpicky but - I think this 10 was a footnote and can be deleted?

snapshots/homicide/2024-10-30/unodc.xlsx.dvc
```diff
@@ -0,0 +1,45 @@
...
+      1. The killing of a person by another person (objective element).
+
+      2. The intent of the perpetrator to kill or seriously injure the victim (subjective element).
+
+      3. The unlawfulness of the killing (legal element).
+
+      For recording purposes, all killings that meet the criteria listed above are to be considered intentional homicides, irrespective of definitions provided by national legislations or practices.  Killings as a result of terrorist activities are also to be classified as a form of intentional homicide.
+
+      In order to compile consistent time series of total homicides back to 1990, in several cases data from multiple sources were combined to expand the number of available years within a country’s time series. Time series adjustments were performed when a country had two sources coveringdifferent year-ranges, which had very similar trends in an overlapping time period, but where these trends were at different levels.
+
```
antea04, line 21: I don't know if this comes straight from the UNODC, but this sentence is a bit confusing - what time series adjustments are we talking about here? What does "there trends were at different levels" mean?

etl/steps/data/grapher/homicide/2024-10-30/unodc.py
```diff
@@ -0,0 +1,24 @@
...
+
+def run(dest_dir: str) -> None:
+    #
+    # Load garden dataset.
+    ds_garden = paths.load_dataset("unodc")
+
+    table_names = ds_garden.table_names
+    # if your data is in long format, you can use `grapher_helpers.long_to_wide_tables`
+    # to get into wide format
+    tables = []
```
antea04, line 16: I'm surprised we don't have a dataset method that returns all tables at once - should this exist maybe?

etl/steps/data/garden/homicide/2024-10-30/unodc.meta.yml
```diff
@@ -0,0 +1,88 @@
...
+            {definitions.metric} - << category >>
+          description_short: |
+           {definitions.metric} of {definitions.sex} {definitions.age} where the weapon was << category.lower() >>
+          unit: |
+            {definitions.unit}
+    by_relationship_to_perpetrator:
+      variables:
+        value:
+          title: |
+            {definitions.metric} - << category >> - sex: << sex >>
```
antea04, line 66: Is there a reason you are not using {definitions.sex} and {definitions.age} in the title as well? Especially for age I think it would make the titles much nicer.

etl/steps/data/garden/homicide/2024-10-30/unodc.meta.yml
```diff
@@ -0,0 +1,88 @@
...
+
+  metric:
+    <% if unit_of_measurement == "Counts" %>
+    Number of homicides
+    <% elif unit_of_measurement == "Rate per 100,000 population" %>
+    Homicide rate per 100,000 population
+    <%- endif -%>
+  sex:
+    <% if sex == "Total" %> all victimes<% elif sex == "Male" %>male victims<% elif sex == "Female" %>female victims<%- endif -%>
+  age:
```
antea04, line 16: nitpick but: why do some age groups start with "aged..." and some don't?

antea04 (approved):
Hi @spoonerf!

The PR looks good! I've left some very minor comments on the metadata, feel free to ignore anything that's too nitpicky.

I've also had a look through the charts, and there's a few things I noticed:

- [This chart](http://staging-site-unodc-2024/grapher/homicide-rate-1990-vs-2020) on Homicide rates between 1990/ 2010 has lost a lot of entries ([this is the version](https://ourworldindata.org/grapher/homicide-rate-1990-vs-2020) before the update). Not sure what happened there.
- [This chart](http://staging-site-unodc-2024/admin/charts/6980/edit) now starts with the "Turks and Caicos Islands", which squishes the whole chart together because the label is so long :D Maybe there's a way to shorten the label or only show the label of the second highest bar?
- Many charts have closed map brackets for the highest bracket. Do we want to open them (as the numbers can theoretically be higher)?

Also I wonder why chart-diff lets you have pending charts and doesn't complain :D Chart-diff is still magic to me.

Let me know if this makes sense and what you think! Thanks for trusting me with this review! 🥰

snapshots/homicide/2024-10-30/unodc.xlsx.dvc
spoonerf: Good catch!

snapshots/homicide/2024-10-30/unodc.xlsx.dvc
spoonerf: I think what you do is very much the correct way to do things. I think with big organizations like WHO and UN, their data is normally just open by default, so they haven't thought too much about the licensing side of things, so it's hard to know what exactly to put. But I think using `© United Nations`, which they at least mention in their copyright, is better than giving a license they don't mention. Thanks for catching it!

etl/steps/data/grapher/homicide/2024-10-30/unodc.py
spoonerf: Good point, I don't _think_ we do, but I'm not very on top of all the helper functions. Could be a good thing to add in a cooldown :)

etl/steps/data/garden/homicide/2024-10-30/unodc.meta.yml
spoonerf: Ooops yep, that's a mistake!

etl/steps/data/garden/homicide/2024-10-30/unodc.meta.yml
```diff
@@ -0,0 +1,88 @@
...
+  common:
+    presentation:
+      topic_tags:
+        - Homicides
+
+  metric:
+    <% if unit_of_measurement == "Counts" %>
+    Number of homicides
+    <% elif unit_of_measurement == "Rate per 100,000 population" %>
+    Homicide rate per 100,000 population
```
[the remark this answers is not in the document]
spoonerf: Hmm yes I notice this sometimes with Jinja templating, I've not been able to figure it out either. I'll make an issue.

etl/steps/data/garden/homicide/2024-10-30/unodc.meta.yml
spoonerf: Good point, I was being a little lazy in trying to match to the existing variable names in the previous dataset. I can make a nicer version for presentation.title_public.

snapshots/homicide/2024-10-30/unodc.xlsx.dvc
spoonerf: Good point, yep it comes straight from UNODC, but I've adjusted it now to: 

"In several cases data from multiple sources were combined to expand the number of available years within a country’s time series, so that a consistent time series of total homicides back to 1990 could be compiled. Time series adjustments were performed when a country had two sources covering an overlapping time period had similar trends but differening values."

spoonerf:
Thanks for the great and thorough review @antea04!

I think I've addressed all your points on the code now, with the outstanding one on the Jinja template issue. I'll make an issue for that. 

---

> [This chart](http://staging-site-unodc-2024/grapher/homicide-rate-1990-vs-2020) on Homicide rates between 1990/ 2010 has lost a lot of entries ([this is the version](https://ourworldindata.org/grapher/homicide-rate-1990-vs-2020) before the update). Not sure what happened there.

Good catch, I accidentally upgraded the wrong indicator here, it should all be okay now.

---
> [This chart](http://staging-site-unodc-2024/admin/charts/6980/edit) now starts with the "Turks and Caicos Islands", which squishes the whole chart together because the label is so long :D Maybe there's a way to shorten the label or only show the label of the second highest bar?

There is a way to shorten this yep, you need to alter the regions.yml, I've made a PR for Turks and Caicos here #3504 

---
> Many charts have closed map brackets for the highest bracket. Do we want to open them (as the numbers can theoretically be higher)?

Good catch, I've fixed these now. 

---
> Also I wonder why chart-diff lets you have pending charts and doesn't complain :D Chart-diff is still magic to me.

I think chart-diff was temporarily not working correctly but it looks like that is fixed now!

snapshots/homicide/2024-10-30/unodc.xlsx.dvc
antea04: Thanks for the input! Yeah, their licensing is a bit confusing sometimes, so it's good to know what you think 👍

etl/steps/data/garden/homicide/2024-10-30/unodc.meta.yml
antea04: Yeah, that makes sense! Thank you! :)

PR #3495, opened 2024-11-05
:tada: Switch to pyarrow dtypes

Marigold:


Marigold:
@larsyencken you mentioned that `pyarrow[string]` is encoded as categorical under the hood (I don't remember whether you meant in memory or just on disk). Do you know if it's on by default? Using `memory_usage` shows that it doesn't efficiently encode it (assuming it works correctly for pyarrow types) and pyarrow recommends to [convert pd.Categorical to Dictionary](https://arrow.apache.org/docs/python/pandas.html#pandas-arrow-conversion)
```python
s = pd.Series(["a" * 1000] * 1000)
print(s.memory_usage(deep=True))
print(s.astype("string[python]").memory_usage(deep=True))
print(s.astype("string[pyarrow]").memory_usage(deep=True))
print(s.astype("category").memory_usage(deep=True))
```
prints
```
1057128
1057128
1008128
2293
```

larsyencken:
> @larsyencken you mentioned that `pyarrow[string]` is encoded as categorical under the hood (I don't remember whether you meant in memory or just on disk). Do you know if it's on by default? Using `memory_usage` shows that it doesn't efficiently encode it (assuming it works correctly for pyarrow types) and pyarrow recommends to [convert pd.Categorical to Dictionary](https://arrow.apache.org/docs/python/pandas.html#pandas-arrow-conversion)

Sorry, my bad. I confirm that `pyarrow[string]` does not have dictionary encoding in memory.

(I learned now by experimenting that Parquet _does have_ automatic dictionary encoding for all strings in its on-disk format, but the encoding is unpacked again on load.)

That means that if we want to get rid of categoricals, we would have to either stop packing to them, or we would need to unpack on load, by default.

It looks like they are an optimisation really meant for the Feather format. If we standardised on Parquet we could choose to stop packing to them and still have a tiny on-disk format.

Marigold:
Closing in favour of https://github.com/owid/etl/pull/3474.

Issue #3457, opened 2024-10-23
Country harmonizer improvements


spoonerf:
I had a couple of ideas for small improvements to the country harmonize: 

* Add an 'exclude' option (similar to `skip` and `custom`) to the harmonizer CLI, which would automatically add the entity to `excluded_countries.json`
  ![Image](https://github.com/user-attachments/assets/a8182ac2-536b-46a1-841e-0e2f854490e0)

* We shouldn't automatically assign entities that happen to have the same name as OWID continents to those entities, e.g., Africa in a new dataset currently automatically becomes OWID Africa, and it may actually be defined differently. Not a huge deal but it is an extra step to remember.

Marigold:
> We shouldn't automatically assign entities that happen to have the same name as OWID continents to those entities, e.g., Africa in a new dataset currently automatically becomes OWID Africa, and it may actually be defined differently. Not a huge deal but it is an extra step to remember.

@spoonerf are you sure this is happening in the CLI? I tried it and `Africa` became `Africa`.

spoonerf:
Ah, sorry, I didn't mean it literally becomes 'OWID Africa'; I just meant that Africa, as defined in many of our input datasets, will _not_ be made up of the same countries as we define it as (and the other continents, too). 

So it would be good if we could treat the continent entities differently in the country harmonizer so that we always have to check if they are the same or if they should be treated differently, e.g., if Africa should be Africa ([DATASET NAME]).

For example, in [this recent PR ](https://github.com/owid/etl/pull/3455/files#diff-38eab89db235167f00c2107b29e41beb502412bbaaf029aca72255bfecb8233f), the data is only regional (including Africa, Asia, Europe), but their regions aren't defined anywhere so we should be careful not to assume that these regions are the same.

Marigold:
Currently, you can use e.g. `--institution UN` which will ask you about continents and suggest `Continent (UN)` as the first choice. Does that help? We could also explicitly ask for an institution in `harmonize` as the first question.

spoonerf:
Ah cool, I didn't know that existed! I just tried it out and it asks you for continents that don't match our existing ones, e.g. it asked me about 'Americas', but not the others e.g. Africa, Asia, Europe. 

I think the ideal situation would be that it always asks about Asia, Europe, North America, South America and Oceania and that these are not automatically mapped. However, if this is tricky, I'd say it's a pretty minor pain point, so maybe not worth spending too much time on.

PR #3496, opened 2024-11-05
 :sparkles: Make OWID continents ambiguous in harmonizer

Marigold:
Implement suggestion from [this issue](https://github.com/owid/etl/issues/3457#issuecomment-2454993299) by @spoonerf 
> I think the ideal situation would be that it always asks about Asia, Europe, North America, South America and Oceania and that these are not automatically mapped. However, if this is tricky, I'd say it's a pretty minor pain point, so maybe not worth spending too much time on.

The Harmonizer CLI now consistently makes OWID continents ambiguous. It's best used in combination with `--institution WB` to prioritize `Asia (WB)` as the first choice.

@lucasrodes are you ok with this change as a harmonizer's father?

### DIFF
--- a/etl/harmonize.py
+++ b/etl/harmonize.py
@@ -144,6 +144,7 @@ class CountryRegionMapper:
     # known aliases of our canonical geo-regions
     aliases: Dict[str, str]
     valid_names: Set[str]
+    owid_continents: Set[str]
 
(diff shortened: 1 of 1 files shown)


spoonerf (approved):
Perfect, thanks!

PR #3499, opened 2024-11-05
📊 energy: Get Eurostat data on energy prices

pabloarosado:
Main changes:
* Add Eurostat data on gas and electricity prices in Europe.
* Add Ember data on wholesale electricity prices in Europe.
* Add IEA data on fossil fuel subsidies.
* Create an mdim data page of energy prices, which depends on data://grapher steps (and doesn't read from DB).
* Add a helper function to explode views in the multidim module.
* Minor fix in PathFinder.

Review:
@lucasrodes and @Marigold, there's no need to review all steps (but feel free to). However, I'd appreciate your inputs on the following aspects:
* [ ] Please have a look at [the mdim data page](http://staging-site-get-eurostat-data-on-energy-prices/admin/grapher/mdd-energy-prices) and give feedback.
* [ ] Please review the proposed alternative way to create mdim steps without accessing DB in `etl/steps/export/multidim/energy/latest/energy_prices.py`.
* [ ] Please review the changes in `etl/multidim.py`.

Thank you very much!

TODO (for Pablo):
* [x] Clarify the meaning of "euros" and "PPS" in the metadata.
* [x] Complete remaining TODOs in the code.
* [x] Approve current drafts in chart diff (and keep them as drafts for now).

**WARNING**: This work is not yet finished, and will need to wait until Hannah is back. But it's probably better to merge what we have now to avoid future conflicts, and improve drafts and the energy prices mdim data page in a separate PR.


etl/steps/export/multidim/energy/latest/energy_prices.py
```diff
@@ -0,0 +1,49 @@
+from etl import multidim
```
Marigold, line 1: For me personally, showing monthly data as default would make much more sense. Yearly data don't show 2024 and drop in prices in 2023. It took me a while to discover there is monthly frequency.

etl/steps/data/garden/eurostat/2024-11-05/gas_and_electricity_prices.meta.yml
```diff
@@ -0,0 +1,1334 @@
...
+        title: Price
+        unit: euro
+        short_unit: "€"
+        description_short: Energy price in euro.
+      price_pps:
+        title: Price (purchasing power)
+        unit: purchasing power standard
+        short_unit: PPS
+        description_short: Energy price in purchasing power standard.
+  gas_and_electricity_price_components_euro_flat:
```
Marigold, line 91: I'm curious how do you create all combinations in this file. I guess it's done programmatically? If yes, I think it would be useful to share the script or notebook in the same folder in case we have to make changes to it in the future.

etl/multidim.py
```diff
@@ -162,3 +167,74 @@ def fetch_variables_from_table(table: str, engine: Engine) -> pd.DataFrame:
     df_dims = pd.DataFrame(dims, index=df.index)
 
     return df.join(df_dims)
+
+
+def generate_views_for_dimensions(
```
Marigold, line 172: Nice! It'd be good to unify `generate_views_for_dimensions` and `expand_views` into a single function in the future. They both do pretty much the same thing - take a list of (flattened) indicator paths (be it from ETL or DB) and create views from it. `expand_views` has more options for aggregations and `generate_views_for_dimensions` better logging, but they are essentially the same.

etl/multidim.py
```diff
@@ -162,3 +167,74 @@ def fetch_variables_from_table(table: str, engine: Engine) -> pd.DataFrame:
...
+        # Handle missing or multiple table matches.
+        if len(relevant_table) == 0:
+            if warn_on_missing_combinations:
+                log.warning(f"Combination {slug_combination} not found in tables")
+            continue
+        elif len(relevant_table) > 1:
+            log.warning(f"Combination {slug_combination} found in multiple tables: {relevant_table}")
+
+        # Construct the indicator path.
+        indicator_path = (
```
Marigold, line 226: Nit - this won't work for long indicator paths, you'd have to use `trim_long_variable_name` on short_name (`slug_combination`).

Marigold (approved):
Looks good! I'm glad you found an easy way how to make it work for flattened data from ETL. I'd be interested in talking about whether to flatten or not and how to handle combinations in metadata. It seems like there are pros and cons to both and it really depends on the dataset and its complexity.

etl/multidim.py
lucasrodes: Yes, I'd be in favour of harmonizing / consolidating functions

etl/multidim.py
```diff
@@ -162,3 +167,74 @@ def fetch_variables_from_table(table: str, engine: Engine) -> pd.DataFrame:
     df_dims = pd.DataFrame(dims, index=df.index)
 
     return df.join(df_dims)
+
+
+def generate_views_for_dimensions(
+    dimensions, tables, dimensions_order_in_slug=None, additional_fields=None, warn_on_missing_combinations=True
```
lucasrodes, line 173: Maybe add a docstring to clarify what each argument is doing?

From the API alone it is not clear to me what, e.g., `additional_fields` is doing, I had to read the function's code.

My understanding is that you are using it as top-level configuration (i.e. applies to _all_ indicators)?

lucasrodes (approved):
> Please have a look at the mdim data page and give feedback.

MDIM page looks good to me!

> Please review the proposed alternative way to create mdim steps without accessing DB in `etl/steps/export/multidim/energy/latest/energy_prices.py`.

Left some comments!

>  Please review the changes in etl/multidim.py.

Left some comments!

---

Also, could you consolidate your changes with [this docs](https://docs.owid.io/projects/etl/guides/data-work/export-data/?h=multi#exporting-data-to-github)? Note that, even if docs are changing constantly, is good to signal it with a warning. Or some signal for others to know how to even start with MDIMs.

etl/steps/export/multidim/energy/latest/energy_prices.py
pabloarosado: The problem with monthly data is that it only has one option: wholesale electricity prices. On the other hand, we have many more options for annual data. Using monthly as default would look like a very uninformative explorer, where most dropdowns have only one option.

etl/steps/data/garden/eurostat/2024-11-05/gas_and_electricity_prices.meta.yml
pabloarosado: I haven't figured out a nice generic method for this. I usually use `etl.helpers.print_tables_metadata_template`, often tweaking it a bit. It's not a particularly sophisticated method... But once the yaml exists, I have total freedom to tweak things individually.

etl/multidim.py
pabloarosado: Yes, totally! I initially thought of harmonizing both functions. But then I thought they are quite different in what they do. If we are moving towards mdim steps that do not read from DB, maybe there's not much gain in adapting `expand_views`. So, until we decide on that, we can keep these two versions.

etl/multidim.py
pabloarosado: Thanks for pointing that out! Fixed.

PR #3504, opened 2024-11-06
🐝  Update regions.yml to shorten Turks and Caicos Islands

spoonerf:
Hey @pabloarosado an easy PR, but I'm not sure if this will trigger a large rebuild of the ETL, or is that something of the past?

### DIFF
--- a/etl/steps/data/garden/regions/2023-01-01/regions.yml
+++ b/etl/steps/data/garden/regions/2023-01-01/regions.yml
@@ -1528,6 +1528,7 @@
     - "Turks & Caicos"
     - "Turks and Caicos Islands (the)"
     - "Turk Island"
+  short_name: "Turks & Caicos"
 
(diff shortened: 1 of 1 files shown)


pabloarosado (approved):
Looks good! I'm a bit unsure about this change, because, in charts where `short_name` is used, we will see countries with `and` (e.g. "Antigua and Barbuda") and this one with `&`, which may feel a little inconsistent. But I don't have a strong opinion, if you think this is a good choice, then go ahead!

Regarding ETL run, yes, as you can see [in buildkite](https://buildkite.com/our-world-in-data/etl-automated-staging-environment/builds/11659#0193015a-a3bb-4ab6-a382-207d10d80b2f), the ETL will rebuild entirely. We can figure out a way around it, but the easiest would be to just wait until tonight to merge. Would that be an option?

PR #3507, opened 2024-11-06
📊 : Add UNSD subregions to regions dataset

antea04:
Adds the [geographic sub-regions](https://unstats.un.org/unsd/methodology/m49/) by the UN Statistics to our regions data set.  

I'm going to use this data to calculate an indicator for [migration within and between subregions](https://github.com/owid/owid-issues/issues/1722) for Simons upcoming article - but I anticipate that these subregions are useful for other indicators as well. 

Reference to the [discussion in Slack](https://owid.slack.com/archives/C6C0MCXU0/p1730823783319079).


pabloarosado (approved):
Looks good! If you are not in a rush, it would be better to merge this at the end of the day, since it will trigger a long ETL run in production.

PR #3511, opened 2024-11-07
📊 : Calculate median migration distance between origin and destination countries 

antea04:


antea04:
Closing this PR as we are using a different method of calculating distances and have no need for the imported data sets any longer.

PR #3516, opened 2024-11-07
Changes for compatibility with pandas 3.0

jorisvandenbossche:
I ran all the tests locally with the env updated to use a local pandas 2.3.0dev checkout instead of the pinned pandas 2.2.2, and then with using the `PANDAS_COPY_ON_WRITE=1` and `PANDAS_FUTURE_INFER_STRING=1` environment variables set to enable the future behaviour.

The changes in this PR is what I needed to get the tests passing (or mostly, some tests are still failing because of needing a fix upstream in pandas). There are some more deprecation warnings that would need to be addressed, but I mostly focused on failures because of the new string dtype. 

Opening this here so you can see what I did. Not sure if this should actually already get merged because it's not tested on CI with the future version (although if it doesn't break anything with current pandas, that might also do no harm), but in any case you have it when you would update in the future.

### DIFF
--- a/lib/datautils/owid/datautils/dataframes.py
+++ b/lib/datautils/owid/datautils/dataframes.py
@@ -108,7 +108,7 @@ def compare(
     # Compare, column by column, the elements of the two dataframes.
     compared = pd.DataFrame()
     for col in columns:
-        if (df1[col].dtype in (object, "category")) or (df2[col].dtype in (object, "category")):
+        if (df1[col].dtype in (object, "category", "string")) or (df2[col].dtype in (object, "category", "string")):
             # Apply a direct comparison for strings or categories
             compared_row = df1[col].values == df2[col].values
         else:
@@ -461,6 +461,13 @@ def map_series(
         # Replace those nans with their original values, except if they were actually meant to be mapped to nan.
         # For example, if {"bad_value": np.nan} was part of the mapping, do not replace those nans back to "bad_value".
 
+        # if we are setting values from the original series, ensure we have the same dtype
+        try:
+            series_mapped = series_mapped.astype(series.dtype, copy=False)
+        except ValueError:
+            # casting NaNs to integer will fail
+            pass
+
         # Detect values in the mapping that were intended to be mapped to nan.
         values_mapped_to_nan = [
             original_value for original_value, target_value in mapping.items() if pd.isnull(target_value)
@@ -537,7 +544,7 @@ def concatenate(objs: List[pd.DataFrame], **kwargs: Any) -> pd.DataFrame:
         uc = union_categoricals([df[col] for df in objs], ignore_order=ignore_order)
         # Change to union category for all dataframes
         for df in objs:
-            df.loc[:, col] = pd.Categorical(df[col].values, categories=uc.categories)
+            df[col] = pd.Categorical(df[col].values, categories=uc.categories)
 
     with warnings.catch_warnings():
         warnings.simplefilter(action="ignore", category=FutureWarning)
@@ -631,6 +638,14 @@ def combine_two_overlapping_dataframes(
     # Align both dataframes on their common indexes.
     # Give priority to df1 on overlapping values.
     combined, df2 = df1.align(df2)
+
+    new_columns = df2.columns.difference(df1.columns)
+    for col in new_columns:
+        try:
+            combined[col] = combined[col].astype(df2[col].dtype, copy=False)
+        except ValueError:
+            # casting NaNs to integer will fail
+            pass
 
     # Fill missing values in df1 with values from df2.
     combined = combined.fillna(df2)
--- a/lib/datautils/tests/test_dataframes.py
+++ b/lib/datautils/tests/test_dataframes.py
@@ -326,13 +326,14 @@ class TestGroupbyAggregate:
             }
         ).set_index("year")
         df_out["value_03"] = df_out["value_03"].astype(object)
-        assert dataframes.groupby_agg(
+        result = dataframes.groupby_agg(
             df_in,
             ["year"],
             aggregations=None,
             num_allowed_nans=None,
             frac_allowed_nans=None,
-        ).equals(df_out)
+        )
+        assert result.equals(df_out)
 
     def test_default_aggregate_with_num_allowed_nans_zero(self):
         df_in = pd.DataFrame(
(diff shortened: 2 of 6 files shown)


lib/datautils/owid/datautils/dataframes.py
```diff
@@ -461,6 +461,13 @@ def map_series(
         # Replace those nans with their original values, except if they were actually meant to be mapped to nan.
         # For example, if {"bad_value": np.nan} was part of the mapping, do not replace those nans back to "bad_value".
 
+        # if we are setting values from the original series, ensure we have the same dtype
+        try:
+            series_mapped = series_mapped.astype(series.dtype, copy=False)
```
jorisvandenbossche, line 466: This might be specific to a corner case in the tests, but essentially what happened here is that if the mapping does not have anything that is in the original `series`, the resulting `series_mapped` ends up being a Series with all NaNs, and thus gets inferred to have a float dtype.  
And then a bit below you do `series_mapped.loc[missing] = series[missing]`, assigning values from the original series into the result. Starting with pandas 3.0, such a setitem operation will preserve the dtype of the series on the left, i.e. a float here because of all-NaN. And at that point, assigning strings into that will give an error (it gives a warning right now with pandas 2)

lib/datautils/owid/datautils/dataframes.py
```diff
@@ -537,7 +544,7 @@ def concatenate(objs: List[pd.DataFrame], **kwargs: Any) -> pd.DataFrame:
         uc = union_categoricals([df[col] for df in objs], ignore_order=ignore_order)
         # Change to union category for all dataframes
         for df in objs:
-            df.loc[:, col] = pd.Categorical(df[col].values, categories=uc.categories)
+            df[col] = pd.Categorical(df[col].values, categories=uc.categories)
```
jorisvandenbossche, line 547: This is equivalent, except that with `df[col]` you are just setting a new column, while with `df.loc[:, col]` values are being set inplace into the existing column. And that is again a case of where a setitem operation will try to preserve the original dtype when setting values inplace (i.e. not when replacing an entire column. And of course with a `:` slice this is a bit ambiguous, as your intention is to replace the entire column)

lib/datautils/tests/test_dataframes.py
```diff
@@ -326,13 +326,14 @@ def test_default_aggregate_with_some_nans_ignored_different_types_and_more_nans(
...
-        assert dataframes.groupby_agg(
+        result = dataframes.groupby_agg(
             df_in,
             ["year"],
             aggregations=None,
             num_allowed_nans=None,
             frac_allowed_nans=None,
-        ).equals(df_out)
+        )
+        assert result.equals(df_out)
```
jorisvandenbossche, line 336: (this change was just to more easily debug a test failure, i.e. so you can check the resulting dataframe in a debugger, can certainly revert changes like this)

lib/repack/owid/repack/__init__.py
```diff
@@ -62,7 +62,7 @@ def repack_series(s: pd.Series) -> pd.Series:
     if s.dtype.name in ("Int64", "int64", "UInt64", "uint64"):
         return shrink_integer(s)
 
-    if s.dtype.name in ("object", "string", "float64", "Float64"):
+    if s.dtype.name in ("object", "str", "string", "float64", "Float64"):
```
jorisvandenbossche, line 65: The new default string dtype has a name of `"str"`, so adding that here to properly catch such series.

lib/repack/tests/test_repack.py
```diff
@@ -179,7 +179,7 @@ def test_repack_float_object_to_float32():
 
 def test_repack_category():
     s = pd.Series(["a", "b", "c", None])
-    assert s.dtype == np.object_
+    assert s.dtype == np.object_ or s.dtype == "str"
```
jorisvandenbossche, line 182: With pandas 3.0, the above series creation will infer the dtype to be string, and no longer object dtype

lib/datautils/owid/datautils/dataframes.py
```diff
@@ -632,6 +639,14 @@ def combine_two_overlapping_dataframes(
     # Give priority to df1 on overlapping values.
     combined, df2 = df1.align(df2)
 
+    new_columns = df2.columns.difference(df1.columns)
+    for col in new_columns:
+        try:
+            combined[col] = combined[col].astype(df2[col].dtype, copy=False)
```
jorisvandenbossche, line 645: The issue here is that if there is a column in `df2` that is not present in `df1`, it gets introduced in `combined` as an all-NaN column with float dtype (that's what `align` does). 
And then in the next step when doing `combined.fillna(df2)`, it will fill those NaN values with the values of the column in `df2`. But if that column is a string column, pandas will use the "common dtype" for the resulting column, and for a combination of `float64` and `string` dtype, that will be `object` dtype, losing the `string` dtype of the column in `df2`.

As long as strings were stored in a `object` dtype column, this didn't matter that the result was object dtype. But now with a proper string dtype, I think ideally that dtype gets preserved.

lib/datautils/owid/datautils/dataframes.py
Marigold: Thanks, it was always a bit unclear to us what's the difference between the two. We use `.loc[:, col]` because it plays better with our subclassed dataframe class `Table` (type checker is confused by `df[col]` and thinks we're casting it back to DataFrame)

Marigold:
@jorisvandenbossche Amazing—thanks so much, Joris! Meeting you has inspired us to proceed with migrating to pandas nullable types and `string[pyarrow]`. I'll use your PR as a foundation for my branch since our CI only runs for team members.

Just one question: I plan to explicitly use the `string[pyarrow]` type in our repacking library. Would using `string[pyarrow_numpy]` along with `pd.options.future.infer_string = True` offer any advantages?

jorisvandenbossche:
> Would using `string[pyarrow_numpy]` along with `pd.options.future.infer_string = True` offer any advantages?

The main advantage is that it will be the default dtype (starting from pandas 3.0, or with 2.3.0 with the option enabled), so you will need less changes to specify you explicitly want to get `string[pyarrow]`. 
For the rest, behaviour and performance-wise, they should be quite similar (there are only some differences in how NaN vs NA is propagated)

Marigold:
Added those changes in https://github.com/owid/etl/pull/3550 and merged. Thanks Joris!

Issue #3436, opened 2024-10-18
anomalist: Open issues and enhancements


pabloarosado:
### One-liner
Open issues and possible enhancements of Anomalist

### Context & details
See more details in https://github.com/owid/etl/issues/3340

### Open issues

* [ ] (Unlikely to happen) If you filter by a certain indicator and then unselect its dataset, Anomalist fails.
![Image](https://github.com/user-attachments/assets/ff5ff53e-63cb-46fb-b181-8fbe7e516e4f)

* [ ] (Unlikely to happen) If you have already started Anomalist, and want to add a new dataset to the list, if you press "Detect anomalies" nothing happens. The only option is to re-scan all datasets (which takes long).
![Image](https://github.com/user-attachments/assets/96c0a17a-3051-45d6-9660-682d9d5d7308)

* [ ] Decide what to do with filters. Currently, we show indicators, but hide types, which I find confusing. Should we have a checkbox of country and indicator?
* [ ] Add a "hide anomaly" button to each anomaly.
* [ ] Improvements on AI summary (see [conversation](https://owid.slack.com/archives/C07Q7DKTWBC/p1729247282469109?thread_ts=1729239016.641919&cid=C07Q7DKTWBC)):
  * It currently speaks about "indicator 987654" blah blah. We'd need to tweak it to always say the indicator title.
  * The returned info is not particularly insightful: "Indicator blah shows spikes" "indicator blah shows anomalies"... We may need to tweak the prompt a bit to make it more useful.
  * It would be more useful if the AI summary was shown on the side, so the user can read it while also interacting with the filters to visualize the results.
* [ ] Improve the Anomalist workflow. Mojmir's working on letting Anomalist be automatically triggered for any new datasets in a staging server. But the UI is not yet adapted accordingly. Also, clarify what happens if the user, e.g. adds a new dataset to the list.

lucasrodes:
Other minor things to work on:

- **Documentation**: Add instructions in https://docs.owid.io/projects/etl
- **AI/LLM**: 
  - Current summary can be too lengthy. We should make the output simpler.
  - Summaries are not stored. We should store them so we don't need to re-generate them multiple times.
  - We could explore if there is a way to show the summary along with the anomalies. Currently, it lives in a modal and not next to the anomaly list.
[...]

PR #3517, opened 2024-11-08
 :hammer: Speedup anomalist

Marigold:
Speed up loading of UN SDG and GHE datasets. Surprisingly, the bottleneck was loading metadata from dictionary with `dataclasses_json`. This PR replaces it by a custom function that is ~10x faster. This should speed up dataset loading in general, not just for anomalist. It still takes a long time (~1min?) to load these huge datasets (with ~10k variables), but it's manageable when it's run by owidbot.

The new `from_dict` function also raises errors when the type is invalid. This uncovered an inconsistency in our of our old datasets.

It looks like others [have noticed](https://github.com/lidatong/dataclasses-json/issues/228#issuecomment-2355422152) slowness of `dataclasses-json` too. I'll check what makes it so slow on our use case.

Also fixes issues with multidimensional datasets and filters.

Reduce number of GP anomalies - only save anomalies with >= 3 points and p-value < 0.1. That should get rid of useless anomalies like [these from MPI](https://github.com/owid/etl/issues/3436#issuecomment-2464252025).

### DIFF
--- a/apps/anomalist/gp_detector.py
+++ b/apps/anomalist/gp_detector.py
@@ -118,8 +118,8 @@ class AnomalyGaussianProcessOutlier(AnomalyDetector):
             # Get the data for the current entity and variable
             group = df_wide.loc[(entity_name, variable_id)]
 
-            # Skip if the series has only one or fewer data points
-            if isinstance(group, pd.Series) or len(group) <= 1:
+            # Skip if the series has only three or fewer data points
+            if isinstance(group, pd.Series) or len(group) <= 3:
                 continue
 
             # Prepare the input features (X) and target values (y) for Gaussian Process
@@ -160,6 +160,10 @@ class AnomalyGaussianProcessOutlier(AnomalyDetector):
         # Convert z-score into p-value
         df_score_long["p_value"] = 2 * (1 - norm.cdf(np.abs(df_score_long["z"])))
 
+        # Anomalies with p-value < 0.1 are not interesting, drop them. This could be
+        # even stricter
+        df_score_long = df_score_long[df_score_long["p_value"] < 0.1]
+
         # Adjust p-values for multiple testing
         df_score_long["adj_p_value"] = df_score_long.groupby(["entity_name", "variable_id"]).p_value.transform(
             lambda p: multipletests(p, method="fdr_bh")[1]
@@ -167,6 +171,8 @@ class AnomalyGaussianProcessOutlier(AnomalyDetector):
 
         # Final score is 1 - p-value
         df_score_long["anomaly_score"] = 1 - df_score_long["adj_p_value"]
+
+        print(df_score_long.sort_values("adj_p_value"))
 
         return df_score_long.drop(columns=["p_value", "adj_p_value", "z"])
 
--- a/apps/anomalist/gp_detector.py
+++ b/apps/anomalist/gp_detector.py
@@ -172,8 +172,6 @@ class AnomalyGaussianProcessOutlier(AnomalyDetector):
         # Final score is 1 - p-value
         df_score_long["anomaly_score"] = 1 - df_score_long["adj_p_value"]
 
-        print(df_score_long.sort_values("adj_p_value"))
-
         return df_score_long.drop(columns=["p_value", "adj_p_value", "z"])
 
     @staticmethod
--- a/lib/catalog/owid/catalog/processing_log.py
+++ b/lib/catalog/owid/catalog/processing_log.py
@@ -69,6 +69,8 @@ class ProcessingLog(List[LogEntry]):
     __args__ = (LogEntry,)
 
     def __init__(self, entries: List[LogEntry | Dict[str, Any]] = []):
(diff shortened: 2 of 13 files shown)


apps/anomalist/gp_detector.py
```diff
@@ -118,8 +118,8 @@ def get_score_df(self, df: pd.DataFrame, variable_ids: List[int], variable_mappi
             # Get the data for the current entity and variable
             group = df_wide.loc[(entity_name, variable_id)]
 
-            # Skip if the series has only one or fewer data points
-            if isinstance(group, pd.Series) or len(group) <= 1:
+            # Skip if the series has only three or fewer data points
+            if isinstance(group, pd.Series) or len(group) <= 3:
```
pabloarosado, line 122: Good idea, maybe this should be applied to all detectors, not just GP (up to you).

apps/anomalist/gp_detector.py
```diff
@@ -160,6 +160,10 @@ def get_score_df(self, df: pd.DataFrame, variable_ids: List[int], variable_mappi
         # Convert z-score into p-value
         df_score_long["p_value"] = 2 * (1 - norm.cdf(np.abs(df_score_long["z"])))
 
+        # Anomalies with p-value < 0.1 are not interesting, drop them. This could be
+        # even stricter
+        df_score_long = df_score_long[df_score_long["p_value"] < 0.1]
```
pabloarosado, line 165: I'm not sure if this helps much. In the end, we can always filter in the UI. The important thing would be that high scores always mean anomalous points. But if you prefer to hide p>0.1, I don't think it poses any risks of missing out important anomalies.

lib/catalog/owid/catalog/processing_log.py
```diff
@@ -68,6 +68,9 @@ class ProcessingLog(List[LogEntry]):
     # hack for dataclasses_json
     __args__ = (LogEntry,)
 
+    def __init__(self, entries: List[LogEntry | Dict[str, Any]] = []):
+        super().__init__([entry if isinstance(entry, LogEntry) else LogEntry.from_dict(entry) for entry in entries])
```
pabloarosado, line 72: This is a bit obscure to me, but I take your word on this :-D

lib/catalog/owid/catalog/utils.py
```diff
@@ -268,3 +269,55 @@ def hash_any(x: Any) -> int:
...
+        elif dataclasses.is_dataclass(field_type):
+            init_args[field_name] = dataclass_from_dict(field_type, v)  # type: ignore
+        elif isinstance(field_type, type):
+            try:
+                init_args[field_name] = field_type(v)
+            except ValueError as e:
+                # HACK 2024-11-08: we got invalid type into one of the fields, remove it when we fix it
+                # on all staging servers
+                if "invalid literal for int() with base 10: '2020" in str(e):
+                    init_args[field_name] = int(v.replace('"', ""))
```
pabloarosado, line 317: I'm not sure what's going on here. Maybe we should create an issue to remember to fix this properly?

pabloarosado (approved):
Thanks @Marigold, I couldn't follow the logic of some parts of the code (maybe it would be good to add a few comments to clarify the intentions), and I haven't tested it myself, but a speed up of 10x when loading variables? Merge this, please! 🎉

apps/anomalist/gp_detector.py
Marigold: > In the end, we can always filter in the UI

That's right, but it must feel weird for a new user to see GP "anomalies" at the top that are not anomalous at all. I'd keep them if they were useful for "exploration" mode, but I don't think they're useful at all. I'd experiment with filtering them and if we miss them in the future, we can just put them back.

lib/catalog/owid/catalog/processing_log.py
Marigold: It accepts both `LogEntry` and `dict` when initializing. Clarified that in a comment.

lib/catalog/owid/catalog/utils.py
Marigold: It comes [from a typo](https://github.com/owid/etl/pull/3517/files#diff-00e51537a22bb68140652296351986cdc3e1d430d97a27b103c4b83b62eb4280) in one old metadata file. I'll remove it the next time we increase ETL_EPOCH.

Marigold:
After more experimentation, I've found that adding [caching to dataclasses-json](https://github.com/lidatong/dataclasses-json/issues/228#issuecomment-2355422152) would get it almost on par with our custom `from_dict` method. I've also tried https://github.com/cakemanny/fastclasses-json which was fast (again, on par with our custom method), but I was getting tons type hints errors from pylance.

I'll leave a comment about revisiting this and possibly using third-party libraries in the future.

apps/anomalist/gp_detector.py
Marigold: I'll leave this decision up to you. GP is not useful with just 3 points, but other detectors might be.

PR #3519, opened 2024-11-08
:bug: un: aggregate only columns I need in HDR dataset

paarriagadap:
@antea04 Just realized that https://github.com/owid/etl/pull/3425 aggregated all the indicators, regardless of rel_cols. I have fixed that. Take a look to see this is still what you were looking for.


antea04 (approved):
Hi @paarriagadap 

Oops - thanks for spotting this and also taking the initiative to fix it! 🙇‍♀️ 

The code looks good to me, runs without problems and the indicators I wanted are still available as aggregates, so feel free to merge!

PR #3520, opened 2024-11-08
Fix multidimension variable load

pabloarosado:
Loading variable data was failing on multidimensional indicators on [this branch](https://github.com/owid/etl/pull/3466) (more details on [this issue](https://github.com/owid/etl/issues/3436#issuecomment-2464252025)).

@Marigold I have quickly brute-forced a solution, but I'm not sure if it does the intended effect. Please let me know if it doesn't make sense.


Marigold:
@pabloarosado I have something similar plus some performance improvements in [this PR](https://github.com/owid/etl/pull/3517). I'd merge my PR first (since it covers more edge cases found in GHE) and then double check that it works for MPI too.

pabloarosado:
> @pabloarosado I have something similar plus some performance improvements in [this PR](https://github.com/owid/etl/pull/3517). I'd merge my PR first (since it covers more edge cases found in GHE) and then double check that it works for MPI too.

Thanks! Feel free to close this PR if you have already a solution for that bug.

Marigold:
Fixed elsewhere.

PR #3521, opened 2024-11-11
 :hammer: indicator upgrader: use real users

Marigold:
## Problem

When the indicator upgrader updates charts, the edits appear to be made by the last person who manually edited the chart, not by the person who ran the upgrade. This can cause confusion, especially when Slack notifications show updates attributed to users who haven’t interacted with the chart.

This PR addresses the issue by assigning updates to the user who runs the indicator upgrader. However, a drawback is that bulk upgrades (e.g., population or WDI) will be attributed to a single user, even if they are unrelated to the charts.

---

## Implementation

- Added a function to detect the user and their `id` in Streamlit.
- Updated the indicator upgrader to use the detected user when making chart updates (instead of attributing them to the last editor).


pabloarosado (approved):
Looks good, thanks for fixing this!

PR #3525, opened 2024-11-12
 :hammer: Turn on parquet co-generation

Marigold:


danyx23:
This is very nice, @Marigold, thanks! Now I can finally just
```bash
duckdb
> from 'data/garden/energy/2024-06-20/primary_energy_consumption/primary_energy_consumption.parquet' limit 10
```

I don't think we need this on the staging servers for now and we can switch once we are asked for it or a need arises?

If we switch this on in production, will this cause any issues with anything in the existing catalog that might rely on feather files?

danyx23 (approved):

Marigold:
> If we switch this on in production, will this cause any issues with anything in the existing catalog that might rely on feather files?

That's very unlikely. We used to publish both for a long time and never ran into any issues.

I'm not going to rebuild ETL catalog yet, but will wait for nullable types that should be ready soon.

PR #3530, opened 2024-11-12
📊 covid: countries reporting vax data

lucasrodes:


spoonerf (approved):
Looks good, I would just ensure you mention they are vaccines for Covid, at the moment that's not clear. 

In the map you can add your custom label to the tooltip, there's a tickbox at the very bottom of the 'Map' tab in chart editor. Then you won't see 1 and 0 on the hover. 

I'm also slightly curious about why the data stops where it does, is that when all countries had reported at least once, maybe you could make that super clear somewhere.

PR #3535, opened 2024-11-13
📊 antibiotics: WHO Antibiotic usage

spoonerf:
Hey @Marigold, I'm running into a bit of Jinja issue that I can't figure out, would you be able to take a look to see if you can spot the issue.

Quite often with Jinja there's missing spaces around index terms e.g. for the `routeofadministration` term in the titles of variables in [this dataset](http://staging-site-who-amu/admin/datasets/6798). 

I've tried to add spaces in the metadata but still the titles look like this: `Defined daily doses ofparentearally administeredAntivirals for systemic use (ATC J05) - antivirals for treatment of hiv infections, combinations (j05ar)  used`, with no spaces around 'parentearally administered'. Any idea what I'm missing?


Marigold:
It was caused by the dash in `<% if`. `<%- if` trims all whitespaces before the block (or after if you use `-%>`). I've set reasonable defaults so that in most cases you never have to use `-` in ifs (which worked in this case). If you have to use them, they can quickly become very confusing...

spoonerf:
> It was caused by the dash in `<% if`. `<%- if` trims all whitespaces before the block (or after if you use `-%>`). I've set reasonable defaults so that in most cases you never have to use `-` in ifs (which worked in this case). If you have to use them, they can quickly become very confusing...

Okay brilliant, thanks! I'll avoid '-' in the future!

PR #3558, opened 2024-11-18
🔨 update grapher schema to latest version

sophiamersmann:
Sibling PR to https://github.com/owid/owid-grapher/pull/4159 that makes changes to the database and chart config schema.

Changes include:
- Added derived column `chart_configs.chartType`
- For chart configs:
    - Removed field `type` that is replaced by `chartTypes`
        - `type` used to refer to a single chart type, like `"LineChart"`
        - `chartTypes` refers to a list of chart types, like `["LineChart", "SlopeChart"]`
        - For now, the only valid chart type combination is `["LineChart", "SlopeChart"]`. We might add support for more chart type combinations in the future, but that will be a slow process. A lot of words to say: It is usually safe to assume the first element in `chartTypes` is the "type" of the chart
    - Removed field `hasChartTab`. If you want to know whether a Grapher has a chart tab, you can check whether `chart_configs.chartType` is null (if it is, then Grapher has no chart tab)

### DIFF
--- a/apps/wizard/app_pages/map_brackets.py
+++ b/apps/wizard/app_pages/map_brackets.py
@@ -138,7 +138,7 @@ def create_default_chart_config_for_variable(metadata: Dict[str, Any]) -> Dict[s
     """Create a default chart for a variable with id `variable_id`."""
     chart_config = {
         "hasMapTab": True,
-        "hasChartTab": False,
+        "chartTypes": [],
         "tab": "map",
         "map": {
             # "timeTolerance": 0,
--- a/apps/wizard/utils/chart_config.py
+++ b/apps/wizard/utils/chart_config.py
@@ -31,7 +31,7 @@ CONFIG_BASE = {
     "hideLegend": False,
     "tab": "chart",
     "logo": "owid",
-    "$schema": "https://files.ourworldindata.org/schemas/grapher-schema.005.json",
+    "$schema": "https://files.ourworldindata.org/schemas/grapher-schema.006.json",
     "showYearLabels": False,
     "id": 807,
     "selectedFacetStrategy": "none",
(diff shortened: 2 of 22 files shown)


Marigold:
@sophiamersmann happy to help you with ETL side if you want. Tests are currently failing because there are a couple of metadata files using `hasChartTab: false` which is not in the new schema. Since it's just three, it makes sense to fix the actual files rather than version them somehow.

sophiamersmann:
@Marigold, thanks for looking into it. I think I got them all but one of the datasets doesn't compile. Do you mind taking over from here? I now added a description of what changed 👆

I can imagine that some of the data tools also refer to `type` somewhere and would need to be updated? Just a thought though (I really don't have a clue, lol)

Marigold:
Sure, I'll take it from here. Just one thing - https://files.ourworldindata.org/schemas/grapher-schema.006.json returns 403. Is that expected?

sophiamersmann:
The new schema (006) will be uploaded automatically when the PR on the Grapher side is merged, but I could upload it earlier if you want me to?

Marigold:
> The new schema (006) will be uploaded automatically when the PR on the Grapher side is merged, but I could upload it earlier if you want me to?

The dataset is failing, because the schema isn't available at the URL. Which means it should work after we merge the Grapher side! (So don't worry about uploading it)

Marigold (approved):
Looks good! CI should be green after the Grapher part is merged.

sophiamersmann:
@Marigold I just merged the PR on the Grapher side and then did another CI run. Building the staging env fails because there is a config that's apparently not valid (though I can't see why it wouldn't be valid). Is there something that needs to be fixed here?

PR #3566, opened 2024-11-19
:bar_chart: Calculate median migration distance between borders

antea04:
Median migration distance between borders for Simon's upcoming article.

Median distance is calculated by using Natural earth data (https://www.naturalearthdata.com/) and geopandas.


etl/steps/data/garden/migration/2024-11-19/migration_distance.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # Read table from meadow dataset.
+    tb = ds_meadow["migrant_stock_dest_origin"].reset_index()
+
+    # read natural earth data
+    world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))  # type: ignore
+    tiny_countries = gpd.read_file(URL_TINY_CTY)
+
+    # use web mercator projection
+    world = world.to_crs("EPSG:3857")[["name", "geometry"]]
```
spoonerf, line 43: This is really tricky, and map projections are a minefield, but I think the Web Mercator projection might not be the best for calculating distances as there is quite a lot of distortion in it, especially close to the poles. 

ChatGPT says that using EPSG:4326 would be more appropriate, and then the distances can be calculated using geodesic distances. 

It suggests this instead:

> from shapely.ops import nearest_points
> from geopy.distance import geodesic
> import pandas as pd
> 
> # Use WGS84 (EPSG:4326) for geographic distances
> world = world.to_crs("EPSG:4326")[["name", "geometry"]]
> tiny_countries = tiny_countries.to_crs("EPSG:4326")[["admin", "geometry"]]
> tiny_countries = tiny_countries.rename(columns={"admin": "name"})
> 
> # Add tiny countries to world
> world = pd.concat([world, tiny_countries], ignore_index=True)
> 
> # Harmonize country names
> world = geo.harmonize_countries(
>     df=world,
>     country_col="name",
>     countries_file=paths.country_mapping_path,
> )
> 
> # Calculate geodesic distance matrix using borders
> def calculate_distance_matrix(world):
>     # Create an empty distance matrix
>     distance_matrix = pd.DataFrame(index=world["name"], columns=world["name"])
> 
>     for i, row1 in world.iterrows():
>         for j, row2 in world.iterrows():
>             if i == j:
>                 distance_matrix.iloc[i, j] = 0  # Distance to itself
>             else:
>                 # Get the nearest points between two geometries
>                 point1, point2 = nearest_points(row1.geometry, row2.geometry)
> 
>                 # Calculate geodesic distance between the nearest points
>                 distance_matrix.iloc[i, j] = geodesic(
>                     (point1.y, point1.x), (point2.y, point2.x)
>                 ).kilometers
> 
>     return distance_matrix
> 
> distance_matrix = calculate_distance_matrix(world)

etl/steps/data/garden/migration/2024-11-19/migration_distance.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    # Add country names to the distance matrix
+    distance_df = distance_matrix.copy()
+    distance_df.index = world["name"]
+    distance_df.columns = world["name"]
+
+    ## Add distances to migration flows table
+    # Remove "Other" and countries without distance data from country destination or country origin columns
+    cty_no_data = [
```
spoonerf, line 69: These are missing from the UN DESA data?

etl/steps/data/garden/migration/2024-11-19/migration_distance.py
```diff
@@ -0,0 +1,131 @@
...
+    # Load inputs.
+    #
+    # Load meadow dataset.
+    ds_meadow = paths.load_dataset("migrant_stock")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow["migrant_stock_dest_origin"].reset_index()
+
+    # read natural earth data
+    world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))  # type: ignore
```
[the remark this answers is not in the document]
spoonerf: Ah it also says this function is deprecated, so I think definitely best to access directly from NE

etl/steps/data/garden/migration/2024-11-19/migration_distance.py
spoonerf: I've had a quick go at rerunning it with this and I think the numbers are a bit more in the region expected. In the current code, I don't think it's dealing with the spherical nature of the earth, e.g., the value for Wallis and Futuna is around 37,000km, basically the world's circumference!

Using the code here above it drops down to around 2250km, which makes more sense as it looks like most [emigrants move to nearby New Caledonia](https://admin.owid.io/admin/variables/999825).

spoonerf (changes requested):
Super interesting code and data! It mostly looks good, but I think the projection you are using is not appropriate for distances. I've suggested some alternative code which I think gives more reasonable values, but see what you think. Also a minor issue regarding NE data and snapshotting it.

I think the max value you get should be <20000km, as that's about half the circumference of the earth.

etl/steps/data/garden/migration/2024-11-19/migration_distance.py
spoonerf: Here's a decent explanation on the difference between planar and geodesic differences. which I think applies here:
https://pro.arcgis.com/en/pro-app/latest/tool-reference/spatial-analyst/geodesic-versus-planar-distance.htm

etl/steps/data/garden/migration/2024-11-19/migration_distance.py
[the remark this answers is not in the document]
antea04: Fixed - actually I found Lucas had imported the natural earth data set before, so I could simply reuse it :D

etl/steps/data/garden/migration/2024-11-19/migration_distance.py
antea04: These are in the UN DESA but missing from the natural earth data! Since they have very small migration flows I've decided to exclude them instead of trying to hunt down another data set.

spoonerf: Ah interesting, it may be because you were using the low res natural earth data, which has much coarser boundaries and because of this might skip small countries. I'm not sure what version Lucas imported, but it might be different.

spoonerf: But I think it is also totally reasonable to skip them for the reason you give!

spoonerf (approved):
Looks excellent! :)

PR #3570, opened 2024-11-19
wildfires test

veronikasamborska1994:


veronikasamborska1994:
hey @Marigold I am going to merge this just bc things seem to have gone wrong with the automatic update of this data - lots of it was just missing. when I ran it locally it seemed to be completely fine so I am just pushing this for now - no idea how to even replicate it tbh but since you were helping me out with setting up a regular update + might have other similar datasets I thought I'd let you know! could be because there was an issue with accessing the api at the time of the update I don't have any way of throwing an error/warning about this in the snapshot step - if you have a good solution for this that wouldn't mess with the fact that this is updated weekly let me know - otherwise I can just add an exception for when it fails tmrw! 
Screenshot 2024-11-19 at 21 08 55

Marigold:
Whoa, good catch, @veronikasamborska1994. I checked the Git history, and it happened a few days ago—their API returned less data than usual. I remember having some issues with their API in the past.

I've [added an error](https://github.com/owid/etl/commit/261145d2706b84540be41df998eeb4bc7ae63686) to trigger if the downloaded dataframe has fewer rows than the existing snapshot. At least we'll be able to see how often it fails.

Issue #3559, opened 2024-11-18
harmonize_ipython doesn't offer to export mapping of countries if no ambiguous countries are found


antea04:
#### One-liner:
The etl function harmonize_ipython (`from etl.harmonize import harmonize_ipython`) doesn't prompt the user to export its mapping of countries if no ambiguous countries are found.

#### Context:
While it is not strictly necessary to include countries which are already their "harmonized" version in the mapping, the default behaviour of the other harmonizing endpoints (cli, wizard) is to do it regardless. It avoids ambiguity and the etl warning of "these countries have been skipped in the harmonization mapping".  

I would like the ipython harmonization tool to do it as well and offer me this option, even if all countries can be automatically mapped.

#### Assignee:
I can try to do it in my next cooldown/ downtime, or @lucasrodes whenever he has time. It's in no way urgent.

lucasrodes:
Thanks for proposing this, @antea04! It makes total sense.

If I understood you well, the harmonization process would output different mappings depending on whether you use CLI/Wizard or iPython. For the former methods, already standardized countries appear in the mapping (e.g. `"Latvia": "Latvia"`). But for the latter, this one would be omitted. Is this correct?

If so, I agree that all methods should output the same mappings.

If it helps, I did some quick checks, and some points:

The function `harmonize_ipython` makes calls to [run_automatic](https://github.com/owid/etl/blob/master/etl/harmonize.py#L312) and [run_interactive_ipython](https://github.com/owid/etl/blob/master/etl/harmonize.py#L424).

My intuition tells me that `run_interactive_ipython` is fine, and probably the source of the discrepancy (between ipython and cli/wizard) is the logic in `run_automatic`. All of them call `Harmonizer.run_automatic`, but differently:

---

**iPython**
[harmonizer.run_automatic(logging="ipython")](https://github.com/owid/etl/blob/master/etl/harmonize.py#L121)

**CLI**
[harmonizer.run_automatic(logging="shell")](https://github.com/owid/etl/blob/master/etl/harmonize.py#L92)

**Wizard**
[harmonizer.run_automatic()](https://github.com/owid/etl/blob/master/apps/wizard/app_pages/harmonizer.py#L191)
(this is equivalent to `logging=None`)

---


So, I think there's something with `run_automatic` when `logging="ipython"`. I quickly checked its code, but couldn't find anything obvious there.

I'd suggest adding `print` statements and keeping track of the value of `self.mapping` as you run the different steps.

[...]

PR #3577, opened 2024-11-20
🐛 harmonize: Fix harmonize ipython when used on already harmonized datasets

pabloarosado:
Fixes https://github.com/owid/etl/issues/3559

I encountered the same issue and patched it quickly, I haven't looked thoroughly into it, but it worked for my case.

### DIFF
--- a/etl/harmonize.py
+++ b/etl/harmonize.py
@@ -120,11 +120,15 @@ def harmonize_ipython(
     # Run automatic harmonization
     harmonizer.run_automatic(logging="ipython")
 
-    # Need user input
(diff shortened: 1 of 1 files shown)


lucasrodes:
@antea04 when you have time, could you try it out and see if it solves the issue for you, too?

pabloarosado:
Thanks @lucasrodes, I'm not sure if the current behavior is what we'd want, but at least now `harmonize_ipython` can be used when all country names are already harmonized. We can revisit it later and tweak it again in a separate PR.

PR #3581, opened 2024-11-20
:bar_chart: include un desa migrant stock flows in grapher

antea04:
adding [un desa migrant stock](https://www.un.org/development/desa/pd/content/international-migrant-stock) [destination and origin]. 

Simon needs this data for his upcoming article on migration (distances), also I think it's nice to have this data for any DIs etc - even if we do not change the Migration Flows explorer to being etl based immediately.


etl/steps/data/grapher/migration/2024-11-20/migrant_stock_flows.py
```diff
@@ -0,0 +1,83 @@
...
+    # Load inputs.
+    #
+    # Load garden dataset.
+    ds_garden = paths.load_dataset("migrant_stock")
+
+    # Read table from garden dataset.
+    tb = ds_garden.read("migrant_stock_dest_origin")
+
+    # grapher cannot handle this many indicators, so we remove male and female columns
+    tb = tb.drop(columns=["migrants_female", "migrants_male"])
```
spoonerf, line 23: Ooh really! That's good to know. I wonder if it would work if they were in separate tables... but I don't think we need to try that out now for this!

spoonerf (approved):
Very nice!

PR #3582, opened 2024-11-21
📊 emissions: Update emissions and energy datasets

pabloarosado:
* Update GCB data after their latest minor update (they produced some corrected files).
* Improve sanity checks after latest changes.
* Refactor emissions steps to create new release of the CO2 dataset.
* Create an export step to push all files to github.
* Create an export step to push all files to digital ocean.
* Remove all code from the co2-data repos (on [a separate PR](https://github.com/owid/co2-data/pull/48)).
* Minor fix in Anomalist.

~TODO: Wait for confirmation from GCP before doing chart diff. They may update the files in their website.~
I decided to merge this to avoid blocking the update for so long. Once they update the links on their website, I'll create a small PR updating the snapshot files (and I'll check that there were no major changes).

@Marigold no need to review all files, but what do you think about the `export` steps? The idea with them is to be able to fully eliminate the code in the co2-data repos, so that all pipeline comes from the ETL. Thanks!


etl/steps/export/s3/co2_data/latest/owid_co2.py
```diff
@@ -0,0 +1,125 @@
...
+from structlog import get_logger
+from tqdm.auto import tqdm
+
+from etl.helpers import PathFinder
+
+# Initialize logger.
+log = get_logger()
+
+# Define S3 base URL.
+S3_URL = "https://nyc3.digitaloceanspaces.com"
```
Marigold, line 35: Our preferred S3 storage is now R2 on Cloudflare, though we don't have a `owid-public` bucket there yet. I could set it up there & add proper permissions if we wanted to.

Although on second though, since we have all our stuff in DigitalOcean, it'd make more sense to migrate everything at once at some point in the future.

etl/steps/export/s3/co2_data/latest/owid_co2.py
```diff
@@ -0,0 +1,125 @@
...
+        # Initialise S3 client.
+        s3 = S3()
+        for file_name in tqdm(["owid-co2-data.csv", "owid-co2-data.xlsx", "owid-co2-data.json"]):
+            # Path to local file.
+            local_file = temp_dir_path / file_name
+            # Path (within bucket) to S3 file.
+            s3_file = Path("data/co2") / file_name
+            tqdm.write(f"Uploading file {local_file} to S3 bucket {S3_BUCKET_NAME} as {s3_file}.")
+            # Upload and make public each of the files.
+            s3.upload_to_s3(
```
Marigold, line 121: I'm thinking about what should be the default automation for export steps. Should we run them all every time, just like the rest of the ETL? It would only work if proper env variables are set in production.

We could also do it by creating a script similar to `update-monkeypox.sh`. Or just keep doing it manually.

Marigold (approved):
These export steps look great! I'm glad we're moving everything to a single place.

etl/steps/export/s3/co2_data/latest/owid_co2.py
pabloarosado: I leave this decision up to you. Please let me know if you want me to change anything, or feel free to change it yourself, thanks.

PR #3591, opened 2024-11-22
🎉 wizard: Create GIF/video from chart URL

pabloarosado:
I've always thought it would be convenient to have an easy way to generate GIFs/Videos from a chart. This PR creates:
* A CLI tool to generate a GIF/mp4 file from a given chart URL.
* A wizard page for that tool.

NOTES:
* If you load a chart that uses population, the range of years includes the entire range of years in the data, instead of those in the chart. Maybe it would be better to get the range of years from DB instead of getting the chart.metadata.json.
* The way we get the PNGs may be a bit aggressive (using the API in parallel), maybe there is a better way to do this.
* It would be good to have an option to download GIF in social media format. But I couldn't figure out how.
* Currently, it generates the PNG images and the GIF in a folder in Downloads. This is convenient, to avoid downloading many images. However, when run on master, this will create unnecessary files, so maybe it would be better to have a temporary folder. Not sure what the best solution is.

Example use of CLI:
```
python apps/chart_animation/cli.py https://ourworldindata.org/grapher/per-capita-energy-use
```


apps/wizard/app_pages/chart_to_gif.py
```diff
@@ -0,0 +1,150 @@
...
+        st.session_state.image_paths = get_images_from_chart_url(
+            chart_url=chart_url,
+            png_folder=st.session_state.images_folder,
+            tab=tab,
+            years=years,
+            year_range_open=year_range_open,
+            max_workers=None,
+            max_num_years=100,
+        )
+        st.session_state.show_gif_settings = True
```
lucasrodes, line 103: As a user, I'm a bit confused by this step. Maybe this can just run after submitting step 1?

I think the user can just grab the link, including its query params, from admin and that's it.

If there really the need to have the options "year range open", "year range" etc. I'd suggest just merging these options with step 3.

Less steps = simpler

lucasrodes (commented):
Hi Pablo! Thanks for doing this, I'm interested in seeing where this brings us.

I agree that being able to export social-media ready charts would be a win, specially since I think these GIFs/animated content would be valuable in social media platforms.

As I've already mentioned you, I've renamed the app to something a bit more general in case we add support for non-gif exports too. That'd be very cool!

I've added some suggestions for changes that you can introduce to make this app a bit friendlier to the users.

Some other general comments:
- I recommend you take a look at method `st_horizontal` (in `apps.wizard.utils`), which allows you to display streamlit widgets in one line. This can be helpful for steps 2/3, where there's lots of options that take lots of vertical space, IMO. You could squeeze them a bit. E.g. I think that one line for a small checkbox is suboptimal.
- I think you can get rid of step 2. Most of its logic can just run after after submitting the char URL in step 1 (it's fast enough I think). Some of the other options can be in step 3 (eventually renamed as step 2).

Also, you may want to look at [st.form](https://docs.streamlit.io/develop/api-reference/execution-flow/st.form) which might be useful to group different widgets. Alternatively, maybe you can use `st.container(border=True)` to visually group widgets.


Once again, thanks for your work!

apps/wizard/app_pages/chart_to_gif.py
```diff
@@ -0,0 +1,150 @@
...
+    year_range_open = st.checkbox(
+        "Year range open",
+        value=True,
+        help="Only relevant for the chart view. If checked, the year range will be open. Uncheck if you want to generate a sequence of bar charts.",
+    )
+
+    # Button to generate images.
+    # TODO: It's convenient to keep images in the folder, to avoid downloading them again.
+    #  However, we should detect if the images are already there and have a button to delete them (in case the settings for image generation change).
+    if st.button("Get images"):
```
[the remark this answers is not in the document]
pabloarosado: I added a button. We may accumulate files there after some time, but it's probably not a big deal. We can rethink it in the future.

apps/wizard/app_pages/chart_to_gif.py
```diff
@@ -0,0 +1,150 @@
...
+            duration=duration,
+            loops=loop_count,
+            remove_duplicate_frames=remove_duplicates,
+            repetitions_last_frame=repetitions_last_frame,
+            duration_of_full_gif=duration_of == "Entire GIF",
+        )
+
+    # GIF preview.
+    if st.session_state.gif_file.exists():
+        st.image(str(st.session_state.gif_file), caption="Preview of the Generated GIF", use_container_width=True)
```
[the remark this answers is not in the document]
pabloarosado: Good point! I removed the download button and explained how to download it in a st.info message.

apps/wizard/app_pages/chart_to_gif.py
pabloarosado: Yes, this could be simplified a bit, but there are some complications. For example, the grapher URL doesn't tell you what the default chart view is (map or chart), or the range of years. We could get all this from DB, but I'll keep it as-is for now, we can improve it in a separate PR.

lucasrodes: in that case, you can just move all these parameters to step 3 (to be renamed 2), and just have two steps.

1) define the chart slug / URL
2) generate gif (by getting images, etc.)

pabloarosado: I think it's good to keep "Get images" explicit. Otherwise you may attempt to get all images automatically from a chart that has many years. It makes sense for now to keep 3 steps.

lucasrodes: mmh, I see. FWIW, it ran quite fast for me.

You could maybe set a threshold, and show a year-filtering option IF the range of years is above X. Otherwise, you could run it without worrying.

I think it'd make the experience smoother

lucasrodes: @pabloarosado unresolving just to make it visible again.

pabloarosado: Now it's just two steps (I've refactored things a bit), what do you think?

lucasrodes:
Hey @pabloarosado! I've added a couple of cosmetic changes, including:

- Changed radio buttons to be `horizontal=True`. It makes better use of vertical space.
- I've grouped widgets in containers, adding a border for better visual comprehension.
- I've moved the map-chart selection widgets to the last "step". It felt more natural to have it there.
- I've moved the 'get images' button to the bottom. It was a bit odd to have it in the middle of the flow, but maybe I'm missing something.
I've changed the button to simply say "Animate," which I think is less confusing. I've left the help message as-is.
- I've improved the error message on 'too many years' to help the user take action. It is good to briefly point to the users to what they can do to solve the raised error.
- I've changed an info message to toast. When you want to show a temporary information message (e.g. the one for 'images deleted') I think `st.toast` is more appropriate.
- I've changed the chart-type radio button to `st.segmented_control`; I think it looks nicer. Same for Video/GIF

lucasrodes:
There is a minor bug. Recreate it as:

1) Add a chart URL and follow all steps. Choose "map" as the chart type
2) GIF is successfully created
3) Then change the map type to "chart".

Nothing seems to happen then. Even after clicking 'animate'. The output remains as the original one, the map.

Related to this bug, I think we should have some flow control for the 'animate' button. Now, whenever you click on something in the third container, it will just run and generate the output. I think we should wait for the user to click on 'animate' right?

PR #3592, opened 2024-11-22
📊 war: ucdp check

lucasrodes:


lucasrodes:
With a first look, I realized that I had only imported events corresponding to August 2024. I hadn't realized that the export by UCDP only included monthly exports.

UCDP clarifies this in [their codebook](https://ucdp.uu.se/downloads/candidateged/ucdp-candidate-codebook1.3.pdf):

image

I'll proceed to add the currently latest available files:

- January - September 2024
- October 2024

image

Issue #3542, opened 2024-11-14
dependencies: support for Python 3.12, 3.13


lucasrodes:
We are currently using Python 3.10, which was released in 2021, in production and support up to Python 3.11 (see project).

We should consider upgrading to Python>3.12.

**Reasons**

- There are some performance improvements in Python 3.12
- There is no longer active support for Python 3.10, and 3.11 (no new features, no bug fixes)
- Security support for Python 3.10 ends on 31 Oct 2025. We should attempt to upgrade before that, and ensure there is no problem with dependencies.

**Recommended Approach:**

- Assess Compatibility: Review your project's dependencies to ensure they are compatible with the target Python version.
- Testing: Implement thorough testing in a staging environment to identify and resolve any issues before deploying to production.

pabloarosado:
@lucasrodes are there any new features that would make this upgrade particularly useful?
We use 3.10 in production.

lucasrodes:
@pabloarosado 
> @lucasrodes are there any new features that would make this upgrade particularly useful?

There is not really a specific feature that I am eager to try. I know that there have been significant implementation improvements in Python 3.11 and 3.12, which could already be a reason.

But my main reason to upgrade to, at least, Python 3.11/3.12 is that Python 3.10, which was released around October 2021, has transitioned from active support to security-only support (no bug fixes/features are being added) [reference](https://peps.python.org/pep-0619/#lifespan). In addition, transitioning to newer versions should make future transitions easier.


> We could let pyproject open the range to include up to 3.14. If all dependency installations work, great. Otherwise, we close the range (and not spend more than 1h on it).

I think it is worth exploring if we can at least support Python 3.11. Little transitions can make future transitions smoother.


In general, I think it is a good practice to keep Python somehow up to date. It doesn't necessarily need to be in sync with the latest release (3.13), but -1 or -2 versions. We've left ~3 years for the dependencies to adapt their code to be compatible with python 3.11 and
beyond.

PR #3593, opened 2024-11-22
 :honeybee: Bump python version

Marigold:
Partly fixes https://github.com/owid/etl/issues/3542.

Increase range of python versions from `>=3.9,<3.12` to `>=3.9,<3.13` (include 3.12). I've tried 3.13 too, but `torch` didn't have a distribution for 3.13, yet (the problem could be pinning torch to 2.2.2).

I've also discovered bug that was causing all unit tests run against 3.10 🤦 (thank you Makefile!). Luckily, only `owid-catalog` with 3.9 was affected.

The simplest way to upgrade your python version is to run `PYTHON_VERSION=3.12 make .venv`.

## TODO
- [x] Add 3.12 to all buildkite jobs

## TODO after merging
- [x] Delete `soft_fail` from all buildkite jobs
- [x] Let others know they can upgrade


lucasrodes (approved):
lgtm! thanks, Mojmir

Issue #3584, opened 2024-11-21
Export step fails to create new branch


pabloarosado:
I was trying to execute the `export://github/co2_data/latest/owid_co2` step, by running:
```
CO2_BRANCH=update-gcb-data etlr github/co2_data --export
```
However, I encountered a few issues:
* `OWIDBOT_ACCESS_TOKEN` needed to be defined in `.env`. This was easy to fix, but I'm wondering if this was necessary. I suppose it should rather be using our `GITHUB_TOKEN`.
* I expected it would create a new branch, but instead it failed because the branch was not found. So I manually created the branch, rerun the export step, and then it worked.

Ideally, this step should:
* Not require the `CO2_BRANCH` variable. Instead, it would create a branch with the same name as the current ETL branch (and, if it already exists, it should fail, for safety).
* Create a branch and a PR.

Marigold:
Fixed by https://github.com/owid/etl/pull/3605

PR #3605, opened 2024-11-25
 :sparkles: Automatically create a branch if it does not exist

Marigold:
Implements https://github.com/owid/etl/issues/3584

- Running `etlr github/co2_data --export` creates a branch with the same name as working ETL branch and commits updated files there.
- If the branch exists, push a new commit. (@pabloarosado's suggestion was to fail, but I think it's pretty safe to add a commit given that we don't overwrite existing ones and we create a PR at the end)
- Prefer `GITHUB_TOKEN` over `OWIDBOT_ACCESS_TOKEN` (in the future, we should clean it up and use only a single env)
- Show a link to create a PR from our branch

### DIFF
--- a/apps/owidbot/cli.py
+++ b/apps/owidbot/cli.py
@@ -9,7 +9,7 @@ from rich import print
 from rich_click.rich_command import RichCommand
 
 from apps.owidbot import anomalist, chart_diff, data_diff, grapher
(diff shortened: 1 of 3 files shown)


pabloarosado (approved):
Thanks, this will be helpful on the next update!

PR #3626, opened 2024-11-26
✨ chart animation: add square-format support

lucasrodes:
Builds on https://github.com/owid/etl/pull/3627/files

- Add square format support.
- Change use of `st.container` to `st.expander`. This way the user can collapse these.
- Improve presentation of output settings:
  - Add help message to most widgets.
  - Add dedicated section for "frame settings"
  - Change duration of last frame to be in `ms` (instead of number of frames)
  - Change "Remove duplicate frames" to "Allow duplicate frames". By default this is not checked.
  - Change label "Year range open" to "Show only current year". I found the previous label a bit confusing.
- Other changes to improve script readability

### DIFF
--- a/apps/wizard/app_pages/chart_animation.py
+++ b/apps/wizard/app_pages/chart_animation.py
@@ -162,9 +162,9 @@ if st.session_state.chart_animation_show_image_settings:
         # If chart, show option to just show single year
         if tab == "chart":
             year_range_open = not st.toggle(
-                "Show only current year",
+                "Show single year",
                 value=not query_parameters["year_range_open"],
-                help="Only relevant for the chart view. If checked, the animated chart will display a sequence of bar charts with current year.",
+                help="Only relevant for the chart view. If checked, the animated chart will only display a single year per frame. For LineCharts, this means a sequence of bar charts. For ScatterCharts, this means a sequence of bubbles (and not vectors).",
             )
         else:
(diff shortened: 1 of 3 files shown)


apps/wizard/app_pages/chart_animation.py
```diff
@@ -62,103 +74,141 @@
...
         )
 
+        # If chart, show option to just show single year
         if tab == "chart":
-            year_range_open = st.toggle(
-                "Year range open",
-                value=query_parameters["year_range_open"],
-                help="Only relevant for the chart view. If checked, the year range will be open. Uncheck if you want to generate a sequence of bar charts.",
+            year_range_open = not st.toggle(
+                "Show only current year",
```
pabloarosado, line 165: "Show only current year" is not exactly the right explanation. For a line chart, it means showing just one year at a time. But for a scatter chart, it means showing bubbles instead of lines. That's way I called it "year range open/close". Maybe you can think of a better name, but "only current year" is a bit misleading.

pabloarosado (approved):
Looks good (except for a minor change in the name of a parameter). Thanks!

apps/wizard/app_pages/chart_animation.py
lucasrodes: Changing it to "show single year" instead. Improved help message.

PR #3628, opened 2024-11-26
 :tada: Indicator semantic search

Marigold:
Prototype for indicator semantic search. This was originally motivated as a proof of concept that `EmbeddingsModel` can handle all indicators in memory, and we don't have to resort to more complex infrastructure (like [faiss](https://github.com/facebookresearch/faiss) for in-memory similarity search or a vector database like [milvus](https://milvus.io/)). This turned out to be fine since similarity search takes an acceptable <1s.

Whether **semantic search** over indicators is that useful is questionable. It seems that the lowest hanging fruit lies in [improving the existing search](https://github.com/owid/etl/issues/3710) in Admin. This is why I've decided to park it for now and didn't spend too much time making it pretty and really accurate. Still, it's a good showcase of what a semantic search has to offer and is worth merging in my opinion.

Embeddings are created from indicator title and description (concatenated). It works decently, but there are some issues (see problematic queries at the end). There are a couple of knobs to make it prettier (well, the UI is still crap... feedback would be much appreciated!). I plan to create a demo and let others know about it in January.

## How to review

Running this locally would have to calculate embeddings for all indicators, which could be time-consuming. I recommend to check out the [app on staging server](http://staging-site-indicator-search/etl/wizard/indicator_search) and try a few queries.

## Problematic queries
- Query "beer" returns "Coffee" as the most similar indicator. That's because indicators about beer are often long, especially when concatenated with description, which has tons of irrelevant information. This could be possibly fixed by "summarizing" title & description into a "fixed" length representation.


lucasrodes:
I had to merge with master & solve couple of complaints from unit tests.

Other than that, I tried running the `Indicator search` page locally, and it is taking a while (~1hour) to generate embeddings:

```
Batches:   3%|█                                   | 168/5620 [01:46<1:05:10,  1.39it/s]
```

(from terminal)

Is this expected?

Marigold:
Yes, it's expected. It takes a while to calculate embeddings for all indicators (a couple of hundred thousands). It's better to test it on staging unless you want to develop it further.

lucasrodes (approved):
lgtm! let's try it and see how it behaves

PR #3631, opened 2024-11-27
 :sparkles: Refactor & speedup insights search

Marigold:
Refactor and speed up insights search. Its functionality **isn't changing**. [Link](http://staging-site-insight-search-cache/etl/wizard/insight_search) to insights search on staging.

## Refactoring

- Move it into its own folder and split into data & app modules
- Use `Insight` object instead of dicts
- Improve `url_persist`

## Performance

Instead of calculating embeddings on every launch, save them to disk and add missing ones on every load. This makes initial loading almost instant and also speeds up search.


lucasrodes (approved):
LGTM, thanks Mojmir! Nice idea abstracting the logic to `Doc` class!!

PR #3653, opened 2024-11-28
🐝 ghe: getting spaces in jinja

spoonerf:
There were missing spaces in many of the variable titles and description_shorts, this fixes that. 

I'm a bit worried that changing the titles is going to dissociate the indicators from their charts, but it looks like it will be okay on staging, do you think it will be okay? 

/schedule 2024-11-28T20:00:00.000Z


lucasrodes (approved):
seems to be rendering just fine!

spoonerf:
Thanks! Let's try out the scheduling for this :)

Issue #3654, opened 2024-11-28
feature: check and remove unwanted whitespaces and linebreaks in metadata


lucasrodes:
It's often the case that we introduce unwanted linebreaks or whitespace in the metadata. These are sometimes rendered in the FASTT or the metadata of an indicator.

## Current workflow
At the moment, to make sure that I do not introduce unwanted characters, I do the following:

- Work on metadata, data steps.
- After importing to the database, I go to a sample indicator's data page preview.
- For instance, I might see the following chart, which has an unwanted linebreak before the period of the subtitle sentence:
    ![Image](https://github.com/user-attachments/assets/48014cb6-fd9a-419f-980d-76a8a4ccedc9)
- Then, I'll open the browser inspector, and in the Network tab, look for the URL with the indicator metadata shown. I'll open it and explore to see if there are any other unwanted characters. 
	```json
	{
	    "id": 971388,
	    "name": "Deaths from leukaemia among females aged 0-4 year olds\n",
	    "unit": "deaths",
	    "createdAt": "2024-08-07T11:04:57.000Z",
	    "updatedAt": "2024-11-28T10:02:03.000Z",
	    "coverage": "",
	    "timespan": "2000-2021",
	    "datasetId": 6662,
	    "shortUnit": "",
	    "columnOrder": 0,
	    "shortName": "death_count__age_group_years0_4__sex_female__cause_leukaemia",
	    "catalogPath": "grapher/who/2024-07-30/ghe/ghe#death_count__age_group_years0_4__sex_female__cause_leukaemia",
	    "dimensions": {...},
	    "descriptionShort": "Estimated number of deaths from leukaemia among females  aged 0-4 year olds\n.\n",
	...
	}
	```
- We observe that not only is there a line break (`\n`) before the period in `descriptionShort`, but also:
  - Linebreaks: One at the end of `name` and `descriptionShort`.
  - Whitespaces: double white space in `descriptionShort`, before "aged 0-4..."

I could also check the metadata JSON file in `data/` folder, but it contains all the Jinja syntax, and is hard to read:

```
"death_count": {
[...]

PR #3657, opened 2024-11-29
:sparkles: Jinja whitespaces and newlines

Marigold:
A couple of improvements to alleviate some pain from https://github.com/owid/etl/issues/3654.

- Added instructions how to preview metadata with dimensions in a notebook to avoid long build-wait-check cycles
- Strip whitespaces from Jinja blocks
- Improve ETL docs

Note that this is still not perfect and doesn't solve `\n.` problem from the issue (we still need to use `<%- elif` to get rid of it). Ideally, we should never have to use `-` and have the Jinja templates as intuitive as possible. To do this properly, we should:
1. Replace `dynamic-yaml` by something simpler and unify saving & loading of metadata files (while keeping it **fast**, there were tons of performance optimizations)
2. Move `jinja` functionality to `owid-catalog` and add method `VariableMeta.render_jinja(dim_dict={"..."})`
3. Add validation for double whitespaces, newlines, `\n.`, etc.

### DIFF
--- a/docs/architecture/metadata/structuring-yaml.md
+++ b/docs/architecture/metadata/structuring-yaml.md
@@ -233,4 +233,41 @@ tables:
           {definitions.conflict_type_estimate}
 ```
 
-Be cautious with line breaks and trailing whitespace when utilizing templates. Despite using good defaults, you might end up experimenting a lot to get the desired result.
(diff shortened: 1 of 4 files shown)


lucasrodes (approved):
This LGTM!

I think this is an improvement of the current status quo. We can continue thinking on how to make Jinja friendlier, ultimately. Maybe you can create an issue or add the future work you suggested there.

Thanks for improving the docs, too!

PR #3667, opened 2024-12-02
✨ Better filtering of data & metadata in chart-diff

Marigold:
- Fixed a bug that displayed `Config change` for unrelated charts.
- Filtered data and metadata changes by checking which files were modified on the staging server.

I intended to use the function `get_datasets_from_version_tracker`, but calling `vt.steps_df` was too slow. Instead, I added a new function, `get_all_changed_catalog_paths`, which identifies the paths of changed steps along with their downstream dependencies. Let me know if a faster function for this purpose already exists.

Parsing catalog paths over and over again with `.split("/")` is tedious and prone to errors. I wanted to create a special object `StepPath` in this PR https://github.com/owid/etl/pull/3165, but wasn't fully satisfied with it and didn't end up merging. It'd be nice to refactor it all at one point.

### DIFF
--- a/apps/wizard/utils/io.py
+++ b/apps/wizard/utils/io.py
@@ -14,7 +14,7 @@ import etl.grapher_model as gm
 from apps.wizard.utils.cached import get_datasets_from_version_tracker
 from etl.git_helpers import get_changed_files
 from etl.grapher_io import get_all_datasets
-from etl.paths import BASE_DIR, STEP_DIR
+from etl.paths import BASE_DIR, SNAPSHOTS_DIR, STEP_DIR
 from etl.steps import filter_to_subgraph, load_dag
 
 # Initialize logger.
@@ -73,8 +73,10 @@ def get_changed_steps(files_changed: Dict[str, Dict[str, str]]) -> List[str]:
             # Skip deleted files.
             continue
 
-        # Identify grapher data steps, and ignore the rest.
-        if file_path.startswith(STEP_DIR.relative_to(BASE_DIR).as_posix()) and file_path.endswith(".py"):
+        # Identify potential recipes for data steps
+        if file_path.startswith(
+            (STEP_DIR.relative_to(BASE_DIR).as_posix(), SNAPSHOTS_DIR.relative_to(BASE_DIR).as_posix())
+        ):
             changed_steps.append(file_path)
         else:
             continue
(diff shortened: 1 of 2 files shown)


apps/wizard/utils/io.py
```diff
@@ -75,13 +75,21 @@ def get_changed_grapher_steps(files_changed: Dict[str, Dict[str, str]]) -> List[
 
         # Identify grapher data steps, and ignore the rest.
         if file_path.startswith(STEP_DIR.relative_to(BASE_DIR).as_posix()) and file_path.endswith(".py"):
```
pabloarosado, line 77: (Please change the comment above, which is no longer accurate).
Wouldn't we want to catch snapshot changes as well? Maybe:
```
if file_path.startswith((STEP_DIR.relative_to(BASE_DIR).as_posix(), SNAPSHOTS_DIR.relative_to(BASE_DIR).as_posix())) and file_path.endswith((".py", ".dvc")):
```

apps/wizard/utils/io.py
```diff
@@ -137,3 +145,27 @@ def get_new_grapher_datasets_and_their_previous_versions(session: Session) -> Di
...
 
     return new_datasets
+
+
+def get_all_changed_catalog_paths(files_changed: Dict[str, Dict[str, str]]) -> List[str]:
+    """Get all changed steps and their downstream dependencies."""
+    dataset_catalog_paths = []
+
+    # Get catalog paths of all datasets with changed files.
+    for step_path in get_changed_steps(files_changed):
```
pabloarosado, line 155: This would not catch snapshots. To do that, you could do, e.g. `if step_path.startswith("snapshots/")` or something like that.

apps/wizard/utils/io.py
```diff
@@ -137,3 +145,27 @@ def get_new_grapher_datasets_and_their_previous_versions(session: Session) -> Di
...
+        abs_step_path = BASE_DIR / Path(step_path)
+        try:
+            # TODO: use StepPath from https://github.com/owid/etl/pull/3165 to refactor this
+            dataset_catalog_paths.append(abs_step_path.relative_to(STEP_DIR / "data").with_suffix("").as_posix())
+        except ValueError:
+            continue
+
+    # Add all downstream dependencies of those datasets.
+    DAG = load_dag()
+    dag_steps = list(filter_to_subgraph(DAG, dataset_catalog_paths, downstream=True).keys())
```
pabloarosado, line 165: This is OK, as it filters down the DAG a little bit. But using VersionTracker.steps_df would be much more precise. You could do:
```
steps_df[(steps_df["step"].isin([...])]["all_active_usages"]
```
And that would give you only the steps that are affected by the changed files. That would be ultimately what we need. But I understand that loading `steps_df` is very slow.

pabloarosado (approved):
Thanks for fixing this!
I left a few suggestions to also be able to catch changes in snapshots.
And to be able to get the list of affected steps, I don't think we have any alternative but using vt.steps_df. Maybe we could invest some time to optimize that process (since it's used in multiple places).
However, what you propose is already an improvement, so feel free to merge as-is.

apps/wizard/utils/io.py
Marigold: Good catch! I made it work for snapshots, too.

apps/wizard/utils/io.py
Marigold: I profiled `steps_df`, but couldn't find any low-hanging fruit that would significantly speed it up. It just does a lot of things, which takes time. We'd have to refactor it a lot to make it both fast enough for such a simple use case as this and flexible for ETL dashboard. Anyway, I copied your comment to code to not get lost.

PR #3674, opened 2024-12-02
🎉 Enable converting from long to wide format in grapher step

Marigold:
Enable the conversion from long to wide format directly in the `data://grapher` step, rather than relying on a `grapher://grapher` step. This is achieved by using the `long_to_wide=True` parameter in the `create_dataset` function.

YAML metadata is updated twice: once for the long format and once for the wide format. This dual update allows for the use of Jinja templating with non-flattened indicator names while also supporting metadata for flattened indicators. A typical use case would involve overwriting specific indicators that fall outside the scope of the Jinja framework.

In practice, adding `long_to_wide=True` changes only the format (long → wide) in which tables are saved. The implementation remains backward compatible with existing steps.

## TODO
- [x] Investigate performance implications—health datasets are currently saved in long format, and saving them in wide format could introduce inefficiencies.
- [x] Set `long_to_wide=True` as the default and increment `ETL_EPOCH` to evaluate differences.
- [x] Undo `ETL_EPOCH`

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -290,7 +290,7 @@ def create_dataset(
             if ds.metadata.channel != "grapher":
                 log.warning("It is recommended to use long_to_wide=True only in the grapher channel")
 
-            dim_names = set(table.index.names) - {"country", "year", "date"}
+            dim_names = set(table.index.names) - {"country", "year", "date", None}
             if dim_names:
                 # First pass to update metadata from YAML
                 if meta_path.exists():
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -284,7 +284,7 @@ def create_dataset(
         # Expand long to wide
         if long_to_wide:
             if ds.metadata.channel != "grapher":
-                log.warning("long_to_wide=True should be only used in grapher channel")
+                log.warning("It is recommended to use long_to_wide=True only in the grapher channel")
 
             dim_names = set(table.index.names) - {"country", "year", "date"}
             if dim_names:
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -283,6 +283,9 @@ def create_dataset(
 
         # Expand long to wide
         if long_to_wide:
+            if ds.metadata.channel != "grapher":
+                log.warning("long_to_wide=True should be only used in grapher channel")
+
             dim_names = set(table.index.names) - {"country", "year", "date"}
             if dim_names:
                 # First pass to update metadata from YAML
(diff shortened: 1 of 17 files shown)


etl/helpers.py
```diff
@@ -270,16 +278,44 @@ def create_dataset(
         if table.metadata.short_name in used_short_names:
             raise ValueError(f"Table short name `{table.metadata.short_name}` is already in use.")
         used_short_names.add(table.metadata.short_name)
+
+        from etl import grapher_helpers as gh
+
+        # Expand long to wide
+        if long_to_wide:
+            if ds.metadata.channel != "grapher":
+                log.warning("long_to_wide=True should be only used in grapher channel")
```
pabloarosado, line 287: I understand that we don't want to store big wide tables both in garden and grapher, but I don't think this should be a requirement. It's up to the user where the widening of the table happens. Or is there another reason why this should only happen in grapher?

pabloarosado (approved):
Thanks a lot for fixing this issue so quickly, @Marigold!
This will help align what we have in the owid-catalog and in our friendly data catalog, making the ETL our only source of data again. It will also let us find individual indicators in the catalog by metadata content (which I suppose would not be possible with long formats). And it will streamline the production of mdim steps, avoiding circularity with DB. So I think this is a movement in the right direction.

I understand that, for big tables, there could be performance issues. But we can simply stick to long formats in those cases, as an exception (and maybe we'll figure out a better solution in the future).

etl/helpers.py
Marigold: It's not a requirement, just a warning. I've added it because I mistakenly used it in the garden step and it took me a while to figure out the problem. I'll change the wording and if we want we can delete it later.

Marigold:
I've compared performance on `time etlr grapher/who/2024-08-06/mortality_database_cancer$ --force --only --grapher` (this isn't the largest dimensional dataset by far, but wide format has 4500 indicators which is representative)

### long-to-wide
- `data://grapher` step takes 101s and produces `mortality_database_cancer.meta.json` with 19mb and `mortality_database_cancer.feather` with 37mb
- `grapher://grapher` step takes 141s

### keeping long format
- `data://grapher` step takes 3s and produces `mortality_database_cancer.meta.json` with 0.4mb and `mortality_database_cancer.feather` with 30mb
- `grapher://grapher` step takes 120s

Converting from long to wide is slower because of serialization to disk and writing JSON metadata file (this could be improved with compression). The reason long format is faster even in the grapher:// step is loading from data and inefficient processing of wide format (sparse columns, etc.).

pabloarosado:
Thanks a lot for the analysis, @Marigold! My intuition is that we may be able to speed up some of those processes or make exceptions for very wide tables if need be. But overall, I agree with the conclusion. Even if it slows down the run in some cases, it does bring important benefits in most cases (where, if I understand correctly, there is no significant slowdown).
I see that you are still on it, but please let me know if you need help or want to discuss anything, thanks!

PR #3679, opened 2024-12-03
📊 wb: Update poverty projections in Shared Prosperity report

paarriagadap:
Main issue https://github.com/owid/owid-issues/issues/1705


paarriagadap:
@lucasrodes Could you please review this one? I might remove the country level, and I will play with the charts, but the data side is done.

There is a unit-tests check failing, but seems to be unrelated to this PR.

etl/steps/data/garden/wb/2024-12-03/poverty_projections.meta.yml
```diff
@@ -0,0 +1,106 @@
...
+          - "{definitions.description_key_income_consumption}"
+          - "{definitions.description_key_nonmarket_income}"
+          - "{definitions.description_key_scenarios}"
+        presentation:
+          title_public: Share of population living in poverty
+          title_variant: $<<povertyline>> a day, <<scenario>>
+        display:
+          name: Share of population living below $<<povertyline>> a day (<<scenario>>)
+          numDecimalPlaces: 1
+          isProjection: {definitions.isprojection_by_scenario}
```
paarriagadap, line 85: This bit might not working, maybe we need to loosen the boolean restriction

etl/steps/data/garden/wb/2024-12-03/poverty_projections.py
```diff
@@ -0,0 +1,125 @@
...
+    ds_meadow = paths.load_dataset("poverty_projections")
+
+    # Read table from meadow dataset.
+    # Define empty table list to store tables.
+    tables = []
+    for table in TABLES:
+        tb = ds_meadow.read(table)
+
+        # Append table to list.
+        tables.append(tb)
```
lucasrodes, line 52: if you want to read all tables in `ds_meadow` (which I think is what you do here), you can simply do:

```suggestion
    # Read tables from meadow dataset.
    tables = list(ds_meadow)
```

lucasrodes (approved):
LGTM! Just left minor suggestion, maybe you were not aware of the trick :3

etl/steps/data/garden/wb/2024-12-03/poverty_projections.py
paarriagadap: Ah, that's good to know, thanks! I don't think I will use it here, because I won't use the country table in Garden. I will correct that

lucasrodes: @paarriagadap If you want to use a subset of tables like in here, you could use a one-liner instead:

```python
tables = [ds_meadow.read(table_name) for table_name in TABLES]
```

paarriagadap: That's true, thank you!

PR #3697, opened 2024-12-05
📊 climate: era5 surface temperature data December 2024

veronikasamborska1994:
hi @Marigold - looks like they've started to only update the data in the grib format so I am making some changes to read the data in the new format - I know there were issues with this code in the past so thought it'd be good if you reviewed this!


Marigold (approved):
Looks good!

PR #3702, opened 2024-12-06
📊 wittgenstein human capital

lucasrodes:
Tracking issue: https://github.com/owid/owid-issues/issues/1771

We have few charts relying on Wittgenstein Centre data. These datasets and charts are outdated and pre-ETL.

This PR adds the pipeline for ingesting Wittgenstein Centre data. In the future, it should be easier to ingest data from this source.


lucasrodes:
TODO for next week:

- Further work on meadow
  - Keep only tables that are relevant (e.g. tons of population can be dropped)
  - See how we can combine
  - Distribute indicators in different tables based on the dimensions they have. E.g. do not put one indicator that does not have `education` with one that has. This way we can optimize memory.

lucasrodes:
In the v3 data, there are only datapoints ≥2020. I've started importing a historical version of the dataset, which extends 1950-2020.

Currently working on importing these into Meadow.

Next tasks:

- [ ] MVP: bring projection-only to Grapher, with metadata
- [ ] Combine with historical: to make this scale, consider only combining relevant indicators.

PR #3708, opened 2024-12-09
🎉 Similar charts

Marigold:
A Streamlit prototype for finding similar charts to the selected one. Initially, it was searching for similar charts based solely on the semantic similarity of chart titles and subtitles, but later, many more "features" were added and combined into a similarity score. It was created to provide an intuition for how well automatic chart recommendations could work. However, the selection will likely evolve into a semi-automatic process (this app could then be leveraged for labeling).

It's not yet clear what "similar chart" means. Should we show very similar charts ("narrow"), a "wide" selection, or charts that are not so similar but perhaps more interesting?

## Scoring

Similarity is calculated as a weighted average of the following sub-scores (all between 0 and 1):
- **title**: semantic similarity of the title to other charts
- **subtitle**: semantic similarity of the subtitle to other charts
- **tags**: 1 if a chart shares at least one tag, 0 otherwise
- **share_indicator**: 1 if a chart shares an indicator, 0 otherwise
- **pageviews**: normalized `log(365d pageviews)` between 0 and 1

Weights were chosen intuitively. Analytical data to make this less subjective would be helpful, but first, we need to agree on what "similar" actually means.

## Diversity

Some recommendations, such as those for [political-regime](http://staging-site-similar-charts/etl/wizard/similar_charts?chart_search_text=political-regime), show many charts that are essentially the same chart, just from different providers. While it could be useful to inform users about different providers, it would be better to present more diverse charts. Similarly, some recommended charts, such as for [armed-forces-personnel](http://staging-site-similar-charts/etl/wizard/similar_charts?chart_search_text=armed-forces-personnel), often include the same indicator, presented as a share of the total population or a similar derivation.

To address this, I asked GPT to find a set of 5 diverse charts among the top 30 similar charts. These charts are marked with a 🤖 symbol, and the reason for their selection is displayed alongside them. This approach helps improve diversity and is relatively inexpensive ($0.01 per query, so doing it for 5000 charts wouldn't be prohibitively costly).

## How to review

Check out a few random charts with `Diversity with GPT` turned on. Are these recommendations reasonable? Is GPT providing additional value? Do you have any ideas for improving either the scoring system or the UI?

## Other use cases

This tool could be used to find "duplicate charts" or modified to assist with reviewing the [least viewed charts](https://github.com/owid/owid-issues/issues/1777).

### DIFF
--- a/apps/wizard/app_pages/similar_charts/app.py
+++ b/apps/wizard/app_pages/similar_charts/app.py
@@ -6,8 +6,7 @@ from structlog import get_logger
 
 from apps.wizard.app_pages.similar_charts import data, scoring
 from apps.wizard.utils import embeddings as emb
-from apps.wizard.utils import url_persist
-from apps.wizard.utils.components import Pagination, st_horizontal, st_multiselect_wider
+from apps.wizard.utils.components import Pagination, st_horizontal, st_multiselect_wider, url_persist
 from etl.config import OWID_ENV
 
 DEVICE = "cpu"
--- a/apps/wizard/app_pages/similar_charts/app.py
+++ b/apps/wizard/app_pages/similar_charts/app.py
@@ -9,8 +9,6 @@ from apps.wizard.utils import embeddings as emb
 from apps.wizard.utils.components import Pagination, st_horizontal, st_multiselect_wider, url_persist
 from etl.config import OWID_ENV
 
-DEVICE = "cpu"
-
 # Initialize log.
 log = get_logger()
 
@@ -26,7 +24,7 @@ st.set_page_config(
 ########################################################################################################################
 
 
-def st_chart_info(chart: data.Chart):
+def st_chart_info(chart: data.Chart) -> None:
     chart_url = OWID_ENV.chart_site(chart.slug)
     title = f"#### [{chart.title}]({chart_url})"
     if chart.gpt_reason:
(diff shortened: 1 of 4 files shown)


lucasrodes:
This looks very interesting. I find that the "diversity" option can be quite useful in the future, nice one!

In the future, I'd like to understand the prototype a bit more so I can help with the UI. E.g. is it mostly for development purposes (e.g. for engineering team) or also for folks in D&R? I may weigh in more on the UI front depending on this (which, FWIW, I'm keen to help with).

Marigold:
We'll discuss the future of this app in January. This was an initial spike intended to prove that it's feasible, not overly complex, and potentially valuable. Now the hard part will be figuring out how much we can get out of it.

> What's the difference between the chart at the top (without borders) and those that follow?

Great UI, right? 😁 The top chart is the one found by the text search on the right. If the input contains a valid slug or ID, we match on that, otherwise we match using semantic search on the title.

> I guess the above point relates to a more general question: what is this app trying to facilitate? It looks to me as if it is attacking two things now: (i) Free-search to find charts similar to the queried thing. (ii) Select a specific chart from a slug and present similar charts.
For (i) I would not show the current top chart, since there is no actual chart "selected".
For (ii), I would use a dropdown selectbox with just a limited list of options (not free-text).
Maybe we can define two modes?

That's a good point. I'm going to remove mode (i) to make it less confusing. Free-search for charts wasn't in the scope and is not that useful (as far as I know).

(If you think this is a good start, could you approve the PR? I wouldn't want to keep it hanging as PR for too long, and we can always delete it.)

lucasrodes (approved):
This is totally good to go, Mojmir!

Read your comment and makes sense to me. 

Thanks for pushing for this.

PR #3714, opened 2024-12-11
📊 animal_welfare: Update animal welfare data

pabloarosado:
### Summary
Update data on the number of farmed and wild-caught fish.

### Review
@lucasrodes no need for a thorough review. But please have a look at the new charts when you have some time, and please let me know if you see anything odd. Thank you!

### TODO (Pablo R):
* [x] Wait for Mood and Brooke's feedback on the new changes.
* [ ] Once merged, redirect https://ourworldindata.org/grapher/global-farmed-finfishes to https://ourworldindata.org/grapher/farmed-fish-killed
* [ ] Once the chart on global farmed finfishes is redirected, archive `data://grapher/animal_welfare/2023-08-08/farmed_finfishes_used_for_food` and dependencies.


etl/steps/data/garden/animal_welfare/2024-12-12/farmed_finfishes_used_for_food.meta.yml
```diff
@@ -0,0 +1,85 @@
...
+            hasMapTab: false
+      n_fish_high:
+        title: Estimated number of farmed finfish killed for food (upper bound)
+        unit: 'fish'
+        short_unit: ''
+        display:
+          numDecimalPlaces: 0
+        presentation:
+          grapher_config:
+            hasMapTab: false
```
lucasrodes, line 39: nit: maybe worth defining this bit in `common`?

Seems like

```yaml
presentation:
  grapher_config:
    hasMapTab: true
```

is more rare.

lucasrodes (approved):
LGTM.

Some comments on the charts:

[Number of wild-caught fish killed for food](http://staging-site-data-animal-welfare-update/grapher/wild-caught-fish)
A bit of thinking-out-loud:
- I see that for years <2019 there are only values for "World". Maybe you can disable entity-selector in the chart view? Countries are still visible in the map.
- I guess one reason not to do what I suggest above is if you want to also show the CI interval for countries?
- Ideally, I guess, you'd have bar charts for countries? Not sure that can be achieved in Grapher.
- You could use the view that you have in the chart in admin (faceted bar charts).

pabloarosado:
Thanks @lucasrodes I'm not sure about [the chart on wild-caught fish](http://staging-site-data-animal-welfare-update/grapher/wild-caught-fish), [I asked around](https://owid.slack.com/archives/C04VBQ00HQR/p1734443109972449) to see if it would be possible to have bar charts when there's only one data point (that's what I'd expect). The faceted view was fine when we had only one year for all entities. But now we have many years worth of global data, which I think is the most important one. And it's also important to show uncertainties. So I'd prefer to show global data by default. Please let me know if you come up with a better solution. Thanks!
(I'm waiting for the author's feedback before merging)

PR #3716, opened 2024-12-11
🐛 Raise warning on staging server when we cannot delete variables used in charts

Marigold:
There's an error that [sometimes comes up](https://owid.slack.com/archives/C46U9LXRR/p1733928261018389) on staging servers
```
ValueError: Variables used in charts will not be deleted automatically:
```

This issue occurs when someone renames the short names of indicators and updates some charts in production. If you then merge the master branch to the staging server, which is still using old charts from the time it was created, you'll encounter this error.

This PR adds a check that converts the error into a warning if the changes on the staging server are unrelated to the charts. The error disappears in production after the merge.

### DIFF
--- a/etl/grapher_import.py
+++ b/etl/grapher_import.py
@@ -31,8 +31,9 @@ from sqlalchemy.orm import Session
 from apps.backport.datasync import data_metadata as dm
 from apps.backport.datasync.datasync import upload_gzip_string
 from apps.chart_sync.admin_api import AdminAPI
+from apps.wizard.app_pages.chart_diff.chart_diff import ChartDiffsLoader
 from etl import config
-from etl.db import get_engine
+from etl.db import get_engine, production_or_master_engine
 
 from . import grapher_helpers as gh
 from . import grapher_model as gm
@@ -496,8 +497,7 @@ def cleanup_ghost_variables(engine: Engine, dataset_id: int, upserted_variable_i
         if rows:
             rows = pd.DataFrame(rows, columns=["chartId", "variableId"])
 
-            # raise an error if on staging server
-            if config.ENV == "staging":
+            if _raise_error_for_deleted_variables(rows):
                 raise ValueError(f"Variables used in charts will not be deleted automatically:\n{rows}")
             else:
                 # otherwise show a warning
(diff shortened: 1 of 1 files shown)


Marigold:
@lucasrodes added those commits 😉

etl/grapher_import.py
```diff
@@ -575,6 +575,19 @@ def cleanup_ghost_variables(engine: Engine, dataset_id: int, upserted_variable_i
...
 
 
+def _raise_error_for_deleted_variables(rows: pd.DataFrame) -> bool:
+    """If we run into ghost variables that are still used in charts, should we raise an error?"""
+    # raise an error if on staging server
+    if config.ENV == "staging":
+        # It's possible that we merged changes to ETL, but the staging server still uses old charts. In
+        # that case, we first check that the charts were really modified on our staging server.
+        modified_charts = ChartDiffsLoader(config.OWID_ENV.get_engine(), production_or_master_engine()).df
+        return bool(set(modified_charts.index) & set(rows.chartId))
```
lucasrodes, line 585: I tried following the logic here, but didn't really come to an understanding. Could you help me understand it?

So, my logic is:

1) Before we had something like: "raise error if there are ghost variables (i.e. `rows`)"

2) No, you are proposing to evaluate `bool(set(modified_charts.index))` in addition. This should return `True` whenever `modified_charts` is not empty, right? (In that case wouldn't it be enough to just use `not modified_charts.empty`?). 

3) Still, if `modified_charts` is not empty, why is that a potential sign to raise an error? Isn't `modified_charts` containing a row per each chart that has been modified (regardless of it being approved, etc.)?

I think I'm missing something here. Thanks!

lucasrodes (approved):
I would say LGTM, but there is one line I don't fully get.

etl/grapher_import.py
Marigold: > No, you are proposing to evaluate bool(set(modified_charts.index)) in addition. This should return True whenever modified_charts is not empty, right? (In that case wouldn't it be enough to just use not modified_charts.empty?).

`set(modified_charts.index)` contains chart ids that have been modified on staging (either via config, data or metadata) and `set(rows.chartId)` contains problematic chart ids. Previously, we raised an error if `set(rows.chartId)` wasn't empty, but the problem was that sometimes it was empty because of race conditions between staging & prod.

`bool(set(modified_charts.index) & set(rows.chartId))` returns True if there are any problematic charts AND those charts have been also modified on the staging server.

I could have also written it as:
```
if rows and any(row.chartId in modified_charts.index):
  return True
```

Perhaps it's me who's missing something and I misunderstood your question.

lucasrodes: oh my god. I miss-read one of the parents. I read `bool(set(modified_charts.index)) & set(rows.chartId)`.

NVM, I got it, makes total sense.

PR #3718, opened 2024-12-12
✨ wizard: create data step, improve ergonomics

lucasrodes:
- New **data stepI**: am trying to simplify the ergonomics of creating a new data step. 
  - Currently, we present the user with three different options at the top level:, Meadow, Garden, Grapher. I think it should be simpler at the high level, and just label these as "data step". Then, the user can choose how many steps they want to create (all of them, as in 'express mode', just a subset, etc.).
- Maintenance: 
  - I'm also re-viewing some bits in Wizard, to keep things updated.
  - Bumping streamlit version to 1.41
- Integrating `harmonizer` into the data step creation.
- Support customization of dependencies in data step creation.
- Aesthetic tweaks to Fast-track.


lucasrodes:
It's unclear why the CI/CD job is failing. I've created a new PR: https://github.com/owid/etl/pull/3720, to see if the issue gets solved there.

PR #3720, opened 2024-12-16
✨ wizard: data step, fasttrack ux, streamlit version 

lucasrodes:
Continues work from https://github.com/owid/etl/pull/3718 (CI/CD stopped working)

- New **data stepI**: am trying to simplify the ergonomics of creating a new data step. 
  - Currently, we present the user with three different options at the top level:, Meadow, Garden, Grapher. I think it should be simpler at the high level, and just label these as "data step". Then, the user can choose how many steps they want to create (all of them, as in 'express mode', just a subset, etc.).
- Maintenance: 
  - I'm also re-viewing some bits in Wizard, to keep things updated.
  - Bumping streamlit version to 1.41
- Integrating `harmonizer` into the data step creation.
- Support customization of dependencies in data step creation.
- Aesthetic tweaks to Fast-track.

### DIFF
--- a/apps/backport/fasttrack_backport.py
+++ b/apps/backport/fasttrack_backport.py
@@ -88,7 +88,7 @@ def _create_client():
     # Add your service account file
     creds = Credentials.from_service_account_file(os.environ["GOOGLE_APPLICATION_CREDENTIALS"], scopes=scope)
 
-    return gspread.authorize(creds)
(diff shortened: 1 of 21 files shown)


lucasrodes:
https://github.com/owid/etl/pull/3718 ended up working ー closing

Issue #3724, opened 2024-12-16
metadata: Jinja templates are being inherited when a variable is used in calculation


spoonerf:
I think this is a new occurrence: When you use an indicator that has Jinja metadata in a calculation (e.g., UN WPP population). The calculated variable inherits this template, and it can cause the step to fail if you don't overwrite the metadata values that are given in the template.

e.g. [here in Buildkite](https://buildkite.com/our-world-in-data/etl-automated-staging-environment/builds/13428#0193d030-7d92-423b-9d19-eaf48c0a112d). Buildkite is applying a template that doesn't exist in the metadata of the step, but it is the metadata for the variable used in a calculation in this step ([etl/steps/data/garden/malnutrition/2024-12-16/malnutrition.py](https://github.com/owid/etl/pull/3721/files#diff-783d40bcb2a49a6c371989c181e2491545f58c248127a017020a6d68e049246b)).

spoonerf:
(Maybe it's intentional, but in that case I guess it might be good to have a function to strip variable metadata for these kind of variables)

Marigold:
Thanks for reporting this. I've improved the error message to make it clear which variable and metadata are causing the problem.

In this specific case, the issue is that `presentation.title_public` is inherited from the population indicator, but it's missing the dimensions required to render the metadata. I'm not sure what the best approach is—should we stop inheriting the `presentation.title_public` field, or would that cause other problems?

spoonerf:
Would a function that strips Variable metadata be a possible way to go (or maybe an option in load_dataset())? 

I figure it's `presentation.title_public` this time, but it could be a different one with another variable (although population will be by far the most frequent one!).

Marigold:
That's a great idea, created a PR https://github.com/owid/etl/pull/3728.

PR #3728, opened 2024-12-17
✨ Add reset_metadata parameter to `.load()`

Marigold:
Fix for https://github.com/owid/etl/issues/3724#issuecomment-2546229678

`ds.load(reset_metadata=True)` lets you load a dataset and reset all its metadata (origins, titles, etc.). This is especially useful when working with population table. Division by population could propagate unwanted metadata such as `presentation.title_public`.


spoonerf:
Nice!

Would it be possible to preserve the origins? I've had a go in [this PR](https://github.com/owid/etl/pull/3731/files). I was thinking we probably still want to record in the origins when we've used population (etc) to create a new variable, but it's the other parts of the VariableMeta() that we would want to be rid of.

Marigold:
Closing in favour of https://github.com/owid/etl/pull/3731#pullrequestreview-2511498649

PR #3731, opened 2024-12-17
Add reset metadata origin option

spoonerf:


lib/catalog/owid/catalog/datasets.py
```diff
@@ -173,6 +183,15 @@ def read(self, name: str, reset_index: bool = True, safe_types: bool = True) ->
                 t.metadata.dataset = self.metadata
                 if safe_types:
                     t = cast(tables.Table, to_safe_types(t))
+                if reset_metadata:  # Handles True and "keep_origins"
+                    t.metadata = TableMeta()
+                    for col in t.columns:
+                        if reset_metadata == "keep_origins":
```
Marigold, line 189: Rather than mixing True / False / string, we could have a parameter `metadata: Literal["keep", "keep_origins", "reset"]` (similar to `errors: Literal["ignore", "warn", "raise"] = "raise"` from pandas)

Marigold (approved):
Looks good!

PR #3746, opened 2024-12-20
✨ Exclude archived steps from VSCode search

Marigold:
## Motivation

Searching in VSCode can be frustrating when we keep old, archived versions of code in the repository. Quite frequently, I find myself yelling at my computer, only to realize that I was editing the wrong dataset (found through VSCode quick search). This problem has already been addressed by the [find-latest-step](https://github.com/owid/etl/tree/master/vscode_extensions/find-latest-etl-step) extension, although it lacks some capabilities of the native VSCode quick search. I don't personally use the `find-latest-etl-step` extension because it's slow on my laptop and I prefer the fuzzy matching in quick search (although other people use it).

## Solution

Identify all inactive steps and exclude them from the VSCode search toolbar and file explorer by adding them to `"files.exclude"` and `"search.exclude"` in `.vscode/settings.json`. Running the script **excluded over 600 datasets**. One disadvantage is that **all VSCode users must be aware that archived files can't be found there** and must instead be opened, for example, via `code etl/path/to/script` from the terminal.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,12 +36,6 @@
     "**/docs/api/*.md",
     "**/docs/architecture/*.md"
   ],
-  "files.exclude": {
-    "**/dataset_*_config.json": true,
-    "**/dataset_*_values.json": true,
-    "**/dataset_*.json.dvc": true,
-    "**/dataset_*.feather.dvc": true
-  },
   "yaml.format.printWidth": 999,
   "ruff.path": [
     ".venv/bin/ruff"
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 #  Makefile
 #
 
-.PHONY: etl docs full lab test-default publish grapher dot watch clean clobber deploy api activate
+.PHONY: etl docs full lab test-default publish grapher dot watch clean clobber deploy api activate vscode-exclude-archived
 
 include default.mk
 
@@ -31,6 +31,7 @@ help:
 	@echo '  make chart-sync 	Start Chart-sync on port 8083'
 	@echo '  make test      	Run all linting and unit tests'
 	@echo '  make test-all  	Run all linting and unit tests (including for modules in lib/)'
+	@echo '  make vscode-exclude-archived  Exclude archived steps from VSCode user settings'
 	@echo '  make watch     	Run all tests, watching for changes'
 	@echo '  make watch-all 	Run all tests, watching for changes (including for modules in lib/)'
 	@echo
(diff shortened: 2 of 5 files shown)


Marigold:
There are two ways we can do it - either save it to the repository `.vscode/settings.json` (VSCode Project settings) so that all users have it enabled by default or let everyone whether to use it or not and write to User settings (`~/Library/Application Support/Code/User/settings.json`). The latter would mean that everyone will have to manually run it once in a while, while with the former we could run it automatically and commit to the repo.

> If I understand your proposal, I need to first run python scripts/exclude_archived_steps.py and then open the VS Code search?

As per above, you wouldn't have to run it if we go with the project settings.

> I would find it nice if there was the option to "unarchive steps" so that I can search for all steps.

If you want to search for a text inside all steps, you can click on the cog button

image

Searching archived files is trickier as they won't show up in quick search (Cmd + P). You'd have to either do it from terminal `find . -type f -path "*garden/who/2024-09-09/flu_test*"` (then run `code etl/steps/data/grapher/who/2024-09-09/flu_test.py` to open in) or from Finder.

I could also make it possible to keep files that are manually commented in `.vscode/settings.json`. For instance
```
"files.exclude": {
    "etl/steps/data/garden/agriculture/2017-03-08": true,
    // "etl/steps/data/garden/agriculture/2023-04-20": true,
   ...
```
wouldn't hide folder `etl/steps/data/garden/agriculture/2023-04-20` and running `exclude_archived_steps.py` wouldn't overwrite it. That way, you could manually comment archived steps you frequently work on.

pabloarosado:
Hi @Marigold thanks for looking into this. I use `find-latest-step` and it takes 2 seconds to launch, and the displayed hits are more useful than the default `cmd+p` bar. So, for finding the latest ETL steps, I find it quite useful:

https://github.com/user-attachments/assets/7dbf7a57-e86b-41b2-a67d-22a63917184b

I don't know why it's so slow for you. But we could improve the extension. For example, it could search only in `etl/steps/data`, `etl/steps/export` and `snapshots`, and only for `*.py` files. Maybe that could speed up things (and exclude yaml and dvc files, if that's more convenient).

But if you think that's not a good approach, then your solution is also fine. Feel free to merge if you think this is convenient for you, thanks!

Marigold:
> I don't know why it's so slow for you

Well, I have a hunch, my old laptop thinks twice about every operation it does.

This way is more personally useful for things like large-scale refactorings (where you don’t want to be fixing archived steps) and reducing clutter when searching for non-step modules. It’s definitely more of a nice-to-have, though.

I’ll keep it open a bit longer before deciding whether to merge it or not. I’m still unsure if the potential confusion from not being able to find archived steps is worth it.

Marigold:
I've found this to be really useful over the past few months. I'm not going to force everyone to use it by adding it to project settings, but everyone can now run `make vscode-exclude-archived` to modify their User settings.

PR #3768, opened 2024-12-29
💄 change dataset title

lucasrodes:
In admin, we now have multiple datasets named "World Population Prospects". Maybe we could modify some so that it is easier to distinguish them?


lucasrodes:
Will merge as this has low-risk.

spoonerf:
Nice one @lucasrodes, makes sense!

PR #3785, opened 2025-01-02
🐛 wildfires: temporary fix

veronikasamborska1994:
hi @spoonerf! was a slightly annoying bug but I think this solution should be a lot more reliable in the future. I am also adding some code for the 2025 data which should hopefully start running smoothly next week!

### DIFF
--- a/etl/steps/data/garden/climate/latest/weekly_wildfires.py
+++ b/etl/steps/data/garden/climate/latest/weekly_wildfires.py
@@ -97,7 +97,7 @@ def run(dest_dir: str) -> None:
         tables=[tb],
         check_variables_metadata=True,
         default_metadata=ds_meadow.metadata,
-        yaml_params={"date_accessed": last_date_accessed(tb)},
+        yaml_params={"date_accessed": last_date_accessed(tb), "year": last_date_accessed(tb)[-4:]},
     )
 
     # Save changes in the new garden dataset.
--- a/etl/steps/data/grapher/climate/latest/wildfires_by_week_average.py
+++ b/etl/steps/data/grapher/climate/latest/wildfires_by_week_average.py
@@ -87,6 +87,6 @@ def run(dest_dir: str) -> None:
         dest_dir,
         tables=[tb],
         default_metadata=ds_grapher.metadata,
-        yaml_params={"date_accessed": last_date_accessed(tb)},
+        yaml_params={"date_accessed": last_date_accessed(tb), "year": last_date_accessed(tb)[-4:]},
     )
     ds_grapher.save()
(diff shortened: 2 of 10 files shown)


etl/steps/data/garden/climate/latest/weekly_wildfires.py
```diff
@@ -97,7 +97,7 @@ def run(dest_dir: str) -> None:
         tables=[tb],
         check_variables_metadata=True,
         default_metadata=ds_meadow.metadata,
-        yaml_params={"date_accessed": last_date_accessed(tb)},
+        yaml_params={"date_accessed": last_date_accessed(tb), "year": last_date_accessed(tb)[-4:]},
```
spoonerf, line 100: Oh nice, I didn't know about this yaml_params argument!

snapshots/climate/latest/weekly_wildfires.py
```diff
@@ -22,7 +22,7 @@
 SNAPSHOT_VERSION = Path(__file__).parent.name
 # Define a range of years for which data will be processed.
 
-YEAR = 2024
+YEARS = [2024, 2025]
```
spoonerf, line 25: Is it worth adding a function to grab the current year so that when we hit 2026 it is added here?

etl/steps/data/grapher/climate/latest/wildfires_by_week_average.py
```diff
@@ -77,6 +77,26 @@ def run(dest_dir: str) -> None:
         if origin_column in tb.columns:  # Check if the origin column exists
             for col in [f"avg_{group}_until_2024", f"upper_bound_{group}", f"lower_bound_{group}"]:
                 tb[col].origins = tb[origin_column].origins
+
```
spoonerf, line 80: Is it worth generalising this function a little more so that it aggregates all data before the current year, rather than hardcoding 2024? If you think that's appropriate, it might be that the downstream charts need a little more manual updating so it makes sense to keep this as it is.

spoonerf (approved):
Looks good! I'm just wondering if it's possible to generalise it even more to avoid hardcoding 2024 anywhere, but you'll know best if that's possible, I know it's not the most straightforward processing!

Also worth unhashing line 15 [here](https://github.com/owid/etl/blob/master/scripts/update-wildfires.sh) to get the autoupdates going again.

snapshots/climate/latest/weekly_wildfires.py
veronikasamborska1994: yep, good idea! I'll add this!

etl/steps/data/grapher/climate/latest/wildfires_by_week_average.py
veronikasamborska1994: hmm it will probably be a bit tricky to sync this with grapher to not cause errors/create some weird subtitles/titles in the grapher chart so I've decided to keep it as it is for now. Also, I am not sure if we want to start showing the new year (e.g.,2025) immediately or wait a couple of months until there is more data for the chart not to look a bit weird. I'll keep this in mind though when I next take a look at it! :) thank you!

PR #3828, opened 2025-01-12
📊 marriages and divorces: add ONS divorce data

veronikasamborska1994:
hi @spoonerf! pretty straightforward PR but just wanted to see what you think about this new way of plotting the data on this [chart](http://staging-site-ons-divorces/admin/charts/4347/edit). I think it's a bit easier to read than the old version but wanted to hear a second opinion.

### DIFF
--- a/dag/families.yml
+++ b/dag/families.yml
@@ -8,3 +8,13 @@ steps:
     - data://meadow/oecd/2024-12-30/family_database
   data://grapher/oecd/2024-12-30/family_database:
     - data://garden/oecd/2024-12-30/family_database
+
+  #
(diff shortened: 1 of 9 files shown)


spoonerf (approved):
Looks good! 

I prefer your new version of the chart, it's much clearer that divorces peaked in marriages from 1995, although I do slightly miss seeing the profile of each cohort e.g. the sudden uptick in the 4th anniversary in 1975 marriages. 

I think having a colour gradient for the years could help improve the look of the new chart too, at the moment I find the colours a little unintuitive, although it's not easy for so many categories.

PR #3829, opened 2025-01-13
🐛 Cleanup _load_population

Marigold:
We've recently deprecated `key_indicators` dataset, but the function `_load_population` was still using them. I removed it and updated `geo.*` functions accordingly. (Some steps were breaking our nightly build)


lucasrodes (approved):
Thanks for looking into this, Mojmir!

Seems like we managed to get rid off `key_indicators` 🤝

PR #3837, opened 2025-01-13
🔨 refactor guinea worm steps

antea04:
While archiving data sets I noticed that the dependencies for the guinea worm steps are a bit messy.

Things changed:

- separated processing between guinea worm case numbers and country certification status
- moved functions into corresponding steps
- removed dependencies on grapher/ old fasttrack steps
- archived all processing steps which are no longer needed

Hopefully this makes the code a bit easier to understand and updating this data easier.


etl/steps/data/garden/who/2024-06-17/guinea_worm.py
```diff
@@ -39,17 +47,16 @@ def run(dest_dir: str) -> None:
...
+    tb["guinea_worm_reported_cases"] = tb["guinea_worm_reported_cases"].fillna(0)
 
-    # fix data types
-    tb["year_certified"] = tb["year_certified"].astype(str)
+    # add rows for current year
+    tb = add_current_year(tb, tb_cases, year=CURRENT_YEAR)
 
     # format index
-    tb = tb.format(["country", "year"])
+    tb = tb.format(["country", "year"], short_name="guinea_worm")
```
spoonerf, line 59: The variables in this table are missing origins, I think that happens in certain processing steps. You can save the origins as an object early in the script and then add them back in at a late stage.

spoonerf (approved):
Good thinking! This looks good to me, just a minor point on origins, which might have been there before anyway, but this could be a good time to fix it.

etl/steps/data/garden/who/2024-06-17/guinea_worm.py
antea04: Hi Fiona! I actually think it is complaining, because both the cases step and the certification step have sources instead of origins (because they are from fasttrack/ old). 

Do you think it makes sense to migrate the information from the "sources" to "origins" in this garden step?

spoonerf: Ah I see! That sounds a bit more fiddly then, I would just leave it for the next update.

PR #3843, opened 2025-01-15
📊 Update WB poverty projections

paarriagadap:
The WB team has updated the poverty projections replication package, due to an error we found


etl/steps/data/meadow/wb/2025-01-15/poverty_projections.py
```diff
@@ -0,0 +1,60 @@
...
+    # Load inputs.
+    #
+    # Retrieve snapshot.
+    snap = paths.load_snapshot("reproducibility_package_poverty_prosperity_planet.zip")
+
+    # Define empty list to store tables.
+    tables = []
+    for table, table_config in TABLE_PARAMETERS.items():
+        # Load data from snapshot.
+        tb = snap.read_in_archive(f"{FILES_DIRECTORY}/{table_config['file']}")
```
paarriagadap, line 33: Hi @lucasrodes! Do you know why this is failing? I only updated the version of the data and it has the same structure, but now it fails with
```
ValueError: File is neither a zip nor a tar file.
```

lucasrodes: I was testing this, and found no error on my end. Can you provide the complete traceback?

FWIW, I've added some new features the `Snapshot` class, so that we can read multiple files from an archive more efficiently. So far, we had to unarchive the file multiple times (one for each read operation)

paarriagadap: Thanks! Yes, let me know when https://github.com/owid/etl/pull/3863 is ready to go! (I am following the PR as well)

lucasrodes: merged! @paarriagadap

paarriagadap:
Hi @lucasrodes! There is another weird thing with this update: when I run this new snapshot in the same pipeline as before I get these long numbers as dimensions ([dataset](http://staging-site-data-wb-poverty-projections/admin/datasets/6893)), when they should be 2.15, 3.65, 6.85. Even weirder is that if I re-include the steps from version `2024-12-03` I get the same error ([dataset](http://staging-site-data-wb-poverty-projections/admin/datasets/6847)). Do you know if this is due to a change in the libraries we use? I haven't been able to fix it anyway.
image

Marigold:
@paarriagadap I'll have a look at those decimals.

A slightly different question for discussion - does it make sense to create a new version of a dataset if there was no change to code and only snapshot has been corrected? I'd personally just update the snapshot and keep all other channels as they are, but others might have a different opinion.

paarriagadap:
Thanks @Marigold! I only applied changes to the garden step, thinking that it was enough to fix it.

I created a new version to use Anomalist and also check the new values in Indicator Upgrader. If there is another way to do this by just replacing the snapshot please let me know.

Marigold:
> I created a new version to use Anomalist and also check the new values in Indicator Upgrader. If there is another way to do this by just replacing the snapshot please let me know.

If you're doing all quality checks, then you need to bump the version. If you only wanted to compare charts, it'd be enough to re-run the snapshot and then check chart-diff with **Show all charts** turned on. You'd see all data or metadata changes side by side.

PR #3847, opened 2025-01-15
✨ link to data set when running grapher step

antea04:
Added a link to the dataset in grapher to the log output when running a grapher step. 

If this is not useful for anybody but me, I am also happy not to merge.

### DIFF
--- a/etl/grapher/to_db.py
+++ b/etl/grapher/to_db.py
@@ -108,6 +108,7 @@ def upsert_dataset(
             "upsert_dataset.upsert_dataset.end",
             short_name=short_name,
             id=ds.id,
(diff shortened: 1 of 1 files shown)


etl/grapher/to_db.py
```diff
@@ -108,6 +108,7 @@ def upsert_dataset(
             "upsert_dataset.upsert_dataset.end",
             short_name=short_name,
             id=ds.id,
+            url=f"http://{engine.url.host}/admin/datasets/{ds.id}",
```
lucasrodes, line 111: I think this is a useful addition! Let's try it out.

lucasrodes (approved):
let's try it.

PR #3853, opened 2025-01-16
📊 Consolidate table on living conditions vs. GDP per capita

paarriagadap:
Related issue: https://github.com/owid/owid-issues/issues/1820

Consolidate table to recreate the following chart from [this article](https://ourworldindata.org/global-economic-inequality-introduction):
Correlates-of-GDP-How-is-life-at-different-levels-of-income-global-ineq-article

We add a 10-year tolerance to compare all the indicators.


etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
```diff
@@ -0,0 +1,210 @@
...
+        columns={
+            "ny_gdp_pcap_pp_kd": "gdp_per_capita",
+            "sh_med_phys_zs": "physicians_per_1000_people",
+            "eg_elc_accs_zs": "access_to_electricity",
+        }
+    )
+
+    # UN WPP
+    tb_un_wpp = tb_un_wpp[
+        (tb_un_wpp["sex"] == "all") & (tb_un_wpp["age"] == "0") & (tb_un_wpp["variant"] == "estimates")
```
antea04, line 65: the 'age' column has integers, so you probably want to do (tb_un_wpp["age"] == 0)

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
```diff
@@ -0,0 +1,210 @@
...
+    ].reset_index(drop=True)
+    tb_un_wpp = tb_un_wpp[["country", "year", "life_expectancy"]]
+
+    # IGME
+    tb_igme = tb_igme[
+        (tb_igme["obs_value"] == "Under five mortality rate")
```
antea04, line 71: this should be tb_igme["indicator"] == "Under-five mortality rate"

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
```diff
@@ -0,0 +1,210 @@
...
+    tb_list = []
+
+    for indicator in indicators:
+        tb_indicator = tb[["country", "year", indicator]].copy()
+
+        # Drop rows with missing values
+        tb_indicator = tb_indicator.dropna(subset=[indicator]).reset_index(drop=True)
+
+        # Define latest year in the dataset
+        latest_year = tb_indicator["year"].max()
```
antea04, line 162: Before this one you should filter for only data points in the past: tb_indicator[tb_indicator["year"]<= 2025]. You have projections in population and might have them in other indicators in the future.

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
```diff
@@ -0,0 +1,210 @@
...
+
+        # Select all the rows with year higher or equal to latest_year - 10
+        tb_indicator = tb_indicator[tb_indicator["year"] >= latest_year - 10].reset_index(drop=True)
```
antea04, line 165: I am not 100% sure this is what you want - Max wrote "Let’s use a tolerance of 10 years — if the latest available datapoint is more than 10 years old then exclude it, but if its less than 10 years old include it.". I would take that as any data point used in this data set/ graphic should not be older than 10 years i.e. be from 2015 latest. Now for some indicators (e.g. average working hours) the latest datapoint is from 2019 - so this includes data points up to 2009. Is that intended?

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.meta.yml
```diff
@@ -0,0 +1,58 @@
+# NOTE: To learn more about the fields, hover over their names.
```
antea04, line 1: I think you can delete this file if this data set does not end up on grapher?

paarriagadap: That's right, thanks! For some reason it failed when I tried to do it before.

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
paarriagadap: Thanks!

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
paarriagadap: Oops, thanks!

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
paarriagadap: That's right, with the dropna I covered all the tables I processed, except for population which I added via the geo function

etl/steps/data/garden/growth/2025-01-16/gdppc_vs_living_conditions.py
paarriagadap: I think yours is the right interpretation. Before I understood like a tolerance of 10 years per indicator.

paarriagadap:
Thank you very much, @antea04! I followed all your suggestions; now I am getting what I should have.

I will check the zero values a bit more, thanks again.

antea04 (approved):
LGTM! ☺️

PR #3854, opened 2025-01-16
🐛 correct tracking of producer views

lucasrodes:
Fixes https://github.com/owid/etl/issues/3855

Producer analytics does not accurately count the views of data from a given producer. It might be over estimating these.

This PR drops the usage of `VersionTracker` and instead uses plain SQL queries to our database. There are two benefits:

- Analytics will be more accurate
- Performance is better. Using `VersionTracker` to load the steps dataframe is significantly slower.
- Add custom wrapper `st_cache_data` to show custom message in cached function call

/schedule


lucasrodes:
Update: the bug might be coming from `VersionTracker`

lucasrodes:
On my computer, using regular queries to DB improves performance significantly.

Old code (with VersionTracker): Took ~8 seconds.
New code (regular queries): Takes ~1 second.

apps/wizard/utils/components.py
```diff
@@ -514,3 +515,32 @@ def _get_params(component, key):
             return float(params)
         else:
             return params
+
+
+def st_cache_data(func=None, *, custom_text="Running...", show_spinner=False, **cache_kwargs):
```
pabloarosado, line 520: This function is 🤯

apps/wizard/app_pages/producer_analytics/data_io.py
```diff
@@ -113,8 +115,68 @@ def get_chart_views_from_bq(
...
+	LEFT JOIN charts c ON c.id = cd.chartId
+	LEFT JOIN chart_configs cf ON cf.id = c.configId
+	LEFT JOIN origins_variables ov ON ov.variableId = cd.variableId
+	LEFT JOIN origins o ON o.id = ov.originId
+	LEFT JOIN variables v ON cd.variableId = v.id
+	LEFT JOIN datasets d ON d.id = v.datasetId
+)
+SELECT * FROM t_base
+WHERE origin_id IS NOT NULL
+AND is_published = true;
```
pabloarosado, line 153: Nice! I assumed such a query would not be possible, but I was not aware of `origins_variables`. Thanks for figuring this out.

pabloarosado:
Excellent work @lucasrodes! When I start up the dashboard it's still quite slow, but I don't think that should be a problem. The main thing is to get the numbers as accurately as possible.

I wouldn't say the previous implementation was buggy, but it was simply based on relationships at the dataset level (mostly based on the DAG). I was quite certain that fetching this info at the indicator level from DB would not be possible, but I see that the DB schema has changed in ways I was not aware of, and now it's indeed possible. So thanks for figuring out the way to get the numbers at the indicator level.

I haven't double-checked your query (I don't have time now, but if you want me to, please let me know). I see the "Various sources (https://ourworldindata.org/population-sources)" producer in the table, not sure if that's what we want. We can discuss it a bit at some point, and maybe tweak it a bit for our purposes. But for now I think these numbers are probably closer to what we want than what we had before based on dataset relationships. So, feel free to merge!

pabloarosado (approved):
Looks good, thanks!

pabloarosado:
I've noticed a small issue on the chart. It seems that now it's not possible to select an individual chart line:

https://github.com/user-attachments/assets/60cf814b-f030-4263-aad0-f09c7d3c01d6

lucasrodes:
> I've noticed a small issue on the chart. It seems that now it's not possible to select an individual chart line:

Good catch, I introduced legend groups while testing couple of things. Not really needed, rolling it back!

Issue #3861, opened 2025-01-17
wizard: meadow step gets reset when progressing through data steps


spoonerf:
### One-liner
Kind of a hard one to explain, but I've noticed a few times that as I am running through the garden stage of the new data step in Wizard, the meadow step gets reset to the default file. I've checked it by doing `cmd + z` on the step and it goes back to the file as I expect. 

### Context & details

I've not been able to pin down when it happens, perhaps during or after the country harmonizing, but here's a video of me pressing `cmd+z` and the expected meadow code coming back.

https://github.com/user-attachments/assets/9db762d5-a603-458e-a47e-2c6eb057f485

Next time I use the data step I'll try and notice if and when this happens and report back here.

lucasrodes:
Hey @spoonerf, how are you creating the steps? Are you creating Meadow & Garden at the same time? Or separately?

lucasrodes:
After briefly talking with Fiona, she remembers this happening after:

1. create steps with express
2. worked on them
3. run steps, it worked
4. imported dataset to DB (without harmonizing)
5. harmonized using the harmonize countries button in wizard

=> Meadow step was reset ❌

Then she tried:

1. create steps with express
2. worked on them
3. run steps, it worked
4. imported dataset to DB (without harmonizing)
5. harmonized using terminal

=> Everything worked just fine ✅ 


**Hypothesis**: Might be something with the Wizard harmonize, or maybe it just works the second time that it is executed? It's very unclear to me what might be going on here.

lucasrodes:
After a more careful look into this issue, we found out that this happens after clicking on the "Harmonize" button shown in the instructions in Data Step (after submitting for step creation). If the user clicks, it looks like the steps are created again.

Solution: encapsulate button logic under a `st.fragment`.

PR #3865, opened 2025-01-20
🐛 harmonizer resets step code

lucasrodes:
fixes https://github.com/owid/etl/issues/3861

In the "data step" page, after creating some data steps, one can go to "Harmonizer" via a modal view. When the user clicks on that button (shown in the image below), the code in the created steps are reset to the templates.

It appears to me as if the "clicking" of the button triggers a new creation of the step files. 

image

/schedule


spoonerf (approved):
Nice one! Just checked it and the button no longer resets meadow for me :) Thanks for fixing!!

PR #3869, opened 2025-01-21
📊 marriages&divorces: add ONS data on people ever married by age

veronikasamborska1994:


etl/steps/data/garden/ons/2025-01-21/marriages.meta.yml
```diff
@@ -0,0 +1,35 @@
...
+  marriages:
+    variables:
+      cumulative_percentage_per_100:
+        title: Proportions of men or women who had ever married by a certain age for << birth_cohort >> birth cohort
+        unit: "%"
+        short_unit: "%"
+        description_short: |-
+          The percentage of men or women born in a specific year who have ever married by a certain age.
+        description_key:
+          - Ages are presented as "exact years," meaning "by age 30" includes marriages up to the day before the 30th birthday.
```
pabloarosado, line 25: I think the comma should be outside of the double quotes:
"exact years,"
but I think this is not specified in the writing and style guide, so up to you. There is a similar case later with "ever married,".

pabloarosado (approved):
Really interesting data! :-) The code and charts look great. I just pointed out a few minor things in the code, feel free to disregard them. Regarding the charts:
* In divorces, you can remove the "The " at the beginning. And there's a typo: "anniversy".
* In marriages, I think the "(x-axis)" may not be necessary, and possibly neither the "(from 17 to 50 years)", since that's clear when looking at the chart.

PR #3871, opened 2025-01-21
🐛 fix monitor

lucasrodes:


lucasrodes:
@pabloarosado I've modified couple of things, should work now.

What happened before, is that the "select steps" button was not showing up.

pabloarosado:
Thanks, now things can be moved from Preview to Selection. But it seems that the buttons to Update/Archive have disappeared too:
Screenshot 2025-01-22 at 17 33 01

lucasrodes:
This is turning to be very incremental, sorry for the constant updates!

I think it should be fine now.

pabloarosado:
Thanks a lot @lucasrodes, I've just tried to do a mock update and it worked as expected (and it looks better than it used to!).

pabloarosado (approved):
I haven't looked through the code, but I just ran it and worked well, so the issue is fixed. Thanks!

PR #3874, opened 2025-01-22
🐛 Use correct GRAPHER_USER_ID from env when working locally

Marigold:
There was a bug when working with indicator upgrader locally that used Admin grapher user id instead of the one defined in `.env`. This was caused by overwriting `GRAPHER_USER_ID` when working on a staging server. This PR adds new variable `ENV_GRAPHER_USER_ID` that doesn't get overwritten.

In addition to that, it slightly refactors function for getting grapher user id from Tailscale IP.

### DIFF
--- a/apps/wizard/utils/cached.py
+++ b/apps/wizard/utils/cached.py
@@ -11,7 +11,7 @@ from sqlalchemy.orm import Session
 
 import etl.grapher.model as gm
 from apps.utils.map_datasets import get_grapher_changes
-from etl.config import OWID_ENV, OWIDEnv
+from etl.config import ENV_GRAPHER_USER_ID, OWID_ENV, OWIDEnv
 from etl.db import get_engine
 from etl.git_helpers import get_changed_files
 from etl.grapher import io as gio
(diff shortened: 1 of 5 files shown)


apps/wizard/utils/cached.py
```diff
@@ -215,21 +215,28 @@ def get_tailscale_ip_to_user_map():
...
+        raise ValueError(f"No Github username found for IP address {user_ip}")
 
     with Session(get_engine()) as session:
-        grapher_user = gm.User.load_user(session, github_user_name)
-
-    if grapher_user:
-        return grapher_user.id
-    else:
-        return None
+        return gm.User.load_user(session, github_username=github_username)
```
lucasrodes, line 242: I think I find the current logic a bit confusing. My concern is that a call like `get_grapher_user()` might behave differently in staging or local environments (e.g. might return `ETL` user in the former, and `Lucas` in local, right?).

My proposed changes would be:

- Split the function into two, e.g. `get_grapher_user_from_ip` and `get_grapher_user_from_env`.
- Automatically extract the user ip when using `get_grapher_user_from_ip`, so that we don't have to call `st.context.headers.get("X-Forwarded-For")` each time we use it. 
- For the above to work, we'd get rid of the `st.cache_decorator` from the function, but we could add it to an internal one.
- Optionally, we could have a function `get_grapher_user` which conditions the use of `get_grapher_user_from_ip` or `get_grapher_user_from_env` based on the where Wizard is being executed.


```python
def get_grapher_user_from_ip():
	"""Get the Grapher user associated with the given Tailscale IP address.

	If no IP is given, it'll try to extract it from the header context.
	"""

	# Get ip if none is given
	user_ip = st.context.headers.get("X-Forwarded-For")

	@st.cache_data
	def _get_grapher_user_from_ip(user_ip: str):

		# Get Tailscale IP-to-User mapping
	    ip_to_user_map = get_tailscale_ip_to_user_map()

	    # Get the Tailscale display name / github username associated with the client's IP address
	    github_username = ip_to_user_map.get(user_ip)

	    if not github_username:
	        raise ValueError(f"No Github username found for IP address {user_ip}")

	    with Session(get_engine()) as session:
	        return gm.User.load_user(session, github_username=github_username)

	return _get_grapher_user_from_ip(user_ip)

```


```python
@st.cache_data
def get_grapher_user_from_env():
	"""Get the Grapher user based on the environment variable.

	This function is typically used when working in localhost.
	"""
	# Use local env variable if user_ip is not provided (when on localhost)
    if user_ip is None:
        with Session(get_engine()) as session:
            assert ENV_GRAPHER_USER_ID, "GRAPHER_USER_ID is not set!"
            return gm.User.load_user(session, id=int(ENV_GRAPHER_USER_ID))
```


Then, if want to condition the use of one or the other based on the current setting we could have something like:

```python
from etl.config import OWID_ENV
def get_grapher_user():
	if OWID_ENV.env_local == "dev":
		return get_grapher_user_from_env()
	else:
		get_grapher_user_from_ip()
```

lucasrodes (approved):
LGTM, I've added a suggestion on how to structure the `get_grapher_user` slightly differently.

PR #3876, opened 2025-01-22
📊 marriages&divorces: historical US rates

veronikasamborska1994:


Marigold (approved):
Looks good, charts are an improvement compared to the previous version.

Issue #3881, opened 2025-01-22
Replace Bugsnag by Sentry


Marigold:
We’ve recently replaced Bugsnag with Sentry in owid-grapher, and there’s no specific reason to continue using Bugsnag in ETL. It’s currently used only for bug reporting and notifications, which Sentry can handle equally well.

## TODO
- [x] Migrate to Sentry
- [x] Add Sentry key to envs
- [x] Send notifications to #analytics-errors

PR #3882, opened 2025-01-22
✨ Switch from bugsnag to sentry

Marigold:
Implements https://github.com/owid/etl/issues/3881

- Switch from Bugsnag to Sentry
- Fix function `enable_sentry_for_streamlit` that's forwarding exceptions from streamlit to Sentry
- Run `enable_sentry_for_streamlit` when importing `apps.wizard.utils` to enable it for all our apps (running it in utils module looks fishy, but I couldn't find a better way. It doesn't work when run from `apps/wizard/app.py`)

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -15,7 +15,7 @@ It doesn't work for old non-ETL datasets and older steps that don't use `create_
 
 ## Observability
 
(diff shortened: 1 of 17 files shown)


lucasrodes (approved):
LGTM!

PR #3892, opened 2025-01-24
🐛 regional data for human rights indicators

antea04:
**_Summary_**: I have added two more regions to our existing UN SDG data set and cleaned up/ polished our SDG chart landscape a bit. 

----
_Longer version:_
[User feedback](https://owid.slack.com/archives/C3RAU3BJ9/p1737563086844659) mentioned that this chart on [confirmed killings of human rights defenders, journalists and trade unionists](https://ourworldindata.org/grapher/cases-of-killed-human-rights-defenders-journalists-trade-unionists) was missing regional data for the middle east. 

Pulling on that thread I have uncovered that we indeed had removed two UN regions `"Northern Africa and Western Asia" and "Australia and New Zealand"` from our SDG data (in the last two updates!). 

For most charts this is not really an issue, because the UN provides more granular regions (e.g. North Africa/ Sub-saharan Africa) or continents (Africa, Asia), but some charts were missing data on the Middle East completely. I have now gone through the charts and added data where necessary. I have also taken that opportunity to clean up the SDG charts a bit.

----
Charts I wanted to flag/ would remove:
- [Share of indicators drawn from country-led frameworks](https://ourworldindata.org/grapher/use-of-crf-tools-by-providers-of-dev-cooperation) - It is really hard to understand what this measures and I don't think rewriting the FASTT etc. would help. It is used in our SDG tracker though and the only visualisation for indicator 17.15.1, so that might be reason enough to keep it. 
- [Number of countries with a national statistical plan fully funded and under implementation, by source of funding](https://ourworldindata.org/grapher/countries-with-stats-plan-funded-by-source) - Same here. This chart also shows multiple different things (source of funding, progress of funding, progress of implementation) and is hard to interpreted. Once again, in the SDG tracker as sole visualisation for indicator 17.18.3. 
- [Share of women aged 20-24 who were married before 18](https://ourworldindata.org/grapher/percentage-of-women-20-24-who-were-married-or-in-union-before-18) We have [this chart](https://ourworldindata.org/grapher/women-married-by-age-18?time=2023) which goes back further and has more data. But Bastian and Veronika might be [on this right now](https://github.com/owid/owid-issues/issues/799). (@veronikasamborska1994?)
- DRAFT: [Share of individuals who own a mobile telephone, by sex](http://staging-site-sdg-human-rights/admin/charts/2047/edit). We have [this scatterplot](https://ourworldindata.org/grapher/share-of-individuals-who-own-a-mobile-telephone-by-sex), which has the same information in a more interpretable way. The chart is apparently referenced in the (DRAFT) SDG explorer, otherwise I would have deleted it already. 
- DRAFT: [Manufacturing jobs as a share of total employment](http://staging-site-sdg-human-rights/admin/charts/8196/edit) should probably be a narrative chart of [this indicator/ chart.](https://ourworldindata.org/grapher/manufacturing-share-of-total-employment). But we can keep it if it is still used.


spoonerf (approved):
Looks good, thanks for catching this!!

spoonerf:
Hey @antea04!

Thanks for thinking about these improvements. 

For the two charts for Goal 17, it's a little tricky as we have made the commitment to show data for SDG indicators where there is data available, but I agree these aren't the easiest to understand. 

Do you think there are many other charts in the SDG Tracker that would fall under this category? I know we have a number where there are very few data points. 

It could be worth rethinking about how we show data for indicators where there _is_ data, but it's not necessarily up to a standard we would want to have a chart for. We could have a broader conversation in D&R about how we could address this. At the moment it feels like we've painted ourselves into a corner! Could be a good topic for a D&R meeting if you are up for pulling something together?

For now, I think what we can do is tag them as 'Unlisted' to limit where they show up (I can do this now) and if you could try and improve the FAUST as much as you can in the meantime that would be great. 

Agree with you and Veronika on the marriage chart and  you can delete the two draft ones (they can be removed from the draft explorer too).

antea04:
Hi @spoonerf!

Thanks for reviewing the PR and the charts! I understand - I had assumed there was an underlying reason for having these charts in the first place.

I have now edited the FASTT and presentation of the two linked charts with the aim to make them easier digestable. I will take another look at the rest of the SDG explorer and if this is a reoccuring problem I'll prepare a discussion for a D&R meeting. If it just affects a handful of indicators maybe we can just rework them/ decide together with Ed how to best handle them.

Thanks for your review & insight! ✨

PR #3894, opened 2025-01-24
📊 famines: add new wpf dataset

veronikasamborska1994:


Marigold (approved):
Code & charts look good. Ignore that error in CI, it's grapher related.

PR #3895, opened 2025-01-27
✨ Add indicator filter to chart-diff

Marigold:
- Add filtering by used indicator to chart-diff
- Fix `Variable.from_id` when `columns` are used (previously it returned only the first number in a row, now it returns Row object that is not `Variable`, but `Row` object that has the same fields)


lucasrodes (approved):
Thanks for addressing this little issue! Works for me.

I added (and rejected) a chart change to test it on more indicators.

PR #3915, opened 2025-02-02
✨ Add suport for python functions in Jinja

Marigold:


Marigold:
Macros turned out to be a better option for our use cases. We can always reopen this.

Issue #3609, opened 2024-11-25
Metadata: improve our source names for WDI indicators


lucasrodes:
### One-liner
Improve the source name for WDI indicators by specifying the actual source (where WB got it from).

### Context & details
We have the [World Development Indicators](https://admin.owid.io/admin/datasets/6543) dataset, which contains a large list of indicators from multiple sources. By default, charts containing indicators from this dataset display the source "Multiple sources compiled by World Bank (2024)":

![Image](https://github.com/user-attachments/assets/1bd7365d-7abb-47af-a8e7-b2e6be3e6ac5)

This is because we currently assign this source name to _all_ indicators from WDI in ETL. However, this is not accurate enough in many cases, where we'd want to cite the actual source of the data. 

To overcome this, we tend to edit the source chart-by-chart. This, however, is sub-optimal as maintenance can be a bit of a burden, especially when there are data updates.

Therefore, we need to edit the metadata of WDI in ETL and cite the actual source depending on the indicator.

lucasrodes:
[Slack discussion](https://owid.slack.com/archives/C04LNF8LBA8/p1732541791473569)

spoonerf:
In the olden days of the `importers` repo, the WDI did have a [source standardizing JSON](https://github.com/owid/importers/blob/master/worldbank_wdi/config/standardized_source_names.json), which was used to create pretty versions of the source provided by the World Bank where possible. It was quite heavily curated, with input from Max and Hannah on what each source should be called.  

Perhaps a version of this could be used in the ETL to help solve this problem?

We also have a similar issue with the UN SDG dataset, which @antea04 did some great work on in the most recent update. You can see the relevant code [here](https://github.com/owid/etl/blob/master/etl/steps/data/grapher/un/2024-08-27/un_sdg.py).

pabloarosado:
@Marigold said he can look into it on the next WDI update. But please anyone speak up if you think this should be fixed earlier than that.

PR #3916, opened 2025-02-02
📊 WDI - use origins from WB metadata instead of 'Multiple sources compiled by WB'

Marigold:
Fixes https://github.com/owid/etl/issues/3609

- DRY origins in WDI and use only the one from snapshot
- Remove "attribution" from origin so that charts don't show "Multiple sources compiled by World Bank" and instead show real source from WB metadata
- Raw names from WB and shortened "name" we use in origins are stored in [wdi.sources.json](https://github.com/owid/etl/blob/master/etl/steps/data/garden/worldbank_wdi/2025-01-24/wdi.sources.json)

## How to review
Check out some charts with changed metadata from [chart-diff](http://staging-site-wdi-fix-origins/etl/wizard/chart-diff?show_all=). `Data source:` in the footer should be more meaningful.

### DIFF
--- a/etl/steps/data/grapher/worldbank_wdi/2025-01-24/wdi.meta.yml
+++ b/etl/steps/data/grapher/worldbank_wdi/2025-01-24/wdi.meta.yml
@@ -1,34 +1,10 @@
-definitions:
-  # NOTE: we'd like to move this to snapshot, but it's impractical because there are
-  # tons of other downstream datasets that still use sources
-  wdi_origin: &wdi_origin
-    producer: World Bank
-    title: World Bank World Development Indicators
-    description: |-
-      The World Development Indicators (WDI) is the World Bank's primary collection of development indicators, compiled from officially-recognized international sources. It presents the most current and accurate global development data available and includes national, regional, and global estimates.
-
-      [Text from [World Bank data catalog](https://datacatalog.worldbank.org/dataset/world-development-indicators)]
-    citation_full: World Development Indicators, The World Bank (2025).
-    url_main: https://datacatalog.worldbank.org/search/dataset/0037712/World-Development-Indicators
-    url_download: |-
-      https://datacatalogfiles.worldbank.org/ddh-published/0037712/DR0045575/WDI_CSV_2024_10_24.zip?versionId=2024-12-18T05:42:41.1131485Z
-    date_accessed: '2025-01-24'
-    date_published: '2025-01-15'
-    license:
-      name: Creative Commons Attribution 4.0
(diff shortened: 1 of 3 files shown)


paarriagadap (approved):
Hi @Marigold! This is awesome! From a quick review the sources now look much more meaningful.

With this [datasette query](http://analytics/private?sql=SELECT%0D%0A++DISTINCT+c.id%2C%0D%0A++c.slug%2C%0D%0A++json_extract%28cf.full%2C+%27%24.sourceDesc%27%29+AS+sourceDesc%0D%0AFROM%0D%0A++charts+c%0D%0A++JOIN+chart_configs+as+cf+ON+c.configId+%3D+cf.id%0D%0A++JOIN+chart_dimensions+cd+ON+c.id+%3D+cd.chartId%0D%0A++JOIN+variables+v+ON+cd.variableId+%3D+v.id%0D%0A++JOIN+datasets+d+ON+v.datasetId+%3D+d.id%0D%0AWHERE%0D%0A++json_extract%28cf.full%2C+%27%24.sourceDesc%27%29+IS+NOT+NULL%0D%0A++AND+v.catalogPath+IS+NOT+NULL%0D%0A++AND+c.publishedAt+IS+NOT+NULL%0D%0A++AND+d.id+%3D+6912%3B) I found the WDI charts with manually overwritten source fields. I have now "chained" to the dataset, but these are not looking nice. Is there a way to quickly change them in the database?

http://staging-site-wdi-fix-origins/admin/charts/5030/edit
http://staging-site-wdi-fix-origins/admin/charts/2437/edit
http://staging-site-wdi-fix-origins/admin/charts/4893/edit
http://staging-site-wdi-fix-origins/admin/charts/2900/edit
http://staging-site-wdi-fix-origins/admin/charts/5029/edit

etl/steps/data/grapher/worldbank_wdi/2025-01-24/wdi.meta.yml
```diff
@@ -1,34 +1,10 @@
-definitions:
```
paarriagadap, line 1: Would it be better to move all this to Garden or are there problems with that?

paarriagadap: I could try and do that, but there is this unusual message in VS Code and I don't want to ruin the PR:
image

Marigold: Good point! I moved it to `garden/wdi.meta.override.yml` since `wdi.meta.yml` is mostly automatically generated.

paarriagadap (approved):
Great, let's merge then!

PR #3917, opened 2025-02-02
✨ Simplify GBD metadata with Jinja macros

Marigold:
## Motivation

GBD metadata is currently non-manageable. Even simple fix such as adding exception to `cause` ballooned into [this monster](https://github.com/owid/etl/pull/3900/files). We need a way how to DRY it.

## Solution

Add support for `shared.meta.yml` where user can add `definitions:` or `macros:` and use them across other YAML files.

Use macros instead of raw jinja. Macros are very similar to python functions and can be parametrized, which allowed me to DRY a lot of duplicated code. I used o1 and o3 heavily when refactoring & DRYing those functions. There's certainly more space for a cleanup, but I didn't have enough time to polish it.

## Alternatives

At first, I thought that [registering custom python functions](https://github.com/owid/etl/pull/3915) would be the best solution, but it was overly complex and didn't really solve the problem. One of the issues was that python functions would have to be defined in the grapher step, while YAML metadata would live in a garden step.

## GBD metadata

Besides DRYing metadata, I fixed a couple of whitespace issues.


lucasrodes:
we should make sure that this doesn't lead to the behavior [here](https://owid.slack.com/archives/C46U9LXRR/p1738598598818899).

spoonerf (approved):
Looks awesome, thanks Mojmir! There is a still a whitespace issue with HIV/AIDs that Lucas noticed, which I've had a go at getting rid of in a separate PR. but I think it requires a sacrifice to the Jinja gods 😅 

Strangely it looks okay in the playground notebook, but will have more of a play tomorrow.

PR #3932, opened 2025-02-04
Update multidim-schema.json

rakyi:
* Add `grapherConfigSchema`
* Add `metadata` to top level and view level
* Add a new way to specify indicators with an object with an optional `display` override
* Remove unused properties `dimensions_title` and `name`
* Use refs instead of reproducing the grapher config

I used [this JSON schema viewer](https://json-schema.app/start) to view the config and confirm it's correct.


schemas/multidim-schema.json
```diff
@@ -740,222 +178,243 @@
         "views"
     ],
     "$defs": {
-        "axis": {
+        "indicatorConfig": {
```
rakyi, line 181: This is a new way to specify the indicator introduced for the display override.

schemas/multidim-schema.json
```diff
@@ -3,10 +3,9 @@
     "additionalProperties": false,
     "type": "object",
     "properties": {
-        "definitions": {
```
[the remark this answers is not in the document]
lucasrodes: The field `definitions` is currently used by some steps as a way to define variables that are referenced from various places in the metadata file. Take this example: https://github.com/owid/etl/blob/master/etl/steps/export/multidim/covid/latest/covid.deaths.yml.

I wonder if we need to define this in the schema file or not though. Maybe so that VS Code doesn't complain? @Marigold

lucasrodes:
@rakyi I think it looks good. I will look at it with more care tomorrow and Friday. I'll test it with the COVID MDIM and see how it feels!

schemas/multidim-schema.json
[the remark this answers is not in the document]
Marigold: I'd keep it, we put reused things like anchors into `definitions` (and could later have more in there, like `definitions.common`)

schemas/multidim-schema.json
```diff
@@ -740,222 +178,243 @@
...
+                        },
+                        "retrievedDate": {
+                            "type": "string"
+                        },
+                        "additionalInfo": {
+                            "type": "string"
+                        }
+                    }
+                },
+                "origins": {
```
Marigold, line 279: We have some shared definitions in [definitions.json](https://github.com/owid/etl/blob/master/schemas/definitions.json#L144), so it should be possible to reuse it with
```
"origins": {
  "type": "array",
  "items": {
    "$ref": "definitions.json#/origin"
  }
}
```
If these origins are somehow different, then feel free to keep yours.

Marigold (approved):
Nice cleanup. I checked our current YAML files, and they show a lot of errors now, but I guess we can fix it when we work on them, right?

schemas/multidim-schema.json
rakyi: I'd like to reuse the existing definition, however in the mdim config we expect the property names to be camel case, and the existing definition in ELT uses snake case. Do you know if/where does the conversion happen?

rakyi:
> I checked our current YAML files, and they show a lot of errors now, but I guess we can fix it when we work on them, right?

That should work. The format on the grapher side shouldn't have a breaking change.

schemas/multidim-schema.json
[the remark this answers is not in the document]
rakyi: I put them back in. For the future I think it would be valuable to keep the external API schema (in this case grapher admin API) separate from the internal one to make the distinction clear.

schemas/multidim-schema.json
Marigold: > Do you know if/where does the conversion happen?

Hmm, that's a bummer. We switch from underscores to camelCase right before upserting objects (origins, indicators) to MySQL. I sincerely don't know if it's better to be consistent with the rest of ETL or with grapher.

Keep it as it is now, and we can revisit this at some point in the future.

Marigold: (I assume that Admin API validates field names, right?)

schemas/multidim-schema.json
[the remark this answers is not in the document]
lucasrodes: I'm happy to make that distinction from now, since we have few MDIMs. What would we have to do ?

Maybe have `schemas/multidim-schema-grapher.json` be the grapher one and, on top of it, then `schemas/multidim-schema.json` be the internal one with added fields, like `definitions`?

lucasrodes:
Hi, as I test some of this, I think that some points of discussion might be worth just tackling in this thread since we are at an early stage of MDIMs, and maybe it's harder in the future to attack these.

From what I've read so far, I see these two issues:

1. `camelCase` vs `snake_case`
  It seems that there is some field-name conversion between ETL and Grapher. The former uses snake case, and the latter uses camel case.
  Should we settle on a solution here? I'm fine either way, but I want to avoid future headaches. My understanding is that our MDIM metadata currently uses snake-case and camelCase property names.
2. ETL schema might have more fields than the strictly necessary ones by Grapher
  At the moment, `schemas/multidim-schema.json` includes Grapher and internal fields, so VS Code's syntax assistance is helpful in the context of ETL.
  - Martin suggested to keep the "grapher strictly necessary"-schema separate from our internal one (which adds more fields, e.g. `definitions`). I agree with this point, and I think we could do this now since we are just kicking this off? 
3. Related to point 2, VS Code now complains about some properties being "invalid". As an example, take `config`, which is highlighted as invalid. Instead, based on the schema, it should be `grapherConfig`, but this doesn't really work. 
    image


I am now just a bit confused about the role of this MDIM schema. It feels that I am missing a piece between an MDIM metadata YAML file and the proposed schema file. Is there something in the middle handling some re-namings? Should we maybe work on having a settled schema for ETL that enriches the Grapher schema? (point 2). I personally think that it would be very valuable for the team when working on MDIMs, otherwise it might be unclear what we can and cannot do.

https://github.com/owid/etl/blob/c3bc53f122ca748c18f5564386f409b9aec1e90e/etl/steps/export/multidim/covid/latest/covid.cases.yml#L64

lucasrodes:
I was trying to set some display settings, and I think we possibly need to change something on the ETL front (@Marigold).

I’m trying to set some display settings (see here)

```yml
indicators:
      y:
        catalogPath: "{definitions.table}#weekly_cases"
        display:
          name: "something"
```

But when I run `etlr multidim/covid/latest/covid --export --private`, I get a:

```
TypeError: dict can not be used as parameter
```

If I look into the complete traceback, the error seems to first originate from calling `multidim.upsert_multidim_data_page`:

```
  File "/home/lucas/repos/etl/etl/steps/export/multidim/covid/latest/covid.py", line 28, in run
    multidim.upsert_multidim_data_page(slug, config, engine)
  File "/home/lucas/repos/etl/etl/multidim.py", line 20, in upsert_multidim_data_page
    validate_multidim_config(config, engine)
  File "/home/lucas/repos/etl/etl/multidim.py", line 88, in validate_multidim_config
    df = read_sql(q, engine, params={"indicators": tuple(indicators)})
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucas/repos/etl/etl/db.py", line 77, in read_sql
    return pd.read_sql(sql, con, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

lucasrodes:
As an update to my [previous message](https://github.com/owid/etl/pull/3932#issuecomment-2643345869), I've tweaked `etl.multidim` to support the dictionary-based definition of indicators in `views`. 

In addition, I've generalized the current approach to not only check for indicators in dimension `y` but also in `x`, `size` and `color`, if existant.

lucasrodes:
I've incorporated the changes from https://github.com/owid/etl/pull/3945 here!

lucasrodes:
This seems good as it is now, and low-risk to merge.

Merging to avoid making this too big.

Thanks for the hard work everyone!

schemas/multidim-schema.json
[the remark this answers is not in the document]
rakyi: Hmm, I'm not sure if it's possible (or a good idea) to inherit/extend schemas, but if it is, we should probably have the base mdim schema defined in `owid-grapher` repo and then extend/reference it here, similarly to how I added the reference to the grapher config schema in this PR.

schemas/multidim-schema.json
rakyi: Unfortunately, we do only minimal validation on the input in the admin, currently.

> We switch from underscores to camelCase right before upserting objects (origins, indicators) to MySQL.

This seems fine, and we could probably just do that everywhere. But the schema is (would be) a bit murky, since it's more about what we do in ETL and less about what the admin API expects. Which might be fine, I just assumed it's more of a contract between ETL and admin, and didn't realize it's mostly used as ETL internal DX affordance.

rakyi:
Thanks for the additional fixes, Lucas! As I wrote [above](https://github.com/owid/etl/pull/3932#discussion_r1950701359), I was somewhat confused about the purpose of the schema. I thought it's more about capturing the contract the admin API provides, but now I understand it's more of an internal DX affordance for ETL.

So we should probably change it to be consistent with the rest of ETL, e.g. by using snake case instead of camel case, reusing the existing `origins` definition, etc. If you think we should do this sooner rather than later, let's discuss it on data chat on Monday.

rakyi:
> 3\. Related to point 2, VS Code now complains about some properties being "invalid". As an example, take `config`, which is highlighted as invalid. Instead, based on the schema, it should be `grapherConfig`, but this doesn't really work.

Did you manage to solve this? There should be no change to the meaning of `config`. The new field is `grapherConfigSchema`, which just specifies the schema of the `config`.

lucasrodes:
@rakyi This is not a blocker, just a better dev experience. I think we should look at this whenever we have a moment, but not high priority. Others in the data team might appreciate it & make it easier to learn how to implement MDIMs.

`config` continues to be highlighted with message `Missing property "dimensions".yaml-schema: file:///home/lucas/repos/etl/schemas/multidim-schema.json`:

image

Happy to discuss it anytime.

rakyi:
I see, that's because I didn't notice the inline copy of the grapher schema wasn't the same as the upstream one I swtiched to using `$ref`. I'll try to fix it.

PR #3942, opened 2025-02-06
🐛 Fix un_members dependency on grapher regions

Marigold:
@lucasrodes I had to revert your archival of grapher regions because there was still one dataset using it.

This PR updates `un_members` dataset to use garden version of regions (plus a list of mappable countries that are in grapher regions).


lucasrodes:
I see, I didn't catch that dependency, thanks for being on top of it!

How stable is this change? I am not sure I understand the nature of that list. Will that variable be the "source of truth" for that? Or how did we obtain that list? Maybe good to add a comment if it was manually extracted, just for future self.

Marigold:
> How stable is this change? I am not sure I understand the nature of that list. Will that variable be the "source of truth" for that? Or how did we obtain that list? Maybe good to add a comment if it was manually extracted, just for future self.

The list `MAPPABLE_COUNTRIES` was copied from [grapher/regions/latest/regions](https://github.com/owid/etl/blob/master/etl/steps/data/grapher/regions/latest/regions.py#L23) step. It was originally created by @pabloarosado. I'll add a note where it comes from, and we can revise whether it's necessary for the UN dataset on next update.

lucasrodes (approved):

pabloarosado:
Thanks @lucasrodes! At some point we should fix version tracker to prevent issues like this (CI should ideally fail if VT fails because of a broken dependency; but currently it fails for various reasons).
Thanks @Marigold, your solution is good. That list of mappable countries is actually an input from grapher. That's the list of countries that appear in grapher charts. It's not ideal to have it hardcoded like that, but it's unlikely that this list will change, so it's probably fine.

PR #3945, opened 2025-02-07
🔨 Refactor mdims

Marigold:
- Allow referencing indicators with `table#indicator` instead of full catalogPath
  - E.g. `grapher/energy/2024-06-20/energy_mix/energy_mix#coal__twh` -> `energy_mix#coal__twh`
  - It's expanded dynamically in `multidim.upsert_multidim_data_page`, but you have to pass `dependencies=paths.dependencies` (I couldn't think a more elegant way, but I'm all ears if you do)
- Add argument `.read(..., load_data: bool)` which doesn't load data when used. This significantly reduces loading times for huge datasets like GBD


lucasrodes:
Thanks for looking into this.

I've added the changes to `etl/multidim.py` and in the main PR: https://github.com/owid/etl/pull/3945. This way, I can also implement these changes in #3949.

lucasrodes (approved):
LGTM!

PR #3946, opened 2025-02-07
📊 climate: monthly era5 surface temperature update

veronikasamborska1994:
hey @Marigold! we could probably set this code to be automatically updated until next year if that's still something you are keen on doing?


Marigold:
Yep! I plan to look into this when I have some capacity.

PR #3947, opened 2025-02-07
🐛 child mortality anomalies

antea04:
Fix for [this issue](https://owid.slack.com/archives/C03NV9Z3YSV/p1738774206362549), where regional aggregates for child mortality rates fluctuate strongly. This is caused by new countries being added to regional aggregates over time, influencing the overall rate.

I have now changed the regional aggregates to only be calculated if data was available for >80% of the population. This smoothes out the lines and removes most data before ca 1965 (different for different regions).

There are two charts I had issues with:
- [Life satisfaction vs. child mortality](http://staging-site-bug-child-mortality-anomalie/admin/charts/250/edit) - For some reason the Cantril ladder score does not pull the latest year into the scatter plot (it is stuck on 2015). I assume it has something to do with [this bug](https://owid.slack.com/archives/C46U9LXRR/p1739242868670479) - I am happy to leave it for now and would try to update it once the bug is resolved. 
- [Child mortality in the past vs. today](http://staging-site-bug-child-mortality-anomalie/admin/charts/867/edit) - This is a draft chart without any entities selected and the same data as in other available line charts. I would delete it.

### DIFF
--- a/etl/steps/data/garden/un/2024-09-11/igme.py
+++ b/etl/steps/data/garden/un/2024-09-11/igme.py
@@ -48,10 +48,13 @@ def run(dest_dir: str) -> None:
     tb = filter_data(tb)
     tb = round_down_year(tb)
 
-    # add regional data for count variables
-    tb = add_regional_totals_for_counts(tb, ds_regions)
-    # add regional population weighted averages for rate variables
-    tb = add_population_weighted_regional_averages_for_rates(tb, ds_population, ds_regions)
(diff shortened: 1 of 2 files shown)


spoonerf (approved):
Looks good! Thanks for sorting this out :)

I fixed the broken life satisfaction vs child mortality chart (just removing both variables and adding them again seemed to do it). I agree the draft chart can be deleted.

PR #3949, opened 2025-02-07
📊 work on covid mdims

lucasrodes:


lucasrodes:
Work resumed!
Currently looking at each MDIM and checking its FAUST.

lucasrodes:
Substantial progress done.

Besides completing the list of pending MDIMs, I need to review if for bar charts (e.g. `covid.covax`) there is a way to sort bars using `table_name#indicator_name`. For that, need to process `config.sortColumnSlug`

lucasrodes:
All MDIMs here are good to go. I'll merge this PR when CI/CD is passed.

As future work:

- In views with more than one indicator:
  - find a short-term (data work) solution to present the relevant metadata.
  - More on the longer-term, we could investigate some re-design of MDIM pages.
- Could be nice to change default entity selections.

PR #3951, opened 2025-02-10
🐛 fix entities in GHO

spoonerf:
Some odd-looking entities were still being included in the data, so just removing those. 

We also had cases where regions from five different sources were being used, so now we just take the top two of ["World Bank", "WHO", "UNICEF", "UN", "UN SDG"], if there are more than two sources of regions being provided for a particular indicator.

### DIFF
--- a/etl/steps/data/garden/who/2024-01-03/gho.excluded_countries.json
+++ b/etl/steps/data/garden/who/2024-01-03/gho.excluded_countries.json
@@ -10,7 +10,6 @@
   "Asia (142)",
   "Least Developed Countries (LDCs)",
   "Northern America (21)",
-  "GlobalGlobal",
   "Europe (150)",
   "Small Island developing States (SIDS)",
   "Small island developing States (SIDS)",
--- a/etl/steps/data/garden/who/2024-01-03/gho.excluded_countries.json
+++ b/etl/steps/data/garden/who/2024-01-03/gho.excluded_countries.json
@@ -10,7 +10,6 @@
   "Asia (142)",
   "Least Developed Countries (LDCs)",
   "Northern America (21)",
-  "Northern America",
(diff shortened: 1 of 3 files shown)


etl/steps/data/garden/who/2024-01-03/gho.excluded_countries.json
```diff
@@ -1,7 +1,19 @@
...
+  "Europe (150) (UN)",
+  "Oceania (9) (UN)",
+  "Africa (2) (UN)",
+  "Asia (142) (UN)",
+  "Africa (2)",
+  "Asia (142)",
+  "Least Developed Countries (LDCs)",
+  "Northern America (21)",
+  "Northern America",
+  "GlobalGlobal",
```
pabloarosado: Isn't GlobalGlobal both in the mapping and in excluded countries?

pabloarosado (approved):
Looks good. Strange, that such regions appear in the data, like "Africa (2)".

etl/steps/data/garden/who/2024-01-03/gho.excluded_countries.json
spoonerf: Ah yes it was! I've removed it from here now.

spoonerf:
> Now that you are at it, wouldn't it make sense to add the content gho_omms.py into gho.py, to simplify code? As far as I can see, no other step is importing that module (but I may be missing context, so feel free to disregard this suggestion).

I'll leave it in for now as this is primarily now Mojmir's code and I am likely missing context too. The main aim was just to fix the issue raised [here](https://owid.slack.com/archives/C03NV9Z3YSV/p1739178761011049). 

@Marigold feel free to review when you're back, I just wanted to put this on your radar in case future region sources are included in the future.

PR #3955, opened 2025-02-10
🎉 Let export steps execute implicit grapher://grapher steps

pabloarosado:
Let the `--export` flag execute the corresponding `grapher://grapher` steps to all `data://grapher` dependencies. This allow us to remove `grapher://grapher` dependencies, and instead have the corresponding `data://grapher` ones.

NOTE: I'm not totally sure if this will work with mdims that depend on private steps (like `export://multidim/health/latest/causes_of_death`). For now, I omitted it from the dag.

### DIFF
--- a/dag/health.yml
+++ b/dag/health.yml
@@ -668,8 +668,8 @@ steps:
     - data://garden/health/2024-08-23/eurostat_cancer
 
   # Multi-dim indicators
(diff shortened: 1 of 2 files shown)


lucasrodes (approved):
LGTM! thanks for the implementation. Tested on COVID, and seems to work just fine

PR #3959, opened 2025-02-11
Fix mdim grapher schema

rakyi:
A fix for the issue(s) created in #3932

### DIFF
--- a/schemas/multidim-schema.json
+++ b/schemas/multidim-schema.json
@@ -162,7 +162,176 @@
                         ]
                     },
                     "config": {
-                        "$ref": "https://files.ourworldindata.org/schemas/grapher-schema.006.json"
+                        "type": "object",
+                        "description": "A subset of the grapher config.",
+                        "properties": {
+                            "$schema": {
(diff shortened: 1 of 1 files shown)


schemas/multidim-schema.json
```diff
@@ -108,13 +108,19 @@
                                     {
                                         "type": "string"
                                     },
+                                    {
+                                        "type": "number"
```
rakyi, line 112: I noticed we use IDs in some existing configs, and it is supported, so I added the option here.

lucasrodes (commented):
Unsure how much we should advertise this support. I wouldn't suggest we use indicator IDs in ETL, unless in a really exceptional situation (can't think of any now).

schemas/multidim-schema.json
rakyi: OK, I removed that change.

lucasrodes: @pabloarosado just to confirm this

lucasrodes (approved):
Works just fine! Thanks, Martin!

PR #3961, opened 2025-02-11
Update `topic_tags` dataset-schema.json enum with current values

ikesau:
## Changes:
- Updates the enum with latest values (alphabetizes the existing list first, so it's easier to compare the diff)
- Updates the datasette query
- Updates all instances of `Antibiotics` to `Antibiotics & Antibiotic Resistance`

Hopefully this will catch any other invalid tags in CI? Not sure if that's how this repo is set up though 🙂


lucasrodes (approved):
Thanks and keeping this up-to-date, Ike! LGTM, Feel free to merge once it passes CI/CD!

I've added some changes to `apps/wizard/etl_steps/utils.py`, which reduces one bit that was hardcoded. 

More details:
**PAST**: If couldn't access DB, we would use a hardcoded list of tags from the above python script.
**NOW**: If can't access the DB, we will use the list defined in the `enum` field from the schema file that you updated. If for some reason we couldn't access nor the DB neither the field in the schema, then we'll use a fallback (and hardcoded) list in the python script.

ikesau:
@lucasrodes presumably this is an expected CI failure and we're safe to merge?

lucasrodes:
Good catch, just a minor styling thing. Fixed!

PR #3965, opened 2025-02-12
📊 climate: add Equatorial Pacific Sea Surface Temperatures (SST) data

veronikasamborska1994:


snapshots/climate/2025-02-12/sst.py
```diff
@@ -0,0 +1,90 @@
...
+    Convert the season string to the corresponding month.
+
+    Parameters:
+    season (str): The season string (e.g., "DJF").
+
+    Returns:
+    int: The corresponding month (1-12).
+    """
+    season_to_month_map = {
+        "DJF": 2,  # December-January-February -> January
```
pabloarosado, line 30: Maybe clarify why you are mapping to the latter month instead of the middle one (which is what I understand from the comment on the right).

etl/steps/data/garden/climate/2025-02-12/sst.py
```diff
@@ -0,0 +1,52 @@
...
+    # Read table from meadow dataset.
+    tb = ds_meadow.read("sst")
+
+    #
+    # Process data.
+    #
+
+    tb["nino_classification"] = tb.apply(classify_nino_anomaly, axis=1)
+    for col in ["nino_classification"]:
+        tb[col].metadata.origins = tb["nino3_4_anomaly"].metadata.origins
```
pabloarosado, line 25: This is a bit slow. An alternative that would propagate metadata automatically would be something like:

```
    # Fill NA values with a number that won't affect classification
    tb["oni_anomaly"] = tb["oni_anomaly"].fillna(0)

    # Classify NINO3.4 and NINO4 anomaly values
    tb["nino_classification"] = tb["oni_anomaly"].copy()
    tb.loc[tb["oni_anomaly"] > 0.5, "nino_classification"] = 1
    tb.loc[tb["oni_anomaly"] < -0.5, "nino_classification"] = 2
    tb.loc[(tb["oni_anomaly"] >= -0.5) & (tb["oni_anomaly"] <= 0.5), "nino_classification"] = 0
    tb["nino_classification"] = tb["nino_classification"].astype(int)
```

etl/steps/data/grapher/climate/2025-02-12/sst_by_month.py
```diff
@@ -0,0 +1,63 @@
...
+        lambda row: row["colour_date"] + 8
+        if (row["nino_classification"] == 1 and 1 <= row["colour_date"] <= 9)
+        else row["colour_date"],
+        axis=1,
+    )
+    tb["colour_date"] = tb.apply(
+        lambda row: 0 if row["nino_classification"] == 0 else row["colour_date"],
+        axis=1,
+    )
+    tb["colour_date"].metadata.origins = tb["oni_anomaly"].metadata.origins
```
pabloarosado, line 42: If metadata was lost along the way, instead of just recuperating origins, you can copy the entire metadata (if that's the intention), using
```
    tb["colour_date"] = tb["colour_date"].copy_metadata(tb["oni_anomaly"])

```

pabloarosado (approved):
Great work, and very good work on the metadata! I left some minor suggestions and small code improvements. Feel free to disregard them if you disagree.
The charts look great, just a few small suggestions for your consideration:
* Remove "(ONI)" from the title.
* Use "and" instead of "&" (I think that's usually the preferred option).
* In the metadata, I see "Contains modified...". This is an unusual phrasing, why not simply "Copernicus Climate..."?
* The chart per month is a bit hard to understand. What about showing fewer months?
* I think there may be spurious line breaks in the subtitle:
Screenshot 2025-02-14 at 13 22 29

pyproject.toml
```diff
@@ -70,6 +70,7 @@ dependencies = [
     "cfgrib>=0.9.15.0",
     "streamlit-aggrid>=1.0.5",
     "sentry-sdk>=2.20.0",
+    "streamlit>=1.41.1",
```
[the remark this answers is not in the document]
veronikasamborska1994: no, this was accidental :) thanks for spotting it!

veronikasamborska1994:
thanks a lot @pabloarosado! very useful! I've incorporated all of your suggestions with the exception of the "Contains modified materials..." bit because that's explicitly what the source wants us to do and we've already had this discussion with Este and agreed we should keep it as it is. 

> Great work, and very good work on the metadata! I left some minor suggestions and small code improvements. Feel free to disregard them if you disagree. The charts look great, just a few small suggestions for your consideration:
> 
> * Remove "(ONI)" from the title.
> * Use "and" instead of "&" (I think that's usually the preferred option).
> * In the metadata, I see "Contains modified...". This is an unusual phrasing, why not simply "Copernicus Climate..."?
> * The chart per month is a bit hard to understand. What about showing fewer months?
> * I think there may be spurious line breaks in the subtitle:
>   Screenshot 2025-02-14 at 13 22 29

veronikasamborska1994:
p.s. looks like we still need `yearIsDay: true`

PR #3966, opened 2025-02-12
📊 emissions: Get air pollution data

pabloarosado:
## Summary
* Create snapshot and data steps for CEDS data on air pollutant emissions by country, pollutant, and sector.
* Make air pollution explorer an indicator-based ETL explorer ([old explorer](https://ourworldindata.org/explorers/air-pollution) | [new explorer](http://staging-site-data-air-pollution/admin/explorers/preview/air-pollution)).
* Minor fix to `helpers.Explorer` object.

## Review
* Please feel free to play around a bit with the new explorer, and let me know if you have any feedback (Hannah is also having a look).
* @lucasrodes could you please have a look at the `helpers` file and the `export/explorers` step? I don't think we should put much effort on this (given our current uncertainty on explorer/mdims), however, I wanted to share with you a few comments, for possible improvements:
    * I think the `Explorer` object currently assumes an explorer file already exists in `owid-content`; maybe it would be good to let a new explorer be created from scratch.
    * It was failing because it expected the name of the explorer `air-pollution` to coincide with the name of the explorer step `air_pollution`. I let the `Explorer` object have a `name` argument to avoid this problem.
    * Another issue I noticed is that, if at any point you break the content of the tsv file (by pushing some content with a bad structure), the export step fails, and there's no easy way to restore the explorer from ETL. I fixed the issue by accessing the server and manually deleting the bad rows, but it would be better to have some easy way to revert/restart the Explorer.
    * The way the explorer was created is quite manual, I'm wondering if some of the new mdim functions could be reused here. If you think that's the case, please let me know.
* @spoonerf could you please have a look at the snapshot/meadow/garden/grapher steps? No need for a thorough review, but please let me know if you spot something odd.

Thank you both!


pabloarosado:
Hey @Marigold (not urgent) do you know if there's any good way to handle key descriptions from jinja? I wanted to have different key points depending on `sector`, but I couldn't figure out how to achieve that, so I did [an ugly workaround](https://github.com/owid/etl/blob/bc66914afc519c753a7809ae7c3d9f9c030a0c85/etl/steps/data/garden/emissions/2025-02-12/ceds_air_pollutants.meta.yml#L37). Thanks!

spoonerf (approved):
Looks super! Just a couple of minor suggestions but feel free to disregard them. 

For the explorer: 

- Under 'All pollutants' it seems as if 'Methane' and 'Nitrous oxides' are not currently shown. Also, I think it would make sense to show the pollutants in the same order as is given in the drop down, alphabetical.

- Tiny nit-pick but I don't think you need the '(BC)' in the black carbon views, unless it's an acronym used more broadly in the research area. AFAICT the other acronyms given are chemical symbols or NMVOCs, which I can imagine is broadly used.

Marigold:
@pabloarosado you can write it like this
```
description_key_combined: &description_key_combined
    - <% if sector in ["All sectors", "Agriculture"] %> {definitions.description_key_agriculture} <% endif %>
    - <% if sector in ["All sectors", "Buildings"] %> {definitions.description_key_buildings} <% endif %>
    - <% if sector in ["All sectors", "Domestic aviation"] %> {definitions.description_key_domestic_aviation} <% endif %>
    - <% if sector in ["All sectors", "Energy"] %> {definitions.description_key_energy} <% endif %>
    - <% if sector in ["All sectors", "Industry"] %> {definitions.description_key_industry} <% endif %>
    - <% if sector in ["All sectors", "International aviation"] %> {definitions.description_key_international_aviation} <% endif %>
    - <% if sector in ["All sectors", "International shipping"] %> {definitions.description_key_international_shipping} <% endif %>
    - <% if sector in ["All sectors", "Transportation"] %> {definitions.description_key_transportation} <% endif %>
    - <% if sector in ["All sectors", "Solvents"] %> {definitions.description_key_solvents} <% endif %>
    - <% if sector in ["All sectors", "Waste"] %> {definitions.description_key_waste} <% endif %>
    - Emissions assigned to "Other" are those that are included in the global total, but cannot be allocated to any specific countries.
```
empty bullet points are deleted before upserting to grapher.

(This "post-processing" should also happen when rendering the YAML file, unfortauntely right now you'd see a list with empty strings. I'll fix it as part of my dimension work. Here's a full example)
```
from etl.grapher import helpers as gh
from etl import paths

table = 'ceds_air_pollutants'

ds = Dataset(paths.DATA_DIR / f"garden/emissions/2025-02-12/ceds_air_pollutants")
tb = ds['ceds_air_pollutants']

dim_dict = dict(zip(tb.index.names, tb.sample(1, replace=True).index[0]))

yaml_path = f"{table}.meta.yml"
d1 = gh.render_yaml_file(yaml_path, dim_dict=dim_dict)
d1['tables']['ceds_air_pollutants']['variables']['emissions']['description_key']
```

pabloarosado:
Thank you for all the suggestions @spoonerf!

And thank you @Marigold for the clarification. I did notice that the empty key description points were not appearing in grapher, but I was wondering if there was a better way to handle such cases in jinja. I suppose the isn't a much better workaround for now.

lucasrodes:
@pabloarosado I'll look at it later when I work on making the COVID explorer closer to what an MDIM would look like.

Feel free to merge for now, otherwise I'll have a look later.

Thanks!

Marigold:
> I suppose the isn't a much better workaround for now.

Do you have an ideal format in mind? I'll see what I can do (though it's a bit tricky)

(I can make the bullet points not appear when rendering Jinja, but I don't have any other ideas)

PR #3982, opened 2025-02-16
✨  MDIM: minor improvements, towards a template

lucasrodes:
I am trying to improve our current tools so that MDIMs can get closer to a template.

### More context
IMO, there are three phases that I can identify:

1. Loading of the MDIM Config:
    - This deals with loading the configuration from a yaml config file.
2. Processing of the MDIM config
    - Expand the indicator paths to complete URIs.
    - Add more views if necessary.
    - Combine metadata in views with multiple indicators (not implemented).
3. Upsering the config to the DB
    - The MDIM (and its config) is pushed to DB. In this process we also check that the referenced indicators exist in the DB!

Currently, most of 2 happens in step 3, which is OK. However, it might be good to decouple this a bit to ease debugging and development: As a temporary solution for this, I've created function `_upsert_multidim_data_page` which deals with the actual DB interaction, and `upsert_multidim_data_page` is the same but includes a first step of adjusting the views (e.g. expand catalog paths)

### Thoughts
- I think we should consider creating some abstraction in python like `MDIMConfig` to control a bit more the MDIM configuration.
- In addition, we could then have objects like `MDIMView`, `MDIMDimension`, etc.
- should have the means to validate itself against the schema when trying to upser to DB.

### More concrete improvements implemented
- No need to load `engine` anymore, it uses `OWID_ENV` or any other environment (type `OWIDEnv`) that is provided
- 🚧  We review all views and adjust the `catalogPath` if needed. There could be however various edits that we could apply to views (e.g. combine metadata if there are multiple indicators).


lucasrodes:
@Marigold Leaving it here, I hope the code, along with today's discussion, is good to understand. Otherwise, please ping me!

My general thought in the code is that a lot of logic could be better encapsulated if we used classes, e.g., `MDIMConfig,` `MDIMView,` etc. E.g. there are some functions that only apply to views, etc.

Marigold (approved):
Beautiful! Your points make all sense. I especially look towards having view and dimension as dataclasses with methods. I'd also consider adding some unit tests as we work on this. We didn't do that for chart diff, and now it's hard to maintain.

PR #3985, opened 2025-02-17
🔨 Use date instead of yearIsDay for SST

Marigold:


Marigold:
@veronikasamborska1994 I saw your message about `date` not automatically adding `yearIsDay`. I tried to replicate it in this PR, but it seems to be working. What was the problem, please?

veronikasamborska1994:
ah right sorry! I didn't realize you need to set "date" as the index and it doesn't always just have to be year now! thanks for taking a look though!

PR #3991, opened 2025-02-17
🔨 Refactor causes of death mdim and remove expand_views_with_access_db

Marigold:
- Remove `expand_views_with_access_db` function, move the grouping of indicators into lists to `group_views`
- Fix some type issues with DAG
- Enable mdim for causes of death again

Here's the [mdim page](http://staging-site-mdim-causes-of-deaths/admin/grapher/mdd-causes-of-death?cause=All+causes&age=All+ages&metric=Number) on a staging server.

### DIFF
--- a/etl/multidim.py
+++ b/etl/multidim.py
@@ -1,4 +1,3 @@
-import json
 import re
 from copy import deepcopy
 from itertools import product
@@ -6,7 +5,6 @@ from typing import Any, Dict, List, Optional, Union
 
 import pandas as pd
 import yaml
-from deprecated import deprecated
 from owid.catalog import Dataset, Table
 from sqlalchemy.engine import Engine
 from structlog import get_logger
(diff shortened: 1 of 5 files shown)


etl/multidim.py
```diff
@@ -875,3 +791,26 @@ def _check_intersection_iters(
         raise ValueError(
             f"Unexpected items: {', '.join([f'`{d}`' for d in items_unexpected])}. Please review `{key_name}`!"
         )
+
+
+def group_views(views: dict[str, Any], by: list[str]) -> list[dict[str, Any]]:
```
lucasrodes, line 796: I have some questions on the "assumptions" that this function does on the input. E.g.

- what if a view contains multiple indicators already?
- what about other dimensions like `x`?

Maybe good to clarify when this would work and when not?

etl/steps/export/multidim/health/latest/causes_of_death.py
```diff
@@ -20,13 +20,20 @@ def run(dest_dir: str) -> None:
...
+    ]
+
+    # Group age and metric views under "Side-by-side comparison of causes"
+    grouped_views = multidim.group_views(config_new["views"], by=["age", "metric"])
+    for view in grouped_views:
+        view["dimensions"]["cause"] = "Side-by-side comparison of causes"
+
+    # Add views to config
+    config["views"] += config_new["views"]
+    config["views"] += grouped_views
```
lucasrodes, line 37: nice! I think we can work out a solution with this that works out-of-the box in most cases.

lucasrodes (approved):
LGTM. Just left some minor comments.

etl/multidim.py
Marigold: Good points. I didn't spend much time polishing it as I thought it's pretty rare case, but since you said you might use it elsewhere, I'll find some time to improve it.

PR #4000, opened 2025-02-19
✨ Switch mdims to catalogPath

Marigold:
ETL part of https://github.com/owid/owid-grapher/pull/4576

- Send mdim config to a new endpoint `/multi-dims/{quote(mdim_catalog_path, safe='')}` with body `{"config": config}`
- Use catalog paths like `energy/latest/energy_prices#energy_prices` or `covid/latest/covid#covid_xm_models`
- (Fix minor issue with --export and private steps)

I ran it against the staging server and tested in the [new admin](http://staging-site-mdim-catalog-path/admin/multi-dims). Everything was smooth!


rakyi (approved):
LGTM! Thanks for the quick turnaround!

PR #4003, opened 2025-02-19
🎉 catalog: Create tables method to export excel with metadata

pabloarosado:
Inspired by [this comment](https://owid.slack.com/archives/C3RAU3BJ9/p1739981330767749?thread_ts=1739975692.480529&cid=C3RAU3BJ9), I thought it would be useful to have a `.to_excel()` method on our `Table` that by default exports a table with a "data" and a "metadata" sheet.
The metadata sheet contains for now very limited info (colum name, title, short description, and producers with urls). I haven't put much thought on what exactly this metadata should look like, but I suppose there's no harm in adding it as-is for now, and we can improve it in the future (unless `.to_excel()` is being used elsewhere).

### DIFF
--- a/lib/catalog/owid/catalog/tables.py
+++ b/lib/catalog/owid/catalog/tables.py
@@ -224,6 +224,7 @@ class Table(pd.DataFrame):
         """
         Return a codebook for this table.
         """
+
         # Define how to show attributions and URLs in the sources column.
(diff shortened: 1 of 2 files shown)


Marigold (approved):
Nice, tried it on cherry blossom, and it looked like a good start.

PR #4004, opened 2025-02-19
🔨 Use expand_config in energy prices mdim

Marigold:
I've started this with the intention of merging `generate_views_for_dimensions` with `expand_config`. The plan was to add information about dimensions to `tb[col].m.additional_info` and then let `expand_config` do its magic.

It turned out that getting dimensions from a flattened indicator is harder than expected. Putting a band-aid on it in the export step didn't feel right. It'd only work if we were setting this information when flattening (i.e. in the garden steps).

Parking this until we get more experience with mdims.

### DIFF
--- a/docs/guides/data-work/export-data.md
+++ b/docs/guides/data-work/export-data.md
@@ -71,7 +71,7 @@ views:
 
 ```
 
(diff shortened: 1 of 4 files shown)


Marigold:
Moved to https://github.com/owid/etl/pull/4022

PR #4009, opened 2025-02-20
📊 US measles deaths

spoonerf:
Reported deaths from measles in the USA since the early 20th century.

This dataset brings together data from multiple public health records, largely national reports from the United States, and for 1950 onwards data from the WHO Mortality database is used. 

I'm not sure how best to handle a variable with so many different sources, but I think what I've done here is reasonable.

It is intended to replace the variable used in this chart: https://ourworldindata.org/grapher/measles-cases-and-death


etl/steps/data/garden/health/2025-02-20/measles_deaths_long_run.py
```diff
@@ -0,0 +1,41 @@
...
+    tb_pres = tb_pres[
+        (tb_pres["cause"] == "Measles")
+        & (tb_pres["age_group"] == "all ages")
+        & (tb_pres["country"] == "United States")
+        & (tb_pres["sex"] == "Both sexes")
+    ]
+    tb_pres = tb_pres[["country", "year", "number"]]
+    tb_pres = tb_pres.rename(columns={"number": "deaths"})
+
+    tb = pr.concat([tb_hist, tb_pres], short_name="measles_deaths_long_run", ignore_index=True)
```
spoonerf, line 31: Hi @pabloarosado - do you have any insight into what I'm doing wrong here? Both tb_hist and tb_pres have metadata prior to this line, but it disappears after pr.concat().

pabloarosado: Hi @spoonerf the resulting concatenated table does have metadata. But I noticed it doesn't have a table title, which then causes an issue in the grapher step because "Dataset must have a title". This happens because you are putting together two origins with different titles, and hence the concatenated table doesn't know what to do, and simply removes the title. You can either manually add a dataset title in the meta.yml file, or add a `default_metadata` to `create_dataset`. Does that make sense?

etl/steps/data/garden/health/2025-02-20/measles_deaths_historical.meta.yml
```diff
@@ -0,0 +1,21 @@
...
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 365
+
+tables:
+  measles_deaths_historical:
+    variables:
+      deaths:
+        title: Number of measles deathsq
```
pabloarosado, line 17: Typo.

snapshots/health/2025-02-20/measles_deaths_historical.csv.dvc
```diff
@@ -0,0 +1,63 @@
...
+      Public Health Reports, Vol. 55, No. 10 - Notifiable Diseases in the United States, 1938: Morbidity and Mortality Summaries for Certain Important Communicable Diseases.
+
+      Public Health Reports, Vol. 56, No. 7 - Notifiable Diseases in the United States, 1939: Morbidity and Mortality Summaries for Certain Important Communicable Diseases.
+
+      Public Health Reports, Vol. 57, No. 7 - Notifiable Diseases in the United States, 1940: Morbidity and Mortality Summaries for Certain Important Communicable Diseases.
+
+      US Census Bureau. Part 2 - Vital Statistics, Statistical Abstract of the United States 1944-45.
+
+    # Files
+    url_main: https://www.jstor.org/stable/4575902; https://www.jstor.org/stable/4576538; https://www.jstor.org/stable/4577735; https://www.jstor.org/stable/4578131; https://stacks.cdc.gov/view/cdc/69651/cdc_69651_DS1.pdf; https://www.jstor.org/stable/4583204; https://www.jstor.org/stable/4583620; https://www.jstor.org/stable/4584013; https://www.census.gov/library/publications/1952/compendia/statab/73ed.html
```
pabloarosado, line 54: Hmm, I think this could be problematic, `url_main` is expected to be just one URL. Let me see how this works out in the end...

snapshots/health/2025-02-20/measles_deaths_historical.csv.dvc
```diff
@@ -0,0 +1,63 @@
+# Learn more at:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+meta:
+  origin:
+    # Data product / Snapshot
+    title: Historical Measles Deaths - United States
```
pabloarosado, line 6: We usually use this nomenclature for title snapshots. So, if the data product was on historical measless deaths, but you were extracting specifically data for the US, then we would use this format. But as far as I understand, the entire data product is on the US, so maybe the title could be "Historical measless deaths in the United States" (no big deal, up to you).

snapshots/health/2025-02-20/measles_deaths_historical.py
```diff
@@ -0,0 +1,123 @@
...
+            DATA_1919,
+            DATA_1921,
+            DATA_1924,
+            DATA_1925,
+            DATA_1937,
+            DATA_1938,
+            DATA_1939,
+            DATA_1940,
+            DATA_1949,
+        ]
```
pabloarosado, line 116: Why not creating a single dataframe by record?
```
data = [
    {"country": "United States", "year": 1919, "deaths": 12992, "source": "https://www.jstor.org/stable/4575902?seq=25"},
    {"country": "United States", "year": 1921, "deaths": 3370, "source": "https://www.jstor.org/stable/4576538"},
    {"country": "United States", "year": 1924, "deaths": 8370, "source": "https://www.jstor.org/stable/4577735"},
    {"country": "United States", "year": 1925, "deaths": 2309, "source": "https://www.jstor.org/stable/4578131"},
    {"country": "United States", "year": 1937, "deaths": 1395, "source": "https://stacks.cdc.gov/view/cdc/69651/cdc_69651_DS1.pdf"},
    {"country": "United States", "year": 1938, "deaths": 3227, "source": "https://www.jstor.org/stable/4583204"},
    {"country": "United States", "year": 1939, "deaths": 1171, "source": "https://www.jstor.org/stable/4583620"},
    {"country": "United States", "year": 1940, "deaths": 681, "source": "https://www.jstor.org/stable/4584013"},
    {"country": "United States", "year": 1949, "deaths": 949, "source": "https://www.census.gov/library/publications/1952/compendia/statab/73ed.html"},
]
df = pd.DataFrame(data)
```

pabloarosado (approved):
Really interesting data, and really interesting ETL case :-)!
I see how this one is tricky, and there's no perfect solution, given our current metadata structure. But I'd suggest the following:
* Ideally, we'd have one snapshot per data product. But this would be a bit annoying, so maybe a compromise would be to have one snapshot for producer "Public Health Reports" (with all reports combined) and another for producer "US Census Bureau". That way, we can separate them as different producers, instead of having a hybrid producer `Public Health Reports; US Census Bureau`.
* Following the same logic, I'd suggest having just one `url_main`, instead of a list. [By design](https://docs.owid.io/projects/etl/architecture/metadata/reference/#originurl_main) that field is meant to have just one URL, so having a list of strings could cause issues down the line. In the case of the Public Health Reports, you could (arbitrarily) pick just the URL of the first report, and include all the rest in the description, as you already do.
* I'd remove the "Our World in Data (2017)" from the charts footers.

Do you agree with these suggestions? If it's too much work we can leave it, or we can discuss a better solution. More generally, we should come up with a better solution for this kind of combined datasets in the future. Thanks!

spoonerf:
> Really interesting data, and really interesting ETL case :-)! I see how this one is tricky, and there's no perfect solution, given our current metadata structure. But I'd suggest the following:
> 
> * Ideally, we'd have one snapshot per data product. But this would be a bit annoying, so maybe a compromise would be to have one snapshot for producer "Public Health Reports" (with all reports combined) and another for producer "US Census Bureau". That way, we can separate them as different producers, instead of having a hybrid producer `Public Health Reports; US Census Bureau`.
> * Following the same logic, I'd suggest having just one `url_main`, instead of a list. [By design](https://docs.owid.io/projects/etl/architecture/metadata/reference/#originurl_main) that field is meant to have just one URL, so having a list of strings could cause issues down the line. In the case of the Public Health Reports, you could (arbitrarily) pick just the URL of the first report, and include all the rest in the description, as you already do.
> * I'd remove the "Our World in Data (2017)" from the charts footers.
> 
> Do you agree with these suggestions? If it's too much work we can leave it, or we can discuss a better solution. More generally, we should come up with a better solution for this kind of combined datasets in the future. Thanks!

Thanks v much @pabloarosado! I agree, I think one snapshot per producer makes sense and shouldn't be too much effort.  Thanks for your comments, they are all v helpful!

spoonerf:
Hey @pabloarosado! 

Sorry to grab your attention again. I've now split the data into two Snapshots, but I'm having issues with some of the etl steps not being recognised. 

When I run `data://meadow/health/2025-02-20/measles_deaths_public_health_reports` it states that "No steps found among dependencies of current ETL step, that match the given specifications. step_name: data://meadow/health/2025-02-20/measles_deaths_public_health_reports", but I have run the snapshot dependency for that step and it definitely exists, so I'm not sure what's going on. It's the same issue causing the BuildKite to fail. Any ideas would be appreciated!

spoonerf:
> Hey @pabloarosado!
> 
> Sorry to grab your attention again. I've now split the data into two Snapshots, but I'm having issues with some of the etl steps not being recognised.
> 
> When I run `data://meadow/health/2025-02-20/measles_deaths_public_health_reports` it states that "No steps found among dependencies of current ETL step, that match the given specifications. step_name: data://meadow/health/2025-02-20/measles_deaths_public_health_reports", but I have run the snapshot dependency for that step and it definitely exists, so I'm not sure what's going on. It's the same issue causing the BuildKite to fail. Any ideas would be appreciated!

Ignore this, it was me not renaming something properly 🤦🏼‍♀️

Issue #3976, opened 2025-02-14
MDIMs: metadata config validation


ikesau:
Following my PR to update the `topic_tags` enum for indicator metadata yaml files (https://github.com/owid/etl/pull/3961), we've discovered that it's still possible for incorrect tags to be set in MDIM metadata (e.g. [this file](https://github.com/owid/etl/blob/fd8e672cca6ab31ee7d929c37e79ee3dad50981f/etl/steps/export/multidim/un/latest/un_wpp_population.yml), added in https://github.com/owid/etl/pull/3975)

MDIM metadata should have its own JSON Schema validation that ensures that incorrect tags can't be committed.

PR #4013, opened 2025-02-21
✨ Validate mdim config

Marigold:
Fixes https://github.com/owid/etl/issues/3976

- Validate mdim config against JSON schema
- Deprecate `TODAY` parameter in YAML files (it has been causing problems in validation and is no longer needed)


lucasrodes (approved):
LGTM, thanks!

PR #4024, opened 2025-02-25
✨ climate: change baseline period in the MET temperature anomalies chart

veronikasamborska1994:
hi @pabloarosado ! I am working on a short article with Hannah and we want to change the baseline period for this chart to "pre-industrial" times. I tried to explain the reasoning behind it in the metadata but let me know if anything is unclear. We also want to change the source name to "Adapted from.." to better reflect that this is what we've done. If you need to know more details you can see the last couple of comments in this [github issue](https://github.com/owid/owid-issues/issues/1857#issuecomment-2682231788).

### DIFF
--- a/etl/steps/data/garden/met_office_hadley_centre/2025-01-21/near_surface_temperature.meta.yml
+++ b/etl/steps/data/garden/met_office_hadley_centre/2025-01-21/near_surface_temperature.meta.yml
@@ -9,6 +9,8 @@ definitions:
 
     description_processing: |-
       We switch from using 1961-1990 to using 1861-1890 as our baseline to better show how temperatures have changed since pre-industrial times. For each region, we calculate the mean temperature anomaly for 1961–1990 and for 1861–1890. The difference between these two means serves as the adjustment factor. This factor is applied uniformly to all temperature readings and their confidence intervals to ensure that both the central values and the associated uncertainty bounds are correctly shifted relative to the new 1861–1890 baseline.
+    display:
+      numDecimalPlaces: 1
   presentation:
     topic_tags:
     - Climate Change
--- a/etl/steps/data/garden/met_office_hadley_centre/2025-01-21/near_surface_temperature.meta.yml
+++ b/etl/steps/data/garden/met_office_hadley_centre/2025-01-21/near_surface_temperature.meta.yml
@@ -4,6 +4,8 @@ definitions:
       - Temperature anomalies are given in degrees Celsius relative to the average temperature over the period 1961-1990.
       - Temperature anomalies are available for the Northern Hemisphere and the Southern Hemisphere.
       - The global mean is calculated by averaging anomalies for northern and southern hemispheres.
+    description_from_producer:  |-
+      The 1961-90 period is most often used because it is the period recommended by the WMO (World Meteorological Organisation). In some cases other periods are used. For example, satellite data are typically only available from 1979 onwards so it would not be possible to construct either a 1961-1990 or a 1971-2000 climatology. For global average temperatures, an 1861-1890 period is sometimes used to show the warming since the "pre-industrial" period.
   presentation:
     topic_tags:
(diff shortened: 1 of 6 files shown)


etl/steps/data/garden/met_office_hadley_centre/2025-01-21/near_surface_temperature.meta.yml
```diff
@@ -1,12 +1,22 @@
...
+    description_processing: |-
+      We switch from using 1961-1990 to using 1861-1890 as our baseline to better show how temperatures have changed since pre-industrial times. For each region, we calculate the mean temperature anomaly for 1961–1990 and for 1861–1890. The difference between these two means serves as the adjustment factor. This factor is applied uniformly to both the temperature anomalies and the confidence intervals to ensure that both the central values and the associated uncertainty bounds are correctly shifted relative to the new 1861–1890 baseline.
+    processing_level: major
+    display:
+      numDecimalPlaces: 1
   presentation:
     topic_tags:
     - Climate Change
+    grapher_config:
+      note: The period between 1861–1890 is used as a baseline to better reflect temperature changes since pre-industrial times, [as recommended by the source](https://www.metoffice.gov.uk/hadobs/indicators/index.html#:~:text=The%201961%2D90%20period%20is,other%20parts%20of%20the%20world.).
```
pabloarosado, line 19: I personally think that this footnote is appropriate, and enough to convey the modification. I'd say the "Adapted from" is not necessary.

etl/steps/data/garden/met_office_hadley_centre/2025-01-21/near_surface_temperature.py
```diff
@@ -12,8 +12,23 @@ def run(dest_dir: str) -> None:
     #
     # Load meadow dataset and read its main table.
     ds_meadow = paths.load_dataset("near_surface_temperature")
-    tb_meadow = ds_meadow["near_surface_temperature"]
+    tb_meadow = ds_meadow["near_surface_temperature"].reset_index()
```
pabloarosado, line 15: You can simply use:
```
tb_meadow = ds_meadow.read("near_surface_temperature")
```

pabloarosado (approved):
Thanks Veronika, I agree with the changes in the processing. Actually, we could consider adopting a pre-industrial baseline in other places too.
I have suggested avoiding the "Adapted from". This is similar to "Our World in Data based on...", which we have discussed multiple times, and reached the conclusion that it's unnecessary, because it takes space from our charts and doesn't necessarily convey the message we want to the user. Having a footnote mentioning an important modification, plus a more detailed explanation in the `description_processing` (as you already did) should be enough.
However, if Hannah or you disagree, I'd suggest:
(1) Adding the "Adapted from..." in the grapher chart admin, rather than the snapshot metadata (given that the snapshot itself is not adapted).
(2) Mentioning that we need to redefine our guidelines with this regard, so we can figure out a better way to express adaptations that do not affect consistency across charts or our internal analytics.

Please let me know if you want to discuss any of this, thanks!

snapshots/met_office_hadley_centre/2025-01-21/near_surface_temperature_global.csv.dvc
```diff
@@ -1,6 +1,6 @@
 meta:
   origin:
-    producer: Met Office Hadley Centre
+    producer: Adapted from Met Office Hadley Centre
```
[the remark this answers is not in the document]
veronikasamborska1994: I'll let Hannah know! This is something she thought was important to do.

veronikasamborska1994:
thanks @pabloarosado! I'll talk to Hannah about this and let you know. For now, I've changed it to the original name of the producer. Let me know if you want me to add this code to other datasets and recalculate the baseline. Happy to do this or if you are planning regular updates of these soon happy to leave it to you.

PR #4030, opened 2025-02-25
🔨 standardize MDIM config files

lucasrodes:
- Start using `short_name.config.yml` as default mdim/explorer config file name.
- Create submodule for mdim and explorer tooling: `etl.collections`
- Migrate COVID explorer to have MDIM-like config

### How to review?
No exhaustive review is needed. But I think is good if you can look at the overall direction (explorers → mdim-like). 

I also want to merge before generating conflicts. Since the COVID explorer is affected, I'll have a careful look before merging, comparing PROD & staging explorers.

**Changes summary:**
- `etl.collections`: New module for MDIM/Explorer tooling
  - `etl.collections.multidim`: I've moved old `etl.multidim`
  - `etl.collections.base`: I've started drafting a very rough model on how we could handle config files for MDIMs and Explorers. The idea is that we would then have, if needed, specific classes inheriting from these (e.g. `MDIMConfig(Config)` or so).
    - `etl.collections.utils`: Util functions.
- `explorers/covid/latest`: I've re-structured the COVID explorer to use MDIM-like configuration. There are some adhoc functions I had to implement, which live in the covid script for now. The idea is that these could be migrated to `etl.collections.explorer` or something alike.
- `multidim/*`: No relevant changes, just adapting imports.


etl/collections/base.py
```diff
@@ -0,0 +1,47 @@
+"""WIP: Drafting a model for dealing with MDIM/Explorer configuration.
+
+This should be aligned with the MDIM schema.
+"""
+
+from typing import Any, Dict, List, Optional
+
+
+class Config:
```
Marigold, line 9: Eventually, you might want to make a dataclass from it with methods for saving / loading dicts, jsons, etc.

You can check out e.g. [Origin class](https://github.com/owid/etl/blob/a7d34a7bc445600d93091fe5b2f8dcf6359cb2a4/lib/catalog/owid/catalog/meta.py#L110-L112) for inspiration.

etl/collections/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""This module is a package for collections of indicators. This includes:
```
Marigold, line 1: I was trying to think of a better name than collections as it collides with the one from standard library... but I couldn't think of any, so go for it :).

etl/collections/multidim.py
```diff
@@ -16,16 +14,23 @@
 import pandas as pd
 import yaml
 from deprecated import deprecated
-from owid.catalog import Dataset, Table
+from owid.catalog import Table
 from sqlalchemy.engine import Engine
 from structlog import get_logger
 
 from apps.chart_sync.admin_api import AdminAPI
+from etl.collections.utils import (
```
Marigold, line 22: Nit - we're importing from the library itself, so using `from .utils import ...` could work, too.

Marigold (approved):
Looks good! I tested it and it worked, will be testing it more when rebasing. Let's merge this as soon as possible 🚀

etl/collections/__init__.py
lucasrodes: yeah, thought that too... can't really think of a better one atm, but im open to rename

PR #4052, opened 2025-02-28
✨ Check variables metadata by default

pabloarosado:
I think by now we can safely use `check_variables_metadata=True` by default (in `create_dataset`), and remove it from wizard data step templates (which takes a lot of room).
Note that, as far as I can tell, there's little risk in setting this as `True` by default (maybe we'll see some additional warnings in old steps that didn't have this flag defined explicitly).


lucasrodes (approved):
LGTM! let's see

PR #4053, opened 2025-03-03
✨ Simplify code for run function

pabloarosado:


lucasrodes (approved):
LGTM.

I think your reasoning makes a lot of sense, and this is a good direction.

I've been thinking of the role of object `PathFinder` for a while now, and not sure where we want to go with it. To me, it is becoming more and more like an "ETL orchestrator" tool... which makes the current name a bit confusing. I'm interested to learn what your thoughts are on it; especially if we should double down and use it as our main interface for read/write operations in ETL.

pabloarosado:
Thanks @lucasrodes! I agree that we could figure out a different name or purpose for `helpers.PathFinder`, but the status quo is reasonable enough: it's a helper object that lets us find the relevant paths to load and create ETL datasets.

etl/helpers.py
```diff
@@ -799,6 +707,50 @@ def load_mdim_config(self, filename: Optional[str] = None, path: Optional[str |
     def load_explorer_config(self, filename: Optional[str] = None, path: Optional[str | Path] = None) -> Dict[str, Any]:
         return self.load_config(filename, path)
 
+    def create_dataset(
+        self,
+        tables: Iterable[catalog.Table],
```
Marigold, line 712: Is there a reason these args are not passed as **kwargs?

pabloarosado: Not a strong one, but I noticed that this way you can see the arguments in the code editor. With kwargs I think you don't see them, right?

pabloarosado: For example, if you are in a data step and you want to use some of the additional arguments of `create_dataset`, but you don't remember the name... Then VSCode will show you all options. But using kwargs (I think?) you would not be able to see them straightaway. Maybe there's a better solution.

Marigold (approved):

etl/helpers.py
Marigold: That's right, kwargs won't expand in autocomplete unless you add some type hinting magic (which looks way worse than propagating arguments). Go with your solution, if it works, we would ideally remove `dest_dir` from old steps anyway and have a single function.

PR #4055, opened 2025-03-03
✨  Adding extra step to modify snap metadata

spoonerf:
Hey @Marigold, 

The data part of this automatic update step has been running fine, but I've also been trying to auto-update the metadata in `grapher/health/latest/measles_long_run.meta.yml` and this hasn't been updating (e.g. [this chart](https://ourworldindata.org/grapher/number-of-measles-cases?tab=chart&country=~USA), the data updated on Friday but the subtitle didn't change.) 

Do you think these changes will ensure the subtitle changes? I've been copying the wildfires snapshot and I realised I didn't have this step in the snapshot py.


spoonerf:
Actually, it seems to be working now, so I'll merge and see what happens this week!

PR #4058, opened 2025-03-04
🎉 Create VSCode extension to execute code inside run

pabloarosado:
## Context
When working on ETL data steps, most of use use VSCode with an Interactive Window. Usually, what I do is `shift+a` to select the entire script, then `shift+Enter` to execute it. An interactive window opens. Then I scroll down to the beginning of the `run()` function, select a block of code, and hit `shift+Enter`.

If I want to re-execute the entire content of `run()` until a certain moment, I need to manually press `shift` and up or down to select the appropriate block of code, and then `shift+Enter` to execute it. This is a bit time consuming.

Also, when working on a snapshot step, if you do `cmd+a` and `shift+Enter`, the interactive window opens, but the code fails (because of the `if __name__ == "__main__"`. So you have to manually select only the code above this block.

So, the extension in this PR makes this workflow faster:
* If you are outside of `run()` (e.g. when you just opened the script) hit `cmd+Enter` to initialize an interactive window, and execute the entire script (to have all definitions in memory). The cursor will be moved to the first line inside `run()`.
* If you are inside of `run()`, hit `cmd+Enter` to execute all code inside `run()` until the current line.
* If you are working on a snapshot, the `if __name__ == "__main__"` will be ignored.

NOTE: You can execute an entire step by simply hitting `cmd+Enter`. First time you press it, you initialize the interactive window and execute all definitions. Then, the cursor moves to the inside of `run()`, and if you hit `cmd+Enter` multiple times, it will run the block line by line (every time running the entire block). This, plus some caching, let's you "debug" through the code.

## Main changes
* Let `make test` install custom VSCode extensions (if `code` is installed, otherwise skip). This will also install the previously created `find-latest-etl-step` extension.
* Create a new VSCode extension that helps when writing ETL data pipelines.
* I added a couple of parameters to `.vscode/settings`. They are some time delays that are used to ensure the extension works well. We should try it on different computers to see if it works. If not, a user could try setting different delays (possibly longer for a slower computer). I'm not sure if this is totally necessary, we may remove it otherwise.

## Discussion
I have been working with this extension in the last couple of days and found it quite useful! But I'm wondering a few things:
1. Is it working by default? E.g. if anyone does `make install` and reloads VSCode, does the `cmd+Enter` shortcut work?
2. Is the default behavior when the cursor is outside `run()` the most convenient one? If you are inside another function, you may think that `cmd+Enter` should run the content of that function until the cursor position. However, what it will do is execute the entire script and move the cursor to the first line of `run()`. This could be confusing.
  * I think the current behavior is the most convenient one for ETL work, but we can redefine it if we find a better option.
4. Is the default behavior when the cursor is inside `run()` the most convenient one? Currently, it selects all code above the cursor (including the current line) and executes it in the interactive window. Maybe it would be more convenient to do that, and remove the selection.
  * I think the current behavior is reasonable.

## Other ideas
Another similar extension (maybe assigned to `ctrl+Enter`, or replace the default `shift+Enter`) could be one that:
  * If there is a selection, execute that.
  * If nothing is selected, selects the current block (based on where the cursor is), and execute it in the interactive window. This would be more similar to how R-studio works (which, in this sense, is a bit smarter than the VSCode interactive window, since here you need to select the full block manually).
Optionally, it could move to the next line of code. This way, you could run an entire script by simply hitting `ctrl+Enter` multiple times (similar to debugging line by line).

I tried creating this extension (and called it `smart-run-current-line`) but I couldn't manage to make it work. There were always some annoying edge cases, so I gave up.


Marigold:
Good luck with this! It'd be amazing if you could make it work. My dream was to be able to set a breakpoint in VSCode and open the debugger in the interactive python window. Users have been [asking for it](https://github.com/microsoft/vscode-jupyter/issues/1278) for years, but there's no progress 😢

pabloarosado:
Hey @Marigold and @lucasrodes, do you want to give it a try? Simply switch to this branch, and run `make test`. Then go to any etl data step and hit `cmd+Enter`. Then move some lines down inside the `run()` function and run `cmd+Enter` again. Please let me know if that works as expected, thanks!

lucasrodes (approved):
LGTM! Worked just as expected.

1) Just go first row, press `⌘`+`⏎` to run all code.
2) Go inside `run` and go to line `N` and press `⌘`+`⏎` again. This ran all lines within `run()` function until `N-1`.

pabloarosado:
Thanks @lucasrodes ! I'm happy it works. But note that the extension should execute all code until the current line (included).

I've noticed now that, on any interactive window, if I type, e.g. `1+1` and hit `shift+Enter`, it opens a new interactive window (which is absurd). Is that happening to you too? If so, it could be an issue with the extension. Otherwise, I just messed up something in my local settings. Please let me know if you have the same issue, thanks!

Marigold:
`1+1` and hitting shift + enter works fine for me.

The extension is super neat and works as expected. I'd merge it as it is and see how it works for us. I'm personally still in favour of playground notebooks on the side (mainly to have something to come back to), but perhaps this will change my mind!

Makefile
```diff
@@ -64,7 +64,7 @@ unittest: .venv
 	@echo '==> Running unit tests'
 	.venv/bin/pytest -m "not integration" tests
 
-test: check-formatting check-linting check-typing unittest version-tracker
+test: check-formatting check-linting check-typing unittest install-vscode-extensions version-tracker
```
Marigold, line 67: How about keeping this separate from `make test`? Installing extensions as part of "testing" is an unexpected side effect.

pabloarosado: That's a good point, but currently we (at least I) use `make test` to set up an environment from scratch. What would you suggest otherwise?

Marigold (approved):
Great idea! I had only one nit - to keep installing extensions separate from `make test`, but I'll leave it up to you.

Makefile
Marigold: Just to let people run both `make test` and `make install-vscode-extensions` separately. Only a few of us will be using the extension, so there's no need to install it by default. (I did the same thing with installing `uv` as part of `make test` and Lars wanted me to put it away... which I haven't done yet 😅).

pabloarosado: On second thought, I think it makes sense to separate the installation of the environment and the extensions (especially when thinking of external users who may just want to try the ETL for a specific thing and are not going to develop any steps). I'll create a PR.

PR #4063, opened 2025-03-04
📊 Update public expenditure on health OMM

paarriagadap:
Recreate the public expenditure on health OMM that combines three datasets to get the indicator as share of GDP in the long run.

Main issue here https://github.com/owid/owid-issues/issues/1594


paarriagadap:
Hi @spoonerf! I have recreated your work in the public expenditure as a share of GDP OMM, so we can maintain it up-to-date. Please take a look if everything makes sense. We now have the [OECD Health Expenditure and Financing Database](https://admin.owid.io/admin/datasets/6959), please take a look if you still think that picking Government/compulsory schemes is the best option. It looks like it, but we also have Government schemes available. You can even play with the categories by changing CATEGORY_OECD in the garden step of health_expenditure_omm
image

paarriagadap:
@spoonerf If helpful, you can also take a look at this comparison between the current OECD data, your version of the OMM and this new version: http://staging-site-data-public-expenditure-heal/admin/charts/8532/edit

etl/steps/data/garden/oecd/2025-03-04/health_expenditure_1993.meta.yml
```diff
@@ -0,0 +1,33 @@
...
+
+
+tables:
+  health_expenditure_1993:
+    variables:
+      share_gdp:
+        title: Public health expenditure as a share of GDP
+        unit: "% of GDP"
+        short_unit: "%"
+        description_short: Public health expenditure divided bt gross domestic product, expressed as a percentage.
```
spoonerf, line 22: small typo "bt"

etl/steps/data/garden/oecd/2025-03-04/health_expenditure_1993.meta.yml
```diff
@@ -0,0 +1,33 @@
...
+  common:
+    presentation:
+      topic_tags:
+        - Healthcare Spending
+
+
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 365
```
spoonerf, line 12: Perhaps should be 0 as I don't expect we'll update this

spoonerf (approved):
Hey @paarriagadap!

The code all looks good to me, much nicer than my previous version! 

As far as whether we should use Government/Compulsory or Government, I'm happy for you to make the call, I think you have a better understanding of the implications than me. For some countries, like Germany (below), it seems like Government/Compulsory is clearly more aligned with the earlier time-series.

I added just Government schemes to your comparison chart and made a scatter plot directly comparing them [here](http://staging-site-data-public-expenditure-heal/admin/charts/8533/edit), just for interest/exploration, but feel free to delete it!

Screenshot 2025-03-05 at 19 56 11

etl/steps/data/garden/oecd/2025-03-04/health_expenditure_1993.meta.yml
paarriagadap: Yes, thanks! The wizard only allows to put a date in there

etl/steps/data/garden/oecd/2025-03-04/health_expenditure_1993.meta.yml
paarriagadap: Thanks!

paarriagadap:
Thanks @spoonerf! Yeah, I think keeping Government/compulsory is our best option to deal with countries that have mixed systems where the State leaves part of the core health system to the privates. Government data from the OECD Health Expenditure and Financing Dataset is much more limited and with clear different levels compared to the previous series in some countries, so I will push with the current configuration.

PR #4079, opened 2025-03-06
📊 Update social expenditure in the long run dataset

paarriagadap:


paarriagadap:
@spoonerf Hi! This works the same as the health expenditure, that's why I ask you to review again. Most of the code is the same, so feel free to skim. Thanks!

spoonerf (approved):
Looks good! 

My only comment would be that the [FAUST on this chart could be improved](http://staging-site-data-social-expenditure-data/admin/charts/610/edit) e.g. it refers to a Source tab that no longer exists and the variable labels are a bit untidy. 

It looks like it's not used anywhere so you could consider deleting it even.

paarriagadap:
Thanks @spoonerf! Yeah, I think that chart should be gone. I wanted to at least make it draft after merging here.

PR #4083, opened 2025-03-07
✨ Add .render method to render Jinja templates

Marigold:
Replace function for rendering metadata with Jinja templates
```
import etl.grapher.helpers as gh
gh.render_variable_meta(tb.emissions.m, dim_dict={'pollutant': 'CO', 'sector': 'Transport'})
```
by
```
tb.emissions.m.render({'pollutant': 'CO', 'sector': 'Transport'})
```

As part of this, I also did:
- Move Jinja functions to owid-catalog
- Refactor dimension rendering in grapher:// step

The other function `render_yaml_file` for rendering Jinja from YAML file will be also moved in the future (to the new "YAMLMeta" object).

One disadvantage of `.render(...)` is that you need to build the garden dataset & load it to render it. It's often easier to work solely with the YAML file and use `render_yaml_file`.

### DIFF
--- a/docs/architecture/metadata/structuring-yaml.md
+++ b/docs/architecture/metadata/structuring-yaml.md
@@ -259,9 +259,7 @@ The most straightforward way to check your metadata is in Admin, although that m
 
 ```python
 import etl.grapher.helpers as gh
-dim_dict = {
-  "age_group": "YEARS0-4", "sex": "Male", "cause": "Drug use disorders"
-}
+dim_dict = {"age_group": "YEARS0-4", "sex": "Male", "cause": "Drug use disorders"}
 d = gh.render_yaml_file("ghe.meta.yml", dim_dict=dim_dict)
 d["tables"]["ghe"]["variables"]["death_count"]
 ```
@@ -270,15 +268,11 @@ An alternative is examining `VariableMeta`
 
 ```python
 from owid.catalog import Dataset
-import etl.grapher.helpers as gh
 from etl import paths
 
-tb = Dataset(paths.DATA_DIR / "garden/who/2024-07-30/ghe")['ghe']
-
-# Sample a random row to get the dimension values
-dim_dict = dict(zip(tb.index.names, tb.sample(1).index[0]))
-
-gh.render_variable_meta(tb.death_count.m, dim_dict=dim_dict)
+ds = Dataset(paths.DATA_DIR / "garden/emissions/2025-02-12/ceds_air_pollutants")
+tb = ds['ceds_air_pollutants']
+tb.emissions.m.render({'pollutant': 'CO', 'sector': 'Transport'})
 ```
 
 ### Jinja Macros
--- a/etl/grapher/helpers.py
+++ b/etl/grapher/helpers.py
@@ -1,10 +1,8 @@
 import copy
-from dataclasses import dataclass, field, is_dataclass
-from functools import lru_cache
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Literal, Optional, Set, Union, cast
 
-import jinja2
 import numpy as np
 import pandas as pd
 import pymysql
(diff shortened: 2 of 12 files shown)


pabloarosado (approved):
This is great, thanks for doing this work! I have some comments and suggestions. Some of them are just to be considered for the future. I leave it up to you if you want to tackle any of them on this PR:
* Instead of an error, `render` could raise a warning when a dimension is not defined, and show the raw jinja content. It's more convenient to see unrendered metadata than to see an error and be forced to spell out all dimension values.
* Maybe `render` could also (optionally) remove DoDs. I created function `utils.remove_details_on_demand` for this (not very sophisticated).
* For now, the `render` solution is good. But ideally, it should not be necessary to use `render`, and every time we show metadata (i.e. everytime we use `_repr_html_`), it would render the appropriate content by default, based on the indexes displayed. But I know that this would potentially involve some more structural changes.
* On the disadvantage you mentioned of using `render`: Now that we can easily execute `create_dataset` inside `run()` (without having to define `dest_dir`), I think it's not that inconvenient. It's not too different to type:
```
ds_garden.read("ceds_air_pollutants").m.render({...})
```
than
```
gh.render_yaml_file(paths.metadata_path, {...})
```
so your proposed solution is more convenient than the status quo. Maybe one convenient thing we could do is to let `Dataset.read()` read by default the first table, if not specified, instead of requiring an argument (most of our datasets have one table).

Feel free to merge and we can discuss some of those suggestions at some point. Thanks!

Marigold:
> Instead of an error, render could raise a warning when a dimension is not defined, and show the raw jinja content. It's more convenient to see unrendered metadata than to see an error and be forced to spell out all dimension values.

Can you clarify this? Let's have this example:
```
from etl import paths

ds = Dataset(paths.DATA_DIR / "garden/emissions/2025-02-12/ceds_air_pollutants")
tb = ds['ceds_air_pollutants']
# tb.emissions.m.render({'pollutant': 'CO', 'sector': 'Transport'})
tb.emissions.m.render({'pollutant': 'CO'})
```
which currently raises an error
```
UndefinedError: 'sector' is undefined

Dimensions:
{'pollutant': 'CO'}

Template:
<% if pollutant == "BC" %>Black carbon <% elif pollutant == "CH₄" %>Methane <% elif pollutant == "CO" %>Carbon monoxide <% elif pollutant == "N₂O" %>Nitrous oxide <% elif pollutant == "NH₃" %>Ammonia <% elif pollutant == "NMVOC" %>Non-methane volatile organic compounds <% elif pollutant == "NOₓ" %>Nitrogen oxides <% elif pollutant == "OC" %>Organic carbon <% elif pollutant == "SO₂" %>Sulfur dioxide <% endif -%> emissions from << sector.lower() >>
```
Are you suggesting to only show this as a warning and upsert the unrendered raw metadata to Admin so that it's visible there?

> utils.remove_details_on_demand

Added, nice idea.

> so your proposed solution is more convenient than the status quo

Yeah, this was more about refactoring jinja and preparing it for dimensions metadata. I still prefer `gh.render_yaml_file(paths.metadata_path, {...})` for development because you don't have to build the dataset.

pabloarosado:
Thanks @Marigold,

> Are you suggesting to only show this as a warning and upsert the unrendered raw metadata to Admin so that it's visible there?

My point was that, while developing, if you want to quickly visualize metadata in the interactive window/notebook, it's inconvenient to have to specify all dimensions. So, if there's ambiguity, `.render()` could raise a warning (e.g. `Warning: "sector" is not defined`, instead of an error) and display the raw content.
I was not thinking of changing anything regarding upserting or admin. But, if the latter is an unavoidable consequence of my proposal, then ignore it, it's not that important.

Marigold:
> So, if there's ambiguity, .render() could raise a warning (e.g. Warning: "sector" is not defined, instead of an error) and display the raw content.

I still don't understand what you mean, it feels to me that displaying _the error_ with raw content is the same as the warning and returning raw content. I'm going to merge this, but feel free to change it to warning if you find it useful.

pabloarosado:
> > So, if there's ambiguity, .render() could raise a warning (e.g. Warning: "sector" is not defined, instead of an error) and display the raw content.
> 
> I still don't understand what you mean, it feels to me that displaying _the error_ with raw content is the same as the warning and returning raw content. I'm going to merge this, but feel free to change it to warning if you find it useful.

My point was: if this is supposed to be a function that we can use to help us visualize metadata, then showing a huge error when you don't spell out all dimensions is inconvenient. Instead, a one-line warning and the full unrendered metadata would be better. But it was just a minor suggestion, it's fine the way it is, thanks!

PR #4089, opened 2025-03-10
🔨 Refactor run(dest_dir)

Marigold:
Decide whether to pass `dest_dir` or not by inspecting function arguments rather than catching TypeError. The latter has the disadvantage of being included in the traceback when step fails.


pabloarosado (approved):
Thanks, this makes sense.
Also, `run_module_run` reminded me of [this amazing song](https://www.youtube.com/watch?v=lmc21V-zBq0).

Issue #4072, opened 2025-03-04
Explorers: move to ETL as `export` steps


lucasrodes:
Our landscape of explorers is very diverse. This is an attempt to study bringing all our public explorers (45) into ETL as `export` steps, to leverage our latest tooling. 

An idea for this is to start backporting some explorers to ETL:
- #4080

## Explorer types

There are three general categories:

1) Some explorers rely on CSVs
2) Some others are built using charts from our site.
3) Some others are based on references to indicators.

However, within each of these, there are some features (one explorer may be a mix of these):

1) CSV-based
- Is it in ETL?
- Is it from GitHub? `owid/owid-datasets`, `owid/importers`, or another?
- Is there processing in `owid-content`?

3. Indicator-based
- Uses variable IDs
- Uses catalog paths
- Is it from ETL (`export://`) or manually created?

I've tried classifying all our explorers in [this spreadsheet](https://docs.google.com/spreadsheets/d/1JV2Am1ivyPGhpZd5aC3XCKa_LA9FU_cS-nsS0o14Nyk/edit?gid=0#gid=0). I've added the number of explorer views and number of user views, which should help us prioritise any work here.

pabloarosado:
Thank you @lucasrodes for this exploration (on explorers!). Given that it's unclear how we'll tackle the transition, maybe we can close this issue (but we keep it as reference). And during the next shaping session we can decide how to move forward.

lucasrodes:
Hi @pabloarosado I rather keep this one open.

I'm currently working on TSV-to-JSON, which I think falls under the umbrella of this issue.

pabloarosado:
All right @lucasrodes, thanks for rephrasing the description. Feel free to close it when you are done with the TSV-to-JSON, or let me know if you need any help.

Marigold:
@lucasrodes FYI this [code](https://github.com/owid/automation/blob/main/automation/mirror_explorers.py) by Lars _could_ be useful (we talked about it yesterday). It parses TSV files and saves them as JSON to MySQL. It could also be useless though ,and we might want to use an entirely different structure, so don't feel any pressure.

lucasrodes:
2 new explorers have been migrated!

- [Population and Demography](https://github.com/owid/etl/tree/master/etl/steps/export/explorers/un/latest)
- [Monkeypox](https://github.com/owid/etl/tree/master/etl/steps/export/explorers/who/latest)

lucasrodes:
[...]

PR #4095, opened 2025-03-10
🔨 refactor: create explorer

lucasrodes:
ref: #4072 

### Some fixes on the way:
- [x] BUG(`expand_config`): all views were being created, even if we filtered only to a subset of dimensions.
- [x] BUG(explorers): The order of dimension choices was not respected in the final explorer presentation.
- [x] BUG(explorers): All widgets are being rendered as dropdowns
- [x] BUG(explorers): Ordering of dimensions and dimension choices in the final explorer presentation is not respected:
  - [x] The order of the dimensions is set based on the ordering of columns.
  - [x] The order of dimension choices is set based on how views appear in `grapher` table. That is, if first view has dimension `sex=Female`, that'll be the first choice shown for the dimension.

### Refactor
- [x] Rename current methods for explorer creation to signal they are legacy.
- [x] Align Explorer creation with MDIMs. Use `paths.create_explorer` and `Explorer.save()` for new YAML-based explorers.
- [x] Remove `etl.explorer_helpers` (it was not in use)
- [x] Move `etl.explorer` to `etl.collections.explorer_legacy`, since it is an interface for TSV-based explorers. `etl.collections.explorers` is a more general interact, which also  deals with YAML-based explorers.
- [x] Rename `Explorer` (formerly in `etl.explorer`) as `ExplorerLegacy` (now in `etl.collections.explorer_legacy`).

### Case examples
I was testing some of these tools in some explorers, but they need some more work. I'll pause the work (commented the steps in the DAG) and resume in a separate PR.
- [ ] Population: automated (continue work in https://github.com/owid/etl/pull/4116)
- [ ] Monkeypox: more of a manual work

**Future, related:**
- [ ] Add support for `display` for indicators
  - [ ] old TSV → new dataclass
  - [ ] new dataclass → TSV (only indicator-based)
- [ ] Put simple functions in place to do migration
- [ ] Add check to just allow migration when possible: 
  - [ ] CSV-based with ETL-generated
  - [ ] Indicator-based


Marigold (approved):
I like the idea of adding the `_legacy` suffix to functions. It's a bit hard to review with the code being moved around, but I didn’t spot any issues.

Issue #3977, opened 2025-02-14
MDIMs: Standardize casing convention for indicator and metadata


ikesau:
[Indicators use snake_case](https://github.com/owid/etl/blob/master/etl/steps/data/garden/antibiotics/2024-12-02/microbe_neonatal_amr.meta.yml)

[MDIMs use camelCase](https://github.com/owid/etl/blob/master/etl/steps/export/multidim/covid/latest/covid.cases.yml)

[Can I make it any more obvious?](https://youtu.be/TIy3n2b7V9k?t=21)

This seems like a footgun. It means we can't easily share JSON schemas between the two, and will probably make mistakes involving copying YAML with the wrong casing into the wrong file.

IMO there should only be one convention, but I'll leave that to @lucasrodes and the rest of the data engineers to decide 🙂

Marigold:
> IMO there should only be one convention

It's hard to say where to draw the line. Python uses under_score and MySQL & TS camelCase. ETL uses mix of both - in most cases underscore, but for instance [`display` is in camelCase](https://github.com/owid/etl/blob/master/etl/steps/data/garden/antibiotics/2024-12-02/microbe_neonatal_amr.meta.yml#L22-L24) because it's directly inserted into db.

As far as I know, this hasn't caused that many problems, as most of them were caught by our YAML linter. I'll leave it up to @lucasrodes and other data managers to decide what is the preferred casing and whether it's worth changing it.

ikesau:
Yeah it's a shame the 2 languages have different conventions here.

I'm more than happy to close this immediately if it's not a problem for anyone - Lucas just suggested I create a ticket for it after I remarked on it in our DMs. I think the actually impactful thing is making sure that topic tags are validated consistently, but consistent casing isn't a requirement for that.

Marigold:
> I'm more than happy to close this immediately if it's not a problem for anyone

Up to you, I don't mind having it on our eyes for some time. Either we figure it out or owidbot closes it.

pabloarosado:
It seems difficult to have a convincing solution that doesn't become extremely complicated. Hopefully autocomplete helps. If this becomes an issue in the future we can reconsider :-/.

PR #4097, opened 2025-03-10
🔨 Unify camelCase and snake_case

Marigold:
Partially implements https://github.com/owid/etl/issues/3977. The main goal is to use the same casing as in ETL YAML files to prevent confusion. For now, the mix of snake_case and camelCase is acceptable.  

This is achieved by updating the schema in `schemas/multidim-schema.json` and replacing the `metadata` block with the schema from `dataset-schema.json#variables`. We work in snake_case and convert to camelCase right before sending it to the Admin API.  

Some fields such as those in `config` are left in camelCase. I don't think we'd gain much by converting them to snake_case.

@pabloarosado, good call on spending time on this—it looks much better than I expected!


pabloarosado (approved):
Thanks a lot @Marigold ! This is going to give us much more peace of mind in the future :-)

Issue #3956, opened 2025-02-11
Ideal data update workflow for explorers and mdim pages


pabloarosado:
## Context
We currently have a nice workflow to review changes in grapher charts after a data update, using Chart Diff.

Explorers, however, are not supported by Chart Diff. We currently have no structured way to review explorers after a data update. And the same applies to upcoming multidimensional pages.

**NOTE: We will discuss this with Lars next Monday (17th of Feb) during the weekly data call, so it would be good to have some ideas prior to the call.**

**NOTE 2: I just remembered that Anomalist doesn't work with mdims either (which has already bitten me at least [once](https://github.com/owid/owid-issues/issues/1616#issuecomment-2651665157)). So, in this issue, when I speak of Chart Diff, we should also think of Anomalist as an integral part of the data update workflow.**

## Impact
In the long term, we expect to have more mdim pages, replacing explorers and possibly also existing charts. If we don't decide soon on a valid workflow, we may encounter one or more of the following issues:
* Reviewing mdim pages becomes manual, tedious, and prone to errors.
* Mdim pages don't get properly reviewed and present data issues.
* We integrate mdims into Chart Diff in a suboptimal way, and the tool becomes less useful and more time-consuming. This can also lead to people not using it, which would also lead to increasing the likelihood of data issues.

## Goal
Figure out a nice data update workflow integrating explorers and mdim pages (probably using Chart Diff, with some changes).

### (A) Explorers are compared as a unit in Chart Diff
An entire explorer will appear as if it was just another chart in Chart Diff. It will be up to the user to explore different configurations left and right.
  * Mdim pages could also follow this option, once they are embeddable (which I believe is planned to happen?)
  * Nice to have: Changing the configuration of one explorer (e.g. the old one on the left) would change the configuration of the other explorer (e.g. the new explorer on the right). But this may be tricky, or even not desirable.

### (B) All views of explorers become individual charts to review in Chart Diff
Dozens, or, in some cases, hundreds of individual charts will need to be reviewed in Chart Diff whenever the data of an explorer is affected. This could force us to rethink the way Chart Diff works:
(B.1) We may not want to enforce that all views need to be reviewed. Instead, the workflow could be more Anomalist-style, where the user is able to filter and sort charts based on some criteria. In fact, this sorting in Chart Diff could use Anomalist' scores.
(B.2) We still enforce that all views need to be reviewed, but we have easy ways to "check as reviewed" in bulk.

pabloarosado:
[...]

PR #4108, opened 2025-03-11
🎉 Explorer diff app

Marigold:
Proof of concept for https://github.com/owid/etl/issues/3956. We proposed multiple approaches there, this PR implements one of them, but it doesn't mean we can't switch to another. (I found this one the easiest.)

I hacked together a minimal version of **explorer-diff**. It lets you select an explorer and then dynamically creates select boxes (explorers are rendered with hidden controls). There’s no approval mechanism or other fancy features like in **chart-diff**, but it could still be useful as a starting point. (I found the "Random view" button especially handy for ad-hoc checks.)

Here's the [app on staging](http://staging-site-explorer-diff/etl/wizard/explorer-diff)

## Missing Features  
- [ ] If a selection is missing from "views," explorers shouldn't render rather than displaying an arbitrary selection.  
- [ ] Show top views based on pageviews.  
- [ ] "Random view" should be sampled proportionally to pageviews.  
- [ ] Missing country selection (changing a selection always shows default countries)

### DIFF
--- a/apps/wizard/app_pages/explorer_diff/app.py
+++ b/apps/wizard/app_pages/explorer_diff/app.py
@@ -99,6 +99,7 @@ def _extract_all_dimensions(explorer_views: list[dict]) -> dict[str, list]:
 
 
 def main():
+    st.warning("This application is currently in beta. We greatly appreciate your feedback and suggestions!")
     st.title(
         ":material/difference: Explorer Diff",
         help=f"""
--- a/etl/collections/model.py
+++ b/etl/collections/model.py
@@ -527,18 +527,6 @@ class Explorer(Collection):
         return mapping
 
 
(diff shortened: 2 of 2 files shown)


lucasrodes (approved):
This seems like a good first approach IMO.

I'll definitely be using it in the coming weeks, so can improve it as I use it if necessary.

Do you know if this comes with some lag bc of deploy? If so, maybe worth stating somewhere that this tool has some limitations and is in beta.

Marigold:
> Do you know if this comes with some lag bc of deploy?

What do you mean by lag because of deploy?

> If so, maybe worth stating somewhere that this tool has some limitations and is in beta.

Oh right, I'll add it to the top.

PR #4115, opened 2025-03-12
📊 Update UN DESA migrant stock

antea04:
[Update UN DESA Migrant Stock (2024)](https://github.com/owid/owid-issues/issues/1907)

Updates charts and migration explorer to the latest numbers.

[Updated explorer on staging](http://staging-site-data-update-un-desa/admin/explorers/preview/migration).


etl/steps/data/garden/un/2025-03-12/migrant_stock.countries.json
```diff
@@ -0,0 +1,381 @@
...
+  "United Kingdom": "United Kingdom",
+  "United States": "United States",
+  "United States Virgin Islands": "United States Virgin Islands",
+  "Vatican": "Vatican",
+  "Venezuela": "Venezuela",
+  "Vietnam": "Vietnam",
+  "WORLD": "World",
+  "Wallis and Futuna": "Wallis and Futuna",
+  "LATIN AMERICA AND THE CARIBBEAN": "Latin America and the Caribbean",
+  "NORTHERN AMERICA": "Northern America",
```
spoonerf, line 317: Remember to add the organisational acronym to regions which aren't defined by us, so Northern America (UN) etc. You can be quite harsh in removing regions you don't think will add much value to users, especially if you are adding OWID regions on top of the ones in the dataset.

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
```diff
@@ -0,0 +1,113 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    presentation:
+      topic_tags:
+        - Migration
+      grapher_config:
+        originUrl: "https://ourworldindata.org/migration"
+    processing_level: minor #minor for most indicators, major for data that has been manipulated (e.g. 5 year change, share of child migrants)
+    unit: ""
```
spoonerf, line 10: Try and add a unit to most of the variables, unless it is really too long, I think it helps understanding a lot in the charts.

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
```diff
@@ -0,0 +1,113 @@
...
+  migrant_stock:
+    variables:
+      # Immigrant stock
+      immigrants_all:
+        title: Total number of international immigrants
+        description_short: People living in a given country who were born in another country.
+        description_from_producer: "{definitions.description_un.migrant_stock}"
+      immigrants_female:
+        title: Total number of international immigrants - female
+        description_short: The total number of female people living in a given country who were born in another country.
```
spoonerf, line 34: I would use 'females' and 'males' instead of 'female people' and 'male people'

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
```diff
@@ -0,0 +1,113 @@
...
+        short_unit: "%"
+        description_short: Immigrants as a share of the total population.
+        description_from_producer: "{definitions.description_un.share_migrant_stock}"
+        display:
+          numDecimalPlaces: 1
+      immigrant_share_of_dest_population_female:
+        title: Share of the population that was born in another country - female
+        unit: "%"
+        short_unit: "%"
+        description_short: Female immigrants as a share of the total (female) population.
```
spoonerf, line 53: I like the clarification in the brackets! I always wonder what exactly the denominator is and now I no longer have to :)

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
```diff
@@ -0,0 +1,113 @@
...
+        unit: "%"
+        short_unit: "%"
+        description_short: The total number of people that have left the country, as a share of the current population.
+        description_processing: We calculated the share of emigrants as the number of emigrants divided by the total population of their home country.
+        display:
+          numDecimalPlaces: 1
+      # data on 5 year changes
+      immigrants_change_5_years:
+        title: Change in the number of international immigrants in the previous five years
+        description_short: The change in the total number of migrants living in a given country compared to five years ago. A positive number indicates there are now more migrants living in a given country; a negative number means there are fewer. For 2024, this is the change compared to 2020.
```
spoonerf, line 90: Does this mean it's actually a four year change?

etl/steps/data/garden/un/2025-03-12/migrant_stock.py
```diff
@@ -0,0 +1,257 @@
...
+    # destination table, total numbers and shares
+    tb_d_total = ds_meadow.read("migrant_stock_dest_total")
+    tb_d_share = ds_meadow.read("migrant_stock_dest_share")
+    tb_d_pop = ds_meadow.read("un_desa_total_population")
+    # origin table
+    tb_o = ds_meadow.read("migrant_stock_origin")
+
+    # Australia & New Zealand shows up twice in the data
+    duplicates = tb_do[tb_do.duplicated(subset=["country_destination", "country_origin"])]
+    assert (
```
spoonerf, line 38: I would just omit Australia and New Zealand from the datasets, we tend to try and avoid having too many non-OWID entities in the dataset as it can get a bit out of hand.

etl/steps/data/garden/un/2025-03-12/migrant_stock.py
```diff
@@ -0,0 +1,257 @@
...
+        },
+        errors="raise",
+    )
+
+    tb = pr.multi_merge([tb_d_total, tb_d_share, tb_o, tb_d_pop], on=["country", "year"], how="outer")
+
+    return tb
+
+
+def calculate_change_over_5_years(tb):
```
spoonerf, line 133: I think as it's only 4 years for 2020-2024 we are best not calculating this for this most recent release. They may release again next year, but if they don't we can revisit this issue. I think it might be okay to drop the five year change from the explorer entirely, what do you think?

spoonerf (approved):
Looks good to me! I just raised a couple of things that it would be good for you to address, namely: 

- I think you can remove many of the smaller and vaguer regions, and I think you should add in the OWID continents

- I think it's a bit confusing to show the five year change, but then it's actually a four year change for the most recent data. I think it's best just to stop at 2020 for now and we can reconsider whether to keep those views when the next release comes out.

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
antea04: Does women & men also work? Females and males always feels a bit....depersonalizing to me, but I can use it if it's standard FAUST.

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
antea04: Does this refer to the last data point (from 2020-2024)? Otherwise, no, it's the change from June 2020 to June 2025 - so spanning 5 years.

etl/steps/data/garden/un/2025-03-12/migrant_stock.py
antea04: Yeah, it's a bit awkward! I'll remove the calculation for 2024!

etl/steps/data/garden/un/2025-03-12/migrant_stock.meta.yml
antea04: done!

etl/steps/data/garden/un/2025-03-12/migrant_stock.countries.json
spoonerf: I think keep any that you think are definitely relevant, but I'd try not to have too many flavours of regions. I think OWID continents, WB income groups, and two other organizational regional groupings are a good limit, perhaps three at a push!

PR #4121, opened 2025-03-13
📊 climate: Fix climate change explorer

pabloarosado:
* Transition [climate change impacts](https://ourworldindata.org/explorers/climate-change) csv-based manually created explorer [into an ETL export/explorer](http://staging-site-data-climate-change-explorer/admin/explorers/preview/climate-change).
* Ensure explorer/mdim config cites only steps that are in the DAG as dependencies of the current step.

@lucasrodes there's no need to inspect the metadata changes (this is not a very prominent explorer, and I have basically copied things from the old into the new explorer, adjusting a few minor things, and just changed one missing indicator). But please have a look at the changes in `PathFinder`. I ensured that paths added to config should be in the DAG (but I'm not sure if there is a better way to achieve that). Thanks!

### DIFF
--- a/etl/helpers.py
+++ b/etl/helpers.py
@@ -234,41 +234,6 @@ class WrongStepName(ExceptionFromDocstring):
 load_dag_cached = cache(load_dag)
 
 
-def extract_indicators_from_config(config):
-    """
-    Recursively extract all strings matching the pattern
-    'channel/namespace/date/dataset/table#indicator'
-    from any field in the explorer/mdim config dictionary.
-
-    Parameters
-    ----------
-    config : dict
-        The explorer/mdim config dictionary to search.
-
-    Returns
-    -------
-    list
-        A list of matching strings.
-
-    """
-    pattern = re.compile(r"\b[\w\-/]+/[\w\-/]+/[\w\-/]+/[\w\-/]+/[\w\-/]+#[\w\-/]+\b")
-    indicators = set()
-
-    def recursive_search(obj):
-        if isinstance(obj, dict):
-            for value in obj.values():
-                recursive_search(value)
-        elif isinstance(obj, list):
-            for item in obj:
-                recursive_search(item)
-        elif isinstance(obj, str):
-            matches = pattern.findall(obj)
-            indicators.update(matches)
-
-    recursive_search(config)
-    return sorted(indicators)
-
-
 class PathFinder:
     """Helper object with naming conventions. It uses your module path (__file__) and
     extracts from it commonly used attributes like channel / namespace / version / short_name or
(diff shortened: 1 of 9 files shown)


etl/helpers.py
```diff
@@ -591,6 +627,15 @@ def load_config(self, filename: Optional[str] = None, path: Optional[str | Path]
...
             path = self.config_path
         config = catalog.utils.dynamic_yaml_to_dict(catalog.utils.dynamic_yaml_load(path))
+
+        # Extract all indicators mentioned in the config.
+        indicators = extract_indicators_from_config(config=config)
+        # List their corresponding datasets.
+        datasets = sorted(set(["data://" + "/".join(indicator.split("/")[:-1]) for indicator in indicators]))
+        # Ensure that their datasets are included as dependencies of the current step.
+        if not set(datasets) <= self.dependencies:
+            raise NoMatchingStepsAmongDependencies(step_name=self.step_name)
```
lucasrodes, line 637: I think this won't work in cases where the catalog path is not complete (i.e. use `table_name#column`). Note that that is often the case (and actually the recommended way).

There is a sanity check on this later in `create_explorer`, which tries to expand the catalog paths so they are complete based on the given dependencies (see function `process_views`). It'll fail if it can't do it (e.g. no step with that table name).

Hence, I suggest to drop this bit

etl/steps/export/explorers/climate/latest/climate_change.config.yml
```diff
@@ -0,0 +1,337 @@
...
+        name: 'true'
+    presentation:
+      type: checkbox
+views:
+  - dimensions:
+      metric: temperature_anomaly
+      long_run_series: false
+    indicators:
+      y:
+        - catalogPath: grapher/met_office_hadley_centre/2025-01-21/near_surface_temperature/near_surface_temperature#temperature_anomaly
```
lucasrodes, line 77: maybe you can consider removing the `grapher/met_office_hadley_centre/2025-01-21/near_surface_temperature/` bit for all indicators? Could help automate future updates.

lucasrodes (changes requested):
Nice job! Very cool to see a migration <3

One note on your suggestions for `PathFinder`: Checking that paths makes sense can't actually be done until we call `create_explorer`, which is where we make sure that catalog paths are complete (they can come in the reduced shape like `table#indicator`).

Other than this, LGTM!

pabloarosado:
Hey @lucasrodes thanks, that makes sense. I was confused and worried about having paths in the config. I'm happy that this is taken care of in `create_explorer`. Then, to avoid hardcoded paths, the migration function `migrate_csv_explorer` could write only `table_names#indicator_names` by default.

PR #4123, opened 2025-03-13
📊 climate: era5 add continents

veronikasamborska1994:
hey @Marigold ! I am introducing some regional averages to this code but also noticed there might be a bug in the snapshot.py script! take a look and see if this fix makes sense..


etl/snapshot.py
```diff
@@ -179,16 +179,15 @@ def dvc_add(self, upload: bool) -> None:
...
-
         # Calculate md5
         md5 = checksum_file(self.path)
 
         # Get metadata file
         with open(self.metadata_path, "r") as f:
             meta = ruamel_load(f)
-
         # If the file already exists with the same md5, skip the upload
-        if meta["outs"] and meta["outs"][0]["md5"] == md5:
```
veronikasamborska1994, line 191: @Marigold I think this was throwing an error for me when I was trying to add a new snapshot.

Marigold: Yes, it was a bug. Sorry for that, it's fixed in master.

Marigold (approved):
Looks good! Sorry for the bug with the snapshot.

PR #4124, opened 2025-03-14
🧹 Deprecate walden

Marigold:
Wipe out Walden from ETL. Also remove `examples` which are outdated and not used by anyone. I had to move functions from `walden.files` to `etl/download_helpers.py`, but that's the only "interesting" thing in this PR.

## TODO

- [x] Let everyone run `rm -rf ~/.owid/walden` to clean old walden datasets (or should it be done as part of `make test`?)


Marigold:
Reviewing this is a formality. I just removed everything walden related.

Marigold:
@pabloarosado would you mind having a look instead of Lucas? It's not urgent at all.

Marigold:
> We would still need to remove walden steps and usages from the dag archive. I'm happy to do that, so feel free to merge this PR to avoid conflicts (there are already some).

Archived steps are not runnable anyway, do you think it's worth it? Anyway, merging this now and we can create a follow-up.

PR #4135, opened 2025-03-18
📊 migration flow mdims

antea04:


etl/steps/export/explorers/migration/latest/migration_flows.config.yml
```diff
@@ -0,0 +1,58 @@
...
+    - United Kingdom
+    - France
+    - Mexico
+    - India
+    - China
+    - Germany
+    - Australia
+    - Brazil
+    - South Africa
+    - World
```
spoonerf, line 25: I'd be tempted to remove World from here as it can dwarf the other countries in the chart view

etl/steps/export/explorers/migration/latest/migration_flows.config.yml
```diff
@@ -0,0 +1,58 @@
...
+
+dimensions:
+  - name: Country
+    slug: country_select
+  - name: Metric
+    slug: metric
+    presentation:
+      type: radio
+    choices:
+      - slug: "immigrants"
```
spoonerf, line 42: I would put Emigrants first as in the current explorer, just as the default view is Afghanistan and the 'Emigrants' slice is much more populated.

etl/steps/export/explorers/migration/latest/migration_flows.config.yml
```diff
@@ -0,0 +1,58 @@
...
+        metric: emigrants
+    # Applies to all views with dimension metric=immigrants, unless overridden
+    - dimensions:
+        metric: immigrants
+
+config:
+  explorerTitle: Migration Flows
+  explorerSubtitle: Explore global migration data, country-by-country.
+  isPublished: true
+  selection:
```
spoonerf, line 14: At the moment I am finding on the 'Immigrants' view that only 'World' is selected, switching to 'Emigrants' brings in all these. Not a big deal but it feels a tiny bit inconsistent and maybe unintended?

etl/steps/export/explorers/migration/latest/migration_flows.config.yml
antea04: I totally understand that - I am a bit hesitant to remove it, as for many countries there is just no data for all the other countries in the selection and then people are greeted with the "no data" screen, which can be confusing. It's a bit tricky, I'll try it for a few countries and see what I think looks better!

antea04: Hm after doing this I think the best way would be to only have world enabled by default. Since for every country different countries are interesting, setting a global default doesn't seem very meaningful here - this also solved the issue of imigrants/ emigrants having different selections depending on which data is available.

etl/steps/export/explorers/migration/latest/migration_flows.config.yml
antea04: this is caused by some countries having no data available for immigrants/ emigrants - selecting only world fixes that.

etl/steps/export/explorers/migration/latest/migration_flows.config.yml
spoonerf: That makes sense! I think having just World is a good default

Issue #4165, opened 2025-03-24
Unify snapshot helper functions under a single `.save(...)` method


Marigold:
Over time, we've created a few wrappers for saving metadata and uploading it to R2. Every now and then, someone gets confused about which one to use.

Wizard currently recommends using:
```python
snap.create_snapshot(upload=upload)
```

This downloads the data from the given link (if set) and uploads it to R2 as a snapshot. However, it does not call `snap.metadata.save()`, which still needs to be done separately.

Ideally, `create_snapshot` (or `snapshot.save`) should be the only entry point, and the use of other methods should be discouraged.

## TODO
- [x] Merge `add_snapshot` into `create_snapshot` and replace it in all snapshots
- [x] Call `.save()` from `create_snapshot`

pabloarosado:
Please have a look at this issue, which may be related: https://github.com/owid/etl/blob/f47fd1663a6d68a78cd8b07958a12297da36ebce/etl/scripts/faostat/update_snapshots_metadata.py#L61

PR #4166, opened 2025-03-24
:hammer: Refactor snapshot methods

Marigold:
Fixes https://github.com/owid/etl/issues/4165

- Merge `add_snapshot` with `create_snapshot`
- Call `.save` in `create_snapshot`

### DIFF
--- a/etl/scripts/faostat/update_snapshots_metadata.py
+++ b/etl/scripts/faostat/update_snapshots_metadata.py
@@ -56,14 +56,8 @@ def run():
         snapshot.metadata.origin.description = description  # type: ignore
         snapshot.metadata.origin.attribution = attribution  # type: ignore
 
-        # Rewrite metadata to dvc file.
-        ################################################################################################################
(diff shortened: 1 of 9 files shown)


pabloarosado (approved):
Thanks! This looks good. I find it a bit unfortunate, that, for data steps, we have `create_dataset`, which initializes a dataset (without saving it), but for snapshots, `create_snapshot` does the saving. This can be a bit confusing. Maybe we could consider adding the handling of snapshots also to PathFinder. But for now, this PR makes sense and it's an improvement.

Marigold:
@pabloarosado you're right, but it's a slightly different kind of "save" than `Dataset.save`. With snapshots, saving typically means downloading the file and updating `outs:` in `*.dvc` file. `self.metadata.save` in `create_snapshot` re-saves metadata in case someone edits them programmatically (which rarely happens).

Perhaps `create_snapshot` should be renamed to `save` and be responsible for both:
- Creating the "snapshot" (=downloading it)
- Updating metadata

I'm not sure if making such a rename is worth it.

PR #4175, opened 2025-03-25
📊 un igme update

spoonerf:


lucasrodes:
nit: you can use the new defaults to have slightly cleaner code (applies to meadow, garden and grapher):

- You don't need to import `create_dataset`
- You should use `paths.create_dataset` instead of `create_dataset`, which doesn't need `dest_dir` as argument
- You don't need to have `run(dest_dir)`, but can just have `run()`


**❌ Old**
```python
from etl.helpers import PathFinder, create_dataset

# Get paths and naming conventions for current step.
paths = PathFinder(__file__)

    # Code
   ds = create_dataset(
        dest_dir,
        tables=tables,
    )
```

**✅ New**

```python
from etl.helpers import PathFinder

# Get paths and naming conventions for current step.
paths = PathFinder(__file__)

def run() -> None:
    # Code
   ds = paths.create_dataset(
        tables=tables,
    )
```

Marigold:
Yeah, it's likely failing because the branch is old. Just ignore the error and review chart-diff & merge as usual.

Marigold:
Let me know if I should do a full review. (Sources still need some work, e.g. [this chart](http://staging-site-data-un-igme/admin/charts/8481/edit) shows population as the primary source and unreadable title. It has something to do with metadata propagation, happy to help with that)
image

PR #4176, opened 2025-03-26
📊 regions: Keep 'defined_by' field in external regions dataset

pabloarosado:
Following [this conversation](https://owid.slack.com/archives/C0193RW5E2J/p1742917803787839), we decided to keep `defined_by` field in the external `regions` dataset, which used to be unnecessary (since its only value was `owid`), but now that it has multiple options it has become useful.


sophiamersmann (approved):
Thank you!

PR #4184, opened 2025-03-27
🐛 (migration flows) explorer doesn't get populated with catalog paths

antea04:


antea04:
This was the problem: https://github.com/owid/etl/issues/4185#issue-2952853052 - Mojmir has fixed the issue this particular time and hopefully moving on from owid-content will make this superfluous.

PR #4191, opened 2025-03-28
📊 agriculture: Fix missing milk yield indicator

pabloarosado:
FAOSTAT QCL used to have yield for milk (i.e. milk per animal, used in [this chart](https://ourworldindata.org/grapher/milk-yields-per-animal)), but it was removed at some point: that combination of item and element is empty in the 2025 release.
Numerically, it seems that their old milk yield indicator coincides reasonably well with production / animals, so I used that as an estimate. For some countries, however, the new data seems to be significantly larger (see e.g. Saudi Arabia or Mexico in [chart diff](http://staging-site-data-milk-yield-fix/etl/wizard/chart-diff)).
I have changed the subtitle to clarify that this is an estimate, and how it's calculated.

### DIFF
--- a/etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
+++ b/etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
@@ -1219,10 +1219,10 @@ def generate_net_exports_as_share_of_supply(tb_fbsc: Table) -> Table:
     return tb
 
 
-def generate_milk_per_cow(tb_qcl: Table) -> Table:
-    # FAOSTAT QCL used to have yield for milk (milk per cow), but it was removed at some point (that combination of item and element is empty in the 2025 release).
-    # Numerically, it seems that their old milk yield indicator coincides with production / slaughtered animals.
-    # So I will reproduce that indicator using their latest data on production and number of cows.
+def generate_milk_per_animal(tb_qcl: Table) -> Table:
+    # FAOSTAT QCL used to have yield for milk (milk per animal), but it was removed at some point (that combination of item and element is empty in the 2025 release).
+    # Numerically, it seems that their old milk yield indicator coincides with production / number of animals.
+    # So I will reproduce that indicator using their latest data on production and number of milk animals.
 
     # Firstly, check if yield is still missing in the data.
     error = "Yield for milk is not missing in the data anymore. Consider reusing it instead of calculating it."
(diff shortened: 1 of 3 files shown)


etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
```diff
@@ -1219,6 +1219,65 @@ def generate_net_exports_as_share_of_supply(tb_fbsc: Table) -> Table:
...
+    error = "Units of milk production or milk animals have changed."
+    assert set(tb_milk["unit"]) == {UNIT_FOR_PRODUCTION, UNIT_FOR_ANIMALS}, error
+
+    # Transpose data and rename columns conveniently.
+    tb_milk = tb_milk.pivot(
+        index=["country", "year"], columns="element_code", values="value", join_column_levels_with="_"
+    ).rename(
+        columns={
+            ELEMENT_CODE_FOR_PRODUCTION: "milk_production",
+            ELEMENT_CODE_FOR_MILK_ANIMALS: "cows_used_for_milk",
```
spoonerf, line 1258: I think it's important to note that cattle aren't the main source of milk in many countries e.g. it appears to be camels in Saudi Arabia, in Jamaica I think it is goats. I don' think you can make an equivalence between 'milk animals' and cows.

etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
```diff
@@ -1219,6 +1219,65 @@ def generate_net_exports_as_share_of_supply(tb_fbsc: Table) -> Table:
     return tb
 
 
+def generate_milk_per_cow(tb_qcl: Table) -> Table:
+    # FAOSTAT QCL used to have yield for milk (milk per cow), but it was removed at some point (that combination of item and element is empty in the 2025 release).
+    # Numerically, it seems that their old milk yield indicator coincides with production / slaughtered animals.
```
spoonerf, line 1224: Do you mean to write slaughtered animals here? I can't see a reference to it in the code?

spoonerf (commented):
Hey @pabloarosado 

I agree with the theory of what you are doing, but I'm not sure you are using the right indicators. 

I may have misunderstood but it seems to me like you are doing:

[Total milk production (from all milk animals) ](http://staging-site-data-milk-yield-fix/admin/variables/1018770) / [Number of cows used for milk production](http://staging-site-data-milk-yield-fix/admin/variables/1024869)

which causes funky values for all countries where cows are not the main producers of milk (e.g. Saudi Arabia where it is camels).

I think you should be doing:

[Total milk production (from all milk animals) ](http://staging-site-data-milk-yield-fix/admin/variables/1018770) / [All milk animals](http://staging-site-data-milk-yield-fix/admin/variables/1018769)

I've manually looked at a couple and it gets you closer to the missing indicators values (for the ones I checked anyway). Hope that helps.

etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
pabloarosado: No, I was just referring to the element as it's presented on their web, namely "Producing Animals/Slaughtered", but here it doesn't make sense.

pabloarosado:
Thanks @spoonerf! What you describe is not exactly what I was doing. I was actually using the item for just "raw milk from cattle", both for production and for number of animals (there was no "raw milk" for all animals). I did that because I thought that using "raw milk" would be more accurate than using "milk" (which has a very broad definition, "Milk, eggs, honey and beeswax are included as products of live animals. [...]" and I thought could potentially include other dairy products).
However, after some further inspection, I think in this case Milk is the right choice, and numerically, the resulting production/animals matches very well with the original FAOSTAT "yield". So this must be the right item indeed!
I've adapted the metadata to mention that this includes other milk animals.
Thanks a lot for looking into it!!

spoonerf (approved):

PR #4192, opened 2025-03-28
🐝 update grapher schema version

sophiamersmann:
Updates the Grapher schema version to 007. `map.projection` was renamed to `map.region`.

Waiting for https://github.com/owid/owid-grapher/pull/4707 to be reviewed and merged.


Marigold (approved):
Looks good. (Next time, try using the same branch name as in owid-grapher. Both PRs will use the same staging server and the CI will be more representative)

PR #4195, opened 2025-03-28
🐛 gbd: add missing dependency on `un_wpp`

marcelgerber:
I noticed this missing dependency when I was running a full ETL rebuild, and it was failing because of this.


spoonerf (approved):
Good catch, thanks @marcelgerber!

PR #4196, opened 2025-03-28
✨ Improve performance of grapher:// step

Marigold:
A couple of optimizations for the `grapher://` step. The main idea is to run DB queries in bulk where possible. There are other areas that could be optimized (like the `update_links` function), but converting that to bulk would add a lot of complexity without a significant performance gain.

At the moment, the main bottleneck for me is network-related — specifically uploading data and metadata JSON files to R2. I've experimented with various approaches to speed it up (e.g., `asyncio`, tweaking boto3 configs), but none have been significantly faster than using threads (we currently use 40).

I’ve left the profiling notebook in `scripts/`, as we’ll likely want to revisit performance optimizations in the future.

All in all, it's about 2x faster for both matching & non-matching checksums.

## Ideas for Improving Performance
- [x] Compare checksums in bulk before creating MySQL connections.  
- [x] Upsert `origins` in bulk before upserting `variables`.  
- [ ] Run `update_links` in bulk after upserting `variables`.  

## Benchmark

Tested on
```
STAGING=performance-grapher FORCE_UPLOAD=1 etlr grapher://grapher/iea/2024-07-04/critical_minerals_demand_by_technology --grapher --only --force
```
with cProfile
```
python -m cProfile -s cumtime etl/command.py grapher://grapher/iea/2024-07-04/critical_minerals_demand_by_technology --grapher --only --force | head -n 100
```

Improvements:
- uploading data & metadata: 47s -> 24s
- skipping upload when checksums match: 13s -> 6s

### DIFF
--- a/etl/grapher/to_db.py
+++ b/etl/grapher/to_db.py
@@ -215,7 +215,7 @@ def check_table(table: Table) -> None:
 
 def _check_upserted_variable(variable: Variable) -> None:
     assert variable.notnull().all(), (
-        "Tables to be upserted must have no null values. Instead they" f" have:\n{variable[variable.isnull()]}"
+        "Tables to be upserted must have no null values. Instead they" f" have:\n{variable.loc[variable.isnull()]}"
     )
     assert not gh.contains_inf(variable), f"Column `{variable.name}` has inf values"
 
--- a/etl/steps/__init__.py
+++ b/etl/steps/__init__.py
@@ -11,7 +11,6 @@ import re
 import subprocess
 import sys
 import tempfile
-import time
 import warnings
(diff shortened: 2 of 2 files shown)


pabloarosado (approved):
This is great! I've just tested it on various datasets, and saw a big speed-up. For example, the grapher FAOSTAT QCL step used to take ~220s, and now it takes ~70s.
I'm worried this is too good to be true! Have you thought of any potential side effects? If not, merge, please! :-)

Marigold:
Thanks! I'll see if I can squeeze something from FAOSTAT QCL specifically.

PR #4200, opened 2025-03-31
📊 child mortality: GBD add indicators with the leading causes of death

veronikasamborska1994:
hi @spoonerf ! what do you think about adding this chart on [leading causes of deaths in children](http://staging-site-child-mort/admin/charts/8579/edit)? I think it tells a pretty interesting story that you can't otherwise see in all our other charts on causes of deaths in children.

### DIFF
--- a/dag/health.yml
+++ b/dag/health.yml
@@ -520,6 +520,12 @@ steps:
   data-private://grapher/ihme_gbd/2024-05-20/gbd_child_mortality:
     - data-private://garden/ihme_gbd/2024-05-20/gbd_child_mortality
 
+  # GBD 2021 - GBD Child Mortality Leading Causes
+  data-private://garden/ihme_gbd/2025-03-31/gbd_child_mortality_leading_causes:
+    - data-private://garden/ihme_gbd/2024-05-20/gbd_child_mortality
+  data-private://grapher/ihme_gbd/2025-03-31/gbd_child_mortality_leading_causes:
+    - data-private://garden/ihme_gbd/2025-03-31/gbd_child_mortality_leading_causes
+
   # GBD 2021 - GBD Health-adjusted Life Expectancy and Life Expectancy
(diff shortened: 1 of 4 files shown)


veronikasamborska1994:
thanks a lot @spoonerf! I've now changed the code to use the owid hierarchy and switched to using the `gbd_cause` dataset :) let me know if you spot any other issues!

spoonerf (approved):
Nice one! 
I just had a couple of further thoughts:

- It might be worth omitting 'Fire, heat and hot substances' from the map (I think you can just hide it from the legend). As far as I can see it only appears for Tuvalu in 2000, I think due to [this fire](https://www.cbsnews.com/news/19-die-in-tuvalu-dormitory-blaze/), which I don't think killed any under fives. So it's probably just a weird modelling quirk from IHME, influenced by this.


- I would choose a slightly different colour palette, it's quite hard to notice the differences between countries in the current one. It's really hard to find one that works with so many categories. I've used [Color Buddy](https://color-buddy.netlify.app/?&bg=lab%28100%25%200%200%29&colors=lab%2847.361%25%20-8.868%20-40.17%29&colors=lab%2867.643%25%2046.034%2072.736%29&colors=lab%2857.955%25%20-49.881%2047.683%29&colors=lab%2847.681%25%2065.84%2046.233%29&colors=lab%2851.149%25%2031.156%20-38.769%29&colors=lab%2842.662%25%2022.052%2016.324%29&colors=lab%2864.633%25%2049.585%20-19.884%29&colors=lab%2853.193%25%200%200%29&colors=lab%2874.678%25%20-12.573%2069.379%29&colors=lab%2870.097%25%20-35.603%20-20.314%29&name=category10&space=lab) before to figure out a palette for these situations, e.g. for the [all ages leading causes of death chart](https://ourworldindata.org/grapher/leading-cause-of-death)

PR #4204, opened 2025-04-01
✨ Improve performance of grapher:// step

Marigold:


Marigold:
I tried optimizing faostat_qcl, but didn't find any low hanging fruit.

PR #4209, opened 2025-04-01
📊 Happiness by age group

antea04:
Happiness by age group scores, taken from the 2024 World Happiness report (https://happiness-report.s3.amazonaws.com/2024/Ch2+Appendix.pdf, page 54ff)

Scraped from the pdf and formatted by hand/ script. Imported from local file.

This is publically accessible data and part of the WHR 2024, so I would think we are able to publish it, but maybe it would be good to check with e.g. Ed about this. 

______
**Added context:** Last year we wanted to get "Happiness by age group" breakdowns from the World Happiness report into the etl (https://github.com/owid/owid-issues/issues/1581). There was a longer discussion back and forth with the researchers about whether we can get this data in a spreadsheet format, at the end of which we didn't get any data. 

There was some discussion about whether the WHR is allowed to publish demographic breakdowns, but I am assuming this was more of a concern for the long run data, while this is a publically accessible snapshot in time.


antea04:
The head of communications at the World Happiness report got back to me, saying:

> You are welcome to reproduce data that is publicly available from [worldhappiness.report](https://worldhappiness.report/) with an appropriate credit.

So this should be fine. I'll go ahead and publish it!

PR #4210, opened 2025-04-01
✨ Add schema for explorer *.config.yml files

Marigold:
This started as a simple PR to add JSONSchema validation for explorers, but as I re-ran the explorers and checked `explorer-diff`, I uncovered all sorts of small issues. After fixing them, I reached **TSV parity** in [explorer-diff](http://staging-site-explorers-schema/etl/wizard/explorer-diff), so the plan is to automatically run `export://explorers` (in addition to `export://multidim`) after we merge this.

I also found that some of our ETL-generated explorers (like `minerals`) have been manually modified in Admin—particularly things like map brackets—and must be run with `reset=False` (i.e. we load the TSV, parse it, and update).  

In the future, we should aim to make `reset=True` the default and ensure all explorers can be created from scratch cleanly.

## Changes
- Add JSONSchema for explorer config files
  - I've created it by copying mdim JSONSchema and adding missing fields I found in config files
  - It's not perfect, but it's useful for catching basic errors in config files (there's no strict validation of the schema!)
- `explorer_migration` now converts strings to booleans, ints and floats
- Renamed `explorer_path` to `catalog_path` and made it similar to the one we use in MDIMs
- `pickerColumnSlugs` in explorers is weird, it's a list, but in a TSV it must be in a single cell, separated by space (unlike `selection` which is separated by `\t` in the TSV)
- Renamed `projection` to `region` in energy_prices


pabloarosado (approved):
Thanks, it looks good! I should have inspected this PR before refactoring the animal-welfare explorer. I may have to adapt that one a bit after merging this.

Regarding map brackets, we can discuss it next Monday (I added it to the list).

PR #4212, opened 2025-04-02
📊 Migrate poverty explorers from owid-content

Marigold:
Move poverty explorers from [owid-content](https://github.com/owid/owid-content) to `scripts/` folder. The only change I made was replacing writing to file by upserting to DB (via Admin). Having it here on a staging server comes with benefits like `explorer-diff` etc.

We could go one step further and copy those scripts to their own `export://explorers` steps (with the only dependency on `data://explorers` step that generates CSV file, but it'd still be just copy & pasted code with `run` method). ETL would then run them automatically if they change. Not sure if it's worth it, though.

## TODO before merging
- [x] Check that TSV files are identical in [explorer-diff](http://staging-site-migrate-poverty-explorers/etl/wizard/explorer-diff)

### DIFF
--- a/apps/wizard/app_pages/explorer_diff/app.py
+++ b/apps/wizard/app_pages/explorer_diff/app.py
@@ -1,5 +1,6 @@
 import json
 import random
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
 import pandas as pd
@@ -35,6 +36,9 @@ CURRENT_DIR = Path(__file__).resolve().parent
 # DB access
 # Create connections to DB
 SOURCE_ENGINE, TARGET_ENGINE = get_engines()
+
+# Params
+MAX_DIFF_LINES = 100
 
 
 def _show_options():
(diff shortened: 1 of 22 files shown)


paarriagadap (approved):
Thanks @Marigold! It looks good!

I only wonder why we can't see more explorers beyond `inequality` in explorer-diff. I would like to check the rest.

Precisely in explorer-diff I see that the tsv was produced differently: the order of tables change. Do you know why? It seems the content is the same though:
image

Regarding moving these to explorer steps, I think it's a good idea, but more when I will slowly migrate them to a new pipeline. Seems overkill to do many of such changes now when I can just run the steps manually (as I am always aware of them with each update)

Marigold:
> I only wonder why we can't see more explorers beyond inequality in explorer-diff. I would like to check the rest.

It only shows explorers with changed TSV config file. You can show all in options:
image
(You're right that it's confusing, I might change this to something else)

> Precisely in explorer-diff I see that the tsv was produced differently: the order of tables changed. Do you know why? It seems the content is the same though:

I refactored the `save` function which could be the reason why it changed, but the other explorers are identical. (This one too, the order doesn't matter)

> Regarding moving these to explorer steps, I think it's a good idea, but more so when I slowly migrate them to a new pipeline. Seems overkill to do many of such changes now when I can just run the steps manually

Yep, agree! I'll merge it as it is then.

PR #4224, opened 2025-04-04
📊 agriculture: Move crop yields explorer to ETL

pabloarosado:
* Migrate crop yields explorer to become an export explorers step in ETL.
  * NOTE: I deleted the old csv-based explorer step, instead of archiving it, to avoid confusion.
* Improve metadata of its dependencies.
* Add a (nasty) parameter `avoid_duplicate_hack` to `Explorer` to avoid the duplicate transform hack. This way, certain important metadata is propagated automatically.
  * NOTE: @Marigold Ideally, instead of adding this parameter, we could detect whether the duplicates transform is even necessary. However, I couldn't figure how to do that easily. So, given that this is a temporary solution anyway, I'm happy with this parameter. If you dislike it too much, please let me know, and we can try to find a better solution.


Marigold (approved):
The new version looks much nicer in explorer diff! I have nothing against `avoid_duplicate_hack`, seems like a reasonable way to get shorter config files.

PR #4226, opened 2025-04-04
🔨 agriculture: Move global-food explorer to ETL

pabloarosado:
Move all the code to generate the global-food explorer from `owid-content` to ETL.
Initially, I move that logic to `etl/scripts/faostat/`, but then I realised I could directly create an `export://explorers` step for global-food. The "weirdness" here is that this step still creates a csv-based explorer! But at least it's handled fully from ETL.

@Marigold I still want to have a go at making the global-food explorer indicator based, but I can foresee that there will be many complications, so this intermediate solution is probably a safer bet for now.

I've also merged the old `faostat_food_explorer` and `food_explorer` steps into one, and improved and simplified things a bit.


etl/steps/export/explorers/faostat/latest/global_food.py
```diff
@@ -0,0 +1,514 @@
...
+    #
+    # Load the template for the explorer tsv file.
+    template = Template(TEMPLATE_CONTENT)
+    # Load the foods data.
+    foods_df = pd.read_csv(StringIO(FOODS_CONTENT), index_col="slug", dtype=str)
+    # Load the views per food data.
+    views_df = pd.read_csv(StringIO(VIEWS_PER_FOOD_CONTENT), dtype=str)
+
+    # Determine whether the script is executed on staging or production.
+    if OWID_ENV.name == "production":
```
Marigold, line 382: This is nice, I'll move it right before sending the TSV to Admin API (but merge as it is).

PR #4235, opened 2025-04-07
🐝 Fix run-until-cursor extension not working after latest VSCode update

pabloarosado:
After a recent VSCode update, the `run-until-cursor` extension was not working. It seems that one of the parameters, `cursorDelay` needs to be increased. I've created a new version where the default values are adapted, and also changed the default values in the local `.vscode/settings.json` file.


Marigold (approved):
Looks good!

PR #4240, opened 2025-04-08
📊 agriculture: FAOSTAT fixes and improvements

pabloarosado:
* Fix issue with the number of live turkeys in EU, Europe, and HIC.
* Fix chart on the hypothetical meat demand per person (which was based on production, but it makes more sense to base it on consumption). Note that this means that the data changes significantly, and also the latest point is 2022 instead of 2023.
  * Also, fix wrong attribution (see [discussion](https://owid.slack.com/archives/C46U9LXRR/p1744109863123159)).
* Create data for a new chart on the hypothetical number of land animals slaughtered if all countries had the same per-capita meat production.


spoonerf (approved):
Looks good to me and makes sense, I would just suggest a different colour palette for the new chart.

PR #4243, opened 2025-04-08
📊  vaccine introductions: make an mdim

spoonerf:
Here's the preview: http://staging-site-data-vaccine-schedule/admin/grapher/health%2Flatest%2Fvaccination_introductions%23mdd-vaccination-introductions-who?description=aP+%28acellular+pertussis%29+vaccine

Is it possible to remove the 'Chart' tab? When I add `"hasChartTab": False` to the config I get this error: `ValueError: Config validation error: data.views[0].config must not contain {'hasChartTab'} properties`


pabloarosado:
@spoonerf I think that, in [the latest version of the grapher schema](https://files.ourworldindata.org/schemas/grapher-schema.007.json) there is no "hasChartTab". There is "chartTypes", which I suppose should be [] to have no chart tab. However, I've tried that (and None and other things), and it doesn't work. I'll ask around.

pabloarosado (approved):
It looks good. I was wondering why you are using "https://files.ourworldindata.org/schemas/grapher-schema.005.json"
instead of 007.
Apart from that, feel free to merge.

PR #4244, opened 2025-04-08
📊 agriculture: Make global-food explorer an indicator-based explorer

pabloarosado:
## Summary
The goal of this PR is to make the global food explorer indicator-based ([old](https://ourworldindata.org/explorers/global-food) | [new](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food)). This greatly improves its maintenance (good for OWID) and also improves the quality of the metadata shown in the explorer (good for users).

Additionally, this PR improves on some of our explorer/mdim tooling.

## Main changes
* Add features to `expand_config` function:
  * Let it ingest multiple tables. NOTE: Initially I thought it would be better to merge all necessary indicators into one table, and pass that to `expand_config`. However, then I realised that we need the full grapher step URI of the original tables, which was lost when merging. That's why I thought it would be better to ingest multiple tables.
  * Let it set the default view (defined as a dictionary).
  * Let it "humanize" the names of dimensions, when loaded programmatically.
* Simplify and improve global food explorer to be a single indicator-based export explorers step.
  * To achieve this on a table that was wide-format from garden, I "manually" insert the `metadata.dimensions` to each column. I thinks this is a reasonable solution to have multidimensional tables that propagated metadata properly.
* Remove unnecessary steps.
* Improve metadata handling of FAOSTAT steps.

## TODO (Pablo):
* [x] Fix wrong name of the first dropdown "Allocated to human food"?
* [x] Add missing indicator in food explorer (on food in grams per day, which is a duplicate transform of another indicator, not sure which way is best).
* [x] Change units of several per capita elements (e.g. production, in kg, instead of tonnes).
* [x] Keep improving remaining metadata fields.
* [x] ~Add `pickerColumnSlugs`.~ This seems to be more complicated than I expected, so, for now, I'll skip it. If someone misses it I'll have another look.
* [x] Share explorer with the team (bug-bash style).
* [x] ~Update documentation of `expand_config`.~ We decided to refactor things a bit.
* [x] Revert changes to `expand_config` and combine tables elsewhere. Or at least rethink the current logic.

## Review
* [ ] @lucasrodes would you mind reviewing the changes in `expand_config`? I want to ensure that these changes do not affect existing explorers or mdims. The way I'm checking that is by ensuring that [explorer diff](http://staging-site-data-global-food-explorer/etl/wizard/explorer-diff) and [mdim diff](http://staging-site-data-global-food-explorer/etl/wizard/mdim-diff) show no differences (except in the global food and animal welfare explorers, that I have modified).
  * NOTE: For some reason, I see a difference in the `democracy` explorer, but I don't think this is related to my changes.


lucasrodes (commented):
Let's book a session to review this together. This is an important change, which also challenges my initial idea for this function: that it should operate at table level. 

I think that keeping this function minimal (operating at single-table-level), and then developing functions to merge configs is a slightly more modularized and scaleable strategy. It also leaves more flexibility to the end user to create different configs for different indicators (e.g. "use this set of choices for age for population, but this other set for life expectancy"). Basically, my approach had been something like:

```py
from etl.collections.beta import combine_Explorers
from etl.collections.common import expand_config

config1 = expand_config(tb1, ...)
config2 = expand_config(tb2, ...)

explorer1 = paths.create_explorer(
        config=config1,
)
explorer2 = paths.create_explorer(
        config=config2,
)

explorer = combine_explorers([explorer1, explorer2], "explorer_name")
```

It's more verbose, but maybe we can work out better ergonomics on a higher level and avoid tweaking the lower level `expand_config`? I am not certain at all, so I am totally open to re-consider, of course.

Some of my concerns are regarding some cases: multiple tables have indicators with the same name (how does `indicator_name` work in those cases?); or what if different tables have different dimensions?

Having said this, I'm open for `expand_config` to accept more than one table, but conditioned on:
- All tables passed to the function have the same indices (i.e. same dimensions).
- Each table has different indicator names. That is, the same indicator name does not appear in multiple tables.

If any of the above conditions is not met, the function should fail to run.


The argument `default_view` only concerns explorers, so I'd leave out of this function. Note that `expand_config` is meant for both MDIMs and explorers.

Finally, I like the idea behind arguments `humanize_dimension_names`, `dimension_names_transformation` and `dimension_names_replacements`. I've worked out a similar solution in the experimental function `create_explorer_experimental` (see argument `choices_renames`), which I was hoping could become the default function to create an explorer in the future. It only addresses renames for choices, and leaves the dimension renames out of its duty (it is expected for the user to rename them in the YAML file).

Let's chat whenever you have a moment.

etl/steps/export/explorers/faostat/latest/global_food.py
```diff
@@ -1,514 +1,442 @@
...
+        humanize_dimension_names=True,
+        common_view_config={"tab": "map"},
+        dimension_names_replacements={
+            "item": "Food",
+            "maize": "Maize (corn)",
+            "area_harvested": "Land use",
+            "feed": "Allocated to animal feed",
+            "food": "Allocated to human food",
+            "other_uses": "Allocated to other uses",
+        },
```
lucasrodes, line 422: Interesting, I thought it would be more intuitive to have this spelled out in a YAML file, but guess we might want to do this programmatically sometimes, too.

pabloarosado: This is just a subset of replacements in special cases. Exploding all dimension names and choices in the yaml would be inconvenient.

pabloarosado:
Thanks a lot @lucasrodes! After our conversation, we decided that:
* Our tooling to create an explorer should let us combine multiple tables. This is necessary because we may have tables with identical structures but different origins (and merging them would cause all origins to be mixed).
* The combination of tables may not need to happen inside `expand_config`, but rather elsewhere. One option would be to add a method to `PathFinder` that automatically combines all tables in the dependencies, and ingests the corresponding config from the yaml.

We have realised that we have implemented similar features in different places, so it would be better to avoid unnecessary complexity and repetition. I'll try to refactor things and possibly revert my changes in `expand_config`, as long as this doesn't delay things too much (and if so, we can refactor in a separate PR).

pabloarosado:
Hey @lucasrodes, thanks for the review!
I reverted to the old `expand_config` and encountered several issues, so I ended up reverting again to the new implementation of `expand_config`, with a few improvements.

This is what I experienced when using the old version:
```
config_new = expand_config(tb_qcl)
```
This raises an error `ValueError: There are multiple indicators, but no `indicator_name` was provided. Please specify at least one!`. I guess this error is unnecessary (-> the new `expand_config` infers the indicators from the table).

To bypass the error, I defined indicator names:
```
config_new = expand_config(
        tb_qcl,
        indicator_names=[column for column in tb_qcl.drop(columns=tb_qcl.m.primary_key).columns],
    )
```
And for some reason, an "indicator" dimension is introduced (even though `indicator_as_dimension` was `False` by default), so I had to manually remove that spurious dimension:
```
for view in config_new["views"]:
    view["dimensions"].pop("indicator", None)
config_new["dimensions"] = [dimension for dimension in config_new["dimensions"] if dimension["slug"] != "indicator"]
```
I see that the issue was in [this line](https://github.com/owid/etl/blob/67b0d167653f25a96367e36fda4ba770c42f655f/etl/collections/common.py#L197), where, if there are multiple indicators, the "indicator" dimension was added, regardless of `indicator_as_dimension`. I'm not sure why this was the case (-> the new `expand_config` respects the user choice of not using indicator as a dimension).

I've also had a look at the `un_wpp` explorer (where partial explorers are created and then combined). I realised that, if you mix explorers with different dimensions, you are only warned at the very end, after having created the explorers and combined them (by the way, that explorer actually raises several warnings already, but they seem to be unrelated). I think it is more convenient to be warned earlier on, without needing to create partial explorers (-> the new `expand_config` now raises a warning straightaway if dimensions differ between input tables).

I've also executed all explorers and mdims with the new `expand_config` and there are no changes. You can see [explorer diff](http://staging-site-data-global-food-explorer/etl/wizard/explorer-diff) and [mdim diff](http://staging-site-data-global-food-explorer/etl/wizard/mdim-diff) (where the differences seem to be related to the variable ids in duplicate transforms, which is to be expected).

So even though I agree we should homogenize these tools (and eventually integrate them in PathFinder), I think the new `expand_config` is an improvement over the old one; the alternative way of combining tables (as in un_wpp) using partial explorers, is a bit cumbersome. We may need to rethink this a bit, but it may take a few days to get it right. So, what do you think, shall we go ahead with these changes, or do you see any other risks? Thanks.

pabloarosado:
Hi again @lucasrodes, after our last discussion, I reverted (again) the changes in `common`, to use the old `expand_config`. And I figured a way to make it work in the global food explorer without having to create and combine temporary explorers. I don't think that combining explorers (when the user intends to combine tables) is the best approach, but you have looked into this much more than me, so, if you think we should head in that direction, I trust your gut. We can discuss the implementation during the offsite.
Please let me know if you agree with the changes, thanks!

etl/steps/export/multidim/health/latest/vaccination_coverage.py
```diff
@@ -24,6 +24,7 @@ def run() -> None:
         indicator_names=["coverage", "unvaccinated", "vaccinated"],
         dimensions=["antigen"],
         indicators_slug="metric",
+        indicator_as_dimension=True,
```
lucasrodes, line 27: totally fine to keep, but not needed

lucasrodes (approved):
lgtm!

thanks for looking into this and figuring out a solution.

let's keep thinking about what's more useful: combining dictionaries vs. combining mdim/explorer objects

etl/steps/export/multidim/health/latest/vaccination_coverage.py
pabloarosado: Oh, right, I think it may not be necessary after reverting changes (it may have been necessary while I was using the alternative `expand_config` function for some reason).

pabloarosado: I've reverted it (nothing changed in mdim diff).

pabloarosado:
Hi @spoonerf, would you be able to have a look at the explorer ([old](https://ourworldindata.org/explorers/global-food) | [new](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food))? There's no need to look through the code (though feel free to do so). I think it would be good to you have your review, bug bash style, to spot potential data issues, as well as small aesthetic issues (e.g. ugly titles). Thanks a lot!

spoonerf (commented):
Hey @pabloarosado, great work tackling this monster. 

Here are some things that I spotted:  

* The description of the Explorer is 'Explore the world's food system crop-by-crop from production to plate.' but the explorer includes more than crops, so it might be nice to change this to be more accurate.
* I would set 'per capita' to 'Per capita' so it matches the other sentence case config options. Screenshot 2025-04-22 at 09 11 47
* The order in the 'Metric' drop-down is different now, is that intentional? It makes sense to have the ones users are most interested in at the top.
* I think for some options, such as 'Production' it could be nice to have a 'Per km2' option, it seems more relevant that 'Per capita' for more of the food production focused variables. I think we have the land area variable in the ETL already.
* This is probably just a FAOSTAT quirk but the [almond data for China is quite jumpy in recent years](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&country=~CHN&Food=Almonds&Metric=Production&per+capita=false&Unit=Grams+per+day). 
* For '[Apples allocated to other uses](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?Food=Apples&Metric=Allocated+to+other+uses&per+capita=false&Unit=Grams+per+day&country=~CHN)' there is only data for very few countries. It might make sense to drop views like this? It seems to be a common issue for this 'Allocated to other uses' metric.
* I'd set the numDecimalPlaces for Production (and maybe other relevant ones e.g. Land Use) to explicitly be 0. I think it largely is in the data but there are some cases where it's not and it's looks slightly off e.g. [Asparagus in Mexico](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&country=~MEX&Food=Asparagus&Metric=Production&per+capita=false&Unit=Grams+per+day).
* A missing 'l' in the '[Chillies and peppers yield](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?Food=Chillies+and+peppers&Metric=Yield&per+capita=false&Unit=Grams+per+day&country=~OWID_WRL)' 
* A slight variation in how ['Chestnuts' are referred to across the three drop-downs](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&Food=Chestnut&Metric=Land+use&per+capita=false&Unit=Grams+per+day&country=~OWID_WRL). I would just go for 'Chestnuts' across all. It might be worth adding that the shell is included in a footnote or something. At the moment including 'in shell' makes it read slightly strangely. 
* The [title of this view is a bit machine-y,](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&Food=Eggs+from+hens&Metric=Eggs+per+bird&per+capita=false&Unit=Grams+per+day&country=~OWID_WRL) I guess it is a bit of a special case. I think 'Eggs per hen' would be the best title. Also for 'Eggs produced' I'd go for 'Eggs laid by hens'.
* [This title is a bit confusing](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&Food=Eggs+from+hens&Metric=Laying&per+capita=false&Unit=Grams+per+day&country=~OWID_WRL). I don't think you need the information in the brackets? 
* Same issues in the 'Eggs from other birds' views. I would add in the subtitle or something what is meant by 'other' birds, e.g. birds that aren't hens. Perhaps you could say something like 'Production of eggs from birds that aren't hens'.
* Not a big deal but for the 'Fat, animal' I would use the singular rather than the plural of the animal. I think it just reads a bit better and would be consistent with the 'Meat, animal' versions. 
* I find the '(e.g. fennel)' a bit odd in the ['Herbs' views ](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&Food=Herbs%2C+e.g.+fennel&Metric=Production&per+capita=false&Unit=Grams+of+protein+per+day&country=~OWID_WRL). I think any clarification on what a herb is should be given in the subtitle and not the title or dropdown. 
* I know it's almost certainly a FAOSTAT weirdness but could it be worth finding a way to drop out the preceding [zeroes in cases like this](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?tab=chart&Food=Leeks&Metric=Production&per+capita=false&Unit=Kilocalories+per+day&country=~GBR). 
* An error in the data for South Korea [here](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?Food=Lemons+and+limes&Metric=Allocated+to+animal+feed&per+capita=false&Unit=Kilocalories+per+day&country=~GBR). I am increasingly convinced that views with very little data like this should be omitted. 
* 'An' is used as a short unit [here](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?Food=Meat%2C+camel&Metric=Producing+or+slaughtered+animals&per+capita=false&Unit=Grams+of+fat+per+day&country=GBR~OWID_HIC~OWID_WRL), which I think is a mistake. Also in per capita view and in other 'Meat, animal' views. 
* Something strange going on in the legend [here](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?Food=Meat%2C+chicken&Metric=Producing+or+slaughtered+animals&per+capita=false&Unit=Grams+of+fat+per+day&country=GBR~OWID_HIC~OWID_WRL) (and also with the other bird meats). It looks like the first two bins are combined, 0 and 1000? 
* I'd switch 'offals' to 'offal' in all cases, I think 'offals' can be technically right when you are referring to the offal of multiple different animals but generally 'offal' is the plural as well as the singular. 
* If possible it would be nice to add some detail to description_short about what some of the more unusual items are e.g. oilcrops (cake equivalent). I know that's not a small task though and is likely on your radar already! Also it would be useful to know what's included in some of the broader groups e.g. 'Roots and tubers'.
* Angry red map for '[Olive oil food available for consumption - grams of protein per day'](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?Food=Olive+oil&Metric=Food+available+for+consumption&per+capita=false&Unit=Grams+of+protein+per+day&country=GBR~OWID_HIC~OWID_WRL). Same for '[Sesame oil'](http://staging-site-data-global-food-explorer/admin/explorers/preview/global-food?time=latest&Food=Sesame+oil&Metric=Food+available+for+consumption&per+capita=false&Unit=Grams+of+protein+per+day&country=GBR~OWID_HIC~OWID_WRL). 
* The inclusion of 'Silk' surprises me, I wouldn't have thought of that as a food, but I support it is an animal product, same for 'Wool'.  
* A more broad design point, I think it would be nice to introduce an organisational layer, so that all fruits are grouped together, all animal products etc.

pabloarosado:
Thanks @spoonerf I've implemented most of your suggestions. I also noticed that we used to have some footnotes in the old explorer that were missing in the new. I adapted some of those footnotes and added them to the short descriptions. Please let me know if you think it's ready to go. Thanks!

spoonerf:
Amazing, looks great to me, thanks @pabloarosado!

PR #4245, opened 2025-04-08
📊 Update World Bank PIP pipeline in preparation of June update

paarriagadap:
Dedicated issue here: https://github.com/owid/owid-issues/issues/1915

Part of this overarching issue: https://github.com/owid/owid-issues/issues/1902

@lucasrodes, @pabloarosado, could you take a look at this? This is a refactoring of the PIP pipeline that creates a dimensional dataset (`world_bank_pip`), which will be the main one now for Grapher, MDims, and the future explorer migration. At the same time, there is a `world_bank_pip_legacy` which is a generalized version of the old PIP pipeline, with less fixed references to address the change in prices coming in June better.

The data is exactly the same as the previous version, this is mainly a test to see how it will work when we have data in 2017 and 2021 prices in June.

How about

- @lucasrodes takes a look at garden in `world_bank_pip` and the MDim
- @pabloarosado takes a look at `world_bank_pip_legacy` (just an overall look, because it will be deprecated some day), and see if changes in scripts/poverty-inequality-explorers keep the same views for explorers

Of course Pablo has been working with me on the MDim, so you can both make comments about it. I am still waiting for feeback from Joe, but maybe there are things we can improve before that.

I know the file poverty_inequality_file takes a good while to execute. It works, though, so I am happy to not think too much about it for now.


lucasrodes:
nit: you can use the new defaults to have slightly cleaner code (applies to meadow, garden and grapher):

- You don't need to import `create_dataset`
- You should use `paths.create_dataset` instead of `create_dataset`, which doesn't need `dest_dir` as argument
- You don't need to have `run(dest_dir)`, but can just have `run()`


**❌ Old**
```python
from etl.helpers import PathFinder, create_dataset

# Get paths and naming conventions for current step.
paths = PathFinder(__file__)

    # Code
   ds = create_dataset(
        dest_dir,
        tables=tables,
    )
```

**✅ New**

```python
from etl.helpers import PathFinder

# Get paths and naming conventions for current step.
paths = PathFinder(__file__)

def run() -> None:
    # Code
   ds = paths.create_dataset(
        tables=tables,
    )
```

etl/steps/data/garden/wb/2025-04-14/world_bank_pip_dimensional.py
```diff
@@ -0,0 +1,1461 @@
+"""
```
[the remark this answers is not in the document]
pabloarosado: Hey, I haven't looked into it in detail, but I can quickly see that `tb["headcount_ratio"].metadata.dimensions["decile"]` is `nan`. I suppose that's what's causing the problem. Does that help? Otherwise I'm happy to help (let's debug together on Slack).

etl/steps/data/garden/wb/2025-04-14/world_bank_pip_dimensional.meta.yml
```diff
@@ -0,0 +1,928 @@
...
+    <% if poverty_line in ["40% of median", "50% of median", "60% of median"] %>
+    YlOrBr
+    <%- else %>
+    OrRd
+    <%- endif %>
+
+  map_brackets_headcount_ratio: &map_brackets_headcount_ratio
+    <% if poverty_line in ["190", "215"] %>
+    [3, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100]
+    <%- elif poverty_line in ["40% of median"] %>
```
paarriagadap, line 264: Hi @Marigold! How can I make these jinja-brackets work? I get an error that this is not an array? Would we need to add an exception for this category, as you've done in the past?

paarriagadap: And I think there is a similar issue with map and hasmap tab, but it doesn't break the code

Marigold: I'll see what I can do.

Marigold: Check this out https://github.com/owid/etl/pull/4369

etl/steps/data/garden/wb/2025-04-14/world_bank_pip_dimensional.meta.yml
```diff
@@ -0,0 +1,1223 @@
...
+            hasMapTab: "{definitions.hasmaptab}"
+            tab: "{definitions.tab}"
+            yAxis:
+              min: 0
+            map:
+              colorScale:
+                baseColorScheme: "OrRd"
+                binningStrategy: "manual"
+                customNumericValues: *map_brackets_share
+                customNumericMinValue: "{definitions.custom_numeric_min_value_share}"
```
paarriagadap, line 1060: @Marigold So all was fixed, except for this one! I am keeping this as zero for now, but I want to put "{definitions.custom_numeric_min_value_share}"

Marigold: Should be fixed.

paarriagadap: Thank you so much again!

etl/steps/data/garden/wb/2025-04-14/world_bank_pip_dimensional.meta.yml
```diff
@@ -0,0 +1,1477 @@
...
+    *faqs_general
+    <% endif %>
+
+
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 180
+
+
```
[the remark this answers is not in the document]
Marigold: Pablo found an elegant solution

```
faqs_test: &faqs_test
  - fragment_id: <% if poverty_line in ["190", "215"] %>poverty-international-poverty-line<%- endif %>
    gdoc_id: "{definitions.gdoc_id}"
  - fragment_id: poverty-international-dollars
    gdoc_id: "{definitions.gdoc_id}"
```

Here's a [more complex one](https://github.com/owid/etl/pull/4381) if we ever need it.

etl/steps/export/multidim/wb/latest/world_bank_pip.py
```diff
@@ -0,0 +1,84 @@
...
+        config_dimensions_yaml=config.get("dimensions", {}),
+    )
+    config["views"] = config_new["views"]
+
+    # Create mdim
+    mdim = paths.create_mdim(
+        config=config,
+        mdim_name="poverty",
+    )
+
```
paarriagadap, line 61: I tried to implement group_views to have several indicators from the views available in a new view. I tried to replicate Lucas' example in mars.py, but doesn't seem to work. Do you know why, @pabloarosado?

paarriagadap: Ah, taking a closer look I guess it combines for all the views in the MDim, which is not what we want.

etl/steps/export/multidim/wb/latest/world_bank_pip.config.yml
```diff
@@ -0,0 +1,103 @@
...
+      #   description: Relative poverty
+      # - slug: multiple_lines_relative
+      #   name: Multiple lines
+      #   description: Relative poverty
+  - name: Household survey data type
+    slug: table
+    choices:
+      - slug: Income or consumption consolidated
+        name: Show data for both income and consumption
+      - slug: Income
```
paarriagadap, line 84: @pabloarosado I wanted to try to have a multi-indicator view from yaml to understand how it works. This is executed, but doesn't change the output. Do you know why?

etl/steps/export/multidim/wb/latest/world_bank_pip.config.yml
```diff
@@ -0,0 +1,1358 @@
...
+  - Bangladesh
+  - Bolivia
+  - Madagascar
+  - India
+  - China
+  - Ethiopia
+
+default_dimensions:
+  indicator: headcount_ratio
+  poverty_line: "215"
```
paarriagadap, line 14: @pabloarosado So this is the line that doesn't work. poverty_line does though

lucasrodes:
@paarriagadap, according to the unittests there seems to be some error in the schema validation. It might be related to your changes to the schema. Could you review it?

lucasrodes (commented):
I've looked at the code at a high level and it looks fine.

I'm a bit impressed by the length of the config of the MDIM. I think that there is room to craft it in a more automated way by using the tools that come out of the box. I am happy to help anytime. I think you could highly benefit from `Collection.group_views`... I've used it in all my war mdims.

The downside of such big config files is that it makes it a bit harder to interpret and to maintain.

///
As a side comment, I'd prefer smaller PRs for review. This one is quite elaborated and with lots of files affected: `lib/catalog`, `schemas/`, `etl/...`, etc.

etl/steps/data/garden/wb/2025-04-08/world_bank_pip_legacy.py
```diff
@@ -0,0 +1,1465 @@
...
+
+
+def create_survey_spells_inc_cons(tb_inc: Table, tb_cons: Table) -> list:
+    """
+    Create table for each indicator and survey spells, to be able to graph them in explorers.
+    This version recombines income and consumption tables to not lose dropped rows.
+    """
+
+    tb_inc = tb_inc.reset_index().copy()
+    tb_cons = tb_cons.reset_index().copy()
```
pabloarosado, line 1304: I've seen various instances of `reset_index().copy()`, and I think the `copy()` can be dropped, since `reset_index()` creates a copy already.

etl/steps/data/garden/wb/2025-04-08/world_bank_pip_legacy.py
```diff
@@ -0,0 +1,1465 @@
...
+
+    # Identify columns to use (ID + indicators)
+    id_cols = ["country", "year"]
+
+    tb_ppp_old = define_columns_for_ppp_comparison(tb=tb_ppp_old, id_cols=id_cols, ppp_version=PPP_YEAR_OLD)
+    tb_ppp_current = define_columns_for_ppp_comparison(tb=tb_ppp_current, id_cols=id_cols, ppp_version=PPP_YEAR_CURRENT)
+
+    # Rename all the non-id columns with the suffix _ppp(year)
+    # (the suffix option in merge only adds suffix when columns coincide)
+    tb_ppp_old = tb_ppp_old.rename(
```
pabloarosado, line 1334: When using `drop` and `rename` programmatically, it's convenient to use `errors="raise"` for safety. Otherwise, explicitly use `errors="ignore"` if that's what you need in that case.

pabloarosado (approved):
The garden step of `world_bank_pip_legacy` alone has nearly 1500 lines! I haven't followed the logic thoroughly, but simply skimmed through it and added some minor code suggestions. If you want me to dig deeper into any part of it, please let me know where to focus, and I'll do that. Otherwise, we can focus on inspecting the outcomes (charts and explorer views).

etl/steps/data/garden/wb/2025-04-14/world_bank_pip.meta.yml
```diff
@@ -0,0 +1,1740 @@
...
+    $10 and $20
+    <%- elif poverty_line == "2000 and 3000" %>
+    $20 and $30
+    <%- elif poverty_line == "3000 and 4000" %>
+    $30 and $40
+    <%- elif poverty_line == "1000 and 3000" %>
+    $10 and $30
+    <%- else %>
+    <<poverty_line>>
+    <%- endif %>
```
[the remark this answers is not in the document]
paarriagadap: Yeah, handling this is tricky because sometimes I want to show the decimals (2.15, 1.90, etc) and sometimes I don't (1, 10, 20). For now it's working but I will think about a way (if Eurostar's internet allows it)

paarriagadap:
Thanks @lucasrodes and @pabloarosado!

Pablo, I thought the description I added was enough, I didn't want much more than that. And yes, the PR is a draft, but only because I didn't press the button. The PR is ready for review anyway.

Sorry for the PR to have grown to this level, it was mostly that I was modifying the pipeline and then Joe wanted MDims, so it was the only way to have them. All the schema changes are done by Mojmir, who included them in this PR to see them working.

As I tried to describe in the main description, I am modifying the pipeline to be more generalized, robust to changes in prices and without too many fixed definitions in the code. The idea with the explorers is them to be identical to the current ones, because I want to check that all the relative changes I am implementing are working the same as the fixed references.

paarriagadap:
@lucasrodes Regarding the error in the schema validation, it started to happen since you merged master. Now it's not recognizing welfare_type dimension. Is it possible that a change in schema produced this, @Marigold?

paarriagadap:
Thanks @Marigold, I fixed that by filling the nan with a text value. I don't use that table in Grapher, but it will have other uses for static charts

paarriagadap:
Thanks all! We just received confirmation that the World Bank will release the new data on 5 June, so I prefer to push this and create a new PR with the modifications needed for the new prices. I might tackle some of the remaining comments there (MDims, simplifying jinja)

PR #4246, opened 2025-04-08
🐛 Fix attribution

Marigold:
See [Slack](https://owid.slack.com/archives/C46U9LXRR/p1744109863123159)


Marigold:
@pabloarosado you wanted to take a stab at this, right?

PR #4248, opened 2025-04-09
🎉 Add Mdim Diff

Marigold:
Copy Explorer diff and adjust it for MDIM cases. It doesn't show an actual mdim chart, but it'll tell you when configs between staging and production differ.


pabloarosado (approved):
Looks good! (nit: there are a few instances of copy pasted "explorers" in the code)
Please merge, I'll use it straight away :)

PR #4251, opened 2025-04-10
🔨 Refactor explorer migration

Marigold:
TSV parsing into a structured format has been moved to the Admin API. Remove the code for parsing TSV files from the explorer migration and simply load the data from the database instead.

### DIFF
--- a/etl/collections/explorer_migration.py
+++ b/etl/collections/explorer_migration.py
@@ -3,179 +3,20 @@
 import re
 from collections import defaultdict
 from copy import deepcopy
-from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


lucasrodes (approved):
LGTM! Nice (literal) move.

I was wondering if maybe `migrate_influenza.ipynb` should be a python script instead. But not crucial.

Thanks!

Marigold:
> I was wondering if maybe migrate_influenza.ipynb should be a python script instead. But not crucial.

It's more flexible to have it in a notebook. Besides, I'm using it as a template for other explorers and updating it constantly.

PR #4254, opened 2025-04-14
✨ Add iframed mdims to mdim-diff

Marigold:
Added side-by-side comparison to `mdim-diff`. It’s not very useful yet, as it only works for published MDIMs.

This PR also refactors parts of both `mdim-diff` and `explorer-diff`. I didn’t go as far as unifying them, since they still differ in a few ways. Some features (like config diffs) are unique to `explorer-diff` and may become unnecessary once we finish migrating CSV-based explorers.

It’s also possible we take a different approach - showing multiple views instead of a single explorer with controls, or even integrating this into `chart-diff`.


lucasrodes:
@Marigold I still need to review it, but just checked the apps and found there is an error being raised in Explorer diff

image

lucasrodes (approved):
LGTM, thanks for the clean up too!

PR #4256, opened 2025-04-14
📊 Migrate water-and-sanitation explorer to ETL

Marigold:
Converted `water-and-sanitation` to an indicator-based explorer. The migration was done by running `migrate_csv_explorer` and then programmatically modifying the config (all code is in `migrate_wash.ipynb`).

I also looked into the rules for setting color properties, but the final decision tree turned out to be quite complex - so I left them hardcoded for now. We should revisit this during the next update.  

[explorer-diff](http://staging-site-water-and-sanitation/etl/wizard/explorer-diff) shows parity with the current explorer.


etl/steps/export/explorers/wash/latest/water_and_sanitation.config.yml
```diff
@@ -0,0 +1,2040 @@
...
+          display:
+            name: No usage of drinking water facilities
+            type: Numeric
+            unit: '%'
+            shortUnit: '%'
+            additionalInfo: |-
+              Having no access to drinking water means people rely on surface water: directly from a river, dam, lake, pond, stream, canal or irrigation canal.
+            colorScaleScheme: YlOrRd
+            colorScaleNumericBins: 2.5,,;5,,;10,,;25,,;50
+            colorScaleEqualSizeBins: false
```
spoonerf, line 189: I think this should be true as default (as it is on Grapher)
Screenshot 2025-04-15 at 16 27 59 at the moment there are funky bins on staging

spoonerf (approved):
Hey @Marigold, thanks for this! It looks all good to me, I only noticed one thing that is off, the bins should always be equal size (visually), and for some of the views they are not.

etl/steps/export/explorers/wash/latest/water_and_sanitation.config.yml
Marigold: Good catch! There were two columns with `colorScaleEqualSizeBins` in production which caused this. Fixed.

PR #4257, opened 2025-04-14
📊 AI: update Epoch datasets

veronikasamborska1994:
hey @Marigold ! It's getting a bit tedious to copy the code and update all of these datasets and charts every month so I am using your solution here :) hope that's ok!


Marigold (approved):
Amazing, reviewing this is so much easier on the eyes. I really should finish that autoupdate PR and try it out on this dataset...

PR #4269, opened 2025-04-16
🐛 Silence sqlalchemy.exc.PendingRollbackError

Marigold:
Sometimes when I open a wizard on a staging server, I get the following error. It disappears after a refresh or two, but it's still annoying. It's hard to replicate, so this fix is a shot in the dark, but it could help (and shouldn't break anything)
image


lucasrodes (approved):
indeed an annoying exception. I think is fine to silence it and have a second retry.

I remember having this after two trials, but way less often.

Good to go!

PR #4283, opened 2025-04-18
📊 energy: adding custom variables to WDI

veronikasamborska1994:
adding custom energy variables to automatically update as a part of WDI as described in the issue [here](https://github.com/owid/owid-issues/issues/1916)


Marigold (approved):
Both code & chart diff look good, good to have it automatically updated as part of WDI.

PR #4299, opened 2025-04-22
Use `transform` in indicators that strictly need it

lucasrodes:
## Context
MDIMs allow for indicator-view display settings. Doing the same in explorers is a bit trickier but doable: Whenver we have different settings for the same indicator, we can define a duplicate and customize it accordingly.

Something along these lines had been implemented, yet was not optimal. We were relying using duplicates for all indicators in use, regardless of whether it was strictly necessary or not. For instance, when there is only one indicator setting, we don't really need to use `transform`!

This PR solves this, by only adding duplicates when really necessary.

## Side effect
As a side effect, we overcome a side bug: For a duplicate indicator to inherit the metadata from a duplicated indicator, we need a reference to the original indicator in the table `graphers`. This is actually something new, fixed in https://github.com/owid/owid-grapher/pull/4797.

Consequently, this PR, together with https://github.com/owid/owid-grapher/pull/4797, should also address https://github.com/owid/owid-grapher/issues/4680 

/schedule


lucasrodes:
@Marigold I'm experiencing some errors in the unit tests in this PR, but couldn't reproduce these errors on my local branch When I run `make test`, it all runs smoothly (i.e. no error is raised).

I was wondering two things:

1) Do you know why the error is not raised locally?
2) Would you mind having a look at this error? I am finding it hard to test it. I think it should be failing because the assertion is probably no longer valid, because the generated explorer config is now different. More context on this in the PR description, but basically `slug` is only used if strictly necessary.

Marigold:
> @Marigold I'm experiencing some errors in the unit tests in this PR, but couldn't reproduce these errors on my local branch When I run `make test`, it all runs smoothly (i.e. no error is raised).
> 
> I was wondering two things:
> 
> 1. Do you know why the error is not raised locally?
> 2. Would you mind having a look at this error? I am finding it hard to test it. I think it should be failing because the assertion is probably no longer valid, because the generated explorer config is now different. More context on this in the PR description, but basically `slug` is only used if strictly necessary.

1. No idea, it was failing for me locally.
2. Fixed in your PR.

Marigold:
Looks good! I reran all explorers with `etlr explorer --export --private` and did a couple of ad-hoc checks in explorer diff. I didn't find any differences, and the TSV file looks definitely nicer.

PR #4303, opened 2025-04-23
📊 igme update 2025

spoonerf:


Marigold:
@spoonerf if you mean the error
```
ValueError: Variables used in charts will not be deleted automatically:
   chartId  variableId
0     6215     1026560
1     7159     1026612
```
that can be fixed by running indicator upgrader **against itself**. It'll remap the old variable ids to the new ones. I did that for your PR to test it out.

Marigold (approved):
Code and chart-diff looks good!

spoonerf:
@Marigold Ah thanks! That's good to know. No worries about the unreviewed charts, I think I only approved a few any way and there are still some fixes to make.

PR #4325, opened 2025-04-30
🎉 Chart diff: Add analytics and anomaly score to prioritize reviews

lucasrodes:
### Summary
Chart diffs can now be sorted by a relevance score, based on the number of chart views and anomalies.

### Context
Each chart diff has scores assigned based on three things:
1) **Number of chart views** (`num_chart_views)`
2) **Number of articles** where the chart is linked (`num_articles_linked`)
3) **Anomaly score**, given by Anomalist. (`score_anomalist`)

Based on these scores, we've defined a "relevance" score, a weighted average of the regularized scores above. 

#### How is the relevance score estimated?
For a given chart _c_, we have:

image

where:
image
image
image
image
image

We've defined the following weights:
- **chart_views**: 5
- **article_refs**: 3
- **anomalies**: 1

### Task summary
- https://github.com/owid/etl/pull/4328)
  - Add new module for analytics in ETL. Various tools rely on analytics, make a module to host them all.
  - Add logic to get anomalies at chart-level in ChartDiff, from anomalist.
- Add analytics and anomaly scores in the chart diff UI
- Other cosmetic changes to chart diff
- Improve performance: reduce DB calls when accessing approval history of a chart diff. In the past, we were accessing the DB for _each_ chart diff. This PR changes this, so that we only access once, and get all the approval histories for all charts in one query.

/schedule


lucasrodes:
@pabloarosado @Marigold Nothing specific to review, just to confirm that you are fine with us merging this.

It includes all changes to chart diff, including those from Pablo's PR: https://github.com/owid/etl/pull/4328.

It does not include changes to MDIM/Explorer diff, but I think it's fine to merge on a separate PR.

I've fixed some performance issues, so that the app runs with acceptable timings.

pabloarosado:
Thanks a lot @lucasrodes! The new chart diff looks good. There are still several things I'd like to add and improve in the new analytics module (e.g. switching to metabase API instead of datasette), but I'm happy to do that in a separate PR. Regarding Explorer/MDIM diff, it would also be good to add the relevant code to the analytics module, but we can do that in a separate PR, as you suggested.

The main thing I'm not sure about is whether the relevance score is the most useful metric to sort charts. Maybe we could only consider the anomaly score if it exceeds a certain threshold. And maybe it would also be good to specify in chart diff where the biggest anomaly is (entity and year). We could discuss it later in the data eng call. We could also wait until Wednesday's data science call to discuss these things with others.

But I leave it up to you, if you prefer to ship this for now, we can improve on these things afterwards.

lucasrodes:
Thanks for weighing in, @pabloarosado.

From your comment I've identified four features (which I've collected in this issue: https://github.com/owid/etl/issues/4339):

1. Analytics code: Use Metabase
2. MDIM/Explorer: Integrate with changes in Chart diff
3. Change how we estimate the relevance score
4. More descriptive anomaly score

- On 1: I think that it'd be _very_ nice to standardize our tooling there and use the best tools we have. But I also think that it won't really affect the user's experience of the app, so I think it is fine to work on it after merging this PR.
- On 2: I think we should work on improving the analytics of MDIM/Explorer diffs, and align it with the backend and front-end of chart diff. I think it is fine to do in a separate PR (e.g. the one already open https://github.com/owid/etl/pull/4327, cc @Marigold)
- On 3: I think we can definitely improve how the relevance score is estimated. I also think that it is a good improvement over the status quo, so I'm pretty fine with having it be the default now. Since the anomaly contribution is minimal (56% is chart views, 33% is article references, and 11% is anomaly) for now, it is not adding much noise. We can look into increasing its contribution, and work on having a threshold in a separate PR.
- On 4: Totally.

3, and 4 may require some re-thinking of Anomalist, or the tables in DB, so it could take a bit longer. Would suggest doing it in a separate PR.

**TL;DR** Happy to merge, we can continue the discussion and click on "Merge" (if we decide so) in our data meeting

Marigold (approved):
Code looks pretty good! I was expecting something much worse, given the time pressure (pageviews in explorer-diff look terrible compared to this).

Just don't forget to reject all charts before merging.

PR #4328, opened 2025-05-01
🎉 Create functions for chart and article analytics

pabloarosado:
* Create useful functions to read analytics from Datasette and Metabase.
* Adapt Chart Diff to use the new functions (minor changes).
* Add Metabase credentials to etl.config and .env.example.


apps/anomalist/anomalist_api.py
```diff
@@ -736,3 +736,75 @@ def _sample_variables(variables: List[gm.Variable], n: int) -> List[gm.Variable]
         new_n=len(sample_ids),
     )
     return [v for v in variables if v.id in sample_ids]
+
+
+def get_anomalies_for_chart_ids(
+    chart_ids: Optional[List[int]] = None, anomaly_types: Optional[Tuple[str, ...]] = ("version_change",)
```
lucasrodes, line 742: @pablorosado I just realized that the correct name for version change anomalies is `upgrade_change`

pabloarosado: Oh, that's right, I just fixed it in this branch, thanks!

pabloarosado:
Hey @bnjmacdonald thanks for the suggestions on using Metabase. I've created a `read_metabase` function in this PR. There's no need for a very thorough review, but feel free to have a look at some of the functions there, in case that:
1. You spot something odd.
2. I'm reinventing the wheel (and some of those metrics already exist in the analytics repos).
3. Some of that logic should be moved to the analytics repos (and here we should simply call those tables).

Thanks!

.env.example
```diff
@@ -27,3 +27,8 @@ DB_HOST=127.0.0.1
 # If your owid-content repos is not in the same folder as etl, define the path here.
 # NOTE: The path should not point to the root, but to the explorers folder, where the tsv files are.
 # EXPLORERS_DIR="/full_path_to_owid_content_folder/explorers/"
+
+# Metabase credentials.
+# Uncomment and fill out the following variables if you want to interact with the Metabase API to gather analytics.
+# METABASE_API_KEY=***
+# METABASE_URL=***
```
lucasrodes, line 34: how can one get these? Also, will staging servers have this set by default? Otherwise some of the tools wouldn't work, right?

etl/analytics/common.py
```diff
@@ -1,63 +1,315 @@
...
+            df = pd.DataFrame()
+        else:
+            # Concatenate all chunks of data.
+            df: pd.DataFrame = pd.concat(dfs, ignore_index=True)  # type: ignore
+
+    return df
+
+
+def read_metabase(sql: str) -> pd.DataFrame:
+    """Retrieve data from the Metabase API using an arbitrary sql query.
```
lucasrodes, line 229: maybe worth having a dedicated module, e.g. `analytics.metabase`, where you can define all metabase-related tools?

etl/analytics/common.py
```diff
@@ -1,63 +1,315 @@
...
+    -------
+    str
+        Cleaned SQL query.
 
-def read_datasette(sql: str, datasette_csv_url: str = ANALYTICS_CSV_URL) -> pd.DataFrame:
+    """
+    return " ".join(sql.strip().rstrip(";").split())
+
+
+def read_datasette(
```
lucasrodes, line 155: maybe worth having a dedicated module for all datasette-related tools? e.g. `analytics.datasette`

lucasrodes (approved):
It lgtm, i just have some suggestions:

- Structure the code in submodules:
  -  e.g. have dedicated modules for different tools, e.g. `analytics.metabase`, `analytics.datasette`.
  - i'd rename `common` to something else, maybe `analytics.core`
  - those functions that are mostly used, i'd surface them in the `__init__.py`, so that they are easily accessible by `from etl.analytics import ...`
- How should one get the METABASE API tokens? Also, are staging servers set up with them?

.env.example
pabloarosado: Yes, good point, staging servers won't have those credentials, I'll look into it.

pabloarosado:
Thanks @lucasrodes! I agree that having the basic functions available in `analytics` would be convenient (in fact, initially I wanted to create just a single `analytics.py` file, but I encountered the problem that the interactive window didn't work, because of the clash of `collections` with the inbuilt Python function).
I'm not sure if splitting the module into smaller submodules is the best idea. The user shouldn't need to know whether a specific piece of analytics comes from Datasette or Metabase. For now, a single `analytics` module may be easier to maintain. If its content grows significantly, then we could split it into submodules. Ideally, any complex logic should be moved to the analytics repos, and here we should simply read things from the semantic layer.
So I think we should rather move towards a single file, `analytics.py`. Do you agree? If you prefer splitting, I don't have a strong opinion.

pabloarosado:
Thanks @lucasrodes, yes, what you just suggested would be my second preferred choice (it's just slightly annoying to have to manually add things to `__all__`, especially for a young module where things may change and be added in the near future). I'd start with just having an `analytics.py` module, and at some point, if needed, move to folders.
* I created https://github.com/owid/etl/pull/4364 to avoid the issue when running the interactive window in side `etl`. After that PR is merged, we can create the `analytics.py` module without any issues (I'd rather merge this and then refactor in a separate PR).
* I created https://github.com/owid/ops/pull/294 so that staging and prod can access Metabase credentials. For now, if you want to play around with this module locally, you can manually add the credentials to your `.env` file.

Again, no strong opinion about the structure. If you prefer folder structures, just say the word and I'll refactor accordingly. Thanks!

lucasrodes:
That makes sense, nice work!

pabloarosado:
I've created `analytics.py`. Now I'm waiting for https://github.com/owid/ops/pull/294 to be merged, and then I'll merge this one.
I suppose chart diff may break in existing staging servers, because it won't be able to find the credentials. In those cases, I can manually add them to the corresponding `.env` files of the affected servers (not sure if there is a better solution, that doesn't involve restarting those staging server).

etl/analytics.py
```diff
@@ -0,0 +1,833 @@
...
+    """
+    df_views = read_metabase(sql=query)
+
+    # To calculate the average daily views, we need to figure out the number of days for which we are counting views.
+    # This will be either the first date since we have analytics (DATE_MIN) or the publication date of the chart (if more recent than that).
+    published_at = pd.to_datetime(df_views["published_at"])
+    start_date = published_at.where(published_at > pd.to_datetime(date_min), pd.to_datetime(date_min))
+
+    # Add a column with the number of days that the views are referring to.
+    df_views["n_days"] = (pd.to_datetime(date_max) - start_date).dt.days
```
bnjmacdonald, line 386: small bug: b/c there is always a lag in analytics (of 1-2 days for page views data, longer for other external data sources like Google Search Console), the max date of the retrieved data might be less than `date_max`. Which means that the denominator in `views_daily` is larger than what it should be.
- e.g. consider a page published on 2025-05-05 with views data for 2025-05-05 to 2025-05-06 (2 days) available in `grapher_views_detailed`. If today is 2025-05-08, then the denominator for computing `views_daily` will be 4 but it should be 2.

Fix would be to retrieve the data's max date (`data_date_max = read_metabase(sql="select max(day) from grapher_views_detailed")`) and then replace `date_max` in this line with `min(date_max, data_date_max)`.

etl/analytics.py
```diff
@@ -0,0 +1,833 @@
...
+        Number of days to look back for views. Default is 30 days.
+
+    Returns
+    -------
+    pd.DataFrame
+        DataFrame containing the number of chart views per chart for the last n_days.
+
+    """
+    # Calculate date range.
+    date_max = str(datetime.today().date())
```
bnjmacdonald, line 418: same problem as previous comment that `date_max` won't be the same as the max date in `grapher_views_detailed`. So the retrieved days of data here will be less than `n_days`. 

Can fix with something like `date_max = read_metabase(sql="select max(day) from grapher_views_detailed")`

etl/analytics.py
```diff
@@ -0,0 +1,833 @@
...
+    where_clauses.append("url NOT LIKE '% %'")
+    where_clauses.append("url IS NOT NULL")
+    where_sql = "WHERE " + " AND ".join(where_clauses) if where_clauses else ""
+
+    query = f"""
+    SELECT
+        url,
+        SUM(views) AS views
+    FROM views_detailed
+    {where_sql}
```
bnjmacdonald, line 471: optional: if the purpose of `{where_sql}` is just to filter down to articles, topic pages, and data insights, then you can do this more cleanly by joining on the [`pages` table](http://metabase.owid.io/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjI2fSwidHlwZSI6InF1ZXJ5In0sImRpc3BsYXkiOiJ0YWJsZSIsInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==) which has a `type` column.

i.e. replace

```
FROM views_detailed
{where_sql}
```

with 

```
FROM views_detailed
JOIN pages USING(url)
WHERE type in ('topic-page', 'data-insight', 'article')
```

etl/analytics.py
```diff
@@ -0,0 +1,833 @@
...
+    GROUP BY url
+    ORDER BY views DESC
+    """
+    df_views = read_metabase(sql=query)
+
+    n_days = (pd.to_datetime(date_max) - pd.to_datetime(date_min)).days
```
bnjmacdonald, line 477: same issue as previous comments that `date_max` might be later than the max date of data available.

etl/analytics.py
```diff
@@ -0,0 +1,833 @@
...
+    df_views["n_days"] = n_days
+    df_views["views_daily"] = df_views["views"] / n_days
+
+    # Fix infs (for charts that were published in the last day).
+    df_views.loc[df_views["views_daily"] == float("inf"), "views_daily"] = 0
+
+    return df_views
+
+
+def _get_post_references_of_charts_and_redirected_charts(
```
bnjmacdonald, line 487: would it be helpful for the [`charts` table](http://metabase.owid.io/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjoyLCJxdWVyeSI6eyJzb3VyY2UtdGFibGUiOjE1fSwidHlwZSI6InF1ZXJ5In0sImRpc3BsYXkiOiJ0YWJsZSIsInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==) in the Semantic Layer to have a `references` array column that lists all the article URLs referencing the chart? 

Made a separate issue for it: https://github.com/owid/analytics/issues/342.

bnjmacdonald (commented):
Thanks a lot Pablo! I think it's fine for these functions to live in `etl` for now, and then consider a refactor down the road that moves `read_metabase` and some other core functions to the analytics repo.

As you work more in Semantic Layer, let me know if you have any requests for additional columns/tables that would make your work easier!

etl/analytics.py
```diff
@@ -0,0 +1,833 @@
...
+                dfs.append(df_chunk)
+                break
+
+        # Concatenate all chunks of data.
+        df = _safe_concat(dfs)
+
+    return df
+
+
+def read_metabase(sql: str) -> pd.DataFrame:
```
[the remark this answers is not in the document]
pabloarosado: That makes a lot of sense, but I'm not sure how we could put that in practice. ETL is public, while analytics is a private repo. We could do it the other way around though. Otherwise we'd need to rethink the current structure of the repos to avoid duplicating code. I haven't thought much about it, please let me know if you have any suggestions.

etl/analytics.py
pabloarosado: Very good point, I've generalised this code to consider this lag.

etl/analytics.py
pabloarosado: Thanks, I'll reply directly in the issue.

etl/analytics.py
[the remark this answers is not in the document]
bnjmacdonald: ah yes good point, I wasn't thinking about the fact etl is public

PR #4335, opened 2025-05-05
📊 R&D: update data on patents and articles

veronikasamborska1994:
hi @edomt! the data hasn't been updated but I am adding this code to the wdi script instead so it can be automatically updated whenever we update the wdi. I think it's a simpler way of doing this but let me know if you disagree? don't need to actually review the code as I can ask someone else to do it


edomt (approved):
All good on my end for the general method of doing this in WDI 🙂 
(should we archive the existing the code and rows in the DAG then?)

Marigold (approved):
Nice, charts look identical.

PR #4336, opened 2025-05-05
📊 oil spills: update ITOPF dataset

veronikasamborska1994:
pretty much the same code as last year


Marigold (approved):
Nice, charts look good!

PR #4357, opened 2025-05-07
🤖 Autoupdate: Surface temperature

Marigold:


veronikasamborska1994 (approved):
LGTM! thanks a lot! I've also reviewed the Epoch ones which also seem good!

on another note - should we add wildfires update to this workflow so we can review the charts every week - I am happy to do this as we've had some issues with the data in the past.

Marigold:
> LGTM! thanks a lot! I've also reviewed the Epoch ones which also seem good!

You can merge it whenever you want. A new PR will be created next day with fresh update.

> on another note - should we add wildfires update to this workflow so we can review the charts every week - I am happy to do this as we've had some issues with the data in the past.

Let me give it a shot to spare you from running into potential bugs.

PR #4358, opened 2025-05-07
🐝 Update grapher schema

Marigold:
Migrate all default schemas (see [slack](https://owid.slack.com/archives/C0193RW5E2J/p1746612292759529)):
> Valid values for the tab field used to be chart, map and table. This will change to Chart, WorldMap and Table


Marigold:
We've decided to park it for now and bundle it together with some bigger migration in the future.

PR #4364, opened 2025-05-08
🔨 Rename `etl.collections` as `etl.collection`

pabloarosado:
The current `etl.collections` module causes issues (e.g. when running an interactive window from within the `etl` folder, because the name clashes with the standard `collections` library. In this PR I renamed `etl.collections` as `etl.collection`.

To see this issue (from the `master` branch), go to any module inside `etl` (e.g. `command.py`), and press `shift+Enter` to initialize an interactive window. You'll see the errors:

```
ImportError: cannot import name 'deque' from 'collections'
The file 'etl/etl/collections/__init__.py' seems to be overriding built in modules and interfering with the startup of the kernel. Consider renaming the file and starting the kernel again. View Jupyter [log](command:jupyter.viewOutput) for further details.
```
This issue doesn't happen in this branch, after renaming `collections` to `collection`.


lucasrodes:
I'm looking at the error with UN explorer step.

lucasrodes (approved):
LGTM! Thanks, Pablo.

In the future, I think we should rename this module to simply `etl.multidim`, but let's see how it goes.

I've hotfixed a separate issue which was causing CI/CD to fail.

Issue #4333, opened 2025-05-02
Instant metadata updates


Marigold:
## Motivation

Changing metadata in a garden YAML file always triggers both garden & grapher steps. This can be slow if the garden step is doing some heavy lifting, but in practice we only need to update a single metadata field.

## Existing solution

There's a flag `--watch` you can use to run `etlr` continuously. Saving a YAML file then triggers both garden & grapher steps. This is still slow if those steps are slow though.

## Proposal

Add a flag `etlr ... --yaml` that will ONLY update dataset metadata from YAML file instead of running the garden file. It'll do it by:

1. Loading a dataset from disk
2. Updating metadata from YAML
3. Detect which variables have been updated
4. Set `GRAPHER_FILTER` to those variables and run grapher step

It can seamlessly work together with the `--watch` flag.

### Limitations

This might fail or give unexpected results in the following cases:

- Garden step doesn't call `paths.create_dataset` as the last command
- Grapher code is doing something with metadata
- Flattening in grapher step could take a long time

Marigold:
Implemented in https://github.com/owid/etl/pull/4372. If used enough, we can think about further improving performance and maybe even detecting when to use instant update automatically.

PR #4372, opened 2025-05-12
🎉 Instant metadata updates

Marigold:
See the problem in https://github.com/owid/etl/issues/4333.

## Solution

Instant metadata updates can be turned on with `INSTANT=1 etlr ...`. If on, instead of recalculating the garden step, it'll **load the existing garden dataset**, apply metadata YAML file and save it again. Then we detect what indicators have changed and save them in `INSTANT_METADATA_DIFF`. This dictionary will be then used in the `grapher://grapher` step to filter indicators that need to be updated.

## Notes
- It's not bulletproof, and it's always recommended to do the final run **without** `INSTANT=1` (it'll be done by the staging server)
- `data://grapher` is always run (so is `long -> wide` flattening) - if that is a bottleneck, we can try to further optimize it (though it could get tricky) with `INSTANT_METADATA_DIFF` (especially flattening)
- We're mixing env variables with `--` arguments. One might use command like `INSTANT=1 etlr ... --watch`. What's more convenient, env variables or cli args? (if we use cli args then we'd still save them into env variables to avoid passing arguments several levels deep)
- I tried it on a couple of (mid-sized) examples. Do you have a good use case I could optimize it against?

## How to test

I'm using `mars` dataset as an example, but please try to use your own dataset 🙏 

1. Run the step `etlr war/2023-09-21/mars --grapher`
2. Edit metadata of a single indicator in garden metadata YAML
3. Re-run with `INSTANT=1 etlr war/2023-09-21/mars --grapher` and verify that it's faster and that your changes are visible in Admin
5. Run `etlr war/2023-09-21/mars --grapher` to ensure the full run is triggered

### DIFF
--- a/docs/architecture/metadata/data_pages_workflow.md
+++ b/docs/architecture/metadata/data_pages_workflow.md
@@ -13,6 +13,7 @@ For illustration, the step `grapher/gcp/2023-07-10/global_carbon_budget` is cons
 
 ### Initial Configuration
 - Add `DEBUG=1` to your `.env.myname` file and any other `.env.*` files in use to expedite ETL operations. `DEBUG=1` runs all steps in a single process and removes 2s overhead from running each step.
+- Add `PREFER_DOWNLOAD=1` to your environment to use cached downloads of files instead of redownloading them each time, which can significantly speed up steps involving large file downloads.
 
 ### File Monitoring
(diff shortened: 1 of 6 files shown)


lucasrodes (approved):
Nice!! Let's try it 🎉 


I guess the next incremental improvement is to be able to "set" `INSTANT` to 1 automatically.

PR #4379, opened 2025-05-12
📊 energy: Update electricity data

pabloarosado:


etl/steps/data/garden/ember/2025-05-12/yearly_electricity.excluded_countries.json
```diff
@@ -0,0 +1,3 @@
+[
+  "EU"
```
antea04, line 2: Is there a reason we include ASEAN or G7 but no the EU?

etl/steps/data/garden/ember/2025-05-12/yearly_electricity.py
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    for table_name in tables:
+        # Prepare table's format.
+        tables[table_name] = tables[table_name].format(short_name=table_name)
+
+    ####################################################################################################################
+    # The data for many regions presents a big drop in the last year, simply because many countries are not informed.
+    # Assert that this drop exists, and remove the last data point for regions.
+    error = (
```
antea04, line 599: I think this might lead to deletion when data is also available (e.g. for Europe)? 

Could an alternative be to set regional aggregates only to be calculated if 90% (80%, 75%, 60%...) of countries have data available?

etl/steps/data/garden/energy/2025-05-12/electricity_mix.meta.yml
```diff
@@ -0,0 +1,536 @@
...
+    presentation:
+      attribution_short: Ember and Energy Institute
+      topic_tags:
+        - "Energy"
+    description_processing: |-
+      - We rely on Ember as the primary source of electricity data. While the Energy Institute (EI) provides primary energy (not just electricity) consumption data and it provides a longer time-series (dating back to 1965) than Ember (which only dates back to 1990), EI does not provide data for all countries or for all sources of electricity (for example, only Ember provides data on electricity from bioenergy). So, where data from Ember is available for a given country and year, we rely on it as the primary source. We then supplement this with data from EI where data from Ember is not available.
+  description_short_twh: &description_short_twh |-
+    Measured in terawatt-hours.
+  description_short_pct: &description_short_pct |-
+    Measured as a percentage of total electricity.
```
antea04, line 13: [...] total electricity production in this region (?)

etl/steps/data/garden/energy/2025-05-12/electricity_mix.meta.yml
```diff
@@ -0,0 +1,536 @@
...
+  description_short_per_capita: &description_short_per_capita |-
+    Measured in kilowatt-hours per person.
```
antea04, line 15: could include the dod

etl/steps/data/garden/energy/2025-05-12/electricity_mix.meta.yml
```diff
@@ -0,0 +1,536 @@
+definitions:
+  common:
+    processing_level: major
+    presentation:
+      attribution_short: Ember and Energy Institute
+      topic_tags:
+        - "Energy"
+    description_processing: |-
```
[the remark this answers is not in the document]
antea04: Maybe you can split this into multiple blocks and only include the relevant ones per indicator?

antea04:
Hey Pablo! 

Really cool update to the energy charts! Thanks so much for doing this, I think these charts are definitely some of the most relevant and (I think) interesting ones on OWID. 

I split the review into code and charts:
### Code review
I left some small comments regarding metadata, but overall this looks really good. I only did an high-level read through, but the code makes sense to me and runs without problems and the resulting data looks good! 👍 

### Charts (& data page)
For this I used the chart diff tool and wrote myself a little check list to go through the review. I went through all charts on chart-diff for any obvious errors or improvements, but only went through the full checklist for the top 10 charts by relevance (as shown in chart-diff).

#### Checklist:

##### Chart-diff:
- [ ]  Are there any changes that look wrong?

##### Chart:
- [ ]  FASTT: Is the FASTT clear - do I immediately understand the metric and trends?
- [ ]  FASTT: Should anything be further explained in DoDs or footnotes that isn’t yet?
- [ ]  Chart tab: Does the chart type and selected entities make sense?
- [ ]  Chart tab: Are entity selection, the change toggle and log/ linear toggles enabled? Should they be?
- [ ]  Chart & Map tab: Does the data look reasonable? Are there any obvious outliers or mistakes?
- [ ]  Map tab: Do the map brackets make sense? Does the color scheme make sense and look good?
- [ ]  Table tab: Do the labels for values match up with their titles?
- [ ]  Are there any other improvements I would suggest?

##### Data page:
- [ ]  Is the data page title clear?
- [ ]  Is “What you should know about this data” filled? Should there be any more information or clarifications or DoDs?
- [ ]  Is “How we processed this data” clear? Does it match with the code?
- [ ]  Is “How is this data described by the producer” filled and nicely formatted? Do we link back to the source of the description?
- [ ]  Are there any other improvements I would suggest?


I'll put the results of the in-depth chart review into another comment. But overall, the charts and code look really good!

antea04:
## Chart review

Like mentioned in the previous comment, the charts look really good. I think when moving towards "Perfect data pages", we can add some more information and context for each grapher chart, so I wrote down all the suggestions I could think of for the "top 10"-charts. Feel free to disregard suggestions that are unnecessery or out of the scope of this update!

### General points I noticed:

- The current processing note should probably be made either more specific for each indicator in the data set or more general (not mention different energy sources or energy consumption)
  - Currently I find it confusing, as it shows up on indicators that do not look at different energy sources or energy consumption. I would try to make this as specific as possible for each indicator, but if it is not possible, just make it general and supplement via the “What you should know about this indicator” field.
- Having three different versions of the regional aggregates for e.g. Europe (EI, Ember and OWID) is pretty confusing, especially as they sometimes do not add up. 
  - I would vote to only keep one, e.g. OWID, but I am sure you wrestled with this already. If the value they add is larger than the confusion, we should keep them, but maybe we should add a note to “What you should know about this data”.
- I wonder if for the FASTT source just “Ember (2025), Energy Institute (2024)” would be enough. Usually we don’t include the specific publication in the FASTT, do we?
- Generally we could add some more information in the fields “What you should know about this indicator and more specific producer metadata/ processing descriptions for each indicator - but that’s a big effort and should maybe be it’s own project? 😅 I have made some suggestions here, but they apply to most charts in this update. 

### [Per capita electricity generation](http://staging-site-data-electricity-update/grapher/per-capita-electricity-generation)
- Data page: I would (in "What you should know about this indicator” or WYSKATI :D) clearly explain that we take total energy consumption and divide it by population - linking out to our population data set (in addition to mentioning it as a source)
- Data page: If the sources change at some point I would also mention that in WYSKATI, including any caveats e.g. “We combine sources from Ember and the Energy Institute to provide this data. Since their reporting differs in some aspects (x, y & z) this can lead to data points before 1990 to be (lower/ higher/ different/ not take x,y,z into account”
- Data page: I think the processing notes could be a bit clearer, e.g.:
        - We rely on Ember as the primary source of electricity data. Where data for a given country and year is not available from Ember, we use the data from the Energy Institute. This is the case for all data points before 1990 and data for (x, y, z countries)
        
### [Share of electricity production from renewables](http://staging-site-data-electricity-update/grapher/share-electricity-renewables)
- On regional aggregates: Some regions (e.g. Europe) don’t have OWID aggregates for 2024, even though most regions seem to have data
- Data page: Same points as above for WYSKATI
        
### [Electricity production by source](http://staging-site-data-electricity-update/grapher/electricity-prod-source-stacked)    
- Now this is an old data page 😀 I don’t know how old/ new data pages are decided and data is moved, but as this is a popular chart, maybe it could get a new data page.
- Chart: Nitpick but maybe we could bring the footnote up in the subtitle, I think it’s pretty useful information and we have space.
        
### [Share of electricity production by source](http://staging-site-data-electricity-update/grapher/share-elec-by-source)
        
- Same as for “Energy production by source”
        
### [Share of electricity production from coal](http://staging-site-data-electricity-update/grapher/share-electricity-coal)
        
- FASTT: In the subtitle “Measured as a percentage of total electricity production [in a given country]”
- Data page: It would be cool to have a WYSKATI with 1-2 bullets on coal energy production:
  - How does it work, what different technologies are there and what are the caveats.
  - Ember has some information on this in their methodology document.
        
### [Electricity generation](http://staging-site-data-electricity-update/grapher/electricity-generation)
- FASTT: Subtitle: “Total electricity generation in each region, measured in terawatt-hours.”
- DoD: I don’t really have a good feeling for how much one watt-hour is, maybe we could add an example?

### [Share of electricity production from nuclear](http://staging-site-data-electricity-update/grapher/share-electricity-nuclear)
- Same points as for “Share of electricity production from coal”
- Chart: Capitalize text in the map buckets: “no nuclear” → “No nuclear”
- I really like the “No nuclear” bucket in the map! Do we want to add this to e.g. the [coal chart](http://staging-site-data-electricity-update/grapher/share-electricity-coal) as well?

### [Share of electricity generated by low-carbon sources](http://staging-site-data-electricity-update/grapher/share-electricity-low-carbon)
- Another really cool chart!
- Data page: We could add a note in WYSKATI why we care about low-carbon sources
    - e.g. mention that electricity production generates a big share of carbon emissions and link out to our co2 emissions topic page?
    

### [Share of electricity production from fossil fuels](http://staging-site-data-electricity-update/grapher/share-electricity-fossil-fuels)
- Same points as for “Share of electricity production from coal”
- FASTT: This should include a list of fossil fuel energy production methods in the subtitle (e.g. Fossil fuels include coal, oil and natural gas).
- Data page: Just like in “low-carbon sources” a bullet about why this matters would be cool. Could even be the same bullet!
- I also really like the color scheme of this chart :D

### [Share of electricity production from wind](http://staging-site-data-electricity-update/grapher/share-electricity-wind)
- Same points as for “Share of electricity production from coal”
- Map: I understand why, but to me the map brackets are pretty confusing. I would expect linear ones because of the percentages and also think that would make the distribution much clearer. But ultimately up to you!
    - Maybe something like a 0% bucket and then 5%, 10%, 15%, 20%, 25%, 30%+?


Once again, take the suggest you deem useful and disregard the rest! Thanks for the update!

etl/steps/data/garden/ember/2025-05-12/yearly_electricity.py
pabloarosado: That's not a bad idea, but I think it's safer to simply drop the last point of region aggregates. Even if 90% of countries are informed, there will always be a spurious systematic decline in the latest year.

PR #4390, opened 2025-05-13
📊 corruption: Corruption Perception Index

veronikasamborska1994:


paarriagadap (approved):
Looks good! I leave some (mostly minor) comments below.

Regarding [this chart](http://staging-site-corruption-index/admin/charts/343/edit), I think viewers generally don't quite grasp when the color palette is inverted. For instance, they don't recommend it [here](https://www.datawrapper.de/blog/colors). How about picking a non-inversed blue/green palette? Or, if it's standardized, maybe a divergent palette would be even better

PR #4407, opened 2025-05-19
🚧 corruption: GCB mdim

veronikasamborska1994:
hey @lucasrodes! do you know if there is a way to override the description short for a grouped view? It looks like it just automatically selects the first one (see below). 
Screenshot 2025-05-19 at 17 43 16


lucasrodes:
hi @veronikasamborska1994! Nice to see this use-case. `Collection` has a native method `grouped_views`, which is the one that should generally be used to add new views to the Collection.

I've tweaked it to accept parameters to create titles and subtitles dynamically based on the view dimensions. In addition, I've added support to also edit the `metadata` of the new views, which is the field you want to edit to tweak fields like `description_short`, `description_key`, etc.

PR #4415, opened 2025-05-19
📊 Update GHO

Marigold:


spoonerf (commented):
Hey @Marigold 

Thanks for working on this. I noticed there is an issue with how we deal with regions from other data providers. 

Previously, this was done based on the value given in the 'region_source' column but this is no longer available. 

I think it's important to find a way to flag which organisation each region is constructed by, as the GHO has data from multiple sources, not just WHO. 

This was previously done with the drop_excess_region_sources() and the old `add_region_source_suffix()` functions. We also wanted to limit the number of different organisational regions used in a chart to two, which is what `drop_excess_region_sources()` did, that would be good to bring back if possible. 

It could be worth spending some time bulking out the `countries.json`, with the hope that different organisations have different names. 

Additionally, I think the `warn_on_missing_countries` flag in the country harmonization should always be set to True. For example, Turkey changed the spelling of its name, which we were missing that, as it was being skipped, so Turkey is currently missing from a bunch of the charts. 

I started making some headway on the `countries.json` [here](https://github.com/owid/etl/pull/4487), but there's still a bunch of unresolved ones, sorry I ran out of time! 

I've not reviewed the rest yet but can do later on.

Marigold:
Good points about the regions. They way they harmonize & present them changed in their new GHO API. There's a full list of regions they're using https://apps.who.int/gho/data/node.searo-metadata.REGION?lang=en.

It's a bit hard to make sense of it. Looks like the default ones like `AFR: Africa` are by WHO, but then there's `WHO_AFRO: Africa Region` which is also by WHO... should I just use `Africa (WHO)` for both? Also, UN regions that used to be there are no longer there. There's only one group of OECD defined regions.

(If you don't have any ideas, I'll just add them as suffixes, and we'll see how the end result looks)

Marigold:
Oh, my bad! I noticed they **do have** different sources in `SpatialDimType` column. I'll try to fix that.

Marigold:
> Awesome, let me know if I can help out at all.

@spoonerf I'd love your opinion if you're up for it :)

After adding all the "region source" suffixes, I ended up with a huge [`countries.json`](https://github.com/owid/etl/pull/4415/files#diff-2bbf2b6e2a635a554fcda88581c0199bfdb2039865787d23d1793db3eef9c4e0) file. I'm wondering whether I should keep everything in the dataset or exclude some providers (like `MGH` or `GBD`). It doesn't affect our charts, so I'm unsure how much energy it's worth.

Once we resolve this, it'll be ready for a proper review!

spoonerf:
Hey @Marigold! 

I moved a bunch of regions into the excluded file in this #4502, hope that's helpful! Broadly we can just ignore any MGH and GBD and the more niche regions like "Eastern Mediterranean region, stratum D (EMR D)" and "Western Asia (exc. Armenia, Azerbaijan, Cyprus, Israel and Georgia) (UN)" etc.

I also think it would be good to bring back `drop_excess_region_sources()`, if a chart has more than a couple of different region types, it just starts to look a bit messy, so it would be good to limit each table/variable to two.

Marigold:
Thanks! 🙏 I've merged your PR and added back the `drop_excess_region_sources` function. I hope it's good for review now.

etl/steps/data/garden/who/2025-05-19/gho.py
```diff
@@ -0,0 +1,425 @@
...
+        match region_source:
+            case "COUNTRY" | "GLOBAL":
+                continue  # No suffix for countries and global
+            case "FAOREGION":
+                suffix = "FAO"
+            case "REGION" | "WHOINCOMEREGION":
+                suffix = "WHO"
+            case "UNREGION":
+                suffix = "UN"
+            case "WORLDBANKINCOMEGROUP" | "WORLDBANKREGION":
```
spoonerf: I think drop "WORLDBANKINCOMEGROUP" as we are trying to consistently call them 'High-income countries", "Low-income countries", so that they work with the fancy new grapher map.

snapshots/who/2025-05-19/gho.py
```diff
@@ -0,0 +1,464 @@
...
+- [Metadata HTML](https://www.who.int/data/gho/indicator-metadata-registry/imr-details/1)
+
+Unfortunately the Athena API does not provide a way to download metadata, so we have to scrape it from
+HTML page with metadata.
+
+Another issue is that about 800 indicators in the list don't have `url` field with indicator ID, even though
+their metadata page at https://www.who.int/data/gho/indicator-metadata-registry/imr-details/1 exists.
+Because of that, we can't download the metadata. The workaround would be to match their names on
+https://www.who.int/data/gho/data/indicators/indicators-index and grab metadata from the links on that page.
+An example of such indicator without `url` field is HEMOGLOBINLEVEL_REPRODUCTIVEAGE_MEAN.
```
spoonerf, line 24: Oh boy, GHO is not fun

spoonerf (approved):
Looks great! The snapshot.py should be hung in the Louvre 🎨!! 

There are some big updates in here on some heavily visited charts, that haven't had an update in a while, e.g. there's six more years of data on many of the obesity charts and there are some pretty big covid spikes in the [maternal mortality ratio chart](http://staging-site-update-gho/grapher/maternal-mortality-ratio-who-gho?tab=chart&country=~USA&mapSelect=~USA). So I'd definitely flag that there are large changes when you announce the update.

etl/steps/data/garden/who/2025-05-19/gho.py
Marigold: Didn't know that, thanks!

PR #4428, opened 2025-05-21
Fix dependency dirty flag logic

lucasrodes:
## Summary
- mark dependencies clean on grapher steps
- add regression tests for `_set_dependencies_to_nondirty`

## Testing
- `make test` *(fails: uv could not install packages and pyright was interrupted)*


Marigold (approved):

Marigold:
@lucasrodes did you test that it works? Was it entirely created by codex?

PR #4431, opened 2025-05-21
📊 war: Update remaining nuclear weapons data

pabloarosado:


pabloarosado:
Hey @lucasrodes, not urgent, but it seems that `etl pr` created a new PR with a branch name that is identical to a branch that was just merged. Hence, the corresponding staging server it is the old one (with the old chart diff and so on).
This is an unlikely scenario, but if this happens again we should probably consider improving that tool to avoid these clashes. For now, I just wanted to raise this possibility.

PR #4450, opened 2025-05-23
✨ Sync DoDs in chart-sync

Marigold:
Synced DODs from staging servers in `chart-sync`. New and updated DODs are now automatically synced, with Slack notifications sent if there are conflicts (i.e., the production DOD was updated after the staging server was created).

Note: There’s no approval process in place, so experimenting with DODs on a staging server is discouraged.

## TODO
- [x] Announce that `chart-sync` now supports DODs
- [ ] Test it against production


apps/chart_sync/cli.py
```diff
@@ -232,10 +233,15 @@ def cli(
...
+            dods_synced = _sync_dods(source_session, target_session, target_api, dry_run, SERVER_CREATION_TIME)
+
     if charts_synced > 0:
         print(f"\n[bold green]Charts synced: {charts_synced}[/bold green]")
-    else:
-        print("\n[bold green]No charts synced[/bold green]")
+    if dods_synced > 0:
+        print(f"[bold green]DODs synced: {dods_synced}[/bold green]")
+    if charts_synced == 0 and dods_synced == 0:
+        print("\n[bold green]No charts or DODs synced[/bold green]")
```
pabloarosado: Minor suggestion: maybe it's better to clarify "No charts or DoDs need to be synced" (otherwise it may be interpreted as "No charts or DoDs were synced because something failed).

apps/chart_sync/cli.py
```diff
@@ -274,6 +280,23 @@ def _notify_slack_chart_create(source_chart_id: int, source: str, dry_run: bool)
         send_slack_message(channel="#data-architecture-github", message=message)
 
 
+def _notify_slack_dod_conflict(
+    dod_name: str, source: str, source_updated_at: Any, target_updated_at: Any, server_creation_time: Any, dry_run: bool
+) -> None:
+    message = f"""
+:warning: *ETL chart-sync: DOD Conflict Not Synced* from `{source}`
+DOD "{dod_name}" has been updated in production after staging server was created.
```
pabloarosado: Nit: for consistency with grapher (and our usual notation) we should write "DoD" everywhere, instead of "DOD".

apps/chart_sync/cli.py
```diff
@@ -311,5 +334,67 @@ def _chart_config_diff(
...
+    dods_synced = 0
+
+    # Get DODs from source with updatedAt timestamp higher than staging server creation time
+    source_dods = source_session.query(gm.Dod).filter(gm.Dod.updatedAt > server_creation_time).all()
+
+    log.info("dod_sync.start", n=len(source_dods), dod_ids=[dod.id for dod in source_dods])
+
+    for source_dod in source_dods:
+        # Check if DOD exists in target
+        target_dod = target_session.query(gm.Dod).filter(gm.Dod.name == source_dod.name).first()
```
pabloarosado: Not sure if it's relevant or if it would ever happen, but maybe we should consider what would happen if someone edited the name of an existing DoD.

etl/grapher/model.py
```diff
@@ -292,6 +292,22 @@ def load_tags_by_names(cls, session: Session, tag_names: List[str]) -> List["Tag
         return tags
 
 
+class Dod(Base):
```
pabloarosado: I guess you could call this DoD, to avoid confusion.

apps/chart_sync/cli.py
Marigold: Renaming a DoD is not possible as far as I know.

apps/chart_sync/cli.py
Marigold: Fixed!

PR #4463, opened 2025-05-26
🎉 VSCode extension to make DAG entries clickable

pabloarosado:
## Summary

This extension will let you directly cmd+click on an entry in the DAG, which will take you to the corresponding script of the data step, or the dvc file of the snapshot.

It will also show an emoji next to the step, namely:
🟢 if a file exists for that step, and the step's version is the latest in the DAG.
🟡 if a file exists for that step, but there is a newer version of that step (in which case, the latest version is shown in the tooltip when hovering over the step).
🔴 if no file is found for that step.
⚠️ if the step is repeated in the DAG.

There were a few rabbit holes, but I think I covered most of them, e.g. private steps, steps without a `.py` file (but a `.ipynb` or a folder). There may be other edge cases I haven't considered.

## Review

@lucasrodes @Marigold there's no need to go through the code, the most useful review would be to simply install it (`make install-vscode-extensions`) and try it out. If you have any additional ideas, or if you find any unexpected behavior or edge cases, please let me know, thanks!


lucasrodes:
@pablorosado, I'd say let's merge and test it directly on some data work!

PR #4478, opened 2025-05-28
📊 UNODC annual update 

spoonerf:
To do:
- [x] Improve variable titles
- [x] Import regional estimates (a separate sheet in the snapshot)


veronikasamborska1994 (approved):
code looks good @spoonerf! I left a couple of super minor comments. 

When it comes to charts, I was wondering if the`description_key` bullet points could be made more concise — they’re currently quite lengthy, and the key points can get lost before the reader reaches them. 

for example, this point doesn't seem very important
`The criminal justice data was collected from national authorities with the annual United Nations Survey of Crime Trends and Operations of Criminal Justice Systems (UN-CTS). National focal points working in national agencies responsible for statistics on crime and the criminal justice system, and nominated by the Permanent Mission to UNODC, are responsible for compiling the data from the other relevant agencies before transmitting the UN-CTS to UNODC.`

or this one `Following the submission, UNODC checks for consistency and coherence with other data sources.`

I'd probably also restructure and bring the definition of homicide closer to the top. You could start with something like this and add any other information you think is particularly important: 

```
- UNODC collects intentional homicide data from criminal justice systems (law enforcement records) and public health systems (death certificates)
- Intentional homicide requires three elements: one person killing another, intent to kill or seriously injure, and the act was illegal (maybe give examples what the "illegal element is here - eg. self-defence is not included). 
- Terrorist killings are included as intentional homicides (seems important)
```

one last thing I was wondering if you tried to calculate the death rates yourself and if they match those provided by the UN? this would mean you'd have 2024 data for some countries since you have the data on homicide numbers up to 2024 but not "rates" in this dataset. just a thought - definitely feel free to ignore this!

spoonerf:
Thanks @veronikasamborska1994, super helpful ideas, particularly with the description_key. For the rates, unfortunately, the UN WPP data only goes up to 2023, so we can't calculate the 2024 rates :(

PR #4479, opened 2025-05-28
🐛 fix slugs in collections

lucasrodes:
- Map empty slugs `""` to `"na"` when exporting.
- We were snake-casing dimension and choice slugs, but forgetting about `choice[].presentation.choice_slug_true`, if existant.

### Checks
- [x] Run `etlr explorers/ --export --private --force --only`
- [x] Run `etlr multidim/ --export --private --force --only`

/schedule


lucasrodes:
@Marigold Just an FYI to keep you in the loop

Marigold (approved):
Works well, thanks for fixing it!

PR #4480, opened 2025-05-28
cleanup: remove `colorScaleEqualSizeBins`

marcelgerber:
We're getting rid of this setting on the grapher side of things, and it was _always_ set to `true` for a long time now anyhow.
There's not a single instance where it wasn't set to true, so...


Marigold (approved):
Looks good, let's get rid of it!

PR #4481, opened 2025-05-28
📊 vdem further work

lucasrodes:
More work in the context of https://github.com/owid/owid-issues/issues/1988

**Counts**
([↗ chart](http://staging-site-data-vdem-work-425912/admin/charts/8672/edit))
- [x] Number of countries that had a female leader
- [x] Number of countries that had a female - democratically elected - leader

**Share**
([↗ chart](http://staging-site-data-vdem-work-425912/admin/charts/8674/edit))
- [x] Share of countries that had a female leader
- [x] Share of countries that had a female - democratically elected - leader


lucasrodes:
To be discussed with Bastian: http://staging-site-data-vdem-work-425912/admin/charts/8673/edit

lucasrodes:
I have the indicators in place to recreate this static chart.


You can see the interactive [chart here](http://staging-site-data-vdem-work-425912/admin/charts/8674/edit).
Countries with Women Leaders


I sanity checked the data with Bastian's, and data seems to check out. Minor discrepancies in recent years, likely due to V-Dem release update:

- **Slovenia**: I count Slovenia as having a female HOE (and you don’t) since 2013. In v15, we have that Slovenia HOE=HOG, and “Alenka Bratušek” was HOG in 2013. This applied to woman ever elected and woman ever democratically elected
- **Tunisia**: You count Tunisia as having a female HOE, but I don’t. I think that’s due to a change in v15, which sets HOE=HOS (instead of HOE=HOG, which I suspect was the case before). This applied to woman ever elected and woman ever democratically elected
- **Thailand**: While you counted Thailand as having ever democratically elected a woman since 2012, I don’t. This seems to check out, since Thailand hasn’t been considered an electoral/liberal democracy by V-Dem since 2005 (ref).

PR #4490, opened 2025-05-29
🐛 fix ModuleNotFoundError in pandas submodule

lucasrodes:


lucasrodes:
seems to have been fixed by re-running.

Hypothesis: commit to master might have happened while I was re-installing the `.venv` in the prod server.

PR #4494, opened 2025-05-30
✨ region names in global findex database

antea04:
PR to combine identical regions in the global findex database. 

### context

Simon was wondering about the "Sub-Saharan Africa (WB) (excluding high income)" region in the global findex database (as SSA does not have high-income regions). 

I agree that this can be a bit confusing, but renaming the region is not straightforward, as both "Sub-Saharan Africa (WB)" and "Sub-Saharan Africa (WB) (excluding high income)" exist in the data set. This logic checks whether two regions have the same data and combines them and will complain if this changes in the future. 

Feel free to reject the PR if this is not a useful addition!

### DIFF
--- a/etl/steps/data/garden/wb/2025-05-21/findex.py
+++ b/etl/steps/data/garden/wb/2025-05-21/findex.py
@@ -1,7 +1,5 @@
 """Load a meadow dataset and create a garden dataset."""
 
-import numpy as np
-import pandas as pd
 from owid.catalog import Table
 from owid.catalog import processing as pr
 
@@ -32,15 +30,11 @@ def run() -> None:
     # Convert to %
     tb["value"] = tb["value"] * 100
 
+    # combine the two series for Sub-Saharan Africa (WB) and Sub-Saharan Africa (excluding high income) (WB)
+    tb = merge_ssa(tb)
+
     # Add metadata to the table.
     tb = add_metadata(tb)
-
-    tb = combine_countries(
-        tb,
-        "Sub-Saharan Africa (WB)",
-        "Sub-Saharan Africa (excluding high income) (WB)",
-        new_country_name="Sub-Saharan Africa (WB)",
-    )
 
     # Improve table format.
     tb = tb.format(["country", "year"])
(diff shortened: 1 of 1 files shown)


veronikasamborska1994 (approved):
hi @antea04! thanks for implementing this and the logic totally makes sense. My preference would be to just create one function that focuses on the two regions specifically (as it's unlikely we'll be reusing it for any other cases). It just makes it a bit easier to understand why we've done this in the future. My suggested way would be to just add this kind of function and run it before the metadata addition stage (as that pivots the dataframe in order to add the descriptions) but feel free to do it your way if you disagree

```

def merge_ssa(tb: Table) -> Table:
    """
    Merge the two Sub-Saharan Africa series in the given table if they are consistent.

    This function checks that the series for:
      - "Sub-Saharan Africa (WB)"
      - "Sub-Saharan Africa (excluding high income) (WB)"
    have identical values for each (year, indicator) pair, allowing both to be NaN.

    If all overlapping non-null values agree, the function:
      1. Combines the two series by taking the first non-null value for each (year, indicator).
      2. Drops the original separate rows.
      3. Assigns the combined rows to "Sub-Saharan Africa (WB)".
      4. Returns the updated table with a new combined series.

    Raises:
        ValueError: If any (year, indicator) pair has differing non-null values between the two series.

    Args:
        tb (Table): A table containing columns ['country', 'value', 'year', 'indicator_name'].

    Returns:
        Table: A new table with the two SSA series merged.
    """
    # Define the two country labels 
    full = "Sub-Saharan Africa (WB)"
    excl = "Sub-Saharan Africa (excluding high income) (WB)"

    # Extract the 'value' series for each label, indexed by (year, indicator_name)
    s_full = (
        tb.loc[tb['country'] == full, ['value', 'year', 'indicator_name']]
          .set_index(['year', 'indicator_name'])['value']
          .sort_index()
    )
    s_excl = (
        tb.loc[tb['country'] == excl, ['value', 'year', 'indicator_name']]
          .set_index(['year', 'indicator_name'])['value']
          .sort_index()
    )

    # Identify mismatches: both non-null and unequal
    mismatch = s_full.notna() & s_excl.notna() & (s_full != s_excl)
    if mismatch.any():
        # List all (year, indicator_name) pairs where values differ
        bad_pairs = mismatch.loc[mismatch].index.tolist()
        raise ValueError(
            f"Mismatched values for '{full}' vs '{excl}' at (year, indicator) pairs: {bad_pairs}"
        )

    # Combine the series by taking the first non-null value
    combined_series = s_full.combine_first(s_excl)

    # Assign the country label
    combined_df = combined_series.reset_index()
    combined_df['country'] = full

    # Remove the original two series rows and append the new combined rows
    tb = tb[~tb['country'].isin([full, excl])]
    tb = pr.concat([tb, combined_df], ignore_index=True)

    return tb 

```

PR #4496, opened 2025-05-30
📊 Update World Bank PIP to 2021 prices (code changes)

paarriagadap:
Update PIP to the new version (5 June), which will use 2021 prices.

Data issue https://github.com/owid/owid-issues/issues/1915
General issue https://github.com/owid/owid-issues/issues/1902

This PR is where I am making the changes in the scripts created in https://github.com/owid/etl/pull/4495. In that PR, they were only created by Step upgrader, so reviewing this PR is simpler


snapshots/wb/2025-06-05/pip_api.py
```diff
@@ -431,6 +436,13 @@ def pip_query_country(
     version = versions[ppp_version]["version"]
     release_version = versions[ppp_version]["release_version"]
 
+    # NOTE: There is a bug for China: when querying specific poverty lines (as for relative poverty), the API doesn't respond
+    # One hacky way to fix it is to call a more generalized query first (with no welfare_type, reporting_level nor release version)
+    if country_code == "CHN":
```
paarriagadap, line 438: @Marigold I think I've found a better solution for this issue of the API not working for some (correct) queries. Weirdly, it works after querying data for the country in a more generalized query, so I am doing that now for China (which seems to be the only offender)

paarriagadap: Weird solutions for weird problems...

etl/steps/data/garden/wb/2025-06-05/world_bank_pip.meta.yml
```diff
@@ -299,11 +310,11 @@ definitions:
     <%- endif %>
```
paarriagadap, line 335: Hi @Marigold! Is there any way to make these lower() functions work? I tried inside and outside the brackets, and also using `|lower`. I have several cases of this in the yaml

etl/steps/data/garden/wb/2025-06-05/world_bank_pip.meta.yml
```diff
@@ -299,11 +310,11 @@ definitions:
...
 
   poverty_line_subtitle: |-
-    <% if poverty_line in ["190", "215"] %>
+    <% if poverty_line in ["215", "300"] %>
```
paarriagadap: @Marigold Another question, is there any way to call these lists from lines 3, 4 and 5? It would be much better than repeating several times. I couldn't make it work

snapshots/wb/2025-06-05/pip_api.py
Marigold: Nice, that's better than having hardcoded files.

etl/steps/data/garden/wb/2025-06-05/world_bank_pip.meta.yml
Marigold: For such a complex use case it's better to use [macros](https://docs.owid.io/projects/etl/architecture/metadata/structuring-yaml/?h=jinja#jinja-macros). At the very top of the YAML you set
```
macros: |-
  <% macro percentiles_thr(decile) -%>
    <% if decile == 1 -%>
    Poorest decile
    <%- elif decile == 2 -%>
    2nd decile
    <%- elif decile == 3 -%>
    3rd decile
    <%- elif decile in [4, 5, 6, 7, 8] -%>
    <<decile>>th decile
    <%- elif decile == 9 -%>
    Richest decile
    <%- endif -%>
  <% endmacro %>
```

and then you use it like this

```
threshold_title: |-
    {macros}
    <% if decile == 5 %>
    Threshold <<welfare_type>> per day marking the <<percentiles_thr(decile) | lower>> (median)
    <%- else %>
    Threshold <<welfare_type>> per day marking the <<percentiles_thr(decile) | lower>>
    <%- endif %>
```

(Here's code that I used for testing in the playground.ipynb notebook, just in case)
```
import etl.grapher.helpers as gh
dim_dict = {
    "decile": 5,
    'welfare_type': 'foo',
    'poverty_line': 215,
    'table': 'Income or consumption',
    'survey_comparability': 'No spells',
    'ppp_version': 2017,
    'percentile': 40,
    'reporting_level': 'National',
}
d = gh.render_yaml_file("world_bank_pip.meta.yml", dim_dict=dim_dict)
d["tables"]["world_bank_pip"]["variables"]["thr"]['presentation']
```

etl/steps/data/garden/wb/2025-06-05/world_bank_pip.meta.yml
Marigold: This is better solved with Jinja rather than dynamic-yaml (i.e. {...}). You can define it like
```
macros: |-
  <% macro percentiles_thr(decile) -%>
    <% if decile == 1 -%>
    Poorest decile
    <%- elif decile == 2 -%>
    2nd decile
    <%- elif decile == 3 -%>
    3rd decile
    <%- elif decile in [4, 5, 6, 7, 8] -%>
    <<decile>>th decile
    <%- elif decile == 9 -%>
    Richest decile
    <%- endif -%>
  <% endmacro %>

  <# Poverty lines levels #>
  <% set poverty_line_ipl = ["215", "300"] %>
  <% set poverty_line_lmic = ["365", "420"] %>
  <% set poverty_line_umic = ["685", "830"] %>
```

and then use

```
poverty_line_subtitle: |-
    {macros}
    <% if poverty_line in poverty_line_ipl %>
```

snapshots/wb/2025-06-05/pip_api.py
paarriagadap: Aghh, I thought that I solved it for all cases, and now I found that it fails for queries like `https://api.worldbank.org/pip/v1/pip-grp?group_by=wb&povline=1.35&country=AFW&year=1981&welfare_type=all&reporting_level=all&ppp_version=2021&version=20250401_2021_01_02_PROD&release_version=20250401&format=csv`

My solution doesn't work for this one (a simpler query also breaks, for example `https://api.worldbank.org/pip/v1/pip-grp?group_by=wb&povline=1.35&country=AFW&year=1981&ppp_version=2021&format=csv`)

etl/steps/data/garden/wb/2025-06-05/world_bank_pip.meta.yml
paarriagadap: Thanks for the help, now I am including it all in this yaml

paarriagadap:
@pabloarosado Hola!

This is the World Bank PIP in 2021 prices I was talking to you about. You can check the code, but it is mostly based on the previous version of the code (as you can see compared to the steps generated with step upgrader), so I guess it's more important to take a look at the charts, to see that everything is working properly.

You can take a look at:
- [chart-diff](http://staging-site-data-world-bank-pip-f7d786/etl/wizard/chart-diff)
- [the MDim](http://staging-site-data-world-bank-pip-f7d786/admin/grapher/wb%2Flatest%2Fworld_bank_pip%23poverty)
- the [poverty explorer](http://staging-site-data-world-bank-pip-f7d786/admin/explorers/preview/poverty-explorer)
- the [inequality explorer](http://staging-site-data-world-bank-pip-f7d786/admin/explorers/preview/inequality)
- the [PPP comparison explorer](http://staging-site-data-world-bank-pip-f7d786/admin/explorers/preview/poverty-explorer-ppp-comparison)

You can also use explorer-diff if you want, whatever works best for you. Thanks!

I see the staging server is failing, but it seems unrelated to this PR.

etl/steps/data/garden/wb/2025-06-05/world_bank_pip_legacy.py
```diff
@@ -40,9 +40,11 @@
...
 
 # Define current International Poverty Line (in cents)
-# NOTE: Modify if poverty lines are updated from source
-# TODO: Modify the lines in 2021 prices
-INTERNATIONAL_POVERTY_LINE_CURRENT = 215
+# Define international poverty lines as the second value in each list in POVLINES_DICT
+INTERNATIONAL_POVERTY_LINES = {ppp_year: poverty_lines[1] for ppp_year, poverty_lines in POVLINES_DICT.items()}
+
+# Define current International Poverty Line (IPL) in the latest prices
+INTERNATIONAL_POVERTY_LINE_CURRENT = INTERNATIONAL_POVERTY_LINES[PPP_VERSIONS[1]]
```
lucasrodes: I would use `PPP_YEAR_CURRENT` instead `PPP_VERSIONS[1]`. I find it easier to read and quickly understand.

etl/steps/data/explorers/poverty_inequality/latest/poverty_inequality.py
```diff
@@ -15,7 +15,7 @@
 # Define PPP versions
 # NOTE: Change this in case of new PPP versions in the future
 # TODO: Change to 2021 prices
-PPP_VERSIONS = [2011, 2017]
+PPP_VERSIONS = [2017, 2021]
```
lucasrodes: I don't think there's any value in having `PPP_VERSIONS` defined, isn't it?

I would just define `PPP_YEAR_OLD` and `PPP_YEAR_CURRENT` directly, i.e.

```python
PPP_YEAR_OLD = 2017
PPP_YEAR_CURRENT = 2021
```

etl/steps/data/explorers/wb/latest/world_bank_pip.py
```diff
@@ -14,12 +14,12 @@
 
 # Define PPP versions
 # NOTE: Change this in case of new PPP versions in the future
-PPP_VERSIONS = [2011, 2017]
+PPP_VERSIONS = [2017, 2021]
```
lucasrodes: I don't think there's any value in having `PPP_VERSIONS` defined, isn't it?

I would just define `PPP_YEAR_OLD` and `PPP_YEAR_CURRENT` directly, i.e.

```python
PPP_YEAR_OLD = 2017
PPP_YEAR_CURRENT = 2021
```

pabloarosado:
Hi @paarriagadap great work!
I'm going through chart diff. A few potential issues (feel free to disregard if you understand them and they are not issues):
* [ ] I see that [this chart](http://staging-site-data-world-bank-pip-f7d786/grapher/mean-income-or-expenditure-per-day-2011-vs-2017-international-) still has the old article in "Related".
* [ ] Nitpicking, [this chart](http://staging-site-data-world-bank-pip-f7d786/grapher/distribution-of-population-between-different-poverty-thresholds-stacke-bar) has a slug "stacke-bar". Consider renaming it (I'd rather do it in production, after merging, to avoid syncing issues).
* [x] ~[This chart](http://staging-site-data-world-bank-pip-f7d786/grapher/poverty-vs-mean-schooling) has 3.65$ in the x-axis.~ Hmm, not anymore
* [x] ~[This chart](http://staging-site-data-world-bank-pip-f7d786/grapher/gdp-per-capita-world-bank-data-vs-penn-world-table) is mixing 2017 and 2021 prices. If that's intended, I suppose the footnote should be adapted.~ I see it's rejected.
* [ ] [This chart](http://staging-site-data-world-bank-pip-f7d786/grapher/number-extreme-poverty-version-comparison) shows the old and new poverty lines, but the description key is still saying "Extreme poverty here is defined as living below the International Poverty Line of $3 per day." Not a big issue, but consider if it should be rephrased (if it's complicated, it may not be worth it). Same with other similar comparison charts.
* [ ] (Nit) the slug in [this chart](http://staging-site-data-world-bank-pip-f7d786/admin/charts/4705/edit) mentions an old poverty line.
* [ ] [This chart](http://staging-site-data-world-bank-pip-f7d786/admin/charts/1667/edit) mentions 2017 ppp.

I'll look now at the mdim.

pabloarosado (approved):
Thanks for creating a baseline branch and a sub-branch for changes! 🧐

lucasrodes (approved):
LGTM overall! Left some minor comments in the code.

On the explorers, some thoughts below.

### (1) Poverty explorer:
I wonder if you could collapse all the "income consumption"-related indicators (mean, median, poorest 10%, richest 10%) into one option in the first dropdown and then have the breakdown in another dropdown? E.g., in the second one somehow? For that, you'd probably need to change the title of the second dropdown.

For the last dropdown, you could shorten the option "Show data from both income and consumption surveys" to just "income and consumption surveys."

### (2) inequality explorer:
It looks good to me. Side note: I think an MDIM could significantly improve its accessibility. For instance, I was wondering about the "Palma ratio." In an MDIM, we could have a brief description under the dropdown choice explaining it.

I've also observed that the indicator description section seems quite long and has a "NOTES ON HOW WE PROCESSED THIS INDICATOR" bit. This is a solution while the explorer remains CSV-based instead of indicator-based. In the future, it'd be nice to break down this into the different metadata fields.

image


### (3) PPP explorer:
The same comment applies as with (1), where it might make sense to collapse the various "income consumption"-related indicators. In addition, I see that the terms change a bit here: "P10 (poorest tenth)" instead of "Income or consumption of the poorest 10%." Regardless of whether we collapse these indicators, we should use the same terms throughout the explorers.

etl/steps/data/garden/wb/2025-06-05/world_bank_pip_legacy.py
paarriagadap: I agree. Thanks!

paarriagadap:
Thanks @lucasrodes! Yes, one of the main points to have migrated to the world_bank_pip pipeline is that we will be able to have explorers/Mdims using full metadata from the ETL, so it will show the fields better.

I am not sure if collapsing fields is a good solution and I understand P10 or P90 do not directly relate with poverty, but this was the outcome after a million of conversations. Happy to redesign when it's due.

There is a reason for the difference between "P10 (poorest tenth)" instead of "Income or consumption of the poorest 10%. It is mostly to have more clarity in the legends for comparisons, but I will have it in mind once we sort how to handle all these explorers.

paarriagadap:
Hi @Marigold! Is there any reason for the staging server breaking in the `Grapher build` step?

I know this is an old pull request, but I try to keep it updated (chart-diff, for example) until some articles are done. I merged master and doesn't fix it.

PR #4517, opened 2025-06-05
🎉  🤖 Add etl snapshot or etls command

### LINKED ISSUE #4337: Add `etl snapshot` command
## Motivation

We run ETL steps as `etl run ...` or `etlr ...`, but snapshots are still run as simple python scripts `python snapshots/.../short_name.py`. This leads to unnecessary repetition in our scripts.

## Proposal

Implement `etl snapshot ...` or `etls ...` that would run a snapshot matching an argument (same as ETL). Three things can happen:

1. Scripts contains `main` function and `if __name__ == "__main__": main()` (all existing snapshot scripts) -> run the module
2. Script contains `run` function, which might accept arguments like `upload: bool` or `path_to_file: str` -> wrap it in CLI and pass arguments `upload` and `path_to_file` to the function
3. Script doesn't exist, only `*.dvc` file does -> run `snap.create_snapshot` against the DVC file (and add support `--path-to-file`)

This removes the need for `*.py` files for simple scripts that either use `url_download` or `path_to_file`.

It also opens up possibilities for new features, like keeping old checksums in `history:` field in DVC.

Marigold:
## ✨ Add `etl snapshot` command for standardized snapshot execution

Fixes #4337

### Overview

This PR introduces the new `etl snapshot` command (with shorthand `etls`) to standardize and streamline how snapshot scripts are executed in the ETL pipeline. Previously, each snapshot had to be run manually using `python script.py`, leading to repetition and inconsistency.

#### 🚀 New CLI Command

The `etl snapshot` command handles three different snapshot scenarios automatically:

1. **Scripts with `main()` function** - Runs the module directly (legacy support)
2. **Scripts with `run()` function** - Wraps in CLI with proper argument passing
3. **DVC-only snapshots** - Runs `snap.create_snapshot()` when only `.dvc` metadata exists

#### 📝 Command Examples

```bash
# Full path
etl snapshot tourism/2024-08-17/unwto_gdp
etls tourism/2024-08-17/unwto_gdp

# Partial path (auto-resolves namespace)
etls 2024-08-17/unwto_gdp

# Full file path
etls snapshots/tourism/2024-08-17/unwto_gdp.py

# With local file upload
etls abs/2024-08-06/employee_earnings --path-to-file data.csv

# Skip upload
etls dataset_name --skip-upload

# DVC-only snapshots (no Python script)
etls artificial_intelligence/2024-06-28/ai_adoption
```

#### 🔧 Enhanced Snapshot Helper

New `Snapshot.from_script()` classmethod automatically detects namespace, version, and filename from script location:

```python
# Before
snap = Snapshot("tourism/2024-08-17/unwto_gdp.csv")

# After  
snap = Snapshot.from_script(__file__)  # Auto-detects from script path
snap = Snapshot.from_script(__file__, "unwto_gdp.xlsx")  # With custom filename
```

#### 🧙‍♂️ Wizard Updates

The snapshot wizard now:
- Uses `etls` command instead of `python script.py` in instructions
- Supports **DVC-only mode** with new toggle option
- Generates contextual help based on selected mode
- Works seamlessly with both script and metadata-only workflows

#### Before
```bash
# Manual execution, different patterns
python snapshots/tourism/2024-08-17/unwto_gdp.py
python snapshots/abs/2024-08-06/employee_earnings.py -f data.csv

# Script had to manually construct snapshot path
snap = Snapshot("tourism/2024-08-17/unwto_gdp.csv")
```

#### After  
```bash
# Unified command interface
etls tourism/2024-08-17/unwto_gdp
etls abs/2024-08-06/employee_earnings --path-to-file data.csv

# Auto-detection of paths
snap = Snapshot.from_script(__file__)
```

### Migration

Existing snapshot scripts continue to work unchanged. The new command provides a standardized interface while maintaining full backward compatibility.


etl/snapshot_command.py
```diff
@@ -0,0 +1,300 @@
...
+import structlog
+
+from etl import paths
+from etl.snapshot import Snapshot
+
+log = structlog.get_logger()
+
+
+@click.command("snapshot")
+@click.argument("dataset_name", type=str)
```
lucasrodes: I think it'd be helpful to clarify what `dataset_name` exactly means? 

Is it a substring of the snapshot URI? From the examples below it looks like it should contain at least `namespace/version/short_name`, or is `namespace` optional?

apps/wizard/etl_steps/forms.py
```diff
@@ -439,6 +440,7 @@ def __init__(self: Self, **data: str | int) -> None:  # type: ignore[reportInval
         data["license_url"] = data["origin.license.url"]
         data["origin_version"] = data["origin.version_producer"]
         data["dataset_manual_import"] = data["local_import"]
+        data["dvc_only"] = data.get("dvc_only", False)
```
lucasrodes: i suspect this line is not really needed, since we already do `super().__init__(**data)` and the class attribute is initialized to `False` already. But no harm in keeping it.

apps/wizard/etl_steps/snapshot.py
```diff
@@ -408,9 +417,15 @@ def run_snap_step() -> None:
             SNAPSHOTS_DIR / form.namespace / form.snapshot_version / f"{form.short_name}.{form.file_extension}.dvc"
         )
 
+        # If DVC-only mode is enabled, remove the generated Python script
+        if getattr(form, "dvc_only", False):
```
lucasrodes: wouldn't it work with `if form.dvc_only` also?

apps/wizard/etl_steps/snapshot.py
```diff
@@ -422,19 +437,38 @@ def run_snap_step() -> None:
         with st.expander("⏭️ **Next steps**", expanded=True):
             # 1/ Verification
             st.markdown("#### 1. Verification")
-            st.markdown("Verify that generated files are correct and update them if necessary.")
+            if getattr(form, "dvc_only", False):
```
lucasrodes: why not use `form.dvc_only`?

apps/wizard/etl_steps/snapshot.py
```diff
@@ -422,19 +437,38 @@ def run_snap_step() -> None:
...
+                command = f"etls {snapshot_path}"
+                if manual_import_instructions:
+                    command += " --path-to-file **relative_path_of_file**"
+            else:
+                # DVC-only mode - use etls with just the identifier
+                snapshot_path = f"{form.namespace}/{form.snapshot_version}/{form.short_name}"
+                command = f"etls {snapshot_path}"
+                if manual_import_instructions:
+                    command += " --path-to-file **relative_path_of_file**"
+
```
lucasrodes: maybe worth encapsulating this logic into a function? `def create_command` or something similar

lucasrodes (approved):
This is very cool! Thanks for implementing this, Mojmir.

I've just left couple of comments.

Also, a couple of other points:
- Do you think we could have `--dry-run` option for `etls`?
- This is more of a future thing, but I wonder if the "Create only DVC" toggle in Wizard should default to `True`.

pabloarosado:
Hey @Marigold thanks for doing this, this is going to be really useful! I haven't yet looked through the changes, but wanted to quickly share a few considerations before merging:
1. Shouldn't we use `PathFinder(__file__)` for snapshot scripts instead of `Snapshot.from_script(__file__)`? That would make them more aligned with data steps.
2. I wonder if `etl run ... --snapshot` could be a better alternative, instead of `etl snapshot ...`. With the former, we could have a one-liner command to execute everything after an update, e.g. `etl run [all snapshots and data steps] --snapshot --grapher`, instead of `etl snapshot [snapshots]; etl run [data steps]`.
3. I wonder if there could be edge cases when there are multiple dvc files for the same script (with a different file name). I haven't thought much about it, but I wonder how etl know whether to execute the default process for the dvc file vs the script (I'll have a look at the changes later, maybe it becomes clearer).

Feel free to disregard these ideas if you think they don't apply or they could be applied in the future, thanks!

pabloarosado (approved):
The code looks good and your implementation makes sense. Please consider my suggestions, but feel free to merge if you disagree or think we can do them in the future, thanks for this work!

apps/wizard/etl_steps/snapshot.py
Marigold: yep!

PR #4520, opened 2025-06-05
🎉 Add module to read from Notion

pabloarosado:


Marigold (approved):
I only skimmed through it, but it looks good.

PR #4522, opened 2025-06-06
📊 education: mdim on school expectancy

veronikasamborska1994:


etl/steps/export/multidim/education/latest/years_of_schooling.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # UNDP expected years of schooling columns
+    cols_undp = [
+        "eys__sex_total",
+        "eys__sex_female",
+        "eys__sex_male",
+        "mys__sex_female",
+        "mys__sex_male",
+        "mys__sex_total",
+    ]
```
lucasrodes: could maybe use:


```python
cols_undp = tb_undp.filter(regex=r"(eys|mys)__sex_").columns
assert len(cols_undp) == 6
```

etl/steps/export/multidim/education/latest/years_of_schooling.py
```diff
@@ -0,0 +1,286 @@
...
+        "school_life_expectancy__primary__both_sexes__years",
+        "school_life_expectancy__primary__female__years",
+        "school_life_expectancy__primary__male__years",
+        "school_life_expectancy__secondary__both_sexes__years",
+        "school_life_expectancy__secondary__female__years",
+        "school_life_expectancy__secondary__male__years",
+        "school_life_expectancy__tertiary__both_sexes__years",
+        "school_life_expectancy__tertiary__female__years",
+        "school_life_expectancy__tertiary__male__years",
+    ]
```
lucasrodes, line 53: if there is no other column matching `school_life_expectancy__`, you could use:


```python
cols_opri = tb_opri.filter(regex=r"school_life_expectancy__").columns
assert len(cols_opri) == 12
```

etl/steps/export/multidim/education/latest/years_of_schooling.py
```diff
@@ -0,0 +1,286 @@
...
+    cols_gender_stats = [
+        "hd_hci_lays",
+        "hd_hci_lays_fe",
+        "hd_hci_lays_ma",
+    ]
+
+    # Select only the relevant columns
+    tb_undp = tb_undp[["country", "year"] + cols_undp]
+    tb_opri = tb_opri[["country", "year"] + cols_opri]
+    tb_gender_stats = tb_gender_stats[["country", "year"] + cols_gender_stats]
```
lucasrodes: In general, I'd suggest using the `.loc` operator. It helps type checker identify the types of the objects (e.g. when you hover over `tb_undp` you'll get `Table`. Otherwise you could get `Unknown`, `Any` or `DataFrame`).

etl/steps/export/multidim/education/latest/years_of_schooling.py
```diff
@@ -0,0 +1,286 @@
...
+    for col in tb.columns:
+        if col in ["country", "year"]:
+            continue
+        tb[col].metadata.original_short_name = "expected_years_schooling"
+        tb[col].metadata.dimensions = {}
+
+        # Set dimensions
+        tb[col].metadata.dimensions["metric_type"] = metric_type_mapping[col]
+        tb[col].metadata.dimensions["level"] = level_mapping[col]
+        tb[col].metadata.dimensions["sex"] = sex_mapping[col]
```
lucasrodes: Similar as in the other PR, I think this can be simplified by (I) using only one loop, and (ii) dropping the dictionaries.


```python
# Iterate and populate mappings
for col in cols_to_add_dimensions:
    # --- Level ---
    level = None
    for key, val in level_keywords.items():
        if key in col:
            level = val
            break
    
    # Default to "all" for aggregate measures
    if level is None and any(x in col for x in ["eys", "mys", "hd_hci_lays"]):
        level = "all"

    # --- Metric Type ---
    metric_type = None
    for key, val in metric_keywords.items():
        if key in col:
            metric_type = val
            break

    # --- Sex ---
    sex = None
    for key, val in sex_keywords.items():
        if f"__{key}__" in col or col.endswith(f"_{key}") or f"__{key}" in col:
            sex = val
            break

    # Set indicator name
    tb[col].metadata.original_short_name = "expected_years_schooling"
    # Set dimensions
    tb[col].metadata.dimensions = {
        "metric_type": metric_type,
        "level": level,
        "sex": sex or "both" # fallback for non-disaggregated vars
    }
```

etl/steps/export/multidim/education/latest/years_of_schooling.py
```diff
@@ -0,0 +1,286 @@
...
+                    "name": "Secondary",
+                }
+            elif "tertiary" in indicator.catalogPath:
+                indicator.display = {
+                    "name": "Tertiary",
+                }
+            elif "all" in indicator.catalogPath or "eys" in indicator.catalogPath or "mys" in indicator.catalogPath:
+                indicator.display = {
+                    "name": "All levels",
+                }
```
lucasrodes: Similar to the other PR, you can have a simpler block like:

```python
if "expectancy__primary" in indicator.catalogPath:
	display_name = "Primary"
elif "expectancy__pre_primary" in indicator.catalogPath:
    display_name = "Pre-primary"
elif "secondary" in indicator.catalogPath:
    display_name =  "Secondary"
elif "tertiary" in indicator.catalogPath:
    display_name = "Tertiary"
elif "all" in indicator.catalogPath or "eys" in indicator.catalogPath or "mys" in indicator.catalogPath:
	display_name = "All levels"

indicator.display = {
    "name": display_name,
}
```

lucasrodes (approved):
Nice work!

Left minor suggestions, some similar as in the other PR.

etl/steps/export/multidim/education/latest/years_of_schooling.py
veronikasamborska1994: hmm a bit tricky as there are many more columns that'd much this that we don't need

PR #4525, opened 2025-06-09
📊 Update all GHO dependencies and archive it

Marigold:
Update all GHO dependencies and archive it

## TODO
- [ ] Archive datasets in admin


spoonerf (approved):
Looks good!

dag/health.yml
```diff
@@ -62,7 +62,7 @@ steps:
 
   # Registered vehicles per 1000 people
```
spoonerf: I think this has been archived in the grapher so we can actually remove it here too

dag/health.yml
```diff
@@ -104,7 +104,7 @@ steps:
 
   # Cholera
   data://garden/who/2023-06-01/cholera:
-    - data://garden/who/2024-01-03/gho
+    - data://garden/who/2025-05-19/gho
```
spoonerf, line 100: It would make sense to run indicator upgrader on this dataset right?

dag/health.yml
Marigold: Good catch

dag/health.yml
Marigold: You mean bumping the version of cholera? I don't think that's necessary. The charts barely changed according to [chart-diff](http://staging-site-update-gho-dependencies/etl/wizard/chart-diff?show_all=&chart-diff-sort-by-results=relevance).

spoonerf: Ah no I just meant running indicator-upgrader and chart diff, it looks like from the above summary for chart-diff that it hasn't been run, but all good!

PR #4527, opened 2025-06-09
✨ aggregates only when X% of countries are available

lucasrodes:


pabloarosado:
Hey @lucasrodes thanks for considering this. It's an issue we have been discussing for a long time, and it would be good to finally tackle it properly! In fact, it would have been very useful to prevent issues like [this very recent one](https://owid.slack.com/archives/C08LZUSC737/p1749573637118339).

One concern about your current approach is that population may not be the a good proxy in some cases (although, if we had to pick one indicator, of course, population would be the most meaningful choice). We could instead decide whether the region aggregate for an indicator-year should be constructed based on the values of that indicator itself. So, for example, we could see the median contribution (over all years in the data) of each country in the region as a percentage of the total aggregate of the region. That would be your X% (instead of population).

This could be helpful at least for our most common use case of creating region aggregates (summing over all countries in the region for a given year). I know there are some technical details here, if you want we could discuss this later today in the data science call.

lucasrodes:
Hey Pablo, thanks for weighing in. 

I've only added support in the `geo` module to be able to do things like "estimate the aggregate only if X% countries in this list are covered". I haven't implemented a generic solution for the case of "above X% of the population in that region".

Note that the heuristic on using population here was decided in the context of democracy work, so I will stick to it for now.

We can discuss your strategy any time, and think of a generic solution for the `geo` module.

PR #4545, opened 2025-06-11
✨ Fix display and presentation propagation

pabloarosado:


pabloarosado:
@Marigold what do you think about this logic? And what would be the easiest way to run the entire ETL to check for changes, shall we increase `ETL_EPOCH`? Thanks!

Marigold (approved):

PR #4550, opened 2025-06-12
📊 Add latest income group classification to grapher

pabloarosado:


etl/steps/data/grapher/wb/2024-07-29/income_groups.py
```diff
@@ -1,22 +1,43 @@
...
+
+    # Read table of latest income groups (a static classification).
+    tb_static = ds_garden.read("income_groups_latest")
+
+    #
+    # Process data.
+    #
+    # Combine both tables.
+    tb = tb_dynamic.merge(tb_static, on=["country"], how="outer", suffixes=("", "_latest"))
+
```
lucasrodes: is this merge needed as a way to "hack" the country values for all years?

pabloarosado: I was not thinking of this as a hack. The goal of this merger is to simply add the latest classification to the combined table.
I suppose that, to be able to use the latest classification in grapher, we need a table with country-year-classification_latest. We can either use the existing table, or create a new one. I don't have a strong opinion. Do you have an alternative proposal?

etl/steps/data/grapher/wb/2024-07-29/income_groups.py
```diff
@@ -1,22 +1,43 @@
...
+    # Sanity check.
+    error = "Classification in the latest year of data should coincide with 'classification_latest'."
+    assert (
+        tb[tb["year"] == tb["year"].max()]["classification"]
+        == tb[tb["year"] == tb["year"].max()]["classification_latest"]
+    ).all(), error
```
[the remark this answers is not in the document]
pabloarosado: I think it would be preferrable to let the assertion fail if something changes in the data.

etl/steps/data/grapher/wb/2024-07-29/income_groups.py
lucasrodes: I was just wondering why we need to have all years defined. Doesn't scatterplots backpropagate the last value in time by default already?

pabloarosado: I don't think we need **all** years defined, but we may need at least **one** year (I haven't checked if it works without a "year" column). Do you see any issue with the current approach?

pabloarosado: For more context, the `tb_static` table contains just two columns, namely `country`, and `classification`. We could create a separate dataset for it, but I suppose that may be more inconvenient than having it as part of the dynamic one, and it may not work as "color" if it doesn't have a "year" column (I haven't checked, and I don't have a strong opinion).

lucasrodes: It's not that I see any issue with your approach. I am just confused, and hence asking. I don't see the point of merging both tables into one. Instead,  can't we just have two tables:

- `tb_dynamic`: as it comes from garden
- `tb_static`: Just would need to add a column with `year=2024`.

pabloarosado: OK, I do see an issue with my approach, which is that, if someone looks at the "Table", it may be confusing to see multiple years from 1987 onwards (with the same classification). So I'll create the two separate tables.

pabloarosado: I've just changed the code to have two separate tables, and remapped the indicators in the charts. And I've just noticed that this is not ideal either: if you see the Table tab of, e.g. [this chart](http://staging-site-data-income-group-grapher/admin/charts/3776/edit) you don't see any income classification (since the latest point in the axes is 2021, which is smaller than 2023, the latest point for which the income classification has data).

To be able to see the classification for all years, we'd need to add all possible country-years with the latest classification.

Anyway, I'll keep it the way you suggested, and we can rethink it in the future.

pabloarosado: Actually, a simple solution to be able to visualize income classifications in the Table tab would be to add a big tolerance, e.g. 100 years. Shall we do that?

pabloarosado: It would look like this:
Screenshot 2025-06-12 at 15 15 49


We can make it even clearer by changing the `display.name` of the indicator, to be, e.g. "World Bank's income classification for 2023". I think that could be a good solution.

lucasrodes: Yeah, I was just thinking of adding tolerance. That might be good enough for now.

I reckon it is still confusing, lots of years displayed.

Do you prefer `World Bank's latest income classification` over `World Bank's 2024 income classification`? I have a slight preference for the latter, since it feels more obvious to me. But, then we would have lots of years shown in your example (1990, 2023, 2024)...

pabloarosado: Good idea, I've also decided in favor of `World Bank's 2024 income classification`.

PR #4551, opened 2025-06-12
✨ Add CLI options for ETL process control

Marigold:


pabloarosado (approved):
Great, thanks a lot! It looks good. I'm just wondering if you could set the default values using config, e.g. `main(..., continue_on_failure=config.CONTINUE_ON_FAILURE, ...)` and then remove the `if continue_on_failure:` clauses. But up to you!

Marigold:
Not really, the thing is that we rely on `config.*` elsewhere in the code (otherwise we'd have to pass it through several layers of functions). It first looks at env variables and then overwrites them with flags.

PR #4557, opened 2025-06-13
💄 corruption: unodc metadata improvements

veronikasamborska1994:


etl/steps/data/garden/unodc/2025-05-12/justice.meta.yml
```diff
@@ -2,9 +2,11 @@
...
+        - Countries employ different approaches to data collection, classification, and reporting that affect comparability. Some nations implement the "principal offence rule" where only the most serious crime is counted when multiple offenses occur together. Data completeness varies significantly across countries, and adherence to international reporting standards like the [International Classification of Crime for Statistical Purposes (ICCS)](https://www.unodc.org/unodc/en/data-and-analysis/statistics/iccs.html) remains inconsistent, leading to definitional and methodological differences.
+        - Apparent anomalies in the data, such as prosecution rates exceeding arrest rates in certain countries, typically reflect these systematic data collection challenges rather than actual procedural irregularities.
```
spoonerf, line 9: There are quite a few caveats here. Do you think the data still shows something that can be understood by the public? I wonder which of arrests, prosecutions, or convictions is likely to have the most solid data collection; it could be worth just focusing on the best variable rather than comparing across them.

spoonerf (approved):
Looks good! I wonder about the quality of the data and if it's better to just focus in on the variables that are more reliable, but you will know best. 

The subtitle for the scatterplot chart doesn't seem to match what is shown in the chart.

etl/steps/data/garden/unodc/2025-05-12/justice.meta.yml
veronikasamborska1994: Yes, @spoonerf, agreed! We’re talking directly to the source (thanks again for sending the contact details), and this is likely how things will unfold in the end. Nicely picked up!

PR #4576, opened 2025-06-17
✨ Improve harmonization of non-owid regions

pabloarosado:
This PR helps identify already existing mappings of non-OWID regions in our harmonize tool.
For example, if a hypothetical World Bank dataset contains the region "East Asia Pacific", our old harmonize tool would show as default options
```
"East Asia Pacific (WB)" (which would be a new entry)
"Asia"
"Tanzania",
... 
```
Even though we do have "East Asia and Pacific (WB)" among our defined non-OWID regions. With this PR, the harmonize tool shows the following options:
```
"East Asia Pacific (WB)" (which would be a new entry)
"East Asia and Pacific (WB)" (existing entry)
"Asia"
"Tanzania",
... 
```
It is still not ideal that the new entry is shown among the default options (ideally, it would be shown in a separate category, possibly "Custom entry". But that involves additional changes. So, for now, this is a small improvement to help harmonize non-OWID regions.

### DIFF
--- a/etl/harmonize.py
+++ b/etl/harmonize.py
@@ -223,10 +223,13 @@ class CountryRegionMapper:
         pairs = pairs[:num_suggestions]
 
         if institution is not None:
-            # Prepend option to include this region as a custom entry for the given institution (if not already there).
-            region_institution = f"{region} ({institution})"
-            if region_institution not in [pair[1] for pair in pairs]:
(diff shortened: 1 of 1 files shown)


lucasrodes (approved):
Not much to add, and don't want to block, so let's go and test this! \

Changes lgtm, and future improvements make sense.

PR #4586, opened 2025-06-18
💄 literacy: OMM source test

veronikasamborska1994:
Hey @spoonerf! I just wanted to get your thoughts on the two options I’ve tried for flagging our sources on the literacy data page.

In the current version on grapher, I’ve included all the source information under the `description_key`, which I think works okay. But since Ed wanted us to add the homepage text accompanying the OMM to each data page, that section has gotten quite lengthy.

In the new version, I’ve tried implementing your idea of using the map of sources. I now reference that map within the _description_key_ of the OMM and only include the source variable on the chart itself. This way, the source chart  can still function as data page, and people can more easily access the main points and descriptions related to the sources. 

I honestly don't know what version would work best. I think for this particular OMM just having everything under the `description_key` within the OMM might work ok but it's unlikely to work as well for datasets where the concatenation of the data is a lot more complicated. 

Let me know what you think :)


spoonerf:
Hey @veronikasamborska1994! 

Thanks for trying this out, it's nice to see some more examples. 

I have a slight preference for the mapped sources, mainly because I think it makes the description_key a bit easier to digest and it makes it super clear which country uses each source for each year, which I you can't currently get from the text for the overlapping years, although I'm sure you could add this if the heuristic is straightforward. 

One thing I would maybe change is the labels used in the source map, I'd perhaps reference each UNESCO source as "UNESCO (year)" and then also use this in the description_key with the additional information about the name of the exact source. 

I think let's keep collecting these comparisons and assess them together in a couple of weeks.

veronikasamborska1994:
won't be using this approach

Issue #4201, opened 2025-03-31
Fix circular dependencies between `population` and `faostat_rl`


pabloarosado:
I wanted to archive the old FAOSTAT steps (version 2024-03-14) now that we have a version 2025-03-17. However, there is still one use of `faostat_rl` that can't be updated, and that's our population dataset (step `garden/demography/2024-07-15/population`). This dependency cannot be upgraded, because we enter a circular dependency:
* Our population dataset includes population density, for which it uses land area from FAOSTAT RL dataset.
* However, the FAOSTAT RL dataset has per capita indicators, for which it uses population.

PR #4602, opened 2025-06-19
📊 Fix circular data dependencies between FAOSTAT RL and population datasets

pabloarosado:
## Summary
Fix circular data dependency between FAOSTAT RL (which depends on population to calculate capita indicators) and population (which uses FAOSTAT RL to calculate population density). To achieve that, I created an auxiliary garden step that loads the data from the FAOSTAT RL snapshot directly, and does some minimal data processing. Then, the population step loads this auxiliary dataset instead of the "official" FAOSTAT RL garden one.

## Changes
* Create a new `faostat_rl_auxiliary` step (using data from the old FAOSTAT version).
* Use this step as one of the dependencies of `population`, to break the circular dependencies with `FAOSTAT`.
* Archive unused FAOSTAT steps (of the old version).

## Review

I checked that the resulting land area table (after adding region aggregates) coincides within 0.1% with the old one, so the data should not be affected. But if you think further checks are needed, please let me know.

@lucasrodes I was thinking of using the latest FAOSTAT RL data (instead of the old one), but my goal here was to try to minimize data changes. If you think this is a good opportunity to update that dependency, please let me know. Thanks!

NOTE (unrelated, just to keep in mind as a potential issue): chart diff is showing changes in population density (as expected), but also on some random charts. There are also apparently random changes in data diff.

Fixes https://github.com/owid/etl/issues/4201


lucasrodes (approved):
lgtm in principle! I think it's fine to just use the 2024 version for now, and look into upgrading this in a future pr, might have some other downstream effects.

i looked into chart diff but just saw two charts with data changes, which seem fine

PR #4606, opened 2025-06-20
🚧 Suggestions WB PIP

Marigold:
@paarriagadap this is an example of using macros for PIP. If writing `{macros}` in every block gets too annoying, I can add it everywhere automatically by default (just let me know).


paarriagadap:
Thanks Mojmir! I didn't see this PR in time and proceeded with my changes manually. Adding {macros} is a bit annoying, but for now, I am OK with adding them. I made many more changes, so let's close this one?

PR #4609, opened 2025-06-20
📊 OECD - Analytical house price indicators

antea04:
Data from OECD analytical house price indicators (http://data-explorer.oecd.org/s/295).

### DIFF
--- a/etl/steps/data/garden/oecd/2025-06-20/housing_prices.py
+++ b/etl/steps/data/garden/oecd/2025-06-20/housing_prices.py
@@ -59,7 +59,9 @@ def run() -> None:
     # Process data.
     #
     # Harmonize country names.
-    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path)
+    tb = geo.harmonize_countries(
+        df=tb, countries_file=paths.country_mapping_path, excluded_countries_file=paths.excluded_countries_path
+    )
 
     # Improve table format.
     tb = tb.format(["country", "year"])


snapshots/oecd/2025-06-20/housing_prices.csv.dvc
```diff
@@ -0,0 +1,31 @@
...
+    url_download: https://sdmx.oecd.org/public/rest/data/OECD.ECO.MPD,DSD_AN_HOUSE_PRICES@DF_HOUSE_PRICES,1.0/all?dimensionAtObservation=AllDimensions&format=csvfilewithlabels
+    date_accessed: '2025-06-20'
+
+    # License
+    license:
+      name: OECD Terms and Conditions
+      url: https://www.oecd.org/en/about/terms-conditions.html
+outs:
+  - md5: a5d0bd5d0e8adc6e54d1634806aa39cc
+    size: 16108330
```
paarriagadap, line 30: Note that if I run the script again it shows a (very slighly) different size (16108338)

etl/steps/data/meadow/oecd/2025-06-20/housing_prices.py
```diff
@@ -0,0 +1,55 @@
...
+            "Measure",
+            "Unit of measure",
+            "TIME_PERIOD",
+            "OBS_VALUE",
+        ]
+    ]
+
+    tb = tb.rename(
+        columns={
+            "Reference area": "country",
```
paarriagadap: You can maybe set the dictionary at the beginning of the code so you call the keys in the previous line and the same dictionary here

etl/steps/data/garden/oecd/2025-06-20/housing_prices.countries.json
```diff
@@ -0,0 +1,53 @@
...
+  "Spain": "Spain",
+  "Sweden": "Sweden",
+  "Switzerland": "Switzerland",
+  "United Kingdom": "United Kingdom",
+  "United States": "United States",
+  "China (People\u2019s Republic of)": "China",
+  "Euro area": "Euro area (OECD)",
+  "Euro area (17 countries)": "Euro area (17) (OECD)",
+  "Korea": "South Korea",
+  "OECD": "OECD Countries (OECD)",
```
paarriagadap: You can leave it as "OECD countries" (also with `countries` in lowercase)

etl/steps/data/garden/oecd/2025-06-20/housing_prices.countries.json
```diff
@@ -0,0 +1,53 @@
...
+  "Slovak Republic": "Slovakia",
+  "Slovenia": "Slovenia",
+  "South Africa": "South Africa",
+  "Spain": "Spain",
+  "Sweden": "Sweden",
+  "Switzerland": "Switzerland",
+  "United Kingdom": "United Kingdom",
+  "United States": "United States",
+  "China (People\u2019s Republic of)": "China",
+  "Euro area": "Euro area (OECD)",
```
paarriagadap: These two Euro area entities need more clarity, maybe with entityAnnotationsMap. But I don't see much details in the OECD data explorer's selector, so I might be inclined to drop one

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
```diff
@@ -0,0 +1,38 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    presentation:
+      topic_tags:
+        - Uncategorized
```
paarriagadap, line 6: Remember to change to Housing when this is done

etl/steps/data/garden/oecd/2025-06-20/housing_prices.py
```diff
@@ -0,0 +1,71 @@
...
+        index=["country", "year"],
+        columns="measure",
+        values="value",
+    ).reset_index()
+
+    # set units for each measure
+    for measure, unit in units.items():
+        tb[measure].metadata.title = measure
+        if unit == "Index":
+            tb[measure].metadata.unit = "Index (2015=100)"
```
paarriagadap, line 49: You maybe need some sort of assertion against the column `BASE_PER`, in case in this dataset changes the reference year in the future

etl/steps/data/garden/oecd/2025-06-20/housing_prices.py
```diff
@@ -0,0 +1,71 @@
...
+    tb = tb.rename(
+        columns={
+            "Nominal house price indices": "nom_house_price_idx",
+            "Price to income ratio": "price_to_income_ratio",
+            "Price to rent ratio": "price_to_rent_ratio",
+            "Real house price indices": "real_house_price_idx",
+            "Rent prices": "rent_prices",
+            "Standardised price-income ratio": "std_price_income_ratio",
+            "Standardised price-rent ratio": "std_price_rent_ratio",
+        }
```
paarriagadap: You can maybe define the dictionary at the beginning of the code, for easier review in the future

etl/steps/data/garden/oecd/2025-06-20/housing_prices.py
```diff
@@ -0,0 +1,71 @@
...
+    # Load meadow dataset.
+    ds_meadow = paths.load_dataset("housing_prices")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow.read("housing_prices")
+
+    # get units for each measure
+    units = {}
+    for measure in tb["measure"].unique():
+        units[measure] = tb[tb["measure"] == measure]["unit"].unique()[0]
```
paarriagadap: Copilot suggests me to replace all of this by
```
units = dict(tb.drop_duplicates("measure")[["measure", "unit"]].values)
```

etl/steps/data/garden/oecd/2025-06-20/housing_prices.py
```diff
@@ -0,0 +1,71 @@
...
+    # Pivot table to have measures as columns.
+    tb = tb.pivot(
+        index=["country", "year"],
+        columns="measure",
+        values="value",
+    ).reset_index()
+
+    # set units for each measure
+    for measure, unit in units.items():
+        tb[measure].metadata.title = measure
```
paarriagadap, line 47: I would define a list of variables I want at the beginning of the code, to assert in case that in future updates new indicators appear. The way it is, if a new indicator appears it will show up in the grapher dataset without description_short, key, etc

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
```diff
@@ -0,0 +1,38 @@
...
+
+
+dataset:
+  update_period_days: 365
+
+
+tables:
+  housing_prices:
+    variables:
+      nom_house_price_idx:
```
paarriagadap, line 17: I would drop this indicator, since I don't really see its use in OWID (because of unadjusted prices). In case of keeping it, this is also a note to not use it in charts

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
```diff
@@ -0,0 +1,38 @@
...
+        description_short: Index of residential property prices for both new and existing homes. This index is normalized to 2015 prices (=100) and is not adjusted for inflation.
+        description_from_producer: Residential property prices, sales of newly-built and existing dwellings, all types of dwellings.
+      real_house_price_idx:
```
paarriagadap, line 20: I would add a description_key mentioning that 
```
This indicator is adjusted for inflation by using the private consumption deflator from OECD national accounts.
```

Or something similar

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
```diff
@@ -0,0 +1,38 @@
...
+        description_short: Index of residential property prices for both new and existing homes. This index is normalized to 2015 prices (=100) and is adjusted for inflation.
+        description_from_producer: Nominal house prices deflated using the private consumption deflator from the national account statistics.
+      rent_prices:
```
paarriagadap, line 28: I wouldn't use this one either, because it's not adjusted for inflation

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
```diff
@@ -0,0 +1,38 @@
...
+        description_short: Index for rents prices of residential property. This index is normalized to 2015 prices (=100) and not adjusted for inflation.
+        description_from_producer: Data from the OECD Main Economic Indicators database or from Eurostat, Consumer Price Indices (CPIs) for Actual rentals for housing (COICOP 04.1)
+      price_to_income_ratio:
+        description_short: Ratio of house prices to household disposable income per person, a measure for how affordable buying a home is. It is calculated as the ratio of the "nominal house price index" to "net household disposable income index".
```
paarriagadap: I would add a description_key along the lines of
```
Household disposable income per person is obtained from OECD national accounts
```

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
```diff
@@ -0,0 +1,38 @@
...
+        description_from_producer: Nominal house prices divided by nominal disposable income per head. Net household disposable income is used. The population data come from the OECD national accounts database. The long-term average is calculated over the whole period available when the indicator begins after 1980 or after 1980 if the indicator is longer. This value is used as a reference value. The ratio is calculated by dividing the indicator source on this long-term average, and indexed to a reference value equal to 100.
+      std_price_rent_ratio:
+        description_short: Ratio of house prices to rent, compared to the long-term average (since 1980). The average is set to 100.
+        description_from_producer: Nominal house prices divided by rent price indices. The long-term average is calculated over the whole period available when the indicator begins after 1980 or after 1980 if the indicator is longer. This value is used as a reference value. The ratio is calculated by dividing the indicator source on this long-term average, and indexed to a reference value equal to 100.
```
paarriagadap: I am not sure if I understand these two indicators. Have you found a document explaining these further? They might seem good for analysis, but I don't quite get the difference between these and the non-standardized ones (because the long-term average explanation is weird).

etl/steps/data/garden/oecd/2025-06-20/housing_prices.countries.json
antea04: Yeah, I haven't found any documentation on the difference between these - I'll drop them both, thanks!

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
antea04: Yeah, once the topic page/ tag is established I will!

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
antea04: Well how I understand it is that the non-standardized ones are indexed to a single year (2015) while in the standardized ones 100% corresponds to the "long term average" (the average of the index from 1980 until today).

The only additional documentation I found is the one on the OECD data explorer for the whole dataset:

"The standardised price-rent and price-income ratios show the current price-rent and price-income ratios relative to their respective long-term averages. The long-term average, which is used as a reference value, is calculated over the whole period available when the indicator begins after 1980 or 1980 if the indicator is available over a longer time period." 

What do you think - should I rather use this description for the "description_from_producer"? I can also add some "description_key" bullets explaining this if it could help.

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
antea04: Yeah, we are planning on using the real prices for all charts. I'd still import the indicator (just to have it), I think with the description it's pretty clear that it is unadjusted and should not be used.

etl/steps/data/garden/oecd/2025-06-20/housing_prices.meta.yml
antea04: as above

PR #4613, opened 2025-06-20
📊 energy: Update renewable energy patents data

pabloarosado:
## Summary
* Update IRENA's data on renewable energy patents.
* Archive unused steps.

## Review
@paarriagadap there were only minor changes in the code with respect to the previous update, so there's no need for a code review. Feel free to have a look at the charts and please let me know if you have any suggestions or see any issues, thanks!


paarriagadap (approved):
Thanks Pablo! The update looks good, but I have these suggestions:

- [Annual patents filed in sustainable energy](http://staging-site-data-renewable-energy-patent/admin/charts/1414/edit): 
  - I would have the chart view as default (given the scarcity of data in the globe)
  -  I would harmonize the data page and chart's title (by removing the title_public, for example)
  - I also feel that the subtitle should be note and viceversa (though I imagine this was done purposely to explain the drops)
- [Annual patents filed for electric vehicle technologies](http://staging-site-data-renewable-energy-patent/admin/charts/6253/edit). Maybe I miss some subtitle explaining what are you referring to when you say technologies. Looks more like uses?
- [Annual patents filed for carbon capture and storage technologies](http://staging-site-data-renewable-energy-patent/admin/charts/6254/edit)
  - Same comments as for the first chart
  - Also, I would explain what carbon capture and storage technologies are in the subtitle

pabloarosado:
Thanks @paarriagadap, I've implemented most of your suggestions. I decided to show the map tab by default, but changing the default year to a few years back, where the data is more complete. Showing the chart tab directs the attention to the hard drop in recent years (caused by data incompleteness), and distracts from anything else. I think the most important aspect of these charts is to show which countries are contributing to certain technologies, and roughly by how much, without focussing on the latest year.

PR #4621, opened 2025-06-23
🔨 Refactor DAG

Marigold:
## Human overview

The way we create steps based on `--private`, `--export` and `--grapher` was too convoluted. I refactored all functions for DAG manipulation so that it only relies on `filter_to_subgraph` with proper `includes` and `excludes`. We load full DAG and then create a subDAG.

The initial motivation was speeding up `StepUpdater` by using subdag instead of full dag, but I couldn't make it work. Anyway, it wasn't as important and I've wanted to refactor DAG for a long time.

## AI Summary

This PR refactors the DAG (Directed Acyclic Graph) architecture in the ETL system to improve modularity, maintainability, and testability. The refactor separates DAG construction, filtering, and step compilation into distinct phases with cleaner interfaces.

### Key Changes

- **🔄 Simplified DAG Types**: Consolidated `DAG` type to `Dict[str, Set[str]]` and removed separate `Graph` type
- **📦 Modular Architecture**: Separated DAG processing into clear phases:
  1. Load DAG from file
  2. Construct full DAG (with grapher steps)  
  3. Filter to subDAG based on includes/excludes
  4. Compile subDAG into Step objects
  5. Execute steps
- **🗑️ Removed Downstream**: Eliminated unused `--downstream/-d` functionality and related complexity
- **🔧 Enhanced Filtering**: Improved subDAG construction with better regex pattern matching and error handling
- **⚡ Performance**: Pre-compiled step lookup for parallel execution
- **✅ Comprehensive Tests**: Added extensive test coverage for new DAG construction functions

### API Changes

- `compile_steps()` signature simplified from multiple filtering parameters to `(dag, subdag)`
- `main()` function renamed `steps` parameter to `includes` and `exclude` to `excludes` (now a list)
- New functions: `construct_full_dag()`, `construct_subdag()`
- Removed: `--downstream/-d` CLI flag and related filtering logic

### Files Modified

- `etl/command.py` - Core DAG processing logic refactored
- `etl/steps/__init__.py` - Step compilation simplified  
- `tests/test_command.py` - Comprehensive test coverage added
- Various CLI tools updated to use new API

The refactor maintains backward compatibility for most use cases while providing a cleaner, more maintainable foundation for future enhancements.

### DIFF
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -100,7 +100,6 @@ def test_construct_subdag():
         "data-private://garden/secret/2023-01-01/secret": {"snapshot-private://meadow/secret/2023-01-01/secret"},
         "private://test/private_step": {"data-private://garden/secret/2023-01-01/secret"},
         "data://meadow/climate/2023-01-01/temperature": {"snapshot://meadow/climate/2023-01-01/temperature"},
-        "grapher://grapher/regions/latest/regions": {"data://grapher/regions/latest/regions"},
     }
 
     # Test basic include filtering
@@ -143,11 +142,6 @@ def test_construct_subdag():
     subdag = cmd.construct_subdag(full_dag, includes=[".*"], private=True)
     private_steps = [step for step in subdag.keys() if step.startswith("private://")]
     assert len(private_steps) > 0
-
-    # Test regions exclusion
-    subdag = cmd.construct_subdag(full_dag, includes=[".*"], grapher=True)
-    regions_step = "grapher://grapher/regions/latest/regions"
-    assert regions_step not in subdag
 
     # Test exact match - should include dependencies by default
     subdag = cmd.construct_subdag(full_dag, includes=["data://garden/happiness/2023-01-01/happiness"], exact_match=True)


etl/command.py
```diff
@@ -248,16 +241,15 @@ def main_cli(
...
         dry_run=dry_run,
         force=force,
         private=private,
         grapher=grapher,
         export=export,
-        downstream=downstream,
         only=only,
         exact_match=exact_match,
-        exclude=exclude,
+        excludes=exclude.split(",") if exclude else None,
```
pabloarosado, line 252: I suppose `include` and `exclude` would be preferrable (shorter, imperative) than `includes` and `excludes`.

pabloarosado: (same in other similar places)

etl/command.py
```diff
@@ -301,21 +300,34 @@ def main(
     if grapher:
         sanity_check_db_settings()
 
-    dag = construct_dag(dag_path, private=private, grapher=grapher, export=export)
+    # Load DAG
+    dag = load_dag(dag_path)
+
+    # Create full DAG
```
pabloarosado, line 307: If I understand the intention here, load_dag should simply load the dag, regardless of our internal ETL logic, and construct_full_dag should apply additional logic. Is that right?
If so, I see that load_dag currently checks for "fasttrack" steps. Shouldn't this be part of construct_full_dag?

etl/command.py
```diff
@@ -301,21 +300,34 @@ def main(
...
+        grapher=grapher,
+        export=export,
+        private=private,
+        only=only,
+        exact_match=exact_match,
+    )
 
-    excludes = exclude.split(",") if exclude else []
+    # Compile subdag into Step objects
+    steps = compile_steps(full_dag, subdag)
```
pabloarosado, line 323: I think the comments and docstrings should clarify what's happening here. I just saw this comment in compile_steps: `# TODO: we need full DAG here to get dependencies of each step`. Instead of a TODO, this should be part of the docstring, since it explains why we need both a full and a filtered dag.

etl/command.py
```diff
@@ -331,68 +343,80 @@ def sanity_check_db_settings() -> None:
...
+    if not grapher and not export:
+        excludes.append("grapher://.*")
+
+    # Exclude private steps
+    if not private:
+        excludes.append("private://.*")
+
+    # Exclude grapher regions, they're fetched by owid-grapher as CSV from catalog
+    # but are not supposed to be in DB
+    excludes.append("grapher://grapher/regions/latest/regions")
```
pabloarosado: This step is now archived, so I suppose we don't need this hack anymore, do we?

pabloarosado (approved):
Thanks for the refactor, it looks good!
I left some minor suggestions. Feel free to merge after considering them.

etl/command.py
Marigold: The thing is that `excludes` is a list once we split `exclude` by `,`. If I were writing it from scratch, I'd have used regex instead of comma separated arguments. I'll keep it as it is, I don't want to risk bugs just for a cosmetic change.

(Anyway, leaving a note for future refactoring.)

etl/command.py
Marigold: That's right, but the fasttrack check doesn't modify the DAG in any way. Besides, it'd be harder to move it there as we lack dag path which is vital for the check.

etl/command.py
Marigold: Good point, fixed.

etl/command.py
Marigold: :tada:

PR #4627, opened 2025-06-24
📊 Housing Knoll

antea04:


etl/steps/data/garden/housing/2025-06-25/knoll_prices.meta.yml
```diff
@@ -0,0 +1,28 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    presentation:
+      topic_tags:
+        - Uncategorized
```
spoonerf: You'll need to upgrade one of the housing tags to a topic tag, by adding a slug. I'm not sure which one would be more appropriate though.

etl/steps/data/garden/housing/2025-06-25/knoll_prices.py
```diff
@@ -0,0 +1,41 @@
...
+    # Load inputs.
+    #
+    # Load meadow dataset.
+    ds_meadow = paths.load_dataset("knoll_prices")
+
+    # Read table from meadow dataset.
+    tb = ds_meadow.read("knoll_prices")
+
+    # Calculate real house prices from nominal house prices and CPI (analogue to Knoll)
+    tb["hpreal"] = tb["hpnom"] / tb["cpi"] * 100
```
spoonerf, line 22: I guess it's okay to deflate house prices like this? It might be worth running past Bertha or Pablo A if you have any uncertainty.

etl/steps/data/garden/housing/2025-06-25/knoll_prices.meta.yml
```diff
@@ -0,0 +1,28 @@
...
+
+
+tables:
+  knoll_prices:
+    variables:
+      hpreal:
+        title: "Real house price index"
+        unit: "Index (1990=100)"
+        short_unit: ""
+        description_short: Long-run data on real house prices, normalized to 1990=100.
```
spoonerf: Maybe a silly question but what's the difference between 'house prices' and 'real house prices'? Is it just that they are normalized, if so I think you can drop the 'real' as you mention the normalization in the subtitle.

spoonerf (approved):
Looks good, just a couple of minor comments!

PR #4635, opened 2025-06-25
🐛 Use createdAt in chart_revisions

Marigold:
We should be using `createdAt`, not `updatedAt` when getting `chart_revisions`. It used to be identical, but recent migration have bumped `updatedAt` of all revisions and broke chart-diff.


lucasrodes (approved):
lgtm! thanks for the fix

PR #4641, opened 2025-06-25
📊 lexical-index-work

lucasrodes:
Add new indicator as per Bastian's suggestion: https://github.com/owid/owid-issues/issues/1978#issuecomment-3005663199

/schedule


lucasrodes:
Apparently this is no longer needed.

PR #4665, opened 2025-06-27
📊 house prices per GDP

antea04:


etl/steps/data/garden/housing/2025-06-25/knoll_prices.py
```diff
@@ -27,7 +30,26 @@ def run() -> None:
     # Process data.
     #
     # Harmonize country names.
-    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path)
+    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path, warn_on_unused_countries=False)
+    tb_gdp = geo.harmonize_countries(
```
paarriagadap: I am not sure why Maddison Project data is re-harmonized, when the harmonization is already done in their respective step

etl/steps/data/garden/housing/2025-06-25/knoll_prices.py
```diff
@@ -27,7 +30,26 @@ def run() -> None:
...
+        df=tb_gdp, countries_file=paths.country_mapping_path, warn_on_unused_countries=False
+    )
+
+    # calculate real house prices per gdp per capita
+    tb = tb.merge(tb_gdp, on=["country", "year"], how="left")
+
+    tb["hp_per_gdp"] = tb["hpreal"] / tb["gdp_per_capita"] * 1000
```
paarriagadap, line 38: I am not quite sure about what's the goal with this indicator. I don't understand what it represents to divide an index by GDP pc

etl/steps/data/garden/housing/2025-06-25/knoll_prices.py
```diff
@@ -27,7 +30,26 @@ def run() -> None:
...
+
+    # index time series to 1990 value for each country
+    for country in tb["country"].unique():
```
paarriagadap, line 41: What I usually do is to create a new table with the data for 1990, and then I merge by country, so I divide by the new column that only has data for 1990. This seems to work just as well

etl/steps/data/garden/housing/2025-06-25/knoll_prices.py
paarriagadap: Note that this index is adjusted for inflation and the GDP per capita for inflation _and_ living costs, so that also makes things more confusing

paarriagadap: It think it would be clearer if the housing prices were measured in monetary values, so then a comparison with GDP (in the same units) would make more sense.

paarriagadap: Also, why GDP per capita?

paarriagadap (commented):
> Hi! I left my comments below. I am not sure what the goal with this indicator is, because it's a weird mix of an index divided by GDP in different units. It is confusing to explain

PR #4674, opened 2025-07-01
📊 Update WB income groups (code changes)

paarriagadap:
Main issue https://github.com/owid/owid-issues/issues/2054


etl/steps/data/garden/wb/2025-07-01/income_groups.meta.yml
```diff
@@ -5,16 +5,17 @@ definitions:
...
 
       Please note: Regions in this table include economies at all income levels. The term country, used interchangeably with economy, does not imply political independence but refers to any territory for which authorities report separate social or economic statistics. Click [here](https://datahelpdesk.worldbank.org/knowledgebase/articles/378834-how-does-the-world-bank-classify-countries) for information about how the World Bank classifies countries.
+
     description_key:
       - The World Bank creates a yearly classification of countries by income, for all countries with population over 30,000.
       - This classification stays the same throughout the fiscal year (from July 1 to June 30) even if the income data for a country changes.
```
marcelgerber: maybe it makes sense to clarify this some more:
```suggestion
      - This classification stays the same throughout the World Bank's fiscal year (from July 1 to June 30) even if the income data for a country changes.
```

paarriagadap: Thank you!

etl/steps/data/garden/wb/2025-07-01/income_groups.meta.yml
```diff
@@ -5,16 +5,17 @@ definitions:
         - Economic Growth
     processing_level: major
     description_from_producer: |-
-      For the current 2025 fiscal year, low-income economies are defined as those with a GNI per capita, calculated using the [World Bank Atlas method](https://datahelpdesk.worldbank.org/knowledgebase/articles/378832-what-is-the-world-bank-atlas-method), of $1,145 or less in 2023; lower middle-income economies are those with a GNI per capita between $1,146 and $4,515; upper middle-income economies are those with a GNI per capita between $4,516 and $14,005; high-income economies are those with more than a GNI per capita of $14,005.
+      For the current 2026 fiscal year, low-income economies are defined as those with a GNI per capita, calculated using the [World Bank Atlas method](https://datahelpdesk.worldbank.org/knowledgebase/articles/378832-what-is-the-world-bank-atlas-method), of $1,135 or less in 2024; lower middle-income economies are those with a GNI per capita between $1,136 and $4,495; upper middle-income economies are those with a GNI per capita between $4,496 and $13,935; high-income economies are those with more than a GNI per capita of $13,935.
```
pabloarosado, line 8: In case you think this could make your life easier on the next update, consider using placeholders, e.g. {YEAR_FISCAL}, and then `paths.create_dataset([tb], yaml_params={"YEAR_FISCAL": year_fiscal})` (where `year_fiscal` and other variable fields could be extracted from the metadata. Or the entire `description_from_producer` could be extracted programmatically. Up to you!

snapshots/wb/2025-07-01/income_groups.xlsx.dvc
```diff
@@ -2,11 +2,17 @@ meta:
   origin:
     producer: World Bank
     title: Income Classifications
-    citation_full: World Bank - Income classifications (2024).
+    description: |-
+      Every year, the World Bank Group classifies the world’s economies into four income groups: low, lower-middle, upper-middle, and high. These classifications, updated each year on July 1, are based on the previous year’s Gross National Income (GNI) per capita, expressed in U.S. dollars using the [Atlas method](https://datahelpdesk.worldbank.org/knowledgebase/articles/77933-what-is-the-world-bank-atlas-method).
+    citation_full: World Bank - Income classifications (2025).
```
pabloarosado: I think "Income classifications" may be a good title to keep (it's much more intuitive than the official one), but for the full citation, I suppose the official title of the dataset should be "World Bank Country and Lending Groups", right?

pabloarosado (approved):
Thanks, the code looks good! I just left minor suggestions.

Regarding the charts, I think you may have accidentally used the dynamic classification in scatter charts. We decided recently we would always use the static classifications, right?

Another small issue: I'm wondering if we should clarify the situation of Ethiopia. I saw that [the main site](https://datahelpdesk.worldbank.org/knowledgebase/articles/906519-world-bank-country-and-lending-groups) says:
> Notes: Venezuela, RB, classified as an upper-middle income country until FY21, has been unclassified since then due to the unavailability of data. Ethiopia is currently in a temporary status of unclassification for FY26.

And indeed, [the current classification file](https://ddh-openapi.worldbank.org/resources/DR0095333/download) has no classification for Ethiopia in FY2026. However, [the historical file](https://ddh-openapi.worldbank.org/resources/DR0095334/download) does have a classification (L).
Maybe the situation of Venezuela and Ethiopia could be mentioned in some of the chart footnotes? I leave it up to you. Thanks!

etl/steps/data/garden/wb/2025-07-01/income_groups.meta.yml
paarriagadap: Yeah, though life's easy with these updates anyway :). As you see, we don't have to change much and it's good to check the thresholds

snapshots/wb/2025-07-01/income_groups.xlsx.dvc
paarriagadap: Yeah, I guess so. Lending groups are another thing, but they are included in the file and the webpage where they describe the income groups. Thanks!

pabloarosado:
Thanks for the fixes.
I see, maybe it's problematic to have the year in the indicator name, when running indicator upgrader. One solution would be to have "World Bank's latest income classification" as the indicator title, and "World Bank's 2025 income classification" as the display name.

PR #4679, opened 2025-07-03
🎉 add gsheet function

spoonerf:


etl/data_helpers/misc.py
```diff
@@ -836,3 +838,243 @@ def humanize_number(number, sig_figs=2):
...
+
+    >>> # Only metadata for specific variables
+    >>> url, sheet_id = export_table_to_gsheet(
+    ...     tb,
+    ...     "My Dataset - Table Name",
+    ...     metadata_variables=["child_mortality_rate", "population"]
+    ... )
+    """
+    # Check if Google API credentials are available
+    if not CLIENT_SECRET_FILE or not CLIENT_SECRET_FILE.exists():
```
lucasrodes: Maybe also good to check if this is running locally.

You can do this with `OWID_ENV.env_local`, which tells what is the environment where the code is actually running.

```python
from etl.config import OWID_ENV

OWID_ENV.env_local
```

This can return `production`, `dev` or `staging`. We should ensure that this returns `dev`, which stands for local execution.

etl/steps/data/garden/un/2025-04-25/long_run_child_mortality.py
```diff
@@ -34,20 +34,24 @@ def run() -> None:
...
 
     max_year = tb_igme["year"].max()
     tb_gap_full = ds_gapminder_v11["under_five_mortality"].reset_index()
     tb_gap_full = tb_gap_full[tb_gap_full["year"] <= max_year].reset_index(drop=True)
     tb_gap_full = tb_gap_full.rename(columns={"child_mortality": "child_mortality_rate"}, errors="raise")
-    tb_gap_full["source"] = "Gapminder"
+    tb_gap_full["source"] = "Gapminder v11"
+    tb_gap_full["source_url"] = (
+        "https://docs.google.com/spreadsheets/d/1Av7eps_zEK73-AdbFYEmtTrwFKlfruBYXdrnXAOFVpM/edit?gid=501532268#gid=501532268"
+    )
```
pabloarosado, line 47: Hmm, there wasn't any v11 in the data, is that unexpected?

etl/google.py
```diff
@@ -890,8 +890,23 @@ def write_dataframe(
...
+                # Create the sheet
+                self.add_sheet(sheet_name)
+            else:
+                # Clear existing sheet data only if sheet exists
+                self.clear_sheet_data(sheet_name)
+
+        except Exception as e:
+            # If we can't check/create sheets, try to create anyway
+            try:
+                self.add_sheet(sheet_name)
```
pabloarosado: Is the previous line ever going to be used? If the previous try failed, I suppose the user doesn't have the right permissions and this will fail too. Or are there any other specific reasons why the previous would fail (if you have identified it, consider catching exactly that in the exception above).

etl/data_helpers/misc.py
```diff
@@ -11,8 +14,30 @@
...
+        "title": "Title",
+        "description_short": "Description",
+        "description_key": "What you should know about this data",
+        "description_processing": "Processing steps",
+        "unit": "Unit",
+        "origins": "Data sources",
+    },
+}
+DEFAULT_TEAM_FOLDER_NAME = "ETL GSheet Exports"
+OWID_SHARED_FOLDER_ID = "1qH0uBtO5KLvdew8X6u-lF75E4uKHSrjp"  # Folder ID
```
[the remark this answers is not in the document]
pabloarosado: I think the folder should not be public (unless there's a good reason why we think it should be).

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+
+@retry_on_network_error()
+def _create_google_sheet(sheet_title: str, folder_id: Optional[str]) -> GoogleSheet:
+    """Create Google Sheet with retry logic."""
+    if folder_id:
+        # Create sheet in the specified folder
+        return GoogleSheet.create_sheet(title=sheet_title, folder_id=folder_id)
+    else:
+        # Create sheet in root
+        return GoogleSheet.create_sheet(title=sheet_title)
```
pabloarosado: I think this function is not necessary, as the logic is already implemented: the method GoogleSheet.create_sheet creates the sheet in the root folder if folder_id=None.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+
+
+def _update_or_create_sheet(
```
pabloarosado: Couldn't this logic be part of GoogleSheet.create_sheet, to avoid having the code scattered between google and misc? Maybe it would be good to also clarify what "sheet" means (both here and in the google module). It's unclear to me whether it means "a google sheet file" or "a sheet inside a google sheet file".

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    except Exception as e:
+        print(f"Warning: Could not update existing sheet: {e}")
+
+    # Fallback to creating new sheet in the specified folder
+    return _create_google_sheet(sheet_title, folder_id)
+
+
+def _write_main_data(sheet: GoogleSheet, df: pd.DataFrame) -> None:
+    """Write main data to the sheet."""
+    sheet.write_dataframe(df)
```
pabloarosado: This one-line function doesn't seem necessary.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+        try:
+            sheet.write_dataframe(metadata_df, sheet_name=sheet_name)
+        except Exception as e:
+            print(f"Note: Could not create metadata sheet '{sheet_name}': {e}")
+
+
+def _set_permissions(sheet_id: str, role: str, general_access: str) -> None:
+    """Set permissions for the sheet."""
+    drive = GoogleDrive()
+    drive.set_file_permissions(file_id=sheet_id, role=role, general_access=general_access)
```
pabloarosado, line 1123: This function doesn't seem necessary.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+            print(f"Failed to create folder: {folder_name}")
+            return None
+
+        # Set permissions for team access
+        drive.set_file_permissions(file_id=folder_id, role="reader", general_access="anyone")
+
+        print(f"Created new folder: {folder_name} (ID: {folder_id})")
+        print(f"Folder URL: https://drive.google.com/drive/folders/{folder_id}")
+
+        return folder_id
```
pabloarosado: As far as I understand, this function will be used to create a folder just once, and then any other time it should simply return the folder ID. I'd suggest removing this function (it could be reduced to a couple of commented lines of code to explain how the folder was created), and simply getting the folder ID from etl.config (which loads the folder ID from .env).

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    # Create main folder only
+    main_folder_id = create_or_get_shared_folder("OWID ETL Exports")
+
+    if main_folder_id:
+        print(f"✅ Team folder created/found: {main_folder_id}")
+        print(f"📁 Folder URL: https://drive.google.com/drive/folders/{main_folder_id}")
+        return main_folder_id
+    else:
+        print("❌ Failed to create team folder")
+        return None
```
pabloarosado: This function seems to be unnecessary (and should be used just once).

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+
+def _get_variables_to_process(table: Table, metadata_variables: Optional[List[str]]) -> List[str]:
+    """Determine which variables to process for metadata export."""
+    if metadata_variables is None:
+        return list(table.columns)
+
+    variables_to_process = [var for var in metadata_variables if var in table.columns]
+    missing_vars = [var for var in metadata_variables if var not in table.columns]
+    if missing_vars:
+        print(f"Warning: Variables {missing_vars} not found in table")
```
pabloarosado: Replace all prints by log.warning or log.error.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    return [[base_display_name, formatted_value]]
+
+
+def _create_metadata_dataframes(
+    table: Table, metadata_variables: Optional[List[str]] = None
+) -> Dict[str, pd.DataFrame]:
+    """Create metadata DataFrames for each variable in the table."""
+    metadata_dfs = {}
+
+    variables_to_process = _get_variables_to_process(table, metadata_variables)
```
pabloarosado: I think the logic of _get_variables_to_process could simply be here, to avoid unnecessary code.

etl/data_helpers/misc.py
pabloarosado: Maybe we could call "a sheet inside a google sheet file" a "tab"?

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+
+
+def _write_main_data(sheet: GoogleSheet, df: pd.DataFrame) -> None:
+    """Write main data to the sheet."""
+    sheet.write_dataframe(df)
+
+
+def _add_metadata_tabs(sheet: GoogleSheet, table: Table, metadata_variables: Optional[List[str]]) -> None:
+    """Add metadata tabs to the sheet."""
+    metadata_dfs = _create_metadata_dataframes(table, metadata_variables)
```
pabloarosado: I think the logic of _create_metadata_dataframes could live here.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+            # Handle special case for origins - split into separate rows
+            if key == "origins":
+                if len(value) == 1:
+                    # Single origin - extract attributes
+                    origin = value[0]
+                    rows.extend(_extract_origin_attributes(origin, display_name))
+                else:
+                    # Multiple origins - extract attributes for each
+                    for i, origin in enumerate(value, 1):
+                        numbered_display_name = f"{display_name} ({i})"
```
pabloarosado: I think this can be simplified. There's no need to make a distinction between one or multiple origins (and simply assume that, in general, there will always be multiple).

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    if not sheet_title or not sheet_title.strip():
+        print("Error: sheet_title cannot be empty")
+        return False
+
+    return True
+
+
+def _prepare_dataframe(table: Table) -> pd.DataFrame:
+    """Convert table to DataFrame and optimize for Google Sheets."""
+    df = table.reset_index()
```
pabloarosado: The resetting will be necessary only if the Table has an index, otherwise this will create a spurious "index" column.
Maybe this logic should live in google.py, rather than here.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    return not CLIENT_SECRET_FILE or not CLIENT_SECRET_FILE.exists() or OWID_ENV.env_local != "dev"
+
+
+def _sanitize_sheet_name(name: str) -> str:
+    """Sanitize sheet name for Google Sheets compatibility."""
+    sanitized = name.replace("/", "_").replace("\\", "_").replace(":", "_")
+    return sanitized[: GSHEET_EXPORT_CONFIG["MAX_SHEET_NAME_LENGTH"]]
+
+
+def _validate_inputs(table: Table, sheet_title: str) -> bool:
```
pabloarosado: I'd kill this function.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+
+        try:
+            metadata_rows = _extract_metadata_rows(
+                table[column_name].metadata, GSHEET_EXPORT_CONFIG["INCLUDED_METADATA_FIELDS"]
+            )
+
+            if metadata_rows:
+                # Create DataFrame with proper column headers
+                metadata_df = pd.DataFrame(metadata_rows, columns=["Property", "Value"])
+                clean_name = _sanitize_sheet_name(column_name)
```
pabloarosado: I don't think we need the function _sanitize_sheet_name. Its logic could simply be here (I'd rather create that function if we need additional logic).

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    else:
+        return ", ".join(str(item) for item in value)
+
+
+def _should_skip_export() -> bool:
+    """Check if Google Sheets export should be skipped."""
+    return not CLIENT_SECRET_FILE or not CLIENT_SECRET_FILE.exists() or OWID_ENV.env_local != "dev"
+
+
+def _sanitize_sheet_name(name: str) -> str:
```
pabloarosado: This function seems unnecessary.

etl/data_helpers/misc.py
```diff
@@ -836,3 +861,425 @@ def humanize_number(number, sig_figs=2):
...
+    """Format list metadata based on key type."""
+    if key in ["description_key", "origins", "licenses"]:
+        return "\n".join(str(item) for item in value)
+    elif len(str(value)) > 100:
+        return f"[List with {len(value)} items] {str(value)[:100]}..."
+    else:
+        return ", ".join(str(item) for item in value)
+
+
+def _should_skip_export() -> bool:
```
pabloarosado: This function seems unnecessary.

pabloarosado (changes requested):
Nice, thanks for doing this! I have some comments in the code and some extra ones here:
* I think a fair amount of code added in misc is unnecessary. Most of the logic seems to be premature generalization and could be deleted. Also, some of the logic is closely related to Google Sheets and it would be better to move it to the google module. But we can do that in a separate PR if you prefer to move on.
* Looking at the example sheet, I see that there is a "Sheet1" and a "Data" sheet. I guess one of them was created by mistake?
* In the metadata sheet, instead of Data sources (1), (2), (3), maybe it would be better to explicitly say the name that is quoted in the Data sheet, e.g. Data source: UN IGME, Data source: Gapminder v7, Data source: Population.
* The name of the metadata sheet is "metadata_child_mortality_rate". If the idea is to host just one indicator, this could simply be called Metadata (to be consistent with the Data sheet).
  * And have you decided on what should be the logic when there are more than one variables, or do you think for now we should only create sheets for single variables?
* Do we want the folder of sheets to be public, or just the files? Currently, the folder is public, but I think the safest option would be to make the folder private (just shared among GCDL) and make only the files public. If you agree, then it would also be good to keep the id of the folder hidden (as a global variable in the .env file, for example); otherwise we could get access requests from random people.
* Don't you think that the URLs in the Data sheet are a bit redundant? I'd rather include that in the Metadata sheet.
* I created a function in owid.catalog.utils called remove_details_on_demand, in case it helps (I think in this case you didn't have any).
* Nitpicking: would it be possible to delete empty rows and columns from the sheets?

Thanks! Let me know if you want to discuss any of my comments.

etl/data_helpers/misc.py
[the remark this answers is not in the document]
spoonerf: Yeh I agree, sorry if this is a silly question, but do we have a shared .env file or would each person have to add this to theirs individually?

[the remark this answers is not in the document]
pabloarosado: It's not a silly question. Each one of us has their own .env file (ignored by git), but we do have a shared .env.example file, where we can add all possible variables (without their values).
But I agree that adding too many things there can start to become annoying, so maybe we can figure out a better solution (like ansible). I'll ask around. For now, feel free to add things to the .env and the .env.example.

etl/steps/data/garden/un/2025-04-25/long_run_child_mortality.py
spoonerf: Ah, there should be once I merge a different PR (#4664), but I think only for 'World'.

etl/data_helpers/misc.py
spoonerf: Good point, I'll hash it out. I'd like to keep it as when I created the folder manually there were permissions errors, but this function seems to work more happily.

etl/data_helpers/misc.py
spoonerf: Good point, I'll use tab when it's a tab inside a sheet.

spoonerf:
Hey @pabloarosado thanks for the thorough review! I think I've addressed most of your points now, so I'd be keen to merge and try it out in the wild, then maybe work on future improvements in a separate PR. Let me know what you think!

PR #4692, opened 2025-07-04
:honeybee: Update all datasets to latest WB income groups

paarriagadap:
Main issue https://github.com/owid/owid-issues/issues/2055


pabloarosado:
> Hi @pabloarosado!
> 
> Just to understand what breaks and what doesn't, I changed all the `income_groups` dependencies to the latest version. Doing that I got a "dependency order" error. Do you know what is it?

Hey Pablo, that's brave, thanks for tackling this! I think the issue here is that there is a circular dependency: population depends on income_groups, and vice versa. It may be tricky to fix (I recently fixed a similar one with FAOSTAT and population). We'd need to think about this one a bit to come up with a good solution.

pabloarosado:
I had a quick look and just thought that there may be a simple solution: we could move the indicators on population counts (`*_income_countries_count`) as a separate dataset. I think those indicators are only used by the income groups grapher step for a few specific charts. Most other datasets in the dag that use the income groups garden step use the `classification` or `classification_latest` columns only.
If you take those `*_income_countries_count` indicators to a separate dataset, then the income groups garden step would not need to depend on `population` anymore.
Does that make sense?

dag/energy.yml
```diff
@@ -86,7 +86,7 @@ steps:
   data://garden/energy/2024-06-20/energy_mix:
     - data://garden/demography/2023-03-31/population
     - data://garden/energy_institute/2024-06-20/statistical_review_of_world_energy
-    - data://garden/wb/2024-03-11/income_groups
+    - data://garden/wb/2025-07-01/income_groups
```
paarriagadap: Is this needed here, @pabloarosado? You don't seem to call it, and it runs fine without it

pabloarosado: Good catch, it's not needed.

etl/steps/data/garden/energy/2025-05-12/electricity_mix.py
```diff
@@ -390,6 +390,7 @@ def fix_discrepancies_in_aggregate_regions(tb_review: Table, tb_ember: Table, co
                         # px.line(compared.melt(id_vars="year"), x="year", y="value", color="variable", markers=True, title=f"{region} - {col}").show()
                         assert compared["year"].min() == 2000, "Minimum year changed."
         error = f"Expected discrepancies between Statistical Review and Ember data for aggregate regions may have changed for region: {region}. Current discrepant indicators: {_remove_combination}. Use this list in 'segments_not_combined'."
+        # TODO: Fix the error here
```
paarriagadap: @pabloarosado Now this breaks: 
```
AssertionError: Expected discrepancies between Statistical Review and Ember data for aggregate regions may have changed for region: Upper-middle-income countries. Current discrepant indicators: ['gas_generation__twh', 'oil_generation__twh']. Use this list in 'segments_not_combined'.
```

pabloarosado: That's the kind of issues I was expecting, when income groups change! I've just pushed a fix.

paarriagadap:
Thanks @pabloarosado! Now all the checks are in green. I think we should now check the [1028 charts](http://staging-site-data-update-dataset-wb/etl/wizard/chart-diff?show_all=&chart-diff-sort-by-results=relevance) with data changes. Is there a more efficient way to check these views with the income groups selected?

paarriagadap:
Thanks @pabloarosado!

I have taken a look to the first 13 pages in [chart-diff](http://staging-site-data-update-dataset-wb/etl/wizard/chart-diff?chart-diff-sort-by-results=relevance&show_all=) and I think this works well. Of course, there are changes in the data, some sudden jumps, lines starting from different years, but that is expected. I left those cases unreviewed for you to check (again, only for the first 13 pages) because they are mostly coming from your data and to have a sense on how critical do you think these changes are.

pabloarosado (approved):
Hi @paarriagadap thanks for having the courage to tackle this issue! Your changes in the code make sense. Some of your changes will be in conflict with my current big energy update, but I'll handle those in my PR.
I've had a very quick look at the rest of the chart diff pages (up to page 40), and yes, there are some significant changes, but nothing unexpected. More important would be to check if there are changes that could be in conflict with the text in referring articles. But unfortunately, there's no feasible way to achieve that.
So I'd suggest:
1. Remove the "hack" I added to visualize income groups in charts.
2. Optionally (up to you) ask around in some of the data* slack channels, in case anyone disagrees with the approach of updating all steps (maybe someone comes up with some special case we should keep in mind). And, otherwise...
3. Merge (for safety, feel free to do that on Monday, in case there are fires)!

Thanks!

PR #4694, opened 2025-07-07
🐝 Add instructions to google module

pabloarosado:
Add instructions to be able to use the google module.

@lucasrodes feel free to try out the instructions (if you think you may use the module, otherwise feel free to skip). Thanks!


lucasrodes (commented):
Instructions look fine, thanks for doing this!


My only slight concern is where we are saving this logic. We save this under `apps/` and not `etl/`. Why is this so?

I think `apps/` should be allowed to import from `etl/` but not vice versa. In my opinion, `apps` stands for tools, and applications powered by `etl/`, which is lower-level and not app-specific. Instead, having our core library depend on an "application-level" module feels like a bad design choice. 

I am saying this because I've observed some imports from `apps.google` in `etl/scripts` modules, for instance (there might be others, for sure). Instead, we could either (i) move `apps.google` to `etl/scripts` or move `etl/scripts` elsewhere.

It's fine if you consider this out of scope for this PR, since the module `apps.google` existed before this PR. Still, I think this should be addressed at some point.

pabloarosado:
Thanks @lucasrodes, I agree. I've moved the google module to `etl` (and by the same logic also the notion module).

PR #4696, opened 2025-07-07
📊  education: proficiency by subject and level

veronikasamborska1994:


etl/steps/export/multidim/education/latest/proficiency.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # Adjust dimensions
+    tb = adjust_dimensions(tb)
+
+    # Create collection
+    c = paths.create_collection(
+        config=config,
+        tb=tb,
+        common_view_config=MULTIDIM_CONFIG,
+    )
```
lucasrodes, line 128: I note that you often use `adjust_dimensions`. To be clear this is totally fine, so no action required!

I'm just saying, in case it has been lost in translation. The preferred and initial idea with dimensions is that they are defined in garden by indexing the tables with appropriate columns (not only `country` and `year`).

I reckon sometimes is tricky, but I think it is generally nice to do there if possible.

veronikasamborska1994: yep, I agree but it'd be quite a pain to do this at the moment. I'll try to restructure in the next update but definitely agree with you

PR #4698, opened 2025-07-08
📊  trade: IMF dataset exploration

veronikasamborska1994:


etl/steps/data/meadow/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,46 @@
...
+    tb = snap.read()
+
+    #
+    # Process data.
+    #
+    # Drop unnecessary columns.
+    tb = tb.drop(columns=["DATASET", "SERIES_CODE", "OBS_MEASURE", "FREQUENCY", "SCALE"])
+
+    # Melt year columns (1948-2024) into rows
+    year_columns = [str(year) for year in range(1948, 2025)]
```
paarriagadap: Wouldn't it be better to extract the years from the Table instead? So you don't need to maintain this in the future.

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+    ds_garden = paths.create_dataset(tables=[tb], default_metadata=ds_meadow.metadata)
+    ds_garden.save()
+
+
+def _harmonize_countries(tb: Table) -> Table:
+    """Harmonize country names for both country and counterpart_country columns."""
+    tb = geo.harmonize_countries(
+        df=tb, countries_file=paths.country_mapping_path, excluded_countries_file=paths.excluded_countries_path
+    )
+    tb = geo.harmonize_countries(
```
paarriagadap, line 163: You can add warn_on_unknown_excluded_countries=False to avoid the warning in the terminal

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+        df=tb,
+        country_col="counterpart_country",
+        countries_file=paths.country_mapping_path,
+        excluded_countries_file=paths.excluded_countries_path,
```
paarriagadap, line 167: I think you should exclude these entities that appear as missing in the function's warning:
* Taiwan Province of China
* Western Hemisphere not specified
* Malaysia, West
* Countries & Areas not specified
* Europe not specified
* Africa not specified
* Middle East and Central Asia not specified
* Asia not specified
* Special Categories

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+    ds_regions = paths.load_dataset("regions")
+    ds_wdi = paths.load_dataset("wdi")
+    ds_population = paths.load_dataset("population")
+
+    # Read table from meadow dataset (LONG format with 'indicator' & 'value')
+    tb_long = ds_meadow.read("trade")
+
+    # --- Harmonize & keep only the two indicators we actually use ----------------
+    tb_long = _harmonize_countries(tb_long)
+    tb_long = tb_long[tb_long["indicator"].isin([EXPORT_COL, IMPORT_COL])].copy()
```
paarriagadap, line 70: Nit: I don't think you need the copy() here

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+        "Germany": 1990,
+        "Latvia": 1991,
+        "Aruba": 2010,
+        "Serbia": 2006,
+        "Montenegro": 2006,
+        "Curacao": 2010,  # dataset spelling varies; harmonization usually fixes diacritics
+        "Sint Maarten (Dutch part)": 2010,
+        "Czechia": 1993,
+        "Slovakia": 1993,
+    }
```
paarriagadap: You can maybe define these at the top of the code, so it's easier to track an update

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+        "East Germany": 1990,
+        "USSR": 1991,
+        "Yugoslavia": 1992,
+        "Czechoslovakia": 1993,
+        "Serbia and Montenegro": 2006,
+        "Netherlands Antilles": 2010,
+    }
+    exclude_members_of_historical_regions = {
+        "Germany": 1990,
+        "Latvia": 1991,
```
paarriagadap: Why not other post-USSR countries?

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+        index_columns=["country", "year", "indicator", "counterpart_country"],
+        country_col="country",
+        regions=REGIONS_OWID + ["World"],
+    )
+
+    # Create "Asia (excl. China)" region
+    asia_members = geo.list_members_of_region(region="Asia", ds_regions=ds_regions)
+    asia_minus_china_members = [c for c in asia_members if c != "China"]
+
+    tb = geo.add_region_aggregates(
```
paarriagadap, line 195: The function documentation recommends to use add_regions_to_table instead? Doesn't that function cover what you want?

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+            c
+            for c in ["total_exports", "total_imports", "total_trade_volume", "bilateral_trade_volume"]
+            if c in combined
+        ],
+        errors="ignore",
+    )
+    return combined
+
+
+def _load_gdp_data(ds_wdi: Dataset, ds_population: Dataset) -> Table:
```
paarriagadap: You could have used the GDP indicator, ny_gdp_mktp_cd. Is there any difference with your method here?

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+    return gdp[["country", "year", "gdp_total"]].copy()
+
+
+def add_china_imports_share_of_gdp(tb: Table, gdp_data: Table) -> Table:
+    """Add China imports as share of GDP (as %) for each country-year; attach to counterpart=World rows."""
+    china_imports = tb[(tb["counterpart_country"] == "China")][["country", "year", IMPORT_COL]].copy()
+    china_imports = china_imports.rename(columns={IMPORT_COL: "china_imports_value"})
+
+    china_gdp = pr.merge(gdp_data, china_imports, on=["country", "year"], how="inner")
+    china_gdp["china_imports_share_of_gdp"] = china_gdp["china_imports_value"] / china_gdp["gdp_total"] * 100
```
paarriagadap: Looks like it is the case, but I would double-check if the original data comes in current USD

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+    china_gdp = china_gdp[["country", "year", "china_imports_share_of_gdp"]].copy()
+    china_gdp["counterpart_country"] = "World"
```
paarriagadap, line 287: It is not clear enough for me why the counterpart country is World here

paarriagadap: I guess it's just a table constraint

etl/steps/data/garden/imf/2025-07-08/trade.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+          numDecimalPlaces: 0
+
+      exports_of_goods__free_on_board__fob__share:
+        title: Exports to << counterpart_country >> as a share of total exports
+        unit: "%"
+        short_unit: "%"
+        description_short: The percentage of a country’s total goods exports, excluding freight and insurance costs, that go to << counterpart_country >>.
+        description_key:
+          - This indicator shows how much of a country’s merchandise exports are sent to << counterpart_country >>, expressed as a percentage of country's total exports.
+          - Exports are valued on an FOB basis (Free on Board). This means the values include the cost of the goods up to the exporter’s border, but exclude transport and insurance costs after that point.
```
paarriagadap: Some general bullet points repeated across indicators I would define them above and call them from there. It's easier to maintain.

etl/steps/data/garden/imf/2025-07-08/trade.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+          - It highlights how central << counterpart_country >> is as a market for a country’s goods.
+          - A higher percentage means the country depends more heavily on that partner for its export earnings.
+          - This is particularly useful for spotting whether a country’s exports are diversified or concentrated.
+        description_processing: Shares were calculated by dividing the value of FOB exports to << counterpart_country >> by the total FOB exports to the World, and multiplying the result by 100.
```
paarriagadap: I think it has been suggested that we write the processing as "we calculated" instead of "x were calculated"

etl/steps/data/garden/imf/2025-07-08/trade.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+          - GDP figures were taken from the World Bank and calculated as GDP per capita multiplied by population, both expressed in current US dollars.
+          - Expressing imports as a share of GDP highlights the scale of economic dependence on Chinese goods relative to the overall size of the economy.
+          - A higher percentage indicates that Chinese imports play a larger role in the economy, while a lower percentage suggests less reliance on China as a supplier.
+        description_processing: Shares were calculated by dividing the value of CIF imports from China by GDP, and multiplying the result by 100. Import data were sourced directly from official trade statistics and valued in current US dollars. GDP was computed from World Bank indicators as (GDP per capita × population), both measured in current US dollars. This approach ensured consistency across countries and years. Calculations were performed for each country–year, but values were only displayed on rows where the counterpart was World, to prevent duplication.
```
paarriagadap: Not sure if the last point about duplication would be clear for readers. It's not that clear for me either: duplication of what?

etl/steps/data/garden/imf/2025-07-08/trade.py
```diff
@@ -0,0 +1,387 @@
...
+
+def get_country_import_ranking(tb: Table, target_country: str) -> Table:
+    """
+    Analyze a target country's ranking as an importer for each country.
+    Returns counterpart_country=<target_country> with an 'import_rank' column.
+    """
+    import_data = tb[tb["indicator"] == IMPORT_COL].copy()
+
+    results = []
+    for (country, year), group in import_data.groupby(["country", "year"], sort=False):
```
paarriagadap: Just wondering if this could have been simpler with a sort by group and then a count or something similar

etl/steps/data/garden/imf/2025-07-08/trade.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+          numDecimalPlaces: 0
+
+      top_partner_category:
+        title: Top import partner category
+        unit: ""
+        short_unit: ""
+        description_short: The country that was the largest source of imports in a given year, based on merchandise import values that include the cost of goods plus freight and insurance.
+        description_key:
+          - This indicator shows which country was the largest source of imports for each country–year, based on CIF (Cost, Insurance, and Freight) import values in current US dollars.
+          - CIF means the value includes not only the price of the goods themselves, but also the costs of shipping and insuring them up to the importing country’s border.
```
paarriagadap: Nit typo `'` instead of `´`

etl/steps/data/garden/imf/2025-07-08/trade.meta.yml
```diff
@@ -0,0 +1,161 @@
...
+          - This measure is useful for assessing the diversification of import sources and for spotting vulnerabilities in supply chains if reliance on a single partner is high.
+        description_processing: Shares were calculated by dividing the value of CIF imports from << counterpart_country >> by the total CIF imports from the World, and multiplying the result by 100. Calculations were performed for each country and year. Values were omitted when the counterpart was World, since numerator and denominator would otherwise be the same.
+        presentation:
+          title_public: Imports from << counterpart_country >> as a share of total imports
+        display:
+          name: << counterpart_country >>
+          numDecimalPlaces: 1
+
+      share_of_total_trade:
+        title: Trade with << counterpart_country >> as a share of total trade
```
paarriagadap: The titles and descriptions using << counterpart_country >> look akward in the case of Intraregional

etl/steps/data/garden/imf/2025-07-08/trade.py
veronikasamborska1994: yep, it was the easiest way though I could also probably keep it in a seperate table too which would avoid this issue

etl/steps/data/garden/imf/2025-07-08/trade.py
veronikasamborska1994: ah nice! I didn't see that we had this indicator

etl/steps/data/garden/imf/2025-07-08/trade.py
veronikasamborska1994: I think these are the only ones that were causing issues in this dataset.. but let me know if I am missing something

veronikasamborska1994:
thanks @paarriagadap ! I'll relay these suggestions regarding charts to the rest of us on the project. they are still work in progress and we don't know which ones we will end up using for sure yet but I think it will be easier for others if the dataset is on grapher rather than on a staging server

PR #4712, opened 2025-07-14
🔨 refactor chart types

sophiamersmann:
For line charts to switch to discrete bar charts when both handles are on the same time, we now need to explicitly include the `DiscreteBar` chart tab in `chartTypes`. If `chartTypes` isn't provided, it defaults to `["LineChart", "DiscreteBar"]`.

In this PR, I’ve added `DiscreteBar` whenever `chartTypes` is set to `LineChart`, but another option would be to just remove the chartTypes field altogether and rely on the default. Let me know which you’d prefer!

### DIFF
--- a/etl/steps/export/explorers/climate/latest/climate_change.config.yml
+++ b/etl/steps/export/explorers/climate/latest/climate_change.config.yml
@@ -81,7 +81,7 @@ views:
       yScaleToggle: false
       yAxisMin: auto
       baseColorScheme: owid-distinct
-      type: LineChart
+      type: LineChart DiscreteBar
       hasMapTab: false
   - dimensions:
       metric: sea_surface_temperature_anomaly
@@ -96,7 +96,7 @@ views:
       yScaleToggle: false
       yAxisMin: auto
       baseColorScheme: owid-distinct
-      type: LineChart
+      type: LineChart DiscreteBar
       hasMapTab: false
   - dimensions:
       metric: ocean_heat_content__top_700m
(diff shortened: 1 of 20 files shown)


etl/steps/export/explorers/climate/latest/climate_change.config.yml
```diff
@@ -150,7 +150,7 @@ views:
       yScaleToggle: false
       yAxisMin: auto
       baseColorScheme: owid-distinct
-      type: LineChart
+      type: LineChart DiscreteBar
```
lucasrodes, line 153: I suspect you are on top of this, but just double-checking: explorers require this field to be plain text with space-separated chart types?

lucasrodes (approved):
Seems good to me! I was only wondering about the schema for explorers (https://github.com/owid/etl/blob/b17b6c8f412d9b7fbc80a967424a4ac03a86f68f/schemas/explorer-schema.json#L432-L445). Is the `enum` valid? My guess is that it's not, bc it limits the explorer config to pick just one value from `enum`, and we know it can actually combine multiple (space-separated). 

It's not something that has changed from your PR, though. And also, the schema is not used to validate anything, but rather to assist us with highlighting in YAML files.

etl/steps/export/explorers/climate/latest/climate_change.config.yml
sophiamersmann: yes, that's the syntax :)

PR #4740, opened 2025-07-22
📊 happiness: Create additional happiness data

pabloarosado:
Add private snapshots and steps for some draft charts on self-reported life satisfaction.


dag/main.yml
```diff
@@ -745,6 +745,21 @@ steps:
   data://grapher/oecd/2025-07-03/affordable_housing:
     - data://garden/oecd/2025-07-03/affordable_housing
 
+  #
+  # Gallup World Poll: Life in Five Years and Life Today
```
antea04, line 749: Tiny thing: Is there a reason these are not grouped as a block like the other datasets?

snapshots/happiness/2025-07-22/life_in_five_years_and_life_today.py
```diff
@@ -0,0 +1,25 @@
+"""Script to create a snapshot of dataset."""
```
antea04, line 1: I would comment where the data file came from and how you accessed it, for documentation & replication purposes

etl/steps/data/garden/happiness/2025-07-22/life_in_five_years_and_life_today.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Read tables from meadow dataset.
+    tb_today = ds_meadow.read("life_satisfaction_today")
+    tb_in_5_years = ds_meadow.read("life_satisfaction_in_5_years")
+
+    #
+    # Process data.
+    #
+    # Combine both tables.
+    tb = pr.concat([tb_today.assign(**{"when": "today"}), tb_in_5_years.assign(**{"when": "in_5_years"})])
```
antea04, line 27: This is just curiosity on my part: is there a difference between df.assign({"new_col":"value"}) and df["new_col"]="value"?

etl/steps/data/garden/happiness/2025-07-22/life_in_five_years_and_life_today.meta.yml
```diff
@@ -0,0 +1,44 @@
...
+
+tables:
+  life_in_five_years_and_life_today:
+    variables:
+      satisfaction_in_5_years:
+        title: Predicted life satisfaction in 5 years
+        unit: ''
+        short_unit: ''
+        description_short: |-
+          People were asked the following question: "Please imagine a ladder with steps numbered from 0 at the bottom to 10 at the top. Suppose we say that the top of the ladder represents the best possible life for you, and the bottom of the ladder represents the worst possible life for you. Just your best guess, on which step do you think you will stand on in the future, say about five years from now?"
```
antea04, line 20: Totally up to you but: We have a cantril-ladder DoD that I prefer using in the description_short, rather than writing out the whole survey question again. Could look something like:

Average of survey responses. The survey respondents rank their life satisfaction as their current position on a hypothetical [ladder](#dod:cantril-ladder), where the best possible life for them is a 10, and the worst possible life is a 0.

antea04 (approved):
LGTM! Feel free to ignore the metadata comments I left if not useful - code looks and runs fine! Thanks Pablo!

dag/main.yml
pabloarosado: Not a very good one, rather a technical limitation: `StepUpdater` handles comments above a step in the DAG, but doesn't know how to group related steps. That's why I usually add a description to each step individually. But I'll remove these ones for consistency with other steps in this file.

snapshots/happiness/2025-07-22/life_in_five_years_and_life_today.py
pabloarosado: Absolutely!

etl/steps/data/garden/happiness/2025-07-22/life_in_five_years_and_life_today.py
pabloarosado: No, that would yield the same result (well, actually it should be `df.assign(**{...})`).

PR #4757, opened 2025-07-28
📊 Clinical trials

antea04:


etl/steps/data/garden/health/2025-07-28/clinical_trials.countries.json
```diff
@@ -0,0 +1,247 @@
...
+  "Togo": "Togo",
+  "Trinidad and Tobago": "Trinidad and Tobago",
+  "Tunisia": "Tunisia",
+  "Turkmenistan": "Turkmenistan",
+  "Uganda": "Uganda",
+  "Ukraine": "Ukraine",
+  "United Arab Emirates": "United Arab Emirates",
+  "United Kingdom": "United Kingdom",
+  "United States": "United States",
+  "United States Minor Outlying Islands": "United States Minor Outlying Islands",
```
spoonerf: I'd pop this one under United States too

etl/steps/data/garden/health/2025-07-28/clinical_trials.meta.yml
```diff
@@ -0,0 +1,275 @@
...
+        description_key:
+          - "{definitions.notes.completed_trials}"
+          - Country location is determined by the country with the most locations listed in the trial. If multiple countries have the same number of locations, the first one listed is used.
+          - The year is the year of the trial's completion date. This is first entered (as a planned completion data) when the trial is first registered, and then updated when the trial is completed.
+  sponsor_per_year:
+    variables:
+      n_studies_sponsor:
+        title: Number of clinical trials by sponsor
+        description_short: |
+          The annual number of clinical trials registered in the [ClinicalTrials.gov](#dod:clinical-trials-gov) database, by sponsor type. ClinicalTrials.gov has four categories for sponsors: Trials funded by the National Institutes of Health (US NIH), other US federal agencies (e.g. FDA or CDC), industry (such as pharmaceutical companies), and other sponsors. Other sponsors can include academic institutions, non-profit organizations and other people or entities.
```
spoonerf: This is quite US-centric, is it worth broadening it? One takeaway from some of the charts is that it's not clear if it's just for the US or includes other countries too, it would be good to make that clearer.

spoonerf: Additionally, do we know if the US is particularly well represented in this data because it is a US-based dataset? I don't have a good sense of how many clinical trials are missing, particularly from non-US countries. It seems like from the website it is encouraged for clinical trials to be recorded here, but it's not mandatory?

snapshots/health/2025-07-28/clinical_trials.py
```diff
@@ -0,0 +1,69 @@
...
+FORMAT = "csv"
+
+# Version for current snapshot dataset.
+SNAPSHOT_VERSION = Path(__file__).parent.name
+
+
+def fetch_full_clinical_trials_data(page_size: int) -> list:
+    chunks = []
+    next_page_token = None
+    pbar = tqdm(total=546474, desc="Fetching data", unit="entries")
```
spoonerf: Curious about this number, is there a way it couldn't be hardcoded? Or a way to set a warning if the actual number is larger next time you run this code?

spoonerf: Or if that's a real pain, I'd make it a universal variable at the top and add a note on what it means and how someone running this code again would know what number to use here.

etl/steps/data/garden/health/2025-07-28/clinical_trials.meta.yml
antea04: Yes, that's a really good point. The the US mandates that all FDA approved treatments have their trials documented in the clinicaltrials database. But for trials in other countries for treatments that do not seek FDA approval, this is voluntary. Many countries/ institutions still add their trials to the database (see also http://staging-site-data-clinical-trials/grapher/number-of-clinical-trials), but it is not required.

As Clinicaltrials is the most comprehensive database wrt clinical trials, it's hard to make a statement on how many trials are "missing" from the database. We try to make this clear in the ClinicalTrials DoD and I have adjusted the description of this particular chart a bit more. Do you have any suggestion for making this clear beyond that?

spoonerf:
Thanks for the rapid reply @antea04!

> You can see how many non-US trials are registered in this [chart](http://staging-site-data-clinical-trials/grapher/number-of-clinical-trials?time=earliest..2024&country=USA~CHN~GBR~DEU~BRA~TUR~CAN~FRA~EGY~ITA~ESP). The US mandates that each clinical trial needs to be registered and all treatments that eventually seek FDA approval will need to be registered, but registration for trials conducted in other countries than the US is voluntary if they don't apply for FDA approval.

I think it is important to acknowledge this somewhere (although there isn't a lot of space!). My takeaway from this chart is that the US is a real powerhouse when it comes to clinical trials, which might be true, but we also know the data collection is better in the US, so it could easily not be true as well. I would add something like: "Clinical trial registration is mandatory for trials in the United States, but voluntary for other countries." to the subtitle.

> Other countries are included in all the charts. Since the ClinicalTrials database is run and maintained by the US NIH, they categorize trials with labels that are useful for them, such as only specifically categorizing US government grants in the sponsor type. I adjusted the description for that specific chart. We also have a reference to non-US countries being included in the ClinicalTrials database in the DoD (which is included in every subtitle).
> 
> Is there another chart where you think this could be problematic or would you like me to make this clearer overall (e.g. with a footnote?)

I think you could make it clear in the subtitle by using 'Global'. I suggest this slightly shorter subtitle for that chart and I think you should use 'global' in the other charts where appropriate. 

"Global number of clinical trials registered in the ClinicalTrials.gov database, by sponsor type. ClinicalTrials.gov has four
sponsor categories: Trials funded by the National Institutes of Health (NIH), other US federal agencies (e.g., FDA), industry (e.g., pharmaceutical companies), and other sponsors (e.g., academic institutions, non-profits, and other people or entities)."


> That's a good question! The law mandates clinical trials to be registered since 1997, but since most of these chart are grouped by completion year, the first 3-6 years most of the increase is probably due to increased documentation. Since then I would assume both increased documention for non US trials and an increase in trials conducted is responsible for the increase. I could add that to the footnote, would that be helpful?

Yes I think a footnote acknowledging this would be helpful, even just briefly.

> I agree! I've reordered the charts where that wasn't the case!

Thanks!

> [Average study length by phase](http://staging-site-data-clinical-trials/grapher/average-study-length-by-phase) Here I disagree - I think it's interesting and important to see how long clinical trials take and therefore how much time passes after a drug has been developed until it is mass-market ready (and e.g. that funding cuts do to not only cut short developments in progress but will affect drug development years/ decades down the line). Since different trial phases study different aspects and take different amounts of time I think splitting it along phases makes a lot of sense.

Interesting, I think the total length of time is more interesting than broken down by phases, but if you would like to keep them maybe a stacked area chart would be a good option, as that would show the total as well. 

> [Average number of participants in clinical trials](http://staging-site-data-clinical-trials/admin/charts/8797/edit) I think this chart is interesting because it shows how many participants have to be recruited for a clinical trial to make valid statements about safety and efficacy of a drug, but I feel less strongly about this one. I can take it out if you don't think it adds much.

Yes I would still lean towards not including this one. 

Also one small last thing I noticed, we tend not to include 'The' at the start of subtitles and just jump straight in, where appropriate could you remove the 'The's.

Happy to discuss anything else further!

PR #4765, opened 2025-07-29
📊  teachers mdim

veronikasamborska1994:


lucasrodes (approved):
LGTM!

One question: Why do we need `wdi.override.meta.yml`? Can we integrate it into the existing `wdi.meta.yml`?

PR #4785, opened 2025-08-01
🐛 Fix bug in chart animation

pabloarosado:


apps/wizard/app_pages/chart_animation.py
```diff
@@ -146,8 +146,7 @@ def get_chart_details(slug):
     help="""Paste the URL of the chart you want to animate.
 
 **Considerations**:
-- Only production links work.
-- You can pass the full URL, or skip the https://ourworldindata.org/ bit.
+- You can pass the full URL of a chart in production, in a staging site, or simply pass the "grapher/chart-slug".
```
lucasrodes, line 149: @pabloarosado currently, if you provide only `grapher` or `/grapher` it will load the chart from PROD. I think this clarification was good to keep.

pabloarosado:
> @pabloarosado Mmh, my second point is easy to miss and forget. It is a relatively simple change, which solves a bug; it would have been nice to solve it in this PR.
> 
> I think that just changing the path where the images are stored to include some reference to the environment (e.g., `DOWNLOAD_DIR / staging_or_prod_identifier / chart_slug`) would have sufficed.

I think it's unlikely someone would be working on the same chart in prod and in staging at the same time. That folder is supposed to be a temporary one, so I think it's fine if things get overwritten. I don't mind either way.

apps/wizard/app_pages/chart_animation.py
pabloarosado: I rather think that the /grapher/slug option is actually not that useful (I think it's easier to copy a full URL than a partial one), and it introduces the problem of not knowing whether the images should be taken from prod or staging. Should we rather remove that option? Again, I don't mind either way.

lucasrodes:
@pabloarosado yup, it is unlikely... Yet _very_ easy to avoid! Surely quicker than the time we've spent discussing here 😄 Anyway, no big deal, just a suggestion.

pabloarosado:
> @pabloarosado yup, it is unlikely... Yet _very_ easy to avoid! Surely quicker than the time we've spent discussing here 😄 Anyway, no big deal, just a suggestion.

Pressing merge was even quicker :-P. But sorry for not addressing your suggestion.

apps/wizard/app_pages/chart_animation.py
lucasrodes: I don't mind, your intuition here is better bc you use this tool more often.

I'd only suggest to have the help message as "helpful" as possible. If we allow for `/grapher`-like text, let's mention in the help message how we would treat that. If we do not allow for such text, let's introduce a validation logic that checks that the input text is an actual URL.

PR #4807, opened 2025-08-05
📜  Data sanity check instructions for CLAUDE.md

spoonerf:
Hey @pabloarosado! 
These are the sanity check instructions I've been giving Claude the last couple of days and I think they are working well, if anything perhaps a bit over the top sometimes. 

No rush, but I'm curious to hear what you think. I also wonder if these might be something that make more sense to keep in our individual level Claude instructions, although it would be beneficial to have some standardisation across the team.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -221,3 +221,307 @@ print(f"Garden null values: {tb.date.isnull().sum()}")
 - **Add assertions**: Include data quality checks that fail fast with clear error messages
 - **Document data issues**: Log warnings about data quality problems found during processing
 - **Fix meadow steps**: Most data cleaning should happen in meadow, not garden steps
+
+## Individual Preferences
+- @~/.claude/instructions/etl.md
+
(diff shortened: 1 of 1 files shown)


Marigold:
This is really cool! I'm so glad we're all starting to contribute to the shared instructions in CLAUDE.md. Just two small suggestions (feel free to ignore them):
- How about putting validation instructions into its own .md file and [referencing them](https://docs.anthropic.com/en/docs/claude-code/memory?utm_source=chatgpt.com#claude-md-imports) from CLAUDE.md? (e.g. to `docs/claude-validation.md`)
- If generated validation code is too long, we could put it in a separate file `validation.py` or `[short_name]_validation.py` to keep the main step clean (and add that file to checksum to make sure any changes rerun the step)

spoonerf:
Thanks for the suggestions @Marigold, I will implement them both! I was thinking the validation code was quite long, so putting it in a separate file is a great idea.

PR #4835, opened 2025-08-11
📊 un world regions

lucasrodes:
### Main work
Fixes https://github.com/owid/owid-issues/issues/2088

- Add all UN M49 definitions: 3 levels, ranging from broader regions (e.g. Americas) to more detailed ("Southern Africa").
  - I've changed the naming of `defined_by` from "unsd" to "un_m49_x", where x is the level (1: broader, 3: more detailed.)
  - [M49 details](https://unstats.un.org/unsd/methodology/m49/)
- In Grapher, I've added more metadata on the indicators created (origins, description, etc.) and added a new indicator with a more generic UN definition.

### Bonus
- Move the hardcoded missing members of regions (e.g. Zanzibar) from owid-grapher [update.ts](https://github.com/owid/owid-grapher/blob/abc76ad76ecbc4d86bdbf7da7bc0ca458e17014b/devTools/regionsUpdater/update.ts#L21-L25) file to the [regions.yml](https://github.com/owid/etl/blob/master/etl/steps/data/garden/regions/2023-01-01/regions.yml) file of the ETL regions garden step.
- Added logic to recursively replace members that are aggregates with their respective members in `regions.yml`.

### DIFF
--- a/etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
+++ b/etl/steps/data/garden/faostat/2025-03-17/additional_variables.py
@@ -1343,7 +1343,7 @@ def run() -> None:
 
     # Load dataset about food balances and load its main (long-format) table.
     ds_fbsc = paths.load_dataset("faostat_fbsc")
-    tb_fbsc = ds_fbsc.read("faostat_fbsc")
+    tb_fbsc = ds_fbsc.read("faostat_fbsc", safe_types=False)
 
     # Load dataset about fertilizers by nutrient and load its main (long-format) table.
     ds_rfn = paths.load_dataset("faostat_rfn")


Marigold:
@lucasrodes that error when running `data://garden/who/latest/monkeypox` is not random. 

```
⏺ Found the root cause! The error occurs because the regions dataset has duplicate region names (like "Africa (UN)" appearing twice with different codes: "UNM49_AFR" and "UN_AFR"). When trying to create a
  mapping using .set_index("name").code, pandas throws an InvalidIndexError because the index is not unique.
```

I don't remember how regions work, but perhaps you need to have unique `name:` accross regions.

PR #4836, opened 2025-08-11
📜 Add GSheets documentation

spoonerf:


lucasrodes (commented):
Hey Fi, thanks for drafting this!

There were couple of wrong document links, I've fixed these.

I've also changed a bit the structure of `create-gsheets-omm.md`, to make it a bit easier on the eye.

I've also updated some other entries to reflect that they are no longer "new".

Should we limit the framing to OMMs? OMMs are definitely an example of where this function should be used, but we might want to use it in other places. Maybe we can remove "OMMs" from the entry title, and make a reference in the entry text to them?

spoonerf:
Thanks @lucasrodes, those are all good points, I think I've incorporated them all now. Let me know if you are happy to merge or if there are any other changes you think would be useful.

mkdocs.yml
```diff
@@ -217,7 +218,7 @@ nav:
           - MDIMs and Explorers: "guides/data-work/mdims.md"
           - Export data: "guides/data-work/export-data.md"
           - Access archived data: "guides/data-work/access-archived-data.md"
-          - GSheets for OMMs: "guides/data-work/create-gsheets-omm.md"
+          - Create a public GSheet for a variable: "guides/data-work/create-gsheet.md"
```
lucasrodes: Sorry to insist here, maybe this one is too long.

I'd use "indicator" instead of "variable", to be aligned with our language.

Maybe something like: "Publish indicator metadata with GSheet".

I find it also clarifies why we would want to use Gsheets at all.

spoonerf: Sure I set it as: 'Publish indicator and metadata with GSheet', still a bit long but I think it's important to note that we're publishing the indicator values as well as the metadata.

lucasrodes (approved):
Other than my comment above, it all looks good to me!

PR #4851, opened 2025-08-14
📊  Add SDG regions to regions.yml

spoonerf:
/schedule 2025-08-20


lucasrodes (approved):
LGTM. Just two comments:

- Is it better to use `(UNSDG)` over `(UN SDG)`? I feel like the latter might make it more obvious that it is UN's
- Shall we create the region indicator for UN SDG in the regions dataset and deprecate the one from ?

spoonerf:
Hey @lucasrodes, both good points, I've implemented them both now! I'll set it to merge overnight.

lucasrodes:
I added the `/schedule` command to the PR description and it seems to work.

That's what the [documentation of the action](https://github.com/marketplace/actions/merge-schedule) and [our docs](https://docs.owid.io/projects/etl/guides/pull-requests/?h=schedule#scheduling-a-pr-merge) suggest doing.

PR #4858, opened 2025-08-18
🤖 Autoupdate: Wildfires

Marigold:


veronikasamborska1994:
@Marigold, could you take a look at the updated snapshot code? The previous version lacked error handling for API failures, which seemed to cause some countries to be skipped in the last update. When I tested locally, the issues seemed related to sending too many requests in rapid succession. I've added timeouts and retries. It's working, but it’s slower, and I’m not sure it’s completely foolproof yet.

snapshots/climate/latest/weekly_wildfires.py
```diff
@@ -51,6 +57,87 @@
 }
 
 
+def fetch_with_curl(url: str) -> dict:
```
Marigold, line 60: @veronikasamborska1994 do we need this fallback? When I ran it, it never used it (usually it succeeds with 2 attempts max)

snapshots/climate/latest/weekly_wildfires.py
```diff
@@ -126,25 +213,33 @@ def main(upload: bool) -> None:
 
                 # Append the processed DataFrame to the list of fire DataFrames.
                 dfs_fires.append(df_all)
+            except Exception as e:
```
Marigold: I wouldn't add this except. We want to know if it stops working (even with retries in place), this just masks potential problems.

snapshots/climate/latest/weekly_wildfires.py
veronikasamborska1994: happened to me sometimes but we could increase the number of attempts and maybe delete this?

snapshots/climate/latest/weekly_wildfires.py
veronikasamborska1994: good point! I've removed it now

snapshots/climate/latest/weekly_wildfires.py
Marigold: If it happened to you, then let's keep it. Feel free to merge this.

PR #4880, opened 2025-08-19
📊 agriculture: Alternative way to calculate food prices in PPP

pabloarosado:


etl/steps/data/garden/wb/2025-08-04/food_prices_for_nutrition.py
```diff
@@ -41,19 +44,129 @@ def adapt_units(tb: Table, tb_wdi: Table) -> Table:
...
+    # https://en.wikipedia.org/wiki/Sierra_Leonean_leone
+    # As of 1 July 2022, the ISO 4217 code is SLE due to a redenomination of the old leone (SLL) at a rate of SLL 1000 to SLE 1.
+    # Correct those costs by dividing by 1000.
+    for column in tb.columns:
+        if column.startswith("cost_of_") and ("in_local_currency_unit" in column):
+            tb.loc[(tb["country"].isin(["Sierra Leone"])), column] /= 1000
+
+    # For Liberia, the resulting cost of a healthy diet is also much higher than other countries (over 500 PPP$).
+    # In this case, it seems to me that the PPP conversion factor (from https://data.worldbank.org/indicator/PA.NUS.PRVT.PP?locations=LR ) may not be given in LCU per international-$, but rather in USD per international-$.
+    # To correct for this, convert those USD to LCU using a conversion factor from another WDI indicator (https://data.worldbank.org/indicator/PA.NUS.FCRF?locations=LR).
```
paarriagadap: Yeah, the case of Liberia is weird. US$ is legal currency there, and seems strange that the PPP factor uses another currency compared to the market exchange rate. Note that you wouldn't adjust for living costs for that country.

It would be good to have more info about this case.

etl/steps/data/garden/wb/2025-08-04/food_prices_for_nutrition.py
```diff
@@ -41,19 +44,129 @@ def adapt_units(tb: Table, tb_wdi: Table) -> Table:
...
+        assert set(check[check["pct"] > 10]["country"]) == {"Palestine"}
+
+    # The cost of a healthy diet, however, is given for multiple years, and it seems that the cost in "ppp_dollars" corresponds to **current*** PPP$ (not constant PPP$).
+    # To confirm this, convert costs in local currency units into PPP$.
+    check = tb.copy()
+    # Drop the old PPP conversion factors for 2021, and add PPP conversion factors for all years.
+    check = check.drop(columns=["pa_nus_prvt_pp"]).merge(
+        tb_wdi[["country", "year", "pa_nus_prvt_pp"]].dropna(), on=["country", "year"], how="left"
+    )
+    check["cost_of_a_healthy_diet_in_current_ppp_dollars_owid"] = (
```
paarriagadap: The right approach would be to divide by the value of the PPP factor in 2021. Using all the factors is not quite the same, because it includes relative inflation between countries and the US.

But I don't quite understand. Wasn't this what you wanted to do right before the sanity checks?

PR #4896, opened 2025-08-21
🎉 Indicator upgrader CLI

Marigold:
## Summary

Implements https://github.com/owid/etl/issues/4909

## Changes

- Deprecate `etl variable-translate-mapping` (it was for translating JSON files with mappings which we no longer use)
- Rename `etl variable-match` to `etl indicator-upgrade match` and save indicator mapping to `wiz__variable_mapping` instead of JSON file
- Add `etl indicator-upgrade upgrade` and `etl indicator-upgrade undo`
- Replace some `st.cache_data` by `functools.cache` to get rid of annoying streamlit warnings
- Submit chart updates in parallel to speed it up

## Usage

1. Run `etl indicator-upgrade match --old-dataset-id 123 --new-dataset-id 456` to create `wiz__variable_mapping` table in MySQL
2. Run `etl indicator-upgrade upgrade` to remap indicators in charts
3. (If you're not satisfied, use `etl indicator-upgrade undo` to roll it back)

You'd typically want to upgrade perfect matches first with
```
etl indicator-upgrade match --old-dataset-id 123 --new-dataset-id 456 --no-interactive
etl indicator-upgrade upgrade
```
then open Indicator Upgrader in Wizard to map the rest.

I've also added a button to Indicator Upgrader that maps perfect matches and then lets you map the rest.

## How to test

I prepared a new version of cherry blossom dataset on branch `update-cherry-blossom`.

1. Run `STAGING=update-cherry-blossom streamlit run apps/wizard/app_pages/indicator_upgrade/app.py` to open indicator upgrader
2. Upgrade cherry blossom dataset (ids 7202 and 7051)
3. Click on `Upgrade Perfectly Matching indicators`
4. Click on Next and map the remaining indicators

Then undo and try it with CLI

1. `STAGING=update-cherry-blossom etl indicator-upgrade undo`
2. `STAGING=update-cherry-blossom etl indicator-upgrade match --old-dataset-id 7051 --new-dataset-id 7202`
3. `STAGING=update-cherry-blossom etl indicator-upgrade upgrade`

### DIFF
--- a/apps/indicator_upgrade/cli.py
+++ b/apps/indicator_upgrade/cli.py
@@ -160,11 +160,24 @@ def undo_command(dry_run: bool) -> None:
 
     log.info(f"Found {len(indicator_mapping)} variable mappings to undo")
 
-    # 2. Invert the mapping (swap old and new IDs to reverse the upgrade)
+    # 2. Check for duplicate new indicator IDs (many-to-one mappings)
+    new_ids = list(indicator_mapping.values())
+    duplicate_new_ids = [id for id in set(new_ids) if new_ids.count(id) > 1]
+
+    if duplicate_new_ids:
+        log.error("Cannot undo upgrade: Multiple old indicators were mapped to the same new indicator(s)")
+        log.error(f"Duplicate new indicator IDs: {duplicate_new_ids}")
+        log.error("This creates ambiguity - we cannot determine which old indicator to restore")
+        log.error("Original mapping:")
+        for old_id, new_id in indicator_mapping.items():
+            log.error(f"  {old_id} -> {new_id}")
+        return
+
+    # 3. Invert the mapping (swap old and new IDs to reverse the upgrade)
     mapping_inverted = {v: k for k, v in indicator_mapping.items()}
     log.info(f"Inverted mapping: {mapping_inverted}")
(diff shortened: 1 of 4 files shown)


pabloarosado (approved):
Fantastic! I followed your instructions and it worked very well. A few minor issues and nice to haves:
- I thought it would have been good to be a bit clearer in the wizard, after you have done the perfect match mapping, on whether some indicators need to be manually mapped (and if so, how many). Right now it seems to be optional to click on next step.
- When I did the manual mapping on wizard, for "Day of the year with peak cherry blossom" I saw two options. However, when I run it on the CLI, I only see one option (and the "Other options" area is empty). This seems to be a minor bug (either there are no options, in which case we shouldn't print "Other options", or the other available option is somehow missing).
- Another minor thing is that, when I type a random option "asdfasdf" and press enter, the title changes to "VARIABLE 2/1", then 3/1, etc. which doesn't make sense.

(See screenshot on the last two points)
Screenshot 2025-09-08 at 15 23 25

None of these things are that important, the main functionality works and it's very useful, so, if these things are trivial, feel free to fix them, otherwise ignore and ship. Thanks a lot!

PR #4900, opened 2025-08-21
🎉 Improve tooling around regions dataset

pabloarosado:


etl/data_helpers/geo.py
```diff
@@ -837,7 +849,10 @@ def add_gdp_to_table(
     return tb_with_gdp
 
 
-def create_table_of_regions_and_subregions(ds_regions: Dataset, subregion_type: str = "members") -> Table:
+@cache
+def create_table_of_regions_and_subregions(
```
Marigold, line 853: @pabloarosado I enabled runtime cache for this function, it was called multiple times with the same params and caching it reduced runtime of
```
etlr data://garden/animal_welfare/2025-01-23/number_of_farmed_crustaceans --force --only
```
by about a second

etl/steps/data/garden/irena/2025-07-18/renewable_energy_statistics.py
```diff
@@ -354,7 +347,7 @@ def run() -> None:
     tb = tb.dropna(subset="capacity").reset_index(drop=True)
 
     # Harmonize country names.
-    tb = geo.harmonize_countries(df=tb, countries_file=paths.country_mapping_path)
+    tb = paths.regions.harmonize_names(tb=tb)
```
Marigold, line 350: We should update the template `apps/wizard/etl_steps/cookiecutter/garden/{{cookiecutter.namespace}}/{{cookiecutter.version}}/{{cookiecutter.short_name}}.py` as well

pabloarosado: Good idea, but maybe let's wait a bit on that, try it secretly, and then decide whether this should be the default.

PR #4912, opened 2025-08-22
📊 energy: Update IRENA renewable generation costs

pabloarosado:
## Summary

* Update IRENA's data from the Renewable Power Generation Costs report.
  * Meadow step was adapted because the format of the spreadsheet changed.
* Given that they dropped data prior to 2010, combine it with the previous dataset.
  * To do that, we use >=2010 data from the latest report, and <2010 data from the previous report.
  * ~We convert 2023 constant US$ to 2024 constant US$ using a factor of 1.03 (using [the US Inflation Calculator](https://www.usinflationcalculator.com/)).~
* Archive unused steps (but keep the old IRENA Renewable Power Generation Costs steps that are still used).

## Review

There's no need for a thorough review. But I would just like to confirm that the conversion from constant 2023 USD to constant 2024 USD (in `etl/steps/data/garden/irena/2025-08-22/renewable_power_generation_costs.py`) in LCOE costs makes sense.
UPDATE: In the end we decided to convert to 2024 USD using the US GDP deflator from WDI.

### DIFF
--- a/etl/steps/data/garden/energy/2025-08-22/photovoltaic_cost_and_capacity.meta.yml
+++ b/etl/steps/data/garden/energy/2025-08-22/photovoltaic_cost_and_capacity.meta.yml
@@ -17,17 +17,26 @@ tables:
         title: Solar photovoltaic module price
         description_short: |-
           This data is expressed in US dollars per watt, adjusted for inflation.
+        unit: "constant {LATEST_YEAR} US$ per watt"
+        short_unit: "$/W"
+        display:
+          name: Solar PV module cost
+          shortUnit: "$"
         presentation:
           title_public: Solar photovoltaic module price
           grapher_config:
             note: "Data is expressed in constant {LATEST_YEAR} US$ per watt."
+        description_processing: |-
+          - Photovoltaic cost data between 1975 and 2003 has been taken from Nemet (2009), between 2004 and 2009 from Farmer & Lafond (2016), and since 2010 from IRENA.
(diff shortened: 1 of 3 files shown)


paarriagadap (approved):
Hi! Looks good, but it would be always good to know which was the original deflator used. It is not completely clear with IRENA, and I assume it's even less clear in the case of the older papers.

It could be the case that the original adjustment for inflation was done with CPI or any other deflator, so rebasing with GDP deflator (or any other indicator different from the original) would not necessarily reflect the right adjustment: you wouldn't "remove" the previous base year before applying the new year

pabloarosado:
Thanks @paarriagadap I couldn't find any mention of how the original [Farmer & Lafond (2016)](https://www.sciencedirect.com/science/article/pii/S0048733315001699) or [Nemet (2009)](https://www.sciencedirect.com/science/article/abs/pii/S0301421508005910) papers deflated the data. The only relevant mention (in the former) is the following:

> The DNA sequencing data is from [...] and corrected for inflation using the US GDP deflator.

They don't mention how any other indicator was deflated, but at least this is how they deflated some of the data. The different in the correction factors in my two approaches (US CPI vs US GDP) are smaller than 5% (see the numbers in [this commit](https://github.com/owid/etl/pull/4912/commits/823072bd71ab77cc3ade5b5a9d28f27d52b22be9) or [this commit](https://github.com/owid/etl/pull/4912/commits/377116217d57b8a360eaa176a436349973c85be9)), so in practice, I don't think it matters. Or are you suggesting an alternative approach to what I've tried so far? Thanks!

PR #4918, opened 2025-08-25
✨ Check regions file for duplicates

Marigold:
- Add "Middle East, North Africa, Afghanistan and Pakistan (WB)" to regions
- Check for duplicates

### DIFF
--- a/dag/demography.yml
+++ b/dag/demography.yml
@@ -344,9 +344,13 @@ steps:
     - data://garden/demography/2024-12-06/wittgenstein_human_capital
 
   #
-  # TODO: add step name (just something recognizable)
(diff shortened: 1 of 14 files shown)


paarriagadap:
@Marigold Ah, OK, I don't see that many scripts using it, and I would like to force the update there. But I guess it makes sense to keep it for now because of that.

Having said that, the regions grapher step still fails because of having both regions:
```
AssertionError: Ordinal variable `wb_region` has extra values that are not defined in field `sort`: {'Middle East and North Africa (WB)'}
```

If you prefer to keep the old region in garden, maybe we need to drop it in grapher?

Issue #4919, opened 2025-08-25
Wizard Step Upgrader displays no table (probably after a Streamlit update)


pabloarosado:
Wizard Step Upgrader displays no table. I noticed this for the first time last week, and I suppose it may be related to a Streamlit update.

<img width="997" height="986" alt="Image" src="https://github.com/user-attachments/assets/f5022301-f0f8-43f8-b037-579550b79f57" />

For now the workaround for me is to use the step updater CLI.

lucasrodes:
When looking into the terminal while running `etlwiz` locally I see the following error:

```
2025-08-28 15:10:08.504 Pandas DataFrame hash failed. Falling back to pickling the object.
Traceback (most recent call last):
  File "/home/lucas/repos/etl/.venv/lib64/python3.12/site-packages/streamlit/runtime/caching/hashing.py", line 452, in _to_bytes
    values_hash_bytes = self.to_bytes(hash_pandas_object(df_obj))
```

lucasrodes:
I think it is actually a bug in `streamlit-aggrid` in version 1.1.18.

<img width="805" height="369" alt="Image" src="https://github.com/user-attachments/assets/e5f0c269-0753-4b1e-badb-153388fbc838" />

PR #4932, opened 2025-08-28
🐛 fix step upgrader

lucasrodes:
Fixes https://github.com/owid/etl/issues/4919 by downgrading `streamlit-aggrid` version from `1.1.8` → `1.1.7`.


pabloarosado (approved):
It seems to work as expected now. Thanks for looking into it, catching the issue, and fixing it so quickly!

PR #4935, opened 2025-08-29
📊 Create auxiliary namespace with owid-deflator dataset

pabloarosado:


snapshots/wb/2025-08-29/wdi_gdp.zip.dvc
```diff
@@ -0,0 +1,23 @@
...
+    title_snapshot: World Development Indicators - GDP deflator
+    description: |-
+      The World Development Indicators (WDI) is the primary World Bank collection of development indicators, compiled from officially-recognized international sources. It presents the most current and accurate global development data available, and includes national, regional and global estimates.
+    description_snapshot: |-
+      This data includes specifically one indicator from WDI, namely "GDP deflator (base year varies by country)" (with code NY.GDP.DEFL.ZS).
+    citation_full: |-
+      World Bank national accounts data, and OECD National Accounts data files. World Development Indicators - World Bank (2025). Indicator [NY.GDP.DEFL.ZS](https://data.worldbank.org/indicator/NY.GDP.DEFL.ZS).
+    version_producer: "122"
+    url_main: https://datacatalog.worldbank.org/search/dataset/0037712/World-Development-Indicators
+    url_download: https://api.worldbank.org/v2/en/indicator/NY.GDP.DEFL.ZS?downloadformat=csv
```
paarriagadap, line 14: I would prefer to pick this one NY.GDP.DEFL.ZS.AD, https://data.worldbank.org/indicator/NY.GDP.DEFL.ZS.AD

From the source:
```
This series has been linked to produce a consistent time series to counteract breaks in series over time due to changes in base years, source data and methodologies.
```

etl/steps/data/garden/wb/2025-08-29/wdi_prices.meta.yml
```diff
@@ -0,0 +1,52 @@
...
+
+          # Long Definition:
+          Index of the prices of consumption goods and services, as compared to a certain reference period (2010=100).
+
+          # Statistical Concept and Methodology:
+          Consumer Prices Indices are compiled in accordance with international standards: Consumer Price Index Manual, 2020 or 2004 version. Specific information on how countries compile their CPI statistics can be found on the IMF website: dsbb.imf.org Statistical concept(s): The conceptual basis of a consumer price index series is to measure the rate at which prices of consumption goods and services are changing from one period to another.
+
+      gdp:
```
paarriagadap: I would name this gdp_deflator instead, for clarity

etl/steps/data/meadow/wb/2025-08-29/wdi_prices.py
```diff
@@ -0,0 +1,65 @@
...
+    # Drop unnecessary columns and rename columns conveniently.
+    tb_cpi = (
+        tb_cpi.drop(columns=["Country Code", "Indicator Name", "Indicator Code"], errors="raise")
+        .rename(columns={"Country Name": "country"}, errors="raise")
+        .melt(id_vars=["country"], var_name="year", value_name="cpi")
+    )
+    tb_gdp = (
+        tb_gdp.drop(columns=["Country Code", "Indicator Name", "Indicator Code"], errors="raise")
+        .rename(columns={"Country Name": "country"}, errors="raise")
+        .melt(id_vars=["country"], var_name="year", value_name="gdp")
```
paarriagadap: I would prefer to name the indicator gdp_deflator, for clarity

etl/steps/data/garden/wb/2025-08-29/wdi_prices.meta.yml
```diff
@@ -0,0 +1,52 @@
...
+        title: GDP deflator
+        unit: ""
+        short_unit: ""
+        description_short: |-
+          Ratio of GDP in current local currency to GDP in constant local currency. The base year varies by country.
+        description_from_producer: |-
+          The GDP implicit deflator is the ratio of GDP in current local currency to GDP in constant local currency. The base year varies by country. This indicator is expressed as a ratio (a÷b).
```
paarriagadap, line 41: As I mentioned, I would prefer the linked series. Maybe we can add it alongside this one, since the non-linked covers a longer time period (though should be used with caution)

paarriagadap (commented):
Hey @pabloarosado!

Looks good. I am not completely sure if this is the best solution, but I agree it's easier to update and more practical for inflation adjustments.

My main comment would be related to upload the linked GDP deflator series, since there is processing from the Bank to have a cleaner series.

It only starts on 1990, though, so it might be convenient to have both.

We can think about including other indicators in the future.

Regarding the income groups, it could be an option, though we already update it outside the WDI, so it wouldn't be much of a problem.

paarriagadap (approved):
Looks good to me!

PR #4940, opened 2025-09-01
🔨 Use owid-public bucket in R2 instead of DigitalOcean

Marigold:


etl/steps/export/github/co2_data/latest/owid_co2.py
```diff
@@ -45,7 +45,7 @@ def prepare_readme(tb: Table) -> str:
 
 ## The complete *Our World in Data* CO2 and Greenhouse Gas Emissions dataset
 
-### 🗂️ Download our complete CO2 and Greenhouse Gas Emissions dataset : [CSV](https://nyc3.digitaloceanspaces.com/owid-public/data/co2/owid-co2-data.csv) | [XLSX](https://nyc3.digitaloceanspaces.com/owid-public/data/co2/owid-co2-data.xlsx) | [JSON](https://nyc3.digitaloceanspaces.com/owid-public/data/co2/owid-co2-data.json)
```
pabloarosado, line 48: Would it be possible to redirect the old link to the new one? If there's no easy way, no big deal. Otherwise, we should delete the old file in digital ocean (to avoid people from using the old data).

etl/steps/export/s3/co2_data/latest/owid_co2.py
```diff
@@ -118,16 +114,24 @@ def run(dest_dir: str) -> None:
...
             tqdm.write(f"Uploading file {local_file} to S3 bucket {S3_BUCKET_NAME} as {s3_file}.")
-            # Upload and make public each of the files.
-            s3.upload_to_s3(
-                local_path=str(local_file),
-                s3_path=f"s3://{S3_BUCKET_NAME}/{str(s3_file)}",
-                public=True,
-            )
+
+            # Read file content and determine content type
+            with open(local_file, "rb") as f:
```
pabloarosado: Wouldn't it be worth it adapting the `upload_to_s3` function? Up to you.

etl/steps/export/github/co2_data/latest/owid_co2.py
Marigold: No, unfortunately now. We have a couple options:
1. Keep updating both
2. Delete the old file
3. Replace the old file by a text file with an explanation that the file has been moved

etl/steps/export/s3/co2_data/latest/owid_co2.py
Marigold: It already exists as `s3_utils.upload`. I used it instead of low-level call.

PR #4942, opened 2025-09-01
📊  Maternal mortality update

spoonerf:


spoonerf:
`BTW, it doesn't fail for me @spoonerf. Try deleting data/garden/un/2025-08-29/maternal_mortality in case it's something related to a corrupted step
`

Yeh it's working for me locally, but the staging server isn't. What do you mean by delete that step?

paarriagadap:
Ah, nevermind then. (I meant deleting the data generated by the step, not the step itself: `etl/data` instead of the `etl/steps` folder)

Marigold:
> This staging keeps failing on the grapher step saying there is a primary key missing. But I've checked and as far as I can see the primary key is correctly set as ['country', 'year']. Any idea what could be going on?

Sorry, this is annoying. The reason you don't see the error locally is that staging has an extra `OWID_STRICT=1` env variable.

The problem is in file `etl/steps/data/grapher/un/2025-08-29/maternal_mortality.py` (it's version 2025-08-29!!). Instead of
```
tb = ds_garden.read("maternal_mortality")
```
call
```
tb = ds_garden.read("maternal_mortality", reset_index=False)
```

I ran into the same issue recently and [created a PR](https://github.com/owid/etl/pull/4873), but haven't made any progress yet. I'll prioritise it to make it less confusing next time.

Marigold:
EDIT: So I checked and OWID_STRICT is enabled by default for all datasets after 2023-06-25. So perhaps the confusion came from the fact that the error was from version `2025-08-29`? (Not sure how it made into ETL in the first place)

PR #4949, opened 2025-09-02
📝 Methodology for "does the news reflect what we die from"

antea04:


pabloarosado (approved):
Hey Tuna, thanks for doing this, it looks great! I think your proposal makes sense, it looks neat and convenient. If we decide to move things around in the future, that's also possible, but the current setup make sense.

PR #4952, opened 2025-09-02
📊 energy: Fix inflation adjustment in IRENAs data

pabloarosado:
Use each country's GDP deflator (linked series) to adjust IRENA's renewable power generation costs (to convert data in constant 2023 USD to constant 2024 USD).

* [x] Fix description from producer (it is from WDI). Consider removing WDI's attribution (given that it's used as an auxiliary dataset), so it doesn't appear in the charts as one of the sources.

### DIFF
--- a/etl/steps/data/garden/irena/2025-08-22/renewable_power_generation_costs.py
+++ b/etl/steps/data/garden/irena/2025-08-22/renewable_power_generation_costs.py
@@ -32,6 +32,9 @@ def adjust_old_data_for_inflation(tb_old, tb_deflator, tb):
         "World",
         "South Korea",
     }
+    # Remove South Korea from the old data, since it's the only country for which we can't adjust prices to 2024 USD.
+    tb_old = tb_old[tb_old["country"] != "South Korea"].reset_index(drop=True)
+
     # Calculate the adjustment factor.
     tb_deflator["adjustment"] = tb_deflator[2024] / tb_deflator[2023]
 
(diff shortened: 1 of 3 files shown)


etl/steps/data/garden/irena/2025-08-22/renewable_power_generation_costs.py
```diff
@@ -8,19 +8,53 @@
...
+        "adjustment"
+    ].item()
+
+    # Column by column in the old table, convert to constant USD of the latest year.
+    for column in tb_old.drop(columns=["country", "year", "adjustment"]).columns:
+        # Sanity check.
+        error = f"Unexpected units for column {column}."
+        assert tb[column].metadata.unit == f"constant {LATEST_YEAR} US$ per kilowatt-hour", error
+        # Fill adjustment factor with 1 for countries for which we can't do the adjustment (namely South Korea).
+        tb_old[column] *= tb_old["adjustment"].fillna(1)
```
paarriagadap: I am not sure if it is the best way to proceed here. The South Korean series will end up with a mix of 2023 and 2024 constant prices, which is hard to explain. Is it too bad to just drop the old series for them?

If you really want to proceed, you need to address that in `description_processing`

paarriagadap (commented):
Hi! The approach is good. I am only not sure about what to do with South Korea. Personally, I would drop if I can't adjust. If you want to keep your method you should explain it in `description_processing`

etl/steps/data/garden/irena/2025-08-22/renewable_power_generation_costs.py
pabloarosado: Good point, I've just removed South Korea from the old table (as a result, we lose a few data points for onshore wind for South Korea).

PR #4966, opened 2025-09-04
📊 Medicine Biotech patents

antea04:
Data (and 2 charts) on granted patents in medicine/ biotech from the WIPO. See also: https://github.com/owid/owid-issues/issues/2075

### DIFF
--- a/etl/steps/data/garden/research_development/2025-10-27/patents.meta.yml
+++ b/etl/steps/data/garden/research_development/2025-10-27/patents.meta.yml
@@ -22,6 +22,8 @@ definitions:
         entityType: filing office
         entityTypePlural: filing offices
         tab: chart
+        xAxis:
+          label: Year granted
     display:
       numDecimalPlaces: 0
       entityAnnotationsMap: |-
@@ -44,7 +46,7 @@ tables:
     variables:
       all_patents:
         title: Life science patents granted, by filing office
-        description_short: Patents granted within the fields of medical technology, biotechnology, and pharmaceuticals by filing office. These patents cover a wide range of innovations, including new drugs, medical devices, diagnostic methods, genetic engineering techniques, and biotechnological processes.
+        description_short: Patents granted within the fields of medical technology, biotechnology, and pharmaceuticals by filing office. These patents cover a wide range of innovations, such as medical devices, diagnostic methods, genetic engineering techniques, biotechnological processes and new drugs.
         unit: patents
         short_unit: ""
       medical_technology_patents:


snapshots/research_development/2025-10-27/historical_patents.py
veronikasamborska1994, line 1: are you using this data for anything? looks like it's in the dag for the meadow step but isn't actually being used

snapshots/research_development/2025-10-27/historical_patents.csv.dvc
```diff
@@ -0,0 +1,36 @@
...
+
+    # Citation
+    producer: World Intellectual Property Organization (WIPO)
+    citation_full: |-
+      World Intellectual Property Organization, Historical Patents, https://www.wipo.int/en/web/ip-statistics, accessed 27. October, 2025
+    attribution_short: WIPO
+
+    # Files
+    url_main: https://www.wipo.int/en/web/ip-statistics
+    #url_download: https://www.wipo.int/documents/2948119/3215563/wipo_ip_historical_data.zip
```
veronikasamborska1994, line 26: I am assuming you'll be using this dataset at some point? otherwise not sure why it's here and why the link is uncommented

etl/steps/data/meadow/research_development/2025-10-27/patents.py
```diff
@@ -0,0 +1,39 @@
...
+    # Retrieve snapshot.
+    snap = paths.load_snapshot("patents.csv")
+
+    # Load data from snapshot and adjust columns:
+    tb = snap.read(header=3)
+
+    temp_metadata = tb["Office"].metadata.copy()
+    temp_cols = tb.columns.copy()
+
+    tb = tb.drop(columns=["2023"]).reset_index()
```
veronikasamborska1994, line 22: not sure why you are dropping the 2023 column and why you need to reset the metadata; maybe add a few comments for yourself in the future for why you had to do this

etl/steps/data/garden/research_development/2025-10-27/patents.py
```diff
@@ -0,0 +1,60 @@
...
+    tb = tb.rename(
+        columns={
+            "13 - Medical technology": "medical_technology_patents",
+            "15 - Biotechnology": "biotechnology_patents",
+            "16 - Pharmaceuticals": "pharmaceutical_patents",
+        }
+    )
+
+    tb["all_patents"] = (
+        tb["medical_technology_patents"].fillna(0)
```
veronikasamborska1994, line 38: do you know that NaNs actually mean 0 here? if so then completely fine; maybe just add a comment on where it's documented

etl/steps/data/garden/research_development/2025-10-27/patents.meta.yml
```diff
@@ -0,0 +1,67 @@
...
+      biotechnology_patents:
+        title: Biotechnology patents granted by filing office
+        description_short: Number of biotechnology patents granted by the filing office in a given year. Examples for biotechnology patents could be genetic engineering methods to improve agriculture or novel processes using microorganisms to produce bio-fuel.
+        unit: patents
+        short_unit: ""
+      pharmaceutical_patents:
+        title: Pharmaceutical patents granted by filing office
+        description_short: Number of pharmaceutical patents granted by the filing office in a given year. Pharmaceutical patents typically cover new drugs, biological treatments or vaccines.
+        unit: patents
+        short_unit: ""
```
veronikasamborska1994, line 66: just FYI I think you could just not write short_unit if there is not one :) I think it doesn't throw an error so just less boring stuff to fill out

etl/steps/data/garden/research_development/2025-10-27/patents.meta.yml
```diff
@@ -0,0 +1,67 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  patents_general: &patents_general |-
+    Patents are a form of intellectual property that give their holders exclusive rights to an invention for a limited period, typically 20 years from the filing date.
```
veronikasamborska1994, line 4: I wonder if we should have a DoD for patents that you have here so we can reuse it for other charts :) but again not v important

etl/steps/data/garden/research_development/2025-10-27/patents.py
antea04: It's a good point! They show NaNs as 0 in their own visualizations (e.g. here: https://www3.wipo.int/ipstats/ips-search/search-result?type=IPS&selectedTab=patent&indicator=24&reportType=11&fromYear=1980&toYear=2023&ipsOffSelValues=ZW&ipsOriSelValues=&ipsTechSelValues=13,16,15 - you have to click to line charts), but it's not explicitly written down anywhere. I've reached out to them, but I think it's fine to go with this in the meantime.

etl/steps/data/garden/research_development/2025-10-27/patents.meta.yml
antea04: Oh that's good to know! Thanks!

etl/steps/data/garden/research_development/2025-10-27/patents.meta.yml
antea04: That's a great point! I'll add one!

snapshots/research_development/2025-10-27/historical_patents.csv.dvc
antea04: Yes - I am not using it for this dataset, but I want to import the whole data at some point and this is in preparation for this. I'll add a comment!

PR #4967, opened 2025-09-04
🐝 Living language tweaks

spoonerf:


veronikasamborska1994:
Hi @spoonerf! This looks good — I guess what I found a bit frustrating is that when looking at the actual website you link to, I could only see an example for one country (I picked Ukraine), and then it asked for a subscription. I wish there were some way to explore this within our data, but I guess that might not be possible. It’s just that when you see such a big number of languages for some countries, you immediately wonder what these are, and our chart doesn’t let us explore it. Also, I was wondering how many of these are unique to a country (not a product of immigration)— e.g. in Ukraine there are some very unique ones like Crimean Tatar, but Yiddish is also listed, which is obviously not specific to Ukraine. It's pretty cool, but I wouldn't know if I didn't look at the linked website. 

One other thing I’ve noticed is that sign languages are included — I’m not sure how much these differ between countries, but it could be interesting to highlight or clarify that aspect.

Anyway, it's mostly just rambling and I just thought it was a v cool chart! thanks for working on this!

veronikasamborska1994 (approved):

spoonerf:
I figured out why they include non-indigenous languages and added it to the metadata. I removed the link to the site, because of the subscription issue you mention. 

It would be really interesting to be able to list the languages in each country, but I'm not sure how we could do it, unfortunately. Could be a cool non-grapher/non-OWID data viz project.

PR #4969, opened 2025-09-04
🔨 update dataset-schema with new colorScale settings

marcelgerber:
The schema sections are directly copied over from https://files.ourworldindata.org/schemas/grapher-schema.009.json.


marcelgerber:
@Marigold yeah, that's the closest equivalent. It won't be doing the exact same thing, but will also result in giving these charts similarly nice brackets.

Marigold (approved):
Looks good! I bumped the schema in a few other places.

Issue #3971, opened 2025-02-13
Preferred way to cite WDI and PIP indicators


pabloarosado:
WDI people contacted us and suggested a very specific way to cite their data (which seems to also apply to PIP). This is what they proposed:

[...] we wanted to give you some feedback on how we would like our WDI and PIP data to be cited. Please see our suggestions below.
 
Short citation (please include):
1. Name of database like World Development Indicators or Poverty and Inequality Platform
2. Name of organization: World bank
3. Indicator codes: this is optional if hyperlink included specifically to the indicator
4. Hyperlink(s) to the indicator page on the world bank data site ([data.worldbank.org](http://data.worldbank.org/)) or PIP platform. Please include links to all indicators if using more than one indicator.
 
Long citation (please include):
1. Name of database like World Development Indicators or Poverty and Inequality Platform
2. Name of organization: World bank
3. Indicator codes
4. Date of data retrieval
5. Source of data for WDI (not applicable for PIP). By this, we mean, if the indicator is "population using internet" which comes from ITU, you should also mention ITU.
6. Hyperlink(s) to the indicator page on the world bank data site ([data.worldbank.org](http://data.worldbank.org/)) or PIP platform. Please include links to all indicators if using more than one indicator.

It's great that they defined very specific guidelines. We need to figure out the best way to adjust our current WDI (and PIP?) pipeline, with our current metadata schema, to their requirements.

lucasrodes:
I wonder how these recommendations affect the chart source footer.

In there, we are limited with space, so ideally it shouldn't be very long.

Also, their suggestion seems to point to something similar to what we had in the past (cite WB but not original source in the footer). 

My understanding is that we were moving away from that, though this is revisable ofc.

My preference is for adding original source with a reference to WB, e.g.:

```

International Telecommunication Union (2025), via World Bank
```

Marigold:
I started looking into this a bit. I'll post a couple of examples here to have something concrete. You may find the [dummy data page](https://admin.owid.io/admin/datapage-preview/810767) useful for seeing where origin's attributes end up.

WDI metadata contains "Source". It can be a bit long sometimes, and we are currently shortening it to fit it into footer. Some examples of their sources (full list is [here](https://github.com/owid/etl/blob/master/etl/steps/data/garden/worldbank_wdi/2025-01-24/wdi.sources.json)):
- `Food and Agriculture Organization, electronic files and web site.`
[...]

PR #4980, opened 2025-09-08
📊 Gender Stats

veronikasamborska1994:


etl/steps/data/garden/wb/2025-09-08/gender_statistics.py
```diff
@@ -0,0 +1,311 @@
...
+                # Column not found in metadata - set defaults
+                meta.title = column
+                meta.unit = ""
+                meta.short_unit = ""
+                meta.description_from_producer = f"World Bank indicator: {column}"
+
+    return tb
+
+
+def clean_source_field(source: str) -> str:
```
Marigold, line 187: This is an interesting function. I was doing something very similar for WDI from WB, they actually gave us quite [specific instructions](https://github.com/owid/etl/issues/3971) how to cite them. I ended up [using AI](https://github.com/owid/etl/blob/master/etl/steps/data/garden/worldbank_wdi/2025-01-24/update_metadata.ipynb) to clean source names (by giving it a lot of good examples). It did a decent job, but I'd be happy to reuse your function in the future.

veronikasamborska1994: nice! I didn't think it was as crucial to go this far, but your solution seems pretty cool too!

PR #4995, opened 2025-09-11
📊 animal_welfare: Add survey data on animal welfare issues

pabloarosado:
Create snapshot and data steps for a chart on the opinions of Brits on common farming practices.

For more context, see [this issue](https://github.com/owid/owid-issues/issues/2128).

NOTE: I'm waiting for Max's review, it's still unclear if we may end up using it or not, so there's no need for a thorough review.

### DIFF
--- a/etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
+++ b/etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
@@ -3,7 +3,7 @@ definitions:
   common:
     unit: ""
     short_unit: ""
-    processing_level: major
+    processing_level: minor
     presentation:
       topic_tags:
         - Animal Welfare
--- a/etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
+++ b/etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
@@ -1,0 +1,41 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    unit: ""
+    short_unit: ""
+    processing_level: major
+    presentation:
+      topic_tags:
+        - Animal Welfare
+
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 365
+  non_redistributable: true
+
+
+tables:
+  acceptability_of_uk_farming_practices:
+    variables:
+      acceptable:
+        title: Respondents who consider a farming practice acceptable
+        display:
+          name: Acceptable
+          numDecimalPlaces: 0
+        presentation:
+          title_public: Respondents who consider a farming practice acceptable
+      not_acceptable:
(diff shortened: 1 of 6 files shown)


etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
```diff
@@ -0,0 +1,41 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    unit: ""
+    short_unit: ""
```
veronikasamborska1994, line 5: is unit not % here?

etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
```diff
@@ -0,0 +1,41 @@
...
+
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
+dataset:
+  update_period_days: 365
+  non_redistributable: true
+
+
+tables:
+  acceptability_of_uk_farming_practices:
```
veronikasamborska1994, line 19: it sounds like this won't be a part of the data catalogue so I guess the metadata isn't important in this case, but if it does end up being you should probably provide a bit more detail than there is currently (esp since it looks like you've done a lot of the processing yourself)

etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.py
```diff
@@ -0,0 +1,91 @@
...
+        tb.groupby(["question", "answer"], as_index=False)
+        .agg({"user": "count"})
+        .rename(columns={"user": ""})
+        .pivot(index="question", columns="answer", join_column_levels_with="")
+        .set_index("question")
+    )
+    # For some reason, indicators' metadata is not propagated, copy it from the original table.
+    for column in tb_counts.columns:
+        tb_counts[column] = tb_counts[column].copy_metadata(tb["answer"])
+        tb_counts[column].metadata.unit = ""
```
veronikasamborska1994: do you need to do this here? since you have these in the yaml file anyway

etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
pabloarosado: The unit is actual respondents (but the chart shows relative numbers by default).

etl/steps/data/garden/survey/2025-09-12/acceptability_of_uk_farming_practices.meta.yml
pabloarosado: Yes, I can add more metadata on the processing if needed. The only relevant processing that affects the data (apart from fixing some spurious names) is removing some entries that were unreasonably quick to be completed.

pabloarosado:
Thanks @veronikasamborska1994! I've added the year to the subtitle (and made the chart a draft for now, before I get Max's feedback).

PR #5000, opened 2025-09-12
📊  UN WPP data for Cologne

spoonerf:
Hey @Marigold, in Cologne, my group is working on some visualisations using the UN WPP data. I wanted to make some different slices of it easily accessible, and thought using the `external` channel would be a good way to do it. Is that okay? We can always remove it after Cologne, or move the pipeline into the ETL if we end up using the outputs.


Marigold (approved):
Sure, go for it! I almost forgot that there's such thing as external :).

And congrats to #5000!!!

PR #5009, opened 2025-09-15
✨ Improve Expert performance

Marigold:
## Summary

- Our Datasette DuckDB plugin doesn't support `DATE(...)` functions and returns **zero rows** (it's probably some workaround to make it work). LLM got confused quite often by this, so I added it to instructions and returning error if we want to use it (`DATE ...` syntax works)
- Comment tool `validate_datasette_query` - LLM was doing extra call for each query, now we do validation as part of `get_data_from_datasette` and return `"error": ` field if query is invalid (surprisingly, almost all queries are valid)
- Comment tools `generate_url_to_datasette` - even though this wasn't as slow, it was an unnecessary tool call. I'm returning Datasette URL in every response
- Deprecate Deep research tools in MCP, no one was using them and it confused LLM
- Agent now works on staging servers (using owid MCP specific to that staging server)

I pulled all user queries from Logfire into this [spreadsheet](https://docs.google.com/spreadsheets/d/1kboUO55PirEA0NV2QNEXdW7nkljj9PC07-OLhq-en8E/edit?gid=0#gid=0) and evaluated them. It works fairly well, though it's a bit slow.

@lucasrodes (not urgent) when you get back, check if my changes are ok. If not, we can revert them.


lucasrodes:
@Marigold Thanks for improving performance!

PR #5025, opened 2025-09-17
📊 energy: Update Lazard LCOE data

pabloarosado:
## Summary

Updated Lazard Levelized Cost of Energy from the LCOE+ (2025) report.

## Review
@paarriagadap there were no significant changes in the code with respect to the previous version, so there's no need for a review.

The only thing that would be good to confirm is the method to correct for inflation (in the garden step). I used the new OWID Deflator dataset, but had to add the extra point for the first half of 2025 using the Gross Domestic Product: Implicit Price Deflator from the Federal Reserve Economic Data ([GDPDEF](https://fred.stlouisfed.org/series/GDPDEF)). Please let me know if you agree with this, thanks!

### DIFF
--- a/etl/steps/data/garden/lazard/2025-09-17/levelized_cost_of_energy.py
+++ b/etl/steps/data/garden/lazard/2025-09-17/levelized_cost_of_energy.py
@@ -11,6 +11,17 @@ paths = PathFinder(__file__)
 BASE_DOLLAR_YEAR_ORIGINAL = 2017
 # Define the base year for the deflator.
 BASE_DOLLAR_YEAR = 2024
+# We use the WDI GDP deflator: linked series, but this is usually missing the latest year.
+# We manually add the missing value here.
+# To get this value, go to:
+# https://fred.stlouisfed.org/series/GDPDEF
+# click on "Edit Graph",
+# set "Modify frequency" to "Annual" if possible, otherwise "Semiannual" (if there's no annual data yet for the last year),
+# aggregation method "Average",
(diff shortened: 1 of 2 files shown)


etl/steps/data/garden/lazard/2025-09-17/levelized_cost_of_energy.py
```diff
@@ -0,0 +1,108 @@
...
+# We manually add the missing value here.
+# To get this value, go to:
+# https://fred.stlouisfed.org/series/GDPDEF
+# click on "Edit Graph",
+# set "Modify frequency" to "Annual" if possible, otherwise "Semiannual" (if there's no annual data yet for the last year),
+# aggregation method "Average",
+# then below, in "Units", select from the dropdown "Index (Scale value to 100 for chosen date)",
+# in the "or" field below, add the base year of the original WDI deflator (which, in the last update, was 2017).
+# NOTE: I used 2017-07-01, unsure if it should be some other day in the year, but the retrieved numbers for previous years coincided well with the given WDI values.
+DEFLATOR_MISSING_VALUES = {2025: 127.1}
```
paarriagadap, line 24: I see the value would be 127.7 using a semiannual frequency and an average aggregation method

pabloarosado: I think the exact value depends on the date you choose in the "Select a date that will equal 100 for your custom index" field. I set it to 2017-07-01 (to be fair, a bit arbitrarily, if you have a good reason to choose a different one, let me know).

PR #5027, opened 2025-09-17
💄 Continue work on labor data (ILOSTAT + WDI + SDG)

paarriagadap:


paarriagadap:
Hi @Marigold!

Sorry, it's me with the WDI again. I have modified the metadata for multiple indicators in the override yaml in garden, linked parameters to the ETL and I can't see any of the changes on the server. Do you know what is this?

It's not working either for local updates and changes after commits.

paarriagadap:
@Marigold Ups, you are right. And no, I didn't have an underline in this case (though I have it for some other cases). Which extension do you think I am missing?

PR #5036, opened 2025-09-19
📊 Exclude Pakistan and Afghanistan from South Asia (WB) region

Marigold:


paarriagadap (approved):
Thanks! I added small modifications in the chart: the color of the new region, the label, and removed the footnote.

We also need to change the reference in [the regions article](https://ourworldindata.org/world-region-map-definitions). I can do it after you merge.

PR #5060, opened 2025-09-25
📊 plastic waste: UN comtrade db

veronikasamborska1994:


Marigold (approved):
Looks good! The only suspicious thing is Indonesia importing 100% [plastic waste by air](http://staging-site-un-comtrade-plastic/admin/charts/8870/edit), but maybe it's right?

PR #5124, opened 2025-10-14
📜 Improve data analysis documentation and its capabilities

pabloarosado:
## Summary
Improve the way we present technical analyses.

NOTE: Initially, this PR was to create a technical document about long-term agricultural indicators, but given how much this one has grown, I'll leave that for a separate PR.

## Review
* @antea04 could you please have a look at the changes I made on the media deaths technical doc?
  * Reviewing notebooks is painful on github. It may be easier from VSCode (with [the github PR extension](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github)), or simply comparing [production](https://docs.owid.io/projects/etl/analyses/) | [staging](http://staging-site-docs-agriculture-indicators/etl/docs/analyses/). You can also have a look locally (run `mkdocs serve` and then go to [the local mkdocs page](http://127.0.0.1:8000/projects/etl/analyses/)).
* @lucasrodes what do you think about the other general changes in the docs? [Here you can access the docs on staging](http://staging-site-docs-agriculture-indicators/etl/docs/). 

Thanks both!

## Main changes
* Allow for expanding/collapsing code blocks in notebooks and .py files shown in Read the docs.
  * Currently, I think this works well for notebooks. Blocks are shown by default, but can be hidden by adding the tag `collapsed` to the code block.
  * However, for .py files, code blocks are defined based on empty lines. Maybe it's better if code blocks are detected with a symbol, e.g. `# %%`. We may decide to change this behavior in the future.
* Add link to Colab.
* Allow for displaying interactive plotly charts.
  * NOTE: After multiple failed attempts (with different plugins, and using different plotly renderers), the only thing that seems to work is disabling the `privacy` plugin. For this documentation site, I understand that this is perfectly fine. But just for transparency, I included a privacy page:
* Add a privacy page, and rearrange the home page a bit, to show (1) the footer with the privacy page, and (2) the public analyses page.
* Some visual improvements on the media deaths analysis notebook, so that:
  * Colab installs the necessary libraries and imports the necessary adjacent modules.
  * Charts are now interactive in the notebook.
  * Output pngs are created by plotly instead of matplotlib (which I think are a bit clearer).
* Created a `sitecustomize.py` file to declare common non-secret environment variables (instead of having to manually define them in `.env`).
* [x] TODO: Replace `docs-agriculture-indicators-doc` -> `master` before merging.


mkdocs.yml
```diff
@@ -9,8 +9,8 @@ site_description: >-
 repo_name: owid/etl
 repo_url: https://github.com/owid/etl
 
-# Copyright
-copyright: Copyright &copy; 2024 Our World in Data
+# License
+copyright: '<a href="https://creativecommons.org/licenses/by/4.0/">CC BY 4.0</a> • <a href="/projects/etl/privacy/">Privacy</a> • <a href="https://ourworldindata.org/">Our World in Data</a>'
```
lucasrodes: the link to Privacy is not working for me!

mkdocs.yml
```diff
@@ -176,33 +177,32 @@ extra_css:
...
-            Find all the details about OWID's data management tool.
+      description: |-
+            This documentation explains our data workflow and tooling — both to help OWID team members work effectively and to give external readers insight into how we manage and publish our data.
+            You can also find selected public analyses related to our work.
       # Gradient Color
       # gradient_hsla: hsla(173, 100%, 24%, 1) # Default Green: hsla(160deg,47%,55%,1)
 
       buttons:
-            - name: Get started
+            - name: Data workflow
```
lucasrodes: I would keep the "Get started" label here; I find it more accurate. When I first saw "Data workflow" I thought you had changed the link to the [Design Principles](http://staging-site-docs-agriculture-indicators/etl/docs/architecture/) entry

mkdocs.yml
```diff
@@ -176,33 +177,32 @@ extra_css:
...
-            - name: Guides
+            # - name: Learn more
+            #   url: architecture
+            #   theme: md-button
+            - name: Internal guides
               url: guides
               theme: md-button
-              # blurb: This psycho navigates to an interesting table of stuff located in "/Folder3/index.md"
-              # icon: image
+            - name: Public analyses
```
lucasrodes: I don't like "Public analyses", I think it's a confusing term. I'd rather use something like "Technical publications", to clearly highlight what these are: publications on technical matters. I don't think we need to use the word "public", as everything on our docs page is indeed public.

docs/analyses/index.md
```diff
@@ -1,9 +1,9 @@
...
+# Public analyses
 
 This section contains a selection of public analyses we want to share. This can include the methodology, replication code, and supplementary data or figures.
 
 ## 📰 Does the news reflect what we die from?
 
 This is the analysis for the data used in the article "[Does the news reflect what we die from?](https://ourworldindata.org/does-the-news-reflect-what-we-die-from)". We compare actual causes of death in the United States in 2023 to the media coverage they receive.
 
-[Article](https://ourworldindata.org/does-the-news-reflect-what-we-die-from) | [Methodology](media_deaths/methodology.md) | [Code notebook](media_deaths/media_deaths_analysis.ipynb) | [Download folder](https://catalog.owid.io/analyses/media-deaths-analysis.zip)
+[Article](https://ourworldindata.org/does-the-news-reflect-what-we-die-from) | [Methodology](media_deaths/methodology.md) | [View notebook](media_deaths/media_deaths_analysis.ipynb) | [Run notebook](https://colab.research.google.com/github/owid/etl/blob/docs-agriculture-indicators-doc/docs/analyses/media_deaths/media_deaths_analysis.ipynb) | [Download folder](https://catalog.owid.io/analyses/media-deaths-analysis.zip)
```
lucasrodes: I find this section slightly confusing and with too many CTAs

From what I see, this page aims at listing all technical publications in our docs. If so, I'd put the link to the technical publication more central to each block. Currently, it is the second link under the label "Methodology". I'd directly link to the technical text from the title. 

I would also get rid of the emoji in the title, as the page could rapidly get overwhelming if there are various reports there.

For the link section, I'd make it easier on the eye. Maybe you can use [material icons](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/)? This will make it faster to scan what's available.

The article at OWID is referenced twice: From the description and from the link section. I'd remove the one from the description and just leave it in the link section. Would then rephrase a bit the description, maybe no need to mention it supports an article bc that's implicit? Maybe we can make it more explicit with the page-level description as suggested above.

docs/analyses/index.md
```diff
@@ -1,9 +1,9 @@
-# Analyses
+# Public analyses
 
 This section contains a selection of public analyses we want to share. This can include the methodology, replication code, and supplementary data or figures.
```
lucasrodes: Maybe good to be a bit clearer on what this page contains

> In this page, we publish technical reports on data projects at Our World in Data. These publications include methodology, replication code, and supplementary data or figures. The publications shared here may support live articles on our site, you can think of them as "technical appendices" of research articles.

(just drafted this quickly)

docs/analyses/index.md
lucasrodes: Another thing that I fear is the number of links that we list here. This is overwhelming and the user will likely not click on most of them. I'd try to reduce them to the critical ones, and maybe move other to other places.

From [my comment above](https://github.com/owid/etl/pull/5124/files#r2451630770), I'd remove the methodology one (and have it linked from the title).

Then, I don't think we need a link for "viewing notebook" and another for "running notebook". Why do we need both? I'd prioritize one. If we still want to link to the other, we could then link to it either from the technical publication page itself, or from one of the notebooks (e.g. "view notebook" links to "run notebook"). But I def wouldn't have both, I actually got confused and missed that we had two options.

lucasrodes (commented):
Thanks for working on the docs, Pablo! I think you did a great job here.

I'm leaving some comments.

One general thought I had, is if we need to make our technical publications so central? Currently, we only have one post... And I am not sure what will be the growth here. I'd maybe put them central once we get to 5 or so. But fine if you want to keep them there.

docs/analyses/index.md
pabloarosado: I've removed the "Article", since it's mentioned in the text above the links.
I'm not sure we will always have a .md and a .ipynb file (my gut feeling is that just having a .ipynb may be better sometimes), so I'm not sure if having the link on the title is a good solution.
Now that it's just 3 links, I think it's fine to keep "View notebook" and "Run notebook". But we also add the Colab link inside the .ipynb. We can refine this later.

mkdocs.yml
pabloarosado: OK, I've reverted to "Get started", but I don't think that's very descriptive. We can rethink it in the future.

docs/analyses/index.md
lucasrodes: > I'm not sure we will always have a .md and a .ipynb file (my gut feeling is that just having a .ipynb may be better sometimes), so I'm not sure if having the link on the title is a good solution.

Mmmh, I see. I wonder if it is still good to have the title link to the "main publication", whatever it is. But sure, I'm just cautious of making it clear where the main CTA is. At the moment it is the link to our own OWID article.

> Now that it's just 3 links, I think it's fine to keep "View notebook" and "Run notebook". But we also add the Colab link inside the .ipynb. We can refine this later.
On "view notebook" vs "run notebook", I still don't fully grasp the motivation of having two identical files, where one is interactive and the other is not. I think this isn't very clear to the user. I'd favour just having one.

---
Assuming you keep the article link within the text, and that you also keep the link to the technical report in the link section, I'd have 3 links: Report; Notebook (just one); Download.

PR #5126, opened 2025-10-15
📊 agriculture: Fetch new yield data

pabloarosado:
## Summary

Add data related to crop productivity in Southern African countries, and other small ETL additions and fixes.

NOTE: Hannah is preparing an article, so it's not yet clear which indicators will be more useful. I just quickly created a few drafts to start playing with. We may improve the metadata and add DoDs if we end up using this data. For more details, see https://github.com/owid/owid-issues/issues/2158.

## Main changes
* Add recent paper data for Southern Africa corn yields from FAO vs crop productivity from satellite data.
* Add USDA data on corn yields (for all countries).
* Add `read_custom` method to `Snapshot`.
* Add a few aliases to the `regions` dataset.
* Fix bug in the way `PathFinder` was instantiating `Regions`, that was ignoring excluded countries file by default.
* Add "Northern Hemisphere" and "Southern Hemisphere" to the regions dataset.


etl/snapshot.py
```diff
@@ -357,6 +357,17 @@ def read_parquet(self, *args, **kwargs) -> Table:
             self.path, *args, metadata=self.to_table_metadata(), origin=self.metadata.origin, **kwargs
         )
 
+    def read_custom(self, read_function: Callable, *args, **kwargs) -> Table:
+        """Read data file using a custom reader function, and return a Table with metadata."""
```
lucasrodes: I'd add a short docstring with the input parameter `read_function`.

dag/agriculture.yml
```diff
@@ -196,3 +196,33 @@ steps:
...
+  #
+  # USDA/FAS - Corn yields
+  #
+  data://garden/agriculture/2025-10-16/corn_yields:
+    - data://meadow/agriculture/2025-10-16/corn_yields
+  #
+  # USDA/FAS - Corn yields
+  #
+  data://grapher/agriculture/2025-10-16/corn_yields:
+    - data://garden/agriculture/2025-10-16/corn_yields
```
lucasrodes, line 229: let's group these together under the same comment





```suggestion
  #
  # USDA/FAS - Corn yields
  #
  data://meadow/agriculture/2025-10-16/corn_yields:
    - snapshot://agriculture/2025-10-16/corn_yields.xls
  data://garden/agriculture/2025-10-16/corn_yields:
    - data://meadow/agriculture/2025-10-16/corn_yields
  data://grapher/agriculture/2025-10-16/corn_yields:
    - data://garden/agriculture/2025-10-16/corn_yields
```

dag/agriculture.yml
```diff
@@ -196,3 +196,33 @@ steps:
...
+  #
+  # Lobell Lab: Crop productivity in southern Africa
+  #
+  data://garden/agriculture/2025-10-15/crop_yields_in_southern_africa:
+    - data://meadow/agriculture/2025-10-15/crop_yields_in_southern_africa
+  #
+  # Lobell Lab: Crop productivity in southern Africa
+  #
+  data://grapher/agriculture/2025-10-15/crop_yields_in_southern_africa:
+    - data://garden/agriculture/2025-10-15/crop_yields_in_southern_africa
```
lucasrodes, line 213: let's group these together under the same comment


```suggestion
  #
  # Lobell Lab: Crop productivity in southern Africa
  #
  data://meadow/agriculture/2025-10-15/crop_yields_in_southern_africa:
    - snapshot://agriculture/2025-10-15/crop_yields_in_southern_africa.zip
  data://garden/agriculture/2025-10-15/crop_yields_in_southern_africa:
    - data://meadow/agriculture/2025-10-15/crop_yields_in_southern_africa
  data://grapher/agriculture/2025-10-15/crop_yields_in_southern_africa:
    - data://garden/agriculture/2025-10-15/crop_yields_in_southern_africa
```

etl/steps/data/garden/agriculture/2025-10-15/crop_yields_in_southern_africa.meta.yml
```diff
@@ -0,0 +1,60 @@
...
+        presentation:
+          title_public: "{tables.crop_yields_in_southern_africa.variables.yield_sif.title}"
+      yield_fao_adapted:
+        title: Corn yields (FAO), aligned for comparison with satellite data
+        display:
+          name: FAO
+        presentation:
+          title_public: "{tables.crop_yields_in_southern_africa.variables.yield_fao_adapted.title}"
+        description_processing: |-
+          - The data has been shifted one year earlier to align harvest-year data with satellite observations. Although not explicitly mentioned in the paper, this seems to be the approach followed to produce the series of Fig. 1.
```
lucasrodes: (just being curious here)
Why you prefer to render description_processing as a bullet list (even if there is one item)?

lucasrodes (commented):
lgtm! minor comments

dag/agriculture.yml
pabloarosado: The grouping makes sense, but I prefer to have a description for each step individually. The initial motivation for this criterion is that step upgrader can't group similar steps (it's actually not trivial to do this grouping, except for isolated datasets like this one); so, programmatically, we can handle the text right above a given step, but not grouped steps.

etl/steps/data/garden/agriculture/2025-10-15/crop_yields_in_southern_africa.meta.yml
pabloarosado: Good point, this is a field that makes sense as bullet points. I thought that (1) our metadata propagation would combine description processing by concatenating strings, that (2) grapher was smart enough to make single-bullet lists become single texts (as it happens with `key_description`), and that (3) writing it as bullet points is what we suggested in our guidelines. But none of those assumptions are true. So I've changed it (and it's good to know!).

dag/agriculture.yml
lucasrodes: mmh, i'm not sure i understand the limitation. is it that step upgrader won't be able to process the steps correctly if grouped under the same comment? Isn't that what we are already doing with most steps?

pabloarosado: The limitation is just about how we write to the DAG (so, nothing important about processing). We have functions that can read the comment right above a step, and write a comment right above a step. But figuring out which comment to fetch when steps are grouped, and how to group similar steps, are difficult tasks.

pabloarosado: If we want these tasks to happen automatically, it's safer to have a human-readable label on top of each step.

lucasrodes: mmh, right. I think I lack context here, but I suspect that it wouldn't be hard to condense and have our tools address the relevant comment. E.g. function `get_commnet(step_uri)` could have logic "get first comment above step uri, before there is an empty linebreak".

Otherwise, I think we are unnecessarily polluting the YAML files with comments that don't provide much information IMO and are repetitive.

If we want to keep like this in the short-term and apply to all steps (I'm not a fan), I'd suggest having these comments (i) shorter and (ii) more concrete. 
(i): Can we use one row instead of three rows?
(ii): Maybe worth adding some information on the step type? Otherwise, the comments are just a copy from other steps? I find this unhelpful.

lucasrodes:
@pabloarosado On NH/SH, I think it is great that you are adding this information! I'd suggest mentioning it to Ed, and maybe add a section to our article https://ourworldindata.org/world-region-map-definitions to clearly present what we consider to be NH and what SH (probably by means of a [chart like the one you drafted](http://staging-site-data-fetch-yield/admin/charts/8891/edit)). I guess it's mostly agreed on, but I'd still be transparent here, otherwise it is unclear "who is defining the groups.

I can imagine someone wondering how we classified countries that have the equator line pass through them. Are we categorizing based on the share of area? share of population? I suspect the former, but it's still good to clarify.

dag/agriculture.yml
pabloarosado: We could improve our functions to read and write yaml files. But my experience is that manipulating yaml files gets complex (or buggy) very quickly. At the time I tried things like ruamel, and it wasn't able to handle comments properly. So, it may be possible to make those functions smarter, but it may take some time to get it right.

The safest option, although possibly a bit redundant, is to have a comment per step. It's a good compromise for a DAG that is often read and written by both machine and humans.

We could certainly use one-row comments, and clean up the DAG a bit. I find the current layout a bit nicer to the human eye, but I don't have a strong opinion.

lucasrodes: I still don't quite understand the motivation or appeal of this solution tbh. As in, I don't know how often we tweak these comments programmatically. But I am probably missing bits here and context, so never mind. Let's have this discussion later on.

I would suggest you bring this up during one data meeting, as this is the first time that I am observing this organization... Maybe others are unaware of this?

pabloarosado: We should aim to have a DAG that is easy to read and write for both humans and processes. There are three ways to write to DAG files:
1) A human manually adds steps: we usually group similar steps in some meaningful way, and add a single comment on top.
2) Wizard creates data steps: it usually adds all steps in one go with a placeholder comment on top (that a human needs to fill out manually).
3) Step upgrader: it reads an existing step with its comment above, and (without any human input) then writes the same step elsewhere. However, the order in which new steps are written in the DAG may not be the same as the order in which the old steps were originally written. The result is, therefore, that new steps are not necessarily grouped in the same way. We could try to make this process smarter, or use AI to group steps in some meanigful way, but we should be careful adding too much complexity when writing to the DAG (errors can be painful to handle).

PR #5148, opened 2025-10-22
📊 ai: update Epoch dataset on model costs

veronikasamborska1994:


etl/steps/data/garden/artificial_intelligence/2025-10-22/epoch_compute_cost.py
```diff
@@ -0,0 +1,55 @@
...
+    # Calculate days since 1949-01-01.
+    tb["days_since_1949"] = (tb["publication_date"] - pd.to_datetime("1949-01-01")).dt.days
+    tb["days_since_1949"] = tb["days_since_1949"].astype(int)
+
+    # Validate index columns have no NaN values.
+    assert not tb[["model", "days_since_1949"]].isnull().any().any(), "Index columns should not have NaN values"
+
+    # Standardize domain names.
+    tb["domain"] = tb["domain"].astype(str)
+    replacements = {
```
antea04: You probably want to replace other domains which nicer/ more readable labels as well! See also current chart https://ourworldindata.org/grapher/hardware-and-energy-cost-to-train-notable-ai-systems

etl/steps/data/garden/artificial_intelligence/2025-10-22/epoch_compute_cost.py
```diff
@@ -0,0 +1,55 @@
...
+        "Language,Image generation": "Language and image generation",
+    }
+    for old_value, new_value in replacements.items():
```
antea04: you could probably do this with:

```suggestion
    tb["domain"] = tb["domain"].replace(replacements)
```

veronikasamborska1994:
thanks for picking up on this @antea04! I've reduced the number of categories but will leave the colours as that doesn't seem that important given that the categories have changed since the last update anyway!

Issue #4420, opened 2025-05-20
Automatically spot edge cases in views of collections using AI


pabloarosado:
We often build the metadata of multidimensional datasets, as well as the metadata of views in collections, programmatically. We do this because it is not scalable to manually write metadata for each possible combination of dimensions.
However, sometimes, [edge cases](https://owid.slack.com/archives/C03NV9Z3YSV/p1747746677877249) arise, which are hard to catch programmatically.

It would be useful to have a tool to help us spot such edge cases, where the resulting metadata of a specific indicator in grapher is nonsensical or misleading.

Lucas suggested to use AI to check in the JSON configs from the local directory export/explorers or export/multidim, and suggest changes.

Este suggested exporting every single thumbnail and passing it to ChatGPT.

We can come up with an once-off solution for now, and in the future figure out a more generic solution that integrates well with our workflow.

pabloarosado:
Daniel suggested starting with a simpler solution, like creating a text file with just titles of an explorer, and see if the AI can catch on odd cases, and then we build from there.
Maybe subtitles would also be useful, but not necessarily all metadata.
This could also help find odd cases because of missing strings.
We could also try manually introducing odd cases.

PR #5166, opened 2025-10-27
🎉 Detect typos, edge cases and inconsistencies

pabloarosado:
## Summary

Create a process that inspects explorers, mdims, charts, and posts (all kinds, including articles, DIs and topic pages), for typos (using codespell and Claude) and other context-dependent issues (using Claude).

The results are in [this spreadsheet](https://docs.google.com/spreadsheets/d/1h4eAYoVOgy47c8KprTvTgLAMHptnsqJ6RdAFbKq9auI/edit?usp=sharing), I'll have a look and either tackle them or create a separate issue.

Fixes https://github.com/owid/etl/issues/4420
Fixes https://github.com/owid/etl/issues/5167

## Results and cost
* Typos found by codespell:

| type     |   count | unique|
|---------|--------|-----|
| chart    |     118 | 68|
| explorer |     478 |17|
| multidim |     0 |0|
| post     |     307 |274|

* Typos and issues found by Haiku:

| type     |   count |
|:---------|--------:|
| chart    |    2761 |
| explorer |    1637 |
| multidim |     669 |
| post     |     320 |

Looking at [the Anthropic console](https://console.anthropic.com/settings/workspaces/default/keys?search=etl), all inspections and tests with Haiku have cost $31. So, running it again without testing and repeating due to errors could be around $20?

## Review

The code has been almost entirely created by Claude, and it very quickly became a monster. But given that it's quite self-contained, it's acceptable, and no thorough review is needed. If in the future we decide to integrate it in wizard or other processes, we should clean it up.

To run it: `etl inspector --help`

Feel free to add a typo or an inconsistency in a random chart, explorer, mdim, or article, and then run `etl inspector --slug [whatever slug]`.
If you don't have an `ANTHROPIC_API_KEY`, you can also run it just for codespell typos, with `etl inspector --skip-issues`

## Possible future improvements

- I implemented a final grouping and pruning API call at the end of the process, but it's proven to be a bit problematic (it's hard to find a balance between false detections and false dismissals; it's also tricky when the number of issues is large). So in the end I decided to turn it off by default. But with some additional effort, it could be useful.
- For now, this script is executed manually. If we like the output, we could store issues in a table with checksums, and automatically inspect when there are changes.
- Objects without slugs are currently not inspected (e.g. unpublished charts).
- Lucas suggested using pydantic, and trying different models.

### DIFF
--- a/apps/inspector/db.py
+++ b/apps/inspector/db.py
@@ -458,12 +458,27 @@ def load_views(
     Returns:
         List of view dictionaries (explorers, multidims, charts, and posts)
     """
-    # Fetch data from all sources (explorers, multidimensional indicators, charts, and posts)
+    # Fetch data from sources based on content_type filter (or all sources if no filter)
     # When slug_list is provided, all queries filter by the same slugs
-    df_explorers = fetch_explorer_data(explorer_slugs=slug_list)
-    df_mdims = fetch_multidim_data(slug_filters=slug_list)
-    chart_configs = fetch_chart_configs(chart_slugs=slug_list)
-    posts = fetch_posts(post_slugs=slug_list)
+    if not content_type or "explorer" in content_type:
+        df_explorers = fetch_explorer_data(explorer_slugs=slug_list)
+    else:
+        df_explorers = pd.DataFrame()
+
+    if not content_type or "multidim" in content_type:
+        df_mdims = fetch_multidim_data(slug_filters=slug_list)
+    else:
+        df_mdims = pd.DataFrame()
+
+    if not content_type or "chart" in content_type:
+        chart_configs = fetch_chart_configs(chart_slugs=slug_list)
+    else:
+        chart_configs = []
+
+    if not content_type or "post" in content_type:
+        posts = fetch_posts(post_slugs=slug_list)
+    else:
+        posts = []
 
     # Check if we got any results
     if df_explorers.empty and df_mdims.empty and not chart_configs and not posts:
@@ -542,22 +557,18 @@ def load_views(
         rprint(f"[cyan]Filtering to slug(s): {slugs_str}[/cyan]")
 
     config_count = len(configs_by_slug)
-    view_count = len(views) - config_count
 
     # Count charts and posts separately (they have view_type="chart" and "post")
     chart_count = len([v for v in views if v.get("view_type") == "chart"])
     post_count = len([v for v in views if v.get("view_type") == "post"])
 
-    breakdown = f"{len(agg_df_explorers)} explorers + {len(agg_df_mdims)} multidims"
+    breakdown = f"{len(agg_df_explorers)} explorer views + {config_count} configs + {len(agg_df_mdims)} multidims"
     if chart_count > 0:
         breakdown += f" + {chart_count} charts"
     if post_count > 0:
         breakdown += f" + {post_count} posts"
 
-    rprint(
-        f"[cyan]Aggregated to {view_count} unique views + {config_count} collection configs "
-        f"({breakdown})...[/cyan]\n"
-    )
+    rprint(f"[cyan]Aggregated to {len(views)} unique objects to inspect " f"({breakdown})...[/cyan]\n")
 
     return views
 
--- a/apps/inspector/inspector.py
+++ b/apps/inspector/inspector.py
@@ -414,9 +414,8 @@ def run(
     total_cache_read_tokens = 0
     total_cost = 0.0
 
-    total_collections = len(other_collections) + len(chart_collections) + len(post_collections)
     rprint(
-        f"\n[bold cyan]Processing {total_collections} collection(s): {len(other_collections)} explorer/multidim(s) + {len(chart_collections)} chart(s) + {len(post_collections)} post(s)...[/bold cyan]\n"
+        f"\n[bold cyan]Processing {len(other_collections)} explorer/multidim(s) + {len(chart_collections)} chart(s) + {len(post_collections)} post(s)...[/bold cyan]\n"
     )
 
     # Process explorers/multidims first


apps/inspector/config.py
```diff
@@ -0,0 +1,61 @@
...
+# Source: https://claude.com/pricing (verified 2025-11-03)
+# Note: We use the regular API, not Batch API (which offers 50% discount but is asynchronous)
+# Sonnet 4.5 pricing: $3/M in, $15/M out for prompts ≤200K tokens (our use case)
+MODEL_PRICING = {
+    "claude-haiku-4-5-20251001": {"input": 1.0, "output": 5.0},
+    "claude-3-5-haiku-20241022": {"input": 1.0, "output": 5.0},  # Legacy
+    "claude-sonnet-4-5-20250929": {"input": 3.0, "output": 15.0},
+    "claude-3-7-sonnet-20250219": {"input": 3.0, "output": 15.0},  # Legacy
+    "claude-opus-4-20250514": {"input": 15.0, "output": 75.0},
+}
```
lucasrodes: I strongly recommend getting the costs from this [YAML file](https://github.com/owid/etl/blob/master/apps/wizard/app_pages/expert_agent/models.yml). This file is in use by the expert.

It'd make sense to move that file somewhere else. Maybe under `apps/utils` and have a variable like MODELS_FILE in `apps/utils` too. But it can be done in a separate issue. If this feels out of scope, I'd then prioritize avoiding hardcoding this and using the existing file.

apps/inspector/db.py
```diff
@@ -0,0 +1,740 @@
...
+                "variable_description_short": lambda x: list(x.dropna()),
+                "variable_description_from_producer": lambda x: list(x.dropna()),
+                "variable_description_key": lambda x: list(x.dropna()),
+                "variable_description_processing": lambda x: list(x.dropna()),
+                "variable_attribution": lambda x: list(x.dropna()),
+                "variable_attribution_short": lambda x: list(x.dropna()),
+            }
+        )
+        .reset_index()
+    )
```
lucasrodes, line 121: possible simplification. Need to double check that `var_cols` is actually fine like I've implemented below? As in, unsure if there are false-positives? Otherwise, this feels much cleaner.


```suggestion
    group_cols = [
    "id", "view_type", "explorerSlug", "dimensions",
    "chartConfigId", "chart_config", "explorer_config",
]
# all columns that start with "variable_".
var_cols = [c for c in df.columns if c.startswith("variable_")]

def nonnull_list(s: pd.Series):
    return s.dropna().tolist()

agg_df = (
    df.groupby(group_cols, dropna=False, sort=False)
      .agg({c: nonnull_list for c in var_cols})
      .reset_index()
)
```

apps/inspector/db.py
```diff
@@ -0,0 +1,740 @@
...
```
lucasrodes, line 238: I'd re-use the logic from above here too.

apps/inspector/detectors.py
```diff
@@ -0,0 +1,1129 @@
...
+
+CRITICAL: When unsure about severity, KEEP the issue in the list. Missing a reputational risk is worse than a false positive.
+Write arrays explicitly [0, 1, 2], NOT Python code."""
+
+    try:
+        client = anthropic.Anthropic(api_key=api_key)
+        response = call_claude(
+            client=client,
+            model=GROUPING_MODEL,  # Use higher-quality model for grouping/pruning
+            max_tokens=3072,
```
lucasrodes: i think this might be hardcoded. I think you had a variable defined under `inspector/config`

lucasrodes (commented):
Thanks for doing this, Pablo! I briefly tried it, and it was a nice user experience. 


As you suggested, I haven't looked into the code in detail, but added some suggestions in a couple of places.

These comments are not critical, and some might be future work, but basically thinking on how to improve the accessability of the tool & usage of it. Congrats on the hard work 👏 


Leaving some general thoughts below:

### Thoughts on the CLI
- Could it be possible to have the model be defined by the user optionally? I think that could be nice in the future. Maybe an argument like `--model-name`.
- On option `--slug`
  - I suggest adding a short version like `-s`, so it is faster to filter. More generally, I think short versions of the option names are useful. 
  - How does this option actually work? Like, it filters our content using exact-matching? Also, how does the tool know if I want it to review a chart, explorer, or an article? Like `population` could be referring to `/grapher/population` or `explorers/population` or just `/population`.]
- On the help message from `etl inspector --help`
  - I find the text a bit confusing. I think there might be some formatting issues. It would be nice to have some clearer examples on how to call the command for a chart, for an explorer, for a group of articles, or using some of the `--skip` options. I think examples are the best way to communicate how to use the tool.

### Model costs 
- Added a comment on re-using the existing file that we have on model costs. It'd be nice if we could all rely on the same one, so all tools can benefit from updates there.

### How to use it?
How do you see us using this tool? Where does it fit in our workflow? 

I see that the tool can generate an output CSV, which then can be stored as a spreadsheet and shared. I think this is a nice format, and can serve as input for some other process. Since the number of suggestions is large, I wonder what the best workflow can be. Something that presents these and asks the user something like "apply suggestion".

For now, I'm thinking of having `owidbot` run this tool once a week, and suggest the changes via a PR every Monday or so. Limit the batch to 500 or so. With time, I could see it looking just at recent things, but for now, it's good to look everywhere, to detect "dark corners" on our site.

But ofc i could also see a simple wizard app that presents suggestions and waits for approval/reject, like chart diff. That would maybe help with changes to DB content, which is harder (if possible at all) to present purely via an ETL PR.

apps/inspector/config.py
pabloarosado: Now inspector will read from that yml. Moving that yaml to a different place is also a good idea, but let's do that in a separate PR.

PR #5181, opened 2025-10-30
📊 bls: calculate preliminary CPI for the current year

veronikasamborska1994:
not planning to display the year in progress on the key chart but want to try using the preliminary monthly data for some AI datasets


snapshots/bls/2025-04-12/us_consumer_prices.csv.dvc
```diff
@@ -23,6 +23,6 @@ meta:
       url: https://www.bls.gov/opub/copyright-information.htm
 
 outs:
-  - md5: b32101a40ebd327030d3f15cdf49f70e
-    size: 330105
+  - md5: d42784b58857bed48e4bcbcb0fe3bd68
+    size: 28060
```
Marigold, line 27: It's removing monthly & semi-annual data, is that expected?

Marigold (approved):
Chart diff looks good

snapshots/bls/2025-04-12/us_consumer_prices.csv.dvc
veronikasamborska1994: yep, we don't actually need these! thanks though!

PR #5209, opened 2025-11-07
📊 ai: explore IEA data on energy and AI

veronikasamborska1994:
hi @pabloarosado! we aren't yet publishing any charts from this dataset yet but will be working on these more in the new year.


snapshots/artificial_intelligence/2025-11-07/energy_ai_iea.xlsx.dvc
```diff
@@ -0,0 +1,33 @@
...
+
+      The data provides insights into the energy requirements and efficiency of AI infrastructure globally.
+
+    # Citation
+    producer: International Energy Agency
+    citation_full: "International Energy Agency (IEA), 'Energy and AI'. Published online at iea.org. Retrieved from: 'https://www.iea.org/data-and-statistics/data-product/energy-and-ai' [online resource]"
+
+    # Files
+    url_main: https://www.iea.org/data-and-statistics/data-product/energy-and-ai
+    url_download: https://www.iea.org/product/download/020879-000396-020495
```
pabloarosado: This link doesn't work, the snapshot loads a local file. So, just to avoid confusion, I'd remove the url_download.

snapshots/artificial_intelligence/2025-11-07/energy_ai_iea.xlsx.dvc
```diff
@@ -0,0 +1,33 @@
...
+    date_accessed: '2025-11-07'
+
+    # License
+    license:
+      name: Public domain
```
pabloarosado: The dataset comes from IEA, not EIA (and it's called Energy and AI, EAI ^^).
IEA datasets are almost certainly copyrighted, not public domain.

snapshots/artificial_intelligence/2025-11-07/energy_ai_iea.py
```diff
@@ -0,0 +1,38 @@
+"""Script to create a snapshot of dataset 'Energy and AI (IEA, 2025)'.
+
+INSTRUCTIONS FOR RUNNING THIS SNAPSHOT:
+
+Since the IEA dataset requires authentication, you need to manually download the file first:
+
+1. Visit: https://www.iea.org/data-and-statistics/data-product/energy-and-ai
+2. Create an IEA account or log in if you already have one
+3. Download the Excel file
+4. Place the downloaded file at: snapshots/artificial_intelligence/2025-11-07/energy_ai_iea.xlsx
```
pabloarosado: I think that you can place the file anywhere, as long as you then pass the same path to --path-to-file.

snapshots/artificial_intelligence/2025-11-07/energy_ai_iea.py
```diff
@@ -0,0 +1,38 @@
+"""Script to create a snapshot of dataset 'Energy and AI (IEA, 2025)'.
```
pabloarosado: You could copy these instructions as commented lines into the companion .dvc file.
Then you can simply remove this .py file, and execute the snapshot with:
```
etls artificial_intelligence/2025-11-07/energy_ai_iea --path-to-file ...
```

dag/artificial_intelligence.yml
```diff
@@ -221,4 +221,14 @@ steps:
...
+    - data://garden/us_census_bureau/2025-11-17/datacenter_construction
+
+
+  # IEA Energy Consumption by AI dataset
+  data://meadow/artificial_intelligence/2025-11-07/energy_ai_iea:
+    - snapshot://artificial_intelligence/2025-11-07/energy_ai_iea.xlsx
+  data://garden/artificial_intelligence/2025-11-07/energy_ai_iea:
+    - data://meadow/artificial_intelligence/2025-11-07/energy_ai_iea
+    - data://garden/energy/2025-06-27/electricity_mix
+  data://grapher/artificial_intelligence/2025-11-07/energy_ai_iea:
```
pabloarosado: Unless you are planning to have multiple energy_ai_* datasets, I think it would be preferrable to simply call the steps `energy_ai` (or `energy_and_ai`, to match the title of the original dataset). That way we avoid hardcoding the producer (in fact, since the electricity mix is included, IEA is no longer the only origin). But if the refactor is too inconvenient, you can leave it this way.

etl/steps/data/garden/artificial_intelligence/2025-11-07/energy_ai_iea.py
```diff
@@ -0,0 +1,239 @@
...
+def duplicate_2024_values_for_scenarios(tb: Table) -> Table:
+    """
+    Duplicate 2024 historical values for each non-historical scenario.
+
+    This ensures that when plotting scenario lines, they all start from the same
+    2024 baseline point, avoiding breaks in the visualization.
+    """
+
+    # Filter for 2024 historical data
+    historical_2024 = tb[(tb["year"] == 2024) & (tb["scenario"] == HISTORICAL_SCENARIO)].copy()
```
pabloarosado: I'd suggest renaming the function and avoiding the hardcoded year. You could get the latest historical year with:
```
tb[tb["scenario"]==HISTORICAL_SCENARIO]["year"].max()
```
This would be safer for future updates.

etl/steps/data/garden/artificial_intelligence/2025-11-07/energy_ai.meta.yml
```diff
@@ -0,0 +1,77 @@
...
+      total_installed_capacity__gw__historical:
+        title: Total installed capacity for data centers
+        unit: gigawatts
+        short_unit: GW
+        description_short: Total installed power capacity of data centers worldwide and by region.
+        description_key: *data_centers_description
+        display:
+          numDecimalPlaces: 1
+
+      total_electricity_consumption__share_of_total_electricity_demand__historical:
```
pabloarosado, line 42: The processing level of this indicator should be major.

etl/steps/data/garden/artificial_intelligence/2025-11-07/energy_ai_iea.meta.yml
```diff
@@ -0,0 +1,77 @@
...
+        description_short: Total installed power capacity of data centers worldwide and by region.
+        description_key: *data_centers_description
+        display:
+          numDecimalPlaces: 1
+
+      total_electricity_consumption__share_of_total_electricity_demand__historical:
+        title: Share of total electricity demand consumed by data centers
+        unit: "%"
+        short_unit: "%"
+        description_short: Data center electricity consumption as a percentage of total global or regional electricity demand.
```
pabloarosado: I think the "total global or regional" is confusing. The "global" applies only to "World", as far as I understand. Maybe something like:
Data center electricity consumption as a percentage of total electricity demand of the country or region.

etl/steps/data/garden/artificial_intelligence/2025-11-07/energy_ai_iea.py
```diff
@@ -0,0 +1,350 @@
...
+    tb = duplicate_2024_values_for_scenarios(tb)
+
+    # Create "Other" country (World - United States - China)
+    tb = create_other_country(tb)
+
+    # Calculate share of total electricity demand
+    tb_share = calculate_share_of_total_demand(tb, tb_electricity)
+
+    ####################################################################################################################
+    # TODO: Alternative method. Uncomment if agreed.
```
pabloarosado: I think that "Rest of the world" is currently ambiguous. For example, currently, given that "North America (IEA)" is among the countries in the table, I'd expect "Rest of the World" to not include any country in North America. However, as it's currently defined (World - United States - China - Europe (IEA) - Asia Pacific excl. China (IEA)), Canada and Mexico are indeed included.
Additionally, in the table tb_share, only China, US, and World are included, and therefore "Rest of the world" means something different.

I'd suggest a more explicit definition that avoids confusion, e.g. "World excl. United States and China". I've added two functions (currently not used, commented out). Feel free to use them, if you agree with the logic.

In the grapher step you could optionally rename the new region as "Rest of the world", for simplicity. But I wouldn't do it here to avoid the issues mentioned above (I hope I didn't misunderstand anything).

pabloarosado (changes requested):
Thanks for this work, it's a really interesting dataset, I'm looking forward to seeing more charts on this!

I've pushed a few minor changes, and left some minor suggestions. There's just one major change I suggested, around the definition of "Rest of the world" and the calculation of shares of electricity demand. Please let me know if anything is unclear or if I misunderstood anything.

Apart from that, it all looks good!

etl/steps/data/garden/artificial_intelligence/2025-11-07/energy_ai_iea.py
pabloarosado: I've also noticed that Ember is not referenced in the chart. This is because the origin is not properly propagated when calculating the share.
The alternative method ensures the share column carries IEA and also Ember origins.

veronikasamborska1994: thanks @pabloarosado ! very helpful comments :) I've accepted these changes and implemented your other suggestions now too.

pabloarosado (approved):
Looks good!

PR #5233, opened 2025-11-11
🎉 Regions improvements

pabloarosado:
* Add `add_population` as a new method of `Regions`, for convenience. With this, we can do `tb = paths.regions.add_population(tb)` within an ETL step, instead of having to import `ds_population` and `geo`.
* Minor improvement for convenience: `regions.get_regions(["USSR", "Africa"], only_subregions=True)` returns the subregions (regardless of whether they are successors or members).
* Add tests and adapt documentation.


etl/data_helpers/geo.py
```diff
@@ -1757,30 +1770,32 @@ def get_region(self, name: str) -> dict:
...
-            region_dict["members"] = list_members_of_region(  # type: ignore
-                region=name,
-                ds_regions=self.ds_regions,
-                # Load income groups only if necessary (and raise an error if not among dependencies).
-                ds_income_groups=self.ds_income_groups if name in INCOME_GROUPS else None,
-                include_historical_regions_in_income_groups=True,
-                unpack_subregions=True,
-            )
+            # For now, use the existing function to extract members (as well as successors, for historical regions, and related countries like overseas territories), which has some additional logic.
+            for subregion_type in ["members", "successors", "related"]:
```
lucasrodes: minor comment: I see that these names of "members", "successors", "related" and others repeated across the code.

We could have a global class defining valid subregion types. I recently did something similar that I liked quite a bit. Namely, the implementation and use of class [`StepType`](https://github.com/owid/analytics/blob/5ff7aa0c1a30a75fc00ef4598aef85ed250f92e2/analytics/duckdb/steps.py#L63-L91) in analytics.

pabloarosado: Thanks, that's a good idea for uses within the `geo` module. However, those names are already hardcoded in many steps. Are you suggesting to use, e.g. `SubregionType.MEMBERS` instead of `"members"` on those steps? I think that may become inconvenient. Maybe I misunderstood your suggestion.

PR #5265, opened 2025-11-20
🎉 Add json as an ETL output format

spoonerf:
Attempt at solving #5266


spoonerf:
Sorry @Marigold, just properly checked out your [other PR](https://github.com/owid/etl/pull/5281), which is perfect, and I think that means this one is not needed?

PR #5273, opened 2025-11-21
📊 Upgrade remaining UN SDG indicators

Marigold:
Upgrade non-trivial indicators from last UN SDG version. The only tricky one was quite popular [material-footprint-per-capita](http://staging-site-un-sdg-upgrade/grapher/material-footprint-per-capita) that shows weird jump in 2020 and 2023 (if you break it down by continents, you'll see huge drop in North and Latin america that's not realistic). I talked to Hannah, and we've decided to update this one in the next release.


spoonerf (approved):
Looks great, thanks @Marigold!

I made some chart tweaks, it's a bit concerning that some variables change so much between releases e.g. Share of mountain land that's degraded, but I guess they must have changed the methodology or something.

PR #5286, opened 2025-11-24
📊 natural_disasters: Correct natural disaster economic damages for inflation

pabloarosado:
## Summary
Add variables for the cost of natural disasters adjusted for inflation.

For more details, see [this conversation](https://owid.slack.com/archives/C094QM38BGQ/p1761817309853539).

## Changes
* Included the original "*_adjusted" columns from EM-DAT, that use the US CPI [to adjust for inflation](https://doc.emdat.be/docs/protocols/economic-adjustment/).
* Created new "_constant_usd" columns calculated using the GDP deflator.

## Review

Hey @paarriagadap I've added new indicators for costs of natural disasters adjusted for inflation. For comparison, I included both the original ones by EM-DAT (using the US CPI for all countries) and ours (using the GDP deflator). For now I'm not using any of these new indicators in charts or in the explorer.
Please have a look at the code just to ensure that my corrections make sense, following our discussions. And then see [this temporary chart](http://staging-site-data-adjust-damages-inflatio/grapher/adjusted-damages-adjusted-using-us-cpi-vs-using-gdp-deflator) to compare the US CPI vs GDP deflator adjustment.
For some country-years, the difference is immense, and the numbers don't make much sense (see e.g. Zimbabwe); so I'm not sure if we should include these indicators without further cleaning. If you prefer, we can continue the discussion on the thread, with Bertha. Thanks!

### DIFF
--- a/etl/steps/data/garden/emdat/2025-10-29/natural_disasters.py
+++ b/etl/steps/data/garden/emdat/2025-10-29/natural_disasters.py
@@ -1,4 +1,7 @@
-"""Process and harmonize EM-DAT natural disasters dataset."""
+"""Process and harmonize EM-DAT natural disasters dataset.
+
+NOTE: Some of the processing is related to inflation adjustments, which we decided to not use in the end. See further comments on this topic below.
+"""
 
 import datetime
 from typing import Any, Dict, List, Tuple
@@ -1010,8 +1013,16 @@ def run() -> None:
     # Get total count of impacts per year (regardless of the specific individual events during the year).
     tb = get_total_count_of_yearly_impacts(tb=tb)
 
+    ####################################################################################################################
+    # After some internal discussion (see https://owid.slack.com/archives/C094QM38BGQ/p1761817309853539 ), we decided to include variables adjusted for inflation, but not use them in charts or the explorer.
+    # EM-DAT provides variables adjusted for inflation using the US CPI for all countries.
+    # We thought that a more appropriate adjustment would be to use the GDP deflator.
+    # The results of both methods are very different, especially for some countries. In fact, for countries with instances of hyperinflation (like Belarus and Zimbabwe), the inflation adjustments lead to unrealistically high economic damages. Those cases would need to be handled separately.
+    # So, at least for now, we decided to not use the inflation-adjusted variables (neither those from EM-DAT not our own). Instead, we do show some of the cost variables as a percentage of GDP, which is a meaningful metric, and avoids some of those issues.
+
     # Add cost variables adjusted for inflation (using the GDP deflator).
     tb = create_cost_variables_adjusted_for_inflation(tb=tb, tb_wdi=tb_wdi)
+    ####################################################################################################################
 
     # Add a new category (or "type") corresponding to the total of all natural disasters.
     tb = create_a_new_type_for_all_disasters_combined(tb=tb)


etl/steps/data/garden/emdat/2025-10-29/natural_disasters.py
```diff
@@ -176,51 +194,56 @@ def get_last_day_of_month(year: int, month: int):
...
 
 
-def prepare_input_data(tb: Table) -> Table:
+def prepare_input_data(tb_meadow: Table) -> Table:
     """Prepare input data, and fix some known issues."""
+    tb_meadow = tb_meadow.copy()
+
     # Select and rename columns.
-    tb = tb[list(COLUMNS)].rename(columns=COLUMNS, errors="raise")
+    tb_meadow = tb_meadow[list(COLUMNS)].rename(columns=COLUMNS, errors="raise")
```
paarriagadap, line 205: BTW, is COLUMNS a dictionary with the same text twice? If that's the case, maybe it's better to keep a list and not use rename

etl/steps/data/garden/emdat/2025-10-29/natural_disasters.meta.yml
```diff
@@ -106,6 +106,23 @@ definitions:
     - *definition-disaster
     - *definition-dead
     - *definition-unknown-deaths
+  # Metadata for variables adjusted for inflation using the GDP deflator.
+  description_short_constant_usd: &description-short-constant-usd |-
+    This data is expressed in US dollars. It is adjusted for inflation but does not account for differences in living costs between countries.
+  footnote_constant_usd: &footnote-constant-usd |-
+    This data is expressed in constant {BASE_YEAR} US$.
+  description_key_inflation_adjustment: &description-key-inflation-adjustment
+    - This variable has been adjusted for inflation using the OWID Deflator dataset (Our World in Data, 2025).
```
paarriagadap: Looks like you're no longer using this dataset. Also, it's weird to express it like that for external readers. They don't have any reference about it.

etl/steps/data/garden/emdat/2025-10-29/natural_disasters.meta.yml
```diff
@@ -106,6 +106,23 @@ definitions:
...
+    - Cost variables, originally expressed in current US dollars, were converted to constant {BASE_YEAR} US dollars using a GDP deflator, to account for the effects of inflation in each country. The deflator data is available from the [World Bank World Development Indicators](https://data.worldbank.org/indicator/NY.GDP.DEFL.ZS.AD).
+  description_processing_inflation_adjustment: &description-processing-inflation-adjustment |-
+    To adjust cost variables for inflation, we use the GDP deflator series (NY.GDP.DEFL.ZS.AD) from the [World Bank World Development Indicators](https://data.worldbank.org/indicator/NY.GDP.DEFL.ZS.AD). Because the deflator is an index whose base year varies by country, we first identify each country's value in {BASE_YEAR} and re-express the entire time series relative to that year. We then convert all cost variables from current US dollars to constant {BASE_YEAR} US dollars by dividing them by the re-based deflator. For the global series, the World Bank does not provide a GDP deflator, so we construct it ourselves using the standard definition: the ratio of world GDP in current US dollars to world GDP in constant {BASE_YEAR} US dollars.
```
paarriagadap: I don't see you calculate a world deflator, but maybe I missed it

paarriagadap (approved):
Looks good!

Note that seeing differences between the methods is expected (because we are basically comparing US against other countries' CPI). The differences are more evident with countries with very high inflation (Zimbabwe, Argentina), so that would also be expected.

Shame that the deflator starts in 1990. There is the [non-linked version starting on 1960](https://admin.owid.io/admin/variables/1107035), but I assume the one you use is the right one, because:
 
> This series has been linked to produce a consistent time series to counteract breaks in series over time due to changes in base years, source data and methodologies.

Let's see if Bertha has anything to add.

etl/steps/data/garden/emdat/2025-10-29/natural_disasters.py
pabloarosado: You're right. I keep this style to make it easy to change names if neeeded.

etl/steps/data/garden/emdat/2025-10-29/natural_disasters.meta.yml
pabloarosado: You're right, this description is outdated. What I did now is to deflate first, and then aggregate to get the global values. I'll remove this explanation.

PR #5291, opened 2025-11-25
📊 emissions: Investigate alternative carbon intensities

pabloarosado:
## Summary

Estimate direct combustion emissions and carbon intensity of electricity generation.

## Context

We currently have [a chart](https://ourworldindata.org/grapher/carbon-intensity-electricity) on the carbon intensity of electricity. And users sometimes ask why our numbers are so high, especially for countries with a good share of renewables in the mix, compared with other sources (e.g. [EEA](https://www.eea.europa.eu/en/analysis/indicators/greenhouse-gas-emission-intensity-of-1)).

The main reason for the discrepancy is that our intensities come from Ember, who (as explained in [their methodology](https://storage.googleapis.com/emb-prod-bkt-publicdata/public-downloads/ember_electricity_data_methodology.pdf)) estimates **lifecycle** emission factors (so, even solar and wind have some emissions per kWh). In this PR, I implemented a simple calculation of **direct combustion CO₂ intensity of electricity generation** (using fixed IPCC-based factors for coal, gas and other fossil fuels).

For more details and caveats, see https://github.com/owid/owid-issues/issues/2165

## Review

Hey @spoonerf, could you please have a look at this PR? Note that it's not clear yet whether we'll actually publish [the new chart](http://staging-site-data-investigate-carbon-inte/grapher/direct-co2-intensity-of-electricity-generation). I'm waiting for Hannah's input on that. Given that, please don't spend too long on the review, but it would be great if you could sanity-check the logic in the garden step of `yearly_electricity`, especially the function `estimate_emissions_and_carbon_intensity_of_direct_combustion`, and flag anything that looks off. Thanks!


spoonerf (approved):
Hey @pabloarosado, 

As far as I can see it looks mostly good and sensible, I just noticed that for a few years (2001-2014) the direct emissions for Luxembourg are higher than the life-cycle ones, which I don't think should be possible? 

I noticed it in this exploratory chart here: http://staging-site-data-investigate-carbon-inte/admin/charts/8953/edit

PR #5323, opened 2025-12-08
✨ Add find_by_indicator semantic search

Marigold:
## Summary

Add a new function to search for indicators using natural language queries via the OWID search API. It tries to stay compatible to the existing interface, where each row (of CatalogFrame) is a table, by returning a table the indicator is part of and enhancing the output with columns `indicator_title`, `indicator`, `score`.

This means that you first have to call and display
```python
from owid.catalog import find_by_indicator

results = find_by_indicator("solar power")
results
```
pick a indicator, then load the table and indicator, e.g.
```python
tb = results.query('indicator_title == "Solar electricity per capita - kWh"').iloc[0].load()
tb['per_capita_solar_generation__kwh']
```
or load it directly with `catalogPath`
```python
from owid.catalog import RemoteCatalog

path = 'grapher/energy/2025-06-27/electricity_mix/electricity_mix'
rc = RemoteCatalog()
tb = rc[path]
indicator = tb["per_capita_solar_generation__kwh"]
indicator
```

I was experimenting with different interfaces for owid-catalog, but didn't find any that would be good enough to justify rewriting it.

One limitation of indicator semantic search that it only searches among `grapher` channel (we could add more channels if we wanted).

### find_by_indicator semantic search

**Features:**
- Semantic search using the OWID search API at `https://search.owid.io/indicators`
- Returns a standard `CatalogFrame` with extra columns: `indicator_title`, `indicator`, `score`
- Full compatibility with `.load()` for fetching tables

**Usage:**
```python
from owid.catalog import find_by_indicator

results = find_by_indicator("solar power")
print(results[["indicator_title", "indicator", "dataset"]])

table = results.iloc[0].load()
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,15 +41,6 @@ help:
 	@echo '  make watch-all 	Run all tests, watching for changes (including for modules in lib/)'
 	@echo
 
-.venv: install-uv .sanity-check
-	@echo '==> Installing packages (with local overrides)'
-	@if [ -n "$(PYTHON_VERSION)" ]; then \
-		echo '==> Using Python version $(PYTHON_VERSION)'; \
-		[ -f $$HOME/.cargo/env ] && . $$HOME/.cargo/env || true && UV_PYTHON=$(PYTHON_VERSION) uv sync --all-extras; \
-	else \
-		[ -f $$HOME/.cargo/env ] && . $$HOME/.cargo/env || true && uv sync --all-extras; \
-	fi
-
 docs.pre: .venv
 	@echo '==> Fetching external documentation files'
 	@.venv/bin/python docs/ignore/pre-build/bake_catalog_api.py
--- a/lib/catalog/Makefile
+++ b/lib/catalog/Makefile
@@ -6,12 +6,3 @@
 include ../../default.mk
 
 SRC = owid tests
-
-.venv: install-uv .sanity-check
-	@echo '==> Installing packages (with local overrides)'
-	@if [ -n "$(PYTHON_VERSION)" ]; then \
-		echo '==> Using Python version $(PYTHON_VERSION)'; \
-		[ -f $$HOME/.cargo/env ] && . $$HOME/.cargo/env || true && UV_PYTHON=$(PYTHON_VERSION) uv sync --all-extras; \
-	else \
-		[ -f $$HOME/.cargo/env ] && . $$HOME/.cargo/env || true && uv sync --all-extras; \
-	fi


lib/catalog/pyproject.toml
```diff
@@ -53,7 +36,8 @@ dev-dependencies = [
     "argh>=0.26.2",
     "pyright==1.1.373",
     "pandas-stubs==1.2.0.62",
-    "ruff==0.8.6"
+    "ruff==0.8.6",
+    "ipykernel>=6.29.0",
```
lucasrodes, line 40: do we need this library?

lib/catalog/owid/catalog/catalogs.py
```diff
@@ -716,6 +719,80 @@ def find_latest(
...
+                "table": table,
+                "namespace": namespace,
+                "version": version,
+                "dataset": dataset,
+                "channel": channel,
+                "path": path_part,
+                "format": "parquet",
+                "is_public": True,
+            }
+        )
```
lucasrodes, line 794: Could we align the ordering of the tables with what the other `find` methods return?

At the moment, `find` method returns (in order):

- 'table'
- 'dataset'
- 'version'
- 'namespace'
- 'channel'
- 'is_public'
- 'dimensions'
- 'path'
- 'formats'

And `find_by_indicator` returns first specific columns:

- 'indicator_title'
- 'indicator'
- 'score'

and then:

- 'table'
- 'namespace'
- 'version'
- 'dataset'
- 'channel'
- 'path'
- 'format'
- 'is_public'

This latter part has a very different order, and also is missing the `dimensions` column. I think we should have both frames consistent (besides the first thre column).

lib/catalog/pyproject.toml
Marigold: Yep, I need it to run notebooks in VSCode (and it's in dev deps anyway).

lib/catalog/owid/catalog/catalogs.py
Marigold: Done.

lucasrodes:
@marigold, yes, let's merge!

@pabloarosado, as we already discussed today, I think that your suggestion is a bit too opinionated at this point, and commits us to a specific design, which I am not sure is a good one. We have lots of content types that we might want to expose in owid-catalog, and good modularization will be our best friend here.

From our conversation today, I take that we should do baby steps, implement good building blocks, and then see where this brings us. I don't think there is a rush. 

We can always add a simpler higher layer later if we needed. This is less true the other way around 🙂

PR #5330, opened 2025-12-09
📊 ai: update robotaxis dataset

veronikasamborska1994:


snapshots/artificial_intelligence/2025-12-09/robotaxis.py
```diff
@@ -0,0 +1,132 @@
...
+from typing import List
+
+import pandas as pd
+import requests
+
+from etl.helpers import PathFinder
+
+paths = PathFinder(__file__)
+
+# List of all zip URLs to download and process (find them here https://www.cpuc.ca.gov/regulatory-services/licensing/transportation-licensing-and-analysis-branch/autonomous-vehicle-programs/quarterly-reporting)
```
Marigold: Nit - it'd be nicer if we were able to extract them from https://www.cpuc.ca.gov/regulatory-services/licensing/transportation-licensing-and-analysis-branch/autonomous-vehicle-programs/quarterly-reporting instead of hardcoding them, but it's not a big deal.

veronikasamborska1994: yep, fair point! I've added a script to do this now!

PR #5352, opened 2025-12-12
📊 Update Luxembourg Income Study to dimensional tables

paarriagadap:
Main issue https://github.com/owid/owid-issues/issues/2137

Pipeline issue: https://github.com/owid/owid-issues/issues/1816

### DIFF
--- a/scripts/poverty-inequality-explorers/README.txt
+++ b/scripts/poverty-inequality-explorers/README.txt
@@ -4,16 +4,8 @@ python -m scripts.poverty-inequality-explorers.wbpip.pip_expanded_poverty_explor
 python -m scripts.poverty-inequality-explorers.wbpip.pip_incomes_across_distribution_explorer
 python -m scripts.poverty-inequality-explorers.wbpip.pip_inequality_explorer
 python -m scripts.poverty-inequality-explorers.wbpip.pip_poverty_explorer
-python -m scripts.poverty-inequality-explorers.wbpip.pip_ppp_comparison_explorer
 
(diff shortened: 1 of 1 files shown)


etl/steps/export/explorers/lis/latest/poverty_lis.py
```diff
@@ -0,0 +1,141 @@
...
+    c.group_views(
+        groups=[
+            {
+                "dimension": "welfare_type",
+                "choice_new_slug": "after_vs_before_tax",
+                "view_config": {
+                    "hideRelativeToggle": "false",
+                    "selectedFacetStrategy": "entity",
+                    "hasMapTab": "false",
+                    "type": "LineChart",
```
paarriagadap, line 68: Why is `type` and not `chartTypes` the option that works here? And only as a string, and not a list? @lucasrodes

For example, I have set chartTypes in this other yaml https://github.com/owid/etl/blob/16d06cbdd69aa36cce6d03301c7d0672bef18e66/etl/steps/export/multidim/worldbank_wdi/latest/ilostat_comparison.config.yml#L140

paarriagadap: Ah, nevermind, it's because of how explorers are set, right?

lucasrodes: Yes, the configuration for explorers and MDIMs is not exactly the same just yet. The differences are mostly in the top-level `config`, and in the view-level config settings.

PR #5365, opened 2025-12-16
🚧 WIP: Extract owid-etl framework library

Marigold:
## Summary

This PR begins extracting core ETL framework functionality into a standalone `owid-etl` package in `lib/etl/`. The goal is to provide reusable ETL infrastructure that can be used independently of OWID-specific code.

### New `lib/etl/` package structure
- `owid/etl/config.py` - ETLConfig dataclass for user configuration
- `owid/etl/dag.py` - DAG loading and graph operations
- `owid/etl/steps.py` - Step protocol and implementations (DataStep, SnapshotStep, etc.)
- `owid/etl/execution.py` - Step execution engine
- `owid/etl/snapshot.py` - Snapshot class for data ingestion
- `owid/etl/cli.py` - Basic CLI runner
- Tests for DAG and config modules

### Step Registry Pattern
Introduced a registry pattern for step types:
```python
@register_step("data")
class DataStep: ...

@register_step("grapher")
class GrapherStep: ...  # Defined in etl/steps/__init__.py (OWID-specific)
```

This allows external code to register custom step types without modifying the core library.

### Re-export wrappers for backwards compatibility
- `etl/steps/__init__.py` - Re-exports from owid.etl.steps + defines GrapherStep locally
- `etl/snapshot.py` - Re-exports from owid.etl.snapshot
- `etl/dag_helpers.py` - Re-exports Graph type

## Current State

**Working:**
- ✅ Step registry pattern with decorators
- ✅ DAG loading and graph operations
- ✅ Basic config dataclass
- ✅ Step compilation and parsing
- ✅ Tests pass for lib/etl (13 tests)
- ✅ Tests pass for steps (12 tests)

**Broken / Incomplete:**
- ❌ Circular import: `lib/etl/owid/etl/steps.py` imports from `etl.helpers` which imports from `etl.snapshot`
- ❌ Type checking errors (pyright can't resolve some Protocol attributes)
- ❌ Heavy coupling: lib still imports from `etl.config`, `etl.paths`, `etl.files`, `etl.helpers`

## What's Left to Do

1. **Break circular imports** - Either:
   - Move `get_metadata_path` to lib/etl
   - Use lazy imports in steps.py
   - Restructure the dependency chain

2. **Decouple from OWID ETL** - The lib currently imports:
   - `etl.config` (for DEBUG, INSTANT, SUBSET flags)
   - `etl.paths` (for BASE_DIR, DATA_DIR, STEP_DIR, etc.)
   - `etl.files` (for checksum_file)
   - `etl.helpers` (for get_metadata_path)
   
   These need to either be moved to lib or made configurable via ETLConfig.

3. **Fix type checking** - Add missing attributes to Step Protocol or use type: ignore

4. **More test coverage** - Currently only DAG and config tests exist

## My Opinion

**Is this a good idea?** It's a mixed bag:

**Pros:**
- Cleaner separation of concerns
- Reusable ETL framework for other projects
- Registry pattern is elegant and extensible
- Forces us to identify and document the core abstractions

**Cons:**
- **Significant effort** - The ETL code is deeply intertwined with OWID-specific paths, config, and helpers
- **Maintenance burden** - Two places to maintain code, potential for drift
- **Circular dependencies** - The current architecture makes clean separation difficult
- **Limited reuse value** - The framework is quite specific to OWID's DAG-based ETL pattern

**Recommendation:** If the goal is truly to make a reusable library, it would require:
1. A config-driven approach (pass paths via ETLConfig, not import from etl.paths)
2. Moving core utilities (checksum_file, etc.) to lib
3. Removing all imports from `etl.*` in lib code

This is ~2-4 more hours of careful refactoring. If the reuse case is uncertain, it may not be worth the investment.

## Target API (when complete)

```python
from owid.etl import ETLConfig, run_dag, set_config

config = ETLConfig(
    base_dir=Path("./"),
    steps_dir=Path("./steps/data"),
    dag_file=Path("./dag.yml"),
)
set_config(config)

run_dag(steps=["data://garden/example/2024/dataset"])
```


lucasrodes:
@Marigold, What do you think about objects like `AdminAPI` (`apps.chart_sync.admin_api`)? Do you think it belongs to this abstraction? It feels more like "ETL-Grapher" interface territory, so maybe out of scope here?

Also, inspecting this submodule, I see it has other `etl/` dependencies: `etl.config`, `etl.grapher.model` (which depends on: `etl.paths`, `etl.db`)

All these modules feel more like "OWID custom workflow"* territory, which, based on what you wrote, might be out of scope for this library:

> [...] The goal is to provide reusable ETL infrastructure that can be used independently of OWID-specific code.

*Workflow here means how we use GitHub, PRs, branches, DBs, and other tools/frameworks.


Still, I'm not convinced of the current location of `AdminAPI`.

I don't have a preference at the moment or clearer idea; I'm wondering where this OWID workflow-specific code should go. Maybe `etl` is fine.

PR #5369, opened 2025-12-17
✨ Improve catalog find() with regex and fuzzy search

Marigold:
## Summary

Enhances the catalog `find()` method with better search capabilities:

- **Case-insensitive by default** - searches now match regardless of case (use `case=True` for case-sensitive)
- **Regex support by default** - `table` and `dataset` parameters support regex patterns (e.g., `find(table="gdp.*capita")`)
- **Fuzzy search** - new `fuzzy=True` parameter for typo-tolerant matching, sorted by relevance
- **Configurable threshold** - `threshold` parameter (0-100) controls fuzzy match strictness

### Usage examples

```python
from owid.catalog import find

# Case-insensitive regex (default)
find(table="gdp.*capita")
find(dataset="world.*bank")

# Fuzzy search - matches similar strings
find(table="forest area", fuzzy=True)       # matches "forest_area", "forest_share"
find(dataset="wrld bank", fuzzy=True)       # matches "worldbank" despite typo

# Strict fuzzy matching
find(table="population", fuzzy=True, threshold=90)

# Case-sensitive literal search
find(table="GDP", case=True, regex=False)
```

## Technical details

- Uses `rapidfuzz.fuzz.WRatio` for intelligent fuzzy matching (combines multiple strategies)
- Fuzzy results are sorted by match score (best matches first)
- Adds `rapidfuzz` as a new dependency
- Bumps version to 0.4.4

## Test plan

- [x] Added tests for case-insensitive search
- [x] Added tests for regex matching
- [x] Added tests for fuzzy matching with threshold control
- [x] Added test for fuzzy result sorting by score
- [x] All existing tests pass

### DIFF
--- a/lib/catalog/owid/catalog/catalogs.py
+++ b/lib/catalog/owid/catalog/catalogs.py
@@ -923,9 +923,7 @@ def save_frame(df: pd.DataFrame, path: str | Path) -> None:
         raise ValueError(f"could not detect what format to write to: {path}")
 
 
-def _match_score(
-    series: pd.Series, query: str, fuzzy: bool, case: bool, regex: bool
-) -> npt.NDArray[np.float64]:
+def _match_score(series: pd.Series, query: str, fuzzy: bool, case: bool, regex: bool) -> npt.NDArray[np.float64]:
     """Calculate match scores for a text column.
 
     Args:


lib/catalog/owid/catalog/catalogs.py
```diff
@@ -92,14 +104,27 @@ def find(
...
+        fuzzy_scores: npt.NDArray[np.float64] | None = None
 
         if table:
-            criteria &= self.frame.table.str.contains(table)
+            if fuzzy:
+                scores = self._fuzzy_scores(self.frame.table, table, case)
+                criteria &= scores >= threshold
+                fuzzy_scores = scores if fuzzy_scores is None else fuzzy_scores + scores
+            else:
+                criteria &= self.frame.table.str.contains(table, case=case, regex=regex)
```
lucasrodes: It's a bit hard for me to follow the logic here. Maybe some minor comments here so it's easier to review in the future?

lib/catalog/owid/catalog/catalogs.py
```diff
@@ -108,7 +133,12 @@ def find(
             criteria &= self.frame.version == version
 
         if dataset:
-            criteria &= self.frame.dataset == dataset
+            if fuzzy:
+                scores = self._fuzzy_scores(self.frame.dataset, dataset, case)
+                criteria &= scores >= threshold
+                fuzzy_scores = scores if fuzzy_scores is None else fuzzy_scores + scores
+            else:
+                criteria &= self.frame.dataset.str.contains(dataset, case=case, regex=regex)
```
lucasrodes: I'm realising this block can be abstracted and re-used

```python
def get_criteria_and_fuzzy_scores(self, object_type: str, obj_value, criteria, fuzzy_scores, threshold, case, regex):
  if fuzzy:
    scores = self._fuzzy_scores(self.frame[object_type], obj_value, case)
    criteria &= scores >= threshold
    fuzzy_scores = scores if fuzzy_scores is None else fuzzy_scores + scores
  else:
    criteria &= self.frame[object_type].str.contains(obj_value, case=case, regex=regex)
```

i'm also realising how many args this function has, so not super nice. But I guess better than duplicating code?

lucasrodes (approved):
Minor comments, looks good to me!

lib/catalog/owid/catalog/catalogs.py
Marigold: Good catch. I simplified it even more. Yeah, those args suck. Let's fix it in the redesign.

PR #5359, opened 2025-12-15
✨ Add CatalogPath helper class for parsing catalog paths

Marigold:
## Summary

Adds a `pathlib.Path`-like class for parsing and manipulating OWID catalog paths, replacing ad-hoc string splitting throughout the codebase.

### Before
```python
path, column = catalog_path.split("#")
parts = path.split("/")
channel, namespace, version, dataset_slug, table_name = parts[0], parts[1], parts[2], parts[3], parts[4]
```

### After
```python
p = CatalogPath.from_str(catalog_path)
p.channel, p.namespace, p.version, p.dataset, p.table, p.indicator
```

### Features
- Parse paths like `grapher/who/2024/gho/table#indicator`
- Properties: `channel`, `namespace`, `version`, `dataset`, `table`, `indicator`
- Computed properties: `dataset_path`, `table_path`, `step_uri`
- Builder methods: `with_version()`, `with_indicator()`, etc.
- Immutable (frozen dataclass) - can be used as dict keys
- `CHANNEL` literal type for type safety
- `indicator` as preferred alias for `variable`
- `get_variables_data()` now returns a `catalog_path` column with pre-parsed objects

## Test plan
- [x] Unit tests for CatalogPath class (35 tests)
- [ ] Manual testing of refactored code paths

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5373, opened 2025-12-17
🎉 owid-catalog v1.0.0rc0 pre-release

lucasrodes:
> [!NOTE]
> Visit http://staging-site-refactor-owidcatalog-apiupgr/etl/docs/libraries/catalog/api/ for quick overview of the changes through the docs.

## Overview

Major upgrade introducing a unified `Client` interface for data access. The API is organized around **Charts**, **Indicators**, and **Tables**.

### 🎯 Unified Client Interface

```python
from owid.catalog import Client

client = Client()

# Access data through dedicated APIs
df = client.charts.get_data("life-expectancy")
results = client.indicators.search("renewable energy")
tables = client.tables.search(table="population", namespace="un")
```

**New APIs**: `ChartsAPI`, `IndicatorsAPI`, `TablesAPI`, `SiteSearchAPI`
**Result types**: `ChartResult`, `IndicatorResult`, `TableResult`, `ResponseSet`

### ⚡ Lazy Loading

```python
# Metadata only (fast)
chart = client.charts.fetch("life-expectancy")
print(chart.title)

# Data loaded on first access
df = chart.data  # Downloads here
df2 = chart.data  # Cached
```

### CatalogPath
Migrated from
- https://github.com/owid/etl/pull/5359

### 🏗️ Model Improvements

**Field standardization:**
- `IndicatorResult.catalog_path` → `path` (matches `TableResult`)
- Logical grouping: identification → location → structural → content → technical

**Path formats:**
- Charts: `"life-expectancy"`
- Tables: `"garden/un/2024-07-12/un_wpp/population"`
- Indicators: `"garden/un/2024-07-12/un_wpp/population#population"`

### 🔧 API Enhancements

- **Fuzzy search** in `TablesAPI.search()` with threshold control
- **Consistent methods**: `.search()`, `.fetch()`, `.get_data()`
- **Result operations**: `.to_frame()`, `.filter()`, `.sort_by()`, `.latest()`

### 📚 Documentation

Reorganized into focused files:
- `catalog/index.md` - Quick start
- `catalog/api.md` - API reference (auto-generated)
- `catalog/structures.md` - Data structures

### 🏗️ Internal Changes

- Simplified `catalogs.py` (897 → 121 lines)
- Removed unused `internal.py` (100 lines)
- Added Pydantic models
- `RemoteCatalog` → `ETLCatalog` rename

## Migration

**Old (deprecated):**
```python
from owid.catalog.charts import Chart
chart = Chart("life-expectancy")
df = chart.get_data()
```

**New:**
```python
from owid.catalog import Client
client = Client()
df = client.charts.get_data("life-expectancy")
```

**Breaking changes:**
- `IndicatorResult.catalog_path` → `IndicatorResult.path`
- `RemoteCatalog` imports need updating

### DIFF
--- a/docs/libraries/catalog/api.md
+++ b/docs/libraries/catalog/api.md
@@ -6,7 +6,7 @@ The Data API provides unified access to OWID's published data through a simple c
 
 !!! warning "This library is under active development"
 
-    This documentation reflects the latest version, v1.0.0rc1, which is currently in Release Candidate stage. To install it run `pip install owid-catalog==v1.0.0rc1`.
+    This documentation reflects the latest version, v1.0.0rc0, which is currently in Release Candidate stage. To install it run `pip install owid-catalog==v1.0.0rc0`.
 
     We are continuously working to enhance its functionality and performance, and expect to release the stable v1.0.0 soon.
 
--- a/lib/catalog/README.md
+++ b/lib/catalog/README.md
@@ -86,7 +86,7 @@ make watch
 
 ## Changelog
 
-### `v1.0.0rc1` (Release Candidate)
+### `v1.0.0rc0` (Release Candidate)
   - **Highlights**
     - **New unified Client API**: Complete API refactor with `owid.catalog.Client` as single entry point
     - **Specialized APIs**: `ChartsAPI`, `IndicatorsAPI`, `TablesAPI`, `SiteSearchAPI` for different data access patterns
--- a/lib/catalog/owid/catalog/__init__.py
+++ b/lib/catalog/owid/catalog/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.0.0rc1"
+__version__ = "1.0.0rc0"
 
 from owid.catalog import api, processing, utils
 from owid.catalog.api import Client


lib/catalog/owid/catalog/api/indicators.py
```diff
@@ -0,0 +1,234 @@
...
+                )
+            )
+
+        return ResultSet(
+            results=results,
+            query=query,
+            total=data.get("total_results", len(results)),
+        )
+
+    def fetch(self, path: str, *, load_data: bool = False) -> IndicatorResult:
```
Marigold: This could benefit from `path: CatalogPath`. I'll try it out in my other PR.

lib/catalog/owid/catalog/api/tables.py
```diff
@@ -0,0 +1,312 @@
...
+            query_parts.append(f"channel={channel!r}")
+        elif len(list(channels)) > 1 or list(channels)[0] != "garden":
+            # Only show channels if non-default
+            query_parts.append(f"channels={list(channels)!r}")
+
+        return " ".join(query_parts) if query_parts else "all tables"
+
+    def search(
+        self,
+        table: str | None = None,
```
Marigold, line 113: An alternative would be to use glob and use `*`, something like `*/my_table` or `garden/*/my_table`, but it could be a bad idea, I'm not sure.

lib/catalog/owid/catalog/api/utils.py
```diff
@@ -0,0 +1,509 @@
...
+        if not case:
+            query = query.lower()
+            series = series.str.lower()
+        return np.array([fuzz.WRatio(query, x) for x in series], dtype=np.float64)
+    else:
+        matches = series.str.contains(query, case=case, regex=regex)
+        return np.where(matches, 100.0, 0.0)
+
+
+class CatalogMixin:
```
Marigold: It seems a bit weird that this is in utils.py and not in something like catalogs.py

etl/steps/data/garden/technology/2024-05-13/computer_memory_storage.py
```diff
@@ -21,7 +21,7 @@ def run(dest_dir: str) -> None:
     #
     # Process data.
     #
-    us_cpi = catalog.find_latest(dataset="us_consumer_prices").reset_index()[["year", "all_items"]]
+    us_cpi = ETLCatalog().find_latest(dataset="us_consumer_prices").reset_index()[["year", "all_items"]]
```
Marigold: We shouldn't be using `ETLCatalog` in steps. This should be a proper dependency and loaded via `paths.` or `Dataset(`.

lib/catalog/owid/catalog/api/indicators.py
lucasrodes: yes, I think `CatalogPath` would be a nice addition to the library

lib/catalog/owid/catalog/api/tables.py
lucasrodes: I agree that for `TablesAPI`, it is the only `search` method where the arguments are more structured, and that feels different compared to the other APIs.

I could see us adding a first argument `query`, which works similarly to the other ones, maybe along the lines of what you suggest. But then still accept keyword arguments like `dataset`, `table`, etc.

lucasrodes:
Hi @Marigold, thanks for your feedback! I'll answer your points below.

> Let's say I want the get the latest population table in my app, how would I do it?

At the moment this is not staightforward, the way to do it would be:


```python
# 1) Need to use regex to force an exact match. I wonder if we should default to this whenever `fuzzy` is set to false?
results = client.tables.search("^population$")

# 2) The returned object is a ResultSet, which has the list of results in attribute results. We need to sort them by version
results_sorted = sorted(results.results, key=lambda x: x.version, reverse=True)

# Then get the first item and load the data
tb = 3) results_sorted[0].data
```

I reckon this is not the best for the user. How could we improve the ergonomics?

In 1), I think we could force exact matches if `fuzzy` is False, what do you think?

in 2), I am not sure how to improve it... This comes from the fact that the returned object by `search` method is no longer a `CatalogFrame`, but a list of things. I can only think of adding either a `sort` argument to `search`, creating a new method to get the latest... Ideally we want something that works also fine with the other APIs... or that at least preserves the model we have now.

> I'd be more aggressive with deprecating, dragging those legacy interfaces with us might get tiring

What do you have in mind? Do you see some particular bits that should be deprecated?

lib/catalog/owid/catalog/api/utils.py
lucasrodes: I've created a module `owid.catalog.api.catalogs` and moved relevant code there.

lib/catalog/owid/catalog/api/indicators.py
lucasrodes: I've gone ahead and added to the library. Copied over from your other PR.

lucasrodes:
@Marigold! Yup, I'm likely to merge it this week. Thanks for reviewing.

It's a release candidate, so we can totally tweak in the future.

Also, the juicy discussion comes with the higher-level layer in https://github.com/owid/etl/pull/5383.

PR #5425, opened 2026-01-07
📊 wash explorer

antea04:
Update of WASH explorer:

- [X] Generate titles and subtitles programmatically from indicator metadata
- [X] Harmonize and improve map brackets
- [X] Update all data to latest release

WASH Indicators and metadata has already been checked in previous PR.

## To check in this PR:
- Explorer step
- DAG archival/ dependencies

Thanks!

### DIFF
--- a/etl/command.py
+++ b/etl/command.py
@@ -274,7 +274,7 @@ def main_cli(
 
 def _find_closest_matches(includes_str: str, dag: DAG) -> None:
     """Find and print closest matches for misspelled step names."""
-    print(f"No steps matched `{includes_str}`. Closest matches:")
+    print(f"No steps matched `{includes_str}`; check the spelling or try with the `--private` flag.\nClosest matches:")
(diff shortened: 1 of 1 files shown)


dag/covid.yml
```diff
@@ -33,7 +33,7 @@ steps:
     - data://garden/wb/2025-10-09/world_bank_pip_legacy
     - data://garden/un/2025-05-07/undp_hdr
     # Health
-    - data://garden/wash/2024-01-06/who
+    - data://garden/wash/2025-12-08/household
```
lucasrodes, line 36: I suppose the answer is yes, but have you checked that covid step runs successfully?

antea04: I've tested that the household data is imported & accessed correctly. There's other things in the script that don't run (on my machine), but it seems unrelated to the changes I made, so I thought it's probably fine.

PR #5453, opened 2026-01-13
🎉 Add popularity to catalog API search results

Marigold:
## Summary

Add popularity scores to catalog API search results, enabling users to see how popular tables and indicators are based on analytics data.

- **Tables API**: Enriches `TableResult` with dataset-level popularity from datasette (`type=dataset`, slug format: `namespace/version/dataset`)
- **Indicators API**: Enriches `IndicatorResult` with indicator-level popularity from datasette (`type=indicator`, slug format: full catalog path)
- New `popularity.py` module handles datasette API calls with proper error handling

### Usage

```python
from owid.catalog import Client

client = Client()

# Tables - now includes popularity
results = client.tables.search(table="population")
for r in results[:3]:
    print(f"{r.dataset}/{r.table}: popularity={r.popularity:.4f}")

# Indicators - now includes popularity  
results = client.indicators.search("GDP per capita")
for r in results[:3]:
    print(f"{r.title}: popularity={r.popularity:.4f}")
```

### Technical Notes

- Popularity is fetched dynamically after search results are returned (~200ms latency per search)
- On fetch failure, popularity defaults to 0.0 (graceful degradation)
- Uses existing `analytics_popularity` table in datasette

### DIFF
--- a/lib/catalog/owid/catalog/api/search.py
+++ b/lib/catalog/owid/catalog/api/search.py
@@ -159,11 +159,7 @@ class SiteSearchAPI:
 
         # Fetch popularity via Datasette API (chart slugs are full URLs)
         hits = data.get("results", [])
-        urls = [
-            f"https://ourworldindata.org/grapher/{hit.get('slug', '')}"
-            for hit in hits
-            if hit.get("slug")
-        ]
+        urls = [f"https://ourworldindata.org/grapher/{hit.get('slug', '')}" for hit in hits if hit.get("slug")]
         popularity_data = (
             self._client._datasette.fetch_popularity(
                 sorted(set(urls)),
--- a/lib/catalog/owid/catalog/api/tables.py
+++ b/lib/catalog/owid/catalog/api/tables.py
@@ -342,9 +342,7 @@ class TablesAPI:
         )
 
         # Fetch popularity via Datasette API (dataset-level: namespace/version/dataset)
-        dataset_slugs = [
-            f"{row['namespace']}/{str(row['version'])}/{row['dataset']}" for _, row in matches.iterrows()
-        ]
+        dataset_slugs = [f"{row['namespace']}/{str(row['version'])}/{row['dataset']}" for _, row in matches.iterrows()]
         popularity_data = (
             self._client._datasette.fetch_popularity(
                 sorted(set(dataset_slugs)),


lucasrodes:
This looks great! Thanks for working on this, @Marigold.

I've created a proposal for a more generic API to Datasette. You can review this here: https://github.com/owid/etl/pull/5457. Let me know what you think, and we can then merge it into this PR.

Other than that, PR looks good. Just couple comments on popularity integration into our various APIs:

- `TablesAPI`: It works perfectly. I like it very much
- `IndicatorsAPI`: I'm not sure what a good solution is here. We are prioritising the `score` value, so it is a bit confusing. There seems to be two forces at play: 'score' (for semantic similarity) and `popularity`. Should we think of combining these at some point? Not sure what's best and consistent with other APIs.
- `ChartsAPI`: Is there a reason we are not adding popularity on charts? might also be helpful on `.search`.

lib/catalog/owid/catalog/api/indicators.py
```diff
@@ -324,12 +330,45 @@ def search(
                 )
             )
 
+        # Enrich results with popularity from datasette
+        self._enrich_with_popularity(results, timeout=timeout)
```
lucasrodes: This is totally fine, but i'm wondering if we could just pass `popularity` when we construct the object `IndicatorResult`? Seems less hacky, but maybe i'm missing something!

lib/catalog/owid/catalog/api/search.py
```diff
@@ -184,12 +184,44 @@ def charts(
             chart._timeout = effective_timeout
             results.append(chart)
 
+        # Enrich results with popularity from datasette
+        self._enrich_charts_with_popularity(results, timeout=effective_timeout)
```
lucasrodes: idem as with `IndicatorResult`, what about passing popularity score when we construct the objects `ChartResult`?

lib/catalog/owid/catalog/api/tables.py
```diff
@@ -365,6 +371,12 @@ def search(
                 )
             )
 
+        # Enrich results with popularity from datasette
+        self._enrich_with_popularity(results, timeout=timeout)
```
lucasrodes: idem as other comments

lib/catalog/owid/catalog/api/indicators.py
Marigold: Agree, much better.

PR #5454, opened 2026-01-13
📊 animal_welfare: Add US poll on opinions on farming practices

pabloarosado:
Add data from Faunalytics on public opinions about standard farming practices in the US.

[The resulting chart](http://staging-site-data-us-poll-farming/admin/charts/8998/edit) is similar to [this other chart for the UK](https://admin.owid.io/admin/charts/8843/edit).

### DIFF
--- a/etl/steps/data/meadow/survey/2026-01-13/acceptability_of_us_farming_practices.py
+++ b/etl/steps/data/meadow/survey/2026-01-13/acceptability_of_us_farming_practices.py
@@ -30,9 +30,10 @@ COLUMNS = [
 def sanity_check_missing_values(tb):
     # There were missing responses in the RData file.
     # They seem to coincide with the answers "Very acceptable".
+    # Load the original data (a csv file that does not contain weights) and check that all instances of "Very acceptable" in that file coincide with all instances of nan in the RData file.
     import pandas as pd
 
-    df = pd.read_csv("~/Downloads/data_cleaned.csv")
+    df = pd.read_csv("https://osf.io/yz79e/?action=download")
     for c in COLUMNS:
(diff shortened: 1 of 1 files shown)


etl/steps/data/meadow/survey/2026-01-13/acceptability_of_us_farming_practices.py
```diff
@@ -0,0 +1,69 @@
...
+    "broilers_3",
+]
+
+
+def sanity_check_missing_values(tb):
+    # There were missing responses in the RData file.
+    # They seem to coincide with the answers "Very acceptable".
+    import pandas as pd
+
+    df = pd.read_csv("~/Downloads/data_cleaned.csv")
```
veronikasamborska1994: I guess you can just read from a link here rather than the Download folder `pd.read_csv("https://osf.io/yz79e/?action=download")`

etl/steps/data/garden/survey/2026-01-13/acceptability_of_us_farming_practices.meta.yml
```diff
@@ -0,0 +1,121 @@
+# NOTE: To learn more about the fields, hover over their names.
+definitions:
+  common:
+    unit: ""
```
veronikasamborska1994, line 4: does grapher now automatically detect if it's in %? if so that's pretty cool because I can't see you define the units anywhere else and it still shows up on the chart

pabloarosado: The indicators have empty units (defined here). I manually customized the chart to "Display relative values", and, when you do that, grapher assigns "%" as a unit.

PR #5457, opened 2026-01-13
✨ DatasetteAPI

lucasrodes:
> [!NOTE]
> To be merged into #5453  

Create a general-purpose light-weight Datasette API to interact with our public tables

**`DatasetteAPI`**
- Incorporates logic of `fetch_popularity` as a class method
- Provides other tools like `list_tables`
- It is available from various APIs, in case we had to get more data from our Datasette public instance
- Future: use it to get metadata from our public instance.

**`DatasetteTable`**
- Minimal class representing a table in Datasette.
- If required, presents table metadata (fields available, description, etc.)

### DIFF
--- a/lib/catalog/owid/catalog/api/datasette.py
+++ b/lib/catalog/owid/catalog/api/datasette.py
@@ -96,33 +96,30 @@ class DatasetteAPI:
         PAGE_SIZE = 1000
         request_timeout = timeout or self.timeout
 
-        try:
-            all_rows: list[dict[str, Any]] = []
-            offset = 0
+        all_rows: list[dict[str, Any]] = []
+        offset = 0
 
-            while True:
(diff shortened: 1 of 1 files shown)


lib/catalog/owid/catalog/api/datasette.py
```diff
@@ -0,0 +1,258 @@
...
+                all_rows.extend(rows)
+
+                # Stop if we got fewer rows than requested (last page) or not paginating
+                if not paginate or len(rows) < PAGE_SIZE:
+                    break
+
+                offset += PAGE_SIZE
+
+            return pd.DataFrame(all_rows)
+        except Exception:
```
Marigold: I'd rather see the raised exception than silently getting no results. (We should add this to CLAUDE.md, it tends to do it quite often)

PR #5460, opened 2026-01-14
💄 Update metadata in ILO data + create shares by status in employment

paarriagadap:
Aplpying updates mentioned in this [Slack post](https://owid.slack.com/archives/C09BQ6EQ2JG/p1768340751841839).

### DIFF
--- a/etl/steps/data/garden/un/2025-08-12/ilostat.meta.yml
+++ b/etl/steps/data/garden/un/2025-08-12/ilostat.meta.yml
@@ -79,13 +79,19 @@ definitions:
 
   description_short_status_in_employment: |-
     <% if classif1 == "Status in employment (ICSE-93): 1. Employees" %>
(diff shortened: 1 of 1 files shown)


paarriagadap:
Hi @Marigold! This PR has one narrative chart that's being updated.

I wonder if Chart diff is showing the right new version of the chart? [Here](http://staging-site-style-ilo-metadata-update-8a/admin/narrative-charts/263/edit) I can see that the data is updated correctly and the relative view has been disabled
image

but we don't see the same in [Chart diff](http://staging-site-style-ilo-metadata-update-8a/etl/wizard/chart-diff?chart_id=8848&show_reviewed=): while the data is migrated, the Settings button is still there. Is there a reason for that? Also, we can't access to the edit button in the narrative chart view. Maybe it's good to have it.
image

PR #5464, opened 2026-01-14
📊 ai: add robot density dataset

veronikasamborska1994:


etl/steps/data/meadow/ifr/2026-01-14/robot_density.py
```diff
@@ -0,0 +1,39 @@
...
+    tb = snap.read_excel(sheet_name="Tabelle1", skiprows=2, names=["country", "notes", "robot_density"])
+    #
+    # Process data.
+    #
+    # Drop the last row which contains source notes
+    tb = tb[tb["country"].notna()].copy()
+
+    # Remove any rows that don't contain actual country data
+    tb = tb[tb["robot_density"].notna()].copy()
+    tb = tb.drop(columns=["notes"])
```
antea04, line 28: In case there is anything important in the "notes" column, we could add it to the metadata.

etl/steps/data/garden/ifr/2026-01-14/robot_density.meta.yml
```diff
@@ -0,0 +1,13 @@
+dataset:
+  update_period_days: 365
+
+tables:
+  robot_density:
+    variables:
+      robot_density:
+        title: Industrial robots per 10,000 employees
+        description_short: Number of industrial robots per 10,000 employees in the manufacturing industry.
+        unit: "robots per 10,000 employees"
```
antea04, line 10: I would maybe add another sentence about why this measure matters (Something like "This indicates how automatized production within this country is."

You could also add a point in description_key about how this data got collected and what it includes (e.g. which industries, what kind of robots?), if you have this data. But it's only a nice-to-have :)

etl/steps/data/meadow/ifr/2026-01-14/robot_density.py
veronikasamborska1994: I already have the sources that are noted there in the dataset description. I don't think the breakdown is particularly important to include anywhere else :)

etl/steps/data/garden/ifr/2026-01-14/robot_density.meta.yml
veronikasamborska1994: I've added a bit in description key on what "industrial" means but otherwise don't have much information on the indicator :)

PR #5468, opened 2026-01-14
📊 plastic waste: add Cottom et al. dataset

veronikasamborska1994:
I am probably going to be changing some of the metadata fields throughout the week but the code otherwise should be ready :) thanks @lucasrodes!

### DIFF
--- a/etl/steps/data/garden/plastic_waste/2026-01-14/cottom_plastic_waste.meta.yml
+++ b/etl/steps/data/garden/plastic_waste/2026-01-14/cottom_plastic_waste.meta.yml
@@ -251,7 +251,7 @@ definitions:
     Low delivery rates can indicate limited capacity or unreliable service, even where collection services exist.
 
   text_no_collection_derived: &text_no_collection_derived
-    It is derived from modelled collection coverage and population estimates used in the SPOT inventory.
+    It is derived from the modelled collection coverage and population estimates produced by the [SPOT model](https://www.nature.com/articles/s41586-024-07758-6), which first fills gaps in municipal waste data using statistical predictions, then simulates how plastic flows through waste systems and quantifies uncertainty in those estimates. The model generates results for around 50,700 municipalities, which are then aggregated to country and regional totals.
 
   desc_key_plastic_emissions_shares_burn: &desc_key_plastic_emissions_shares_burn
     - This indicator shows the estimated share of global plastic waste emitted via open burning from municipal sources such as households, shops, and offices, attributed to each country or region.


snapshots/plastic_waste/2026-01-14/cottom_plastic_waste.zip.dvc
```diff
@@ -0,0 +1,32 @@
...
+    url_main: https://www.nature.com/articles/s41586-024-07758-6
+    # url_download: https://datadryad.org/downloads/file_stream/3338334
+    # Bot protection on the website means we should manually download the file and upload it to our S3 bucket.
+    # To add the file run: etls cottom_plastic_waste --path-to-file <path_to_downloaded_file>
+    date_accessed: '2026-01-14'
+
+    # License
+    license:
+      name: CC0 1.0
+      url: https://creativecommons.org/publicdomain/zero/1.0/
```
lucasrodes: Maybe I'm missing something, but it looks like it is licensed under CC BY 4.0, doesn't it?

Image

They also share [this other link](https://s100.copyright.com/AppDispatchServlet?title=A%20local-to-global%20emissions%20inventory%20of%20macroplastic%20pollution&author=Joshua%20W.%20Cottom%20et%20al&contentID=10.1038/s41586-024-07758-6&copyright=The%20Author(s)&publication=0028-0836&publicationDate=2024-09-04&publisherName=SpringerNature&orderBeanReset=true&oa=CC%20BY)... Not sure what prevails

etl/steps/data/meadow/plastic_waste/2026-01-14/cottom_plastic_waste.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if not national_file or not regional_file:
+            raise FileNotFoundError(
+                "Could not find the expected Excel files in the zip archive. "
+                f"Found files: {list(temp_path.rglob('*.xlsx'))}"
+            )
+        # Load data from Excel files (data is in 03_Outputs sheet)
+        # Header is in row 1 (0-indexed)
+        df_national = pd.read_excel(national_file[0], sheet_name="03_Outputs", header=1)
+        df_regional = pd.read_excel(regional_file[0], sheet_name="03_Outputs", header=1)
```
lucasrodes: I would suggest using the methods `Snapshot.extracted()` and then `read` (more details [here](https://docs.owid.io/projects/etl/guides/etl-tips/#reading-from-zipped-snapshots)). You can simply have:

```python
with snap.extracted() as archive:
    table1 = archive.read("filename1.csv")
```

`read` accepts any argument, so in your case you can use `sheet_name`, `header`, etc.

I see that in your case you calculate the value of the path dynamically actually, it's not really defined in the code, but rather you have a list (`national_file`) and take the first item. I'd suggest hardcoding it here, I think it is totally fine to clearly define which file you are reading in here. But maybe I'm missing something and this is fine/needed.

etl/steps/export/multidim/plastic_waste/latest/mismanaged_plastic_waste.py
```diff
@@ -0,0 +1,287 @@
...
+            "measure": "total",
+        }
+        tb["plas_recy_em"].m.original_short_name = "emissions"
+    if "plas_recy_em_per_cap" in tb.columns:
+        tb["plas_recy_em_per_cap"].m.dimensions = {
+            "emission_type": "total_burned_debris",
+            "emissions_source": "rejects",
+            "measure": "per_person",
+        }
+        tb["plas_recy_em_per_cap"].m.original_short_name = "emissions"
```
lucasrodes: This feels a bit hardcoded. I'd suggest defining a dictionary with all the dimension information, and then iterate over it and fill the dimensions.

The dimension details could look like this dictionary below

```python
DIMENSIONS_DETAILS = {
      "plas_em": {
            "emission_type": "total_burned_debris",
            "emissions_source": "all",
            "measure": "total",
        },
...
}
```

And then you just

```python
for col, dimensions in DIMENSIONS_DETAILS.items():
    if col in tb.columns:
        tb[col].m.dimensions = dimensions

File: etl/steps/export/multidim/plastic_waste/latest/mismanaged_plastic_waste.py [hunk 4]
```diff
@@ -0,0 +1,287 @@
...
+            "emissions_source": "rejects",
+            "measure": "total",
+        }
+        tb["plas_recy_em"].m.original_short_name = "emissions"
+    if "plas_recy_em_per_cap" in tb.columns:
+        tb["plas_recy_em_per_cap"].m.dimensions = {
+            "emission_type": "total_burned_debris",
+            "emissions_source": "rejects",
+            "measure": "per_person",
+        }
```
lucasrodes: This feels a bit hardcoded. I'd suggest defining a dictionary containing all the dimension (and short_name) information, then iterate over it to fill the dimensions.

The dimension details could look like this dictionary below

```python
COLUMN_DETAILS = {
	"plas_em": {
		"dimensions": {
			"emission_type": "total_burned_debris",
		    "emissions_source": "all",
		    "measure": "total",
		},
		"short_name": "...",
	},
...
}
```

And then you just

```python
for col, details in DIMENSIONS_DETAILS.items():
    if col in tb.columns:
        tb[col].m.dimensions = details["dimensions"]
        tb[col].m.original_short_name = details["short_name"]
```

etl/steps/export/multidim/plastic_waste/latest/mismanaged_plastic_waste.py
```diff
@@ -0,0 +1,287 @@
...
+    """Generate subtitle based on measure."""
+    measure = view.dimensions.get("measure")
+
+    if measure == "total":
+        return "Breakdown of the estimated total amount of plastic waste released to the environment each year through debris and open burning from municipal sources such as households, shops, and offices."
+    elif measure == "per_person":
+        return "Breakdown of the estimated total amount of plastic waste released to the environment per person each year through debris and open burning from municipal sources such as households, shops, and offices."
+    else:
+        return "Breakdown of the estimated total amount of plastic waste released to the environment through debris and open burning from municipal sources such as households, shops, and offices."
+
```
lucasrodes: FWIW, you can use syntax like

```python
if view.matches(measore="total"):
  ...
if view.matches(measore="per_person"):
  ...
else:
  ...
```

PR #5470, opened 2026-01-15
🎉 Auto-detect new datasets in indicator-upgrade CLI

pabloarosado:
Let indicator upgrade CLI auto-detect datasets whose indicators need to be upgraded in charts.

I implemented it with Claude and used it to indicator-upgrade all datasets in the latest climate update. I just ran `etl indicator-upgrade auto` and it worked well.

The CLI has the same options as match and upgrade, but I haven't tried what happens when there isn't a perfect mapping of indicators.


Marigold (approved):
Neat feature! I haven't tested it myself, but the logic is sound. As long as it works for you, that's all that matters.

PR #5475, opened 2026-01-15
📊 demography: Fix population metadata

pabloarosado:
## Summary

Simplify the metadata of the auxiliary `population#population` indicator, so that spurious population metadata does not appear in e.g. per capita indicators (e.g. [this chart](https://ourworldindata.org/grapher/per-capita-nitrous-oxide)).

Screenshot 2026-01-16 at 10 53 46

## Context

We use our population data both as (1) a primary indicator (e.g. to create charts about population growth), and (2) an auxiliary indicator (e.g. to create charts of per capita indicators).

Our population data is built as a combination of multiple origins. When using population as (1), we want to show all those origins in our charts. However, when using population as (2), we don't want those origins to pollute the limited space we have to display sources. For this reason, we decided to create two separate `population` indicators in our population garden dataset, to cater for those two use cases:

(1) `population_original#population`. This indicator has various origins (Hyde, Gapminder, UN WPP). These origins are what we see in charts about population, e.g. [Population by world region](https://ourworldindata.org/grapher/population-regions-with-projections). In fact, the grapher dataset that generates [our Population grapher dataset](https://admin.owid.io/admin/datasets/6621) only uses `population_original`.
  - Note that in most of these charts, population can also be considered as "auxiliary" (e.g. used to define the size of the bubbles in scatter charts). However, in all these cases, population is still shown as a primary indicator, with its own metadata. In other words, **the metadata of `population` is shown directly in our charts, as a primary indicator**.
population-regions-with-projections

(2) `population#population`. This indicator has one collapsed origin, with attribution "Population based on various sources (2024)". This is what we see e.g. in our chart [Per capita electricity demand](https://ourworldindata.org/grapher/per-capita-electricity-demand).
  - Importantly, in these charts, **the metadata of `population` is always shown indirectly in our charts, propagated to other indicators**.
per-capita-electricity-demand

So, when population is used as (2), the only metadata we need to surface is its origin metadata (so that charts can show the attribution "Population based on various sources (2024)", and we can [link users to how we build our population data](https://ourworldindata.org/population-sources)). Any other metadata field of that indicator does not need to be shown publicly.

However, because our metadata is automatically propagated, we have often encountered the problem that the metadata of population shows up in per capita indicators. To avoid this, we have been forced to manually remove certain metadata fields when adding population to a table:
```
tb["population"].m.presentation.attribution = None
```

And there are still cases (probably many) where population metadata is still showing up in per capita indicators. For example, this chart on [per capita nitrous oxide emissions](https://ourworldindata.org/grapher/per-capita-nitrous-oxide), shows the `description_key` of population, which is confusing for users.

## Proposed solution

A possible solution we have discussed on occasions is to to tweak our function `geo.add_population_to_table`, so that it removes population metadata before adding it to a table.

However, I realised that our `population#population` indicator simply doesn't need to have any metadata (except origins). So, a much cleaner solution is simply to remove that metadata in the first place.

## Main changes

In this PR:
- I rearranged the metadata fields in the garden population dataset, so that the auxiliary `population#population` indicator has minimal metadata.
  - Currently, it keeps its full collapsed origin, as well as its `title`, `description_short`, `unit`,  and `short_unit`. I kept these fields for clarity, given that (I think) they don't cause any issues. ~But maybe it's better to remove those as well, and maybe keep only `title`.~ Actually, I think it's convenient to have them, since `population` appears in some datasets as an auxiliary indicator (e.g. our [energy](https://github.com/owid/energy-data/blob/6da75463b2389bb5a7563cf85a9cea942cbeaf80/owid-energy-codebook.csv#L5) and [co2](https://github.com/owid/co2-data/blob/a499dd34c1372468f2335a370c5dd13cc3a72d90/owid-co2-codebook.csv#L5) datasets).
  - We could consider keeping the `processing_level: major`. By construction, any operation with an indicator with major processing leads to an indicator with major processing. However, it may be safer to have a less opinionated solution, and simply have no processing level for `population#population`.
- I removed all those instances where we were manually removing population `presentation.attribution` (otherwise the steps would fail).

With these changes, we solve cases where population metadata leaks into per capita indicators, e.g.
- [Old per capita nitrous oxide emissions](https://ourworldindata.org/grapher/per-capita-nitrous-oxide) (where description key is about population) -> [New per capita nitrous oxide emissions](http://staging-site-data-inspector-fix/grapher/per-capita-nitrous-oxide?tab=map&time=latest&country=FRA~DEU~CAN~ZAF) (with no description key).

## Review:
Hey @lucasrodes please let me know if you agree on the solution proposed here. It would be good to review `etl/steps/data/garden/demography/2024-07-15/population.meta.yml` thoroughly, to ensure I haven't messed up anything. My goal was to remove all unnecessary meetadata from `population#population` while keeping all other indicators untouched. I hope I haven't missed anything important.

I was also wondering something: Our `population_original#population` (shown in [this chart](https://ourworldindata.org/grapher/population-regions-with-projections)) doesn't have a `description_key`, whereas `historical#population_historical` (shown in [this chart](https://ourworldindata.org/grapher/population-marimekko)) does have a `description_key`. Why is that? My intuition is that the former should also have a `description_key` (if unclear, we could do that in a separate PR).


etl/steps/data/garden/demography/2024-07-15/population.meta.yml
```diff
@@ -109,29 +96,42 @@ tables:
...
         description_short: |
           Population per square kilometer by country, available from 10,000 BCE to 2100, based on data and estimates from different sources.
+        processing_level: major
         description_processing: |-
           We have estimated the population density by using population estimates from multiple sources and land area estimates by the Food and Agriculture Organization of the United Nations.
 
           We obtain it by dividing the population estimates by the land area estimates.
 
-          {definitions.common.description_processing}
+          {definitions.others.different_sources_processing}
```
lucasrodes: maybe let's referenfce

```
{tables.population_original.common.description_processing}
```

to avoid duplicating text

etl/steps/data/garden/demography/2024-07-15/population.meta.yml
```diff
@@ -195,7 +209,16 @@ tables:
...
+        topic_tags:
+          - Population Growth
         attribution: UN WPP (2024)
+      processing_level: major
+      description_processing: |-
+        {definitions.others.different_sources_processing}
+
+        - 2024-2100: Projections based on Medium variant by the UN World Population Prospects (2024 revision).
+
+        {definitions.others.regional_aggregates_processing}
```
lucasrodes: maybe let's referenfce

```
{tables.population_original.common.description_processing}
```

to avoid duplicating text

PR #5520, opened 2026-01-19
📊 plastic pollution: Anshassi and Townsend waste systems dataset

veronikasamborska1994:


etl/steps/data/garden/papers/2026-01-20/anshassi_waste_management.meta.yml
```diff
@@ -0,0 +1,235 @@
...
+    - *text_data_sources
+    - *text_gap_filling
+
+
+tables:
+  anshassi_waste_management:
+    variables:
+      final_classification:
+        title: Waste management system level
+        unit: "1 (best) - 4 (poorest)"
```
spoonerf: Nit but I would use "worst" instead of "poorest", it feels more like the natural opposite of "best"

spoonerf: Plus it saves 3 characters :D

etl/steps/data/garden/papers/2026-01-20/anshassi_waste_management.meta.yml
```diff
@@ -0,0 +1,235 @@
...
+    Municipal solid waste incineration is controlled burning of waste in purpose-built incinerators. These facilities typically have emissions controls to reduce air pollution, and some also recover energy from the burning process.
+
+  text_composting_definition: &text_composting_definition
+    Composting is a biological process that breaks down organic waste (such as food scraps and garden waste) to produce compost, which can be used to improve soil quality.
+
+  text_recycling_definition: &text_recycling_definition
+    Recycling is the process of collecting and reprocessing waste materials (such as paper, plastic, glass, and metals) to create new products.
+
+  text_classification_definition: &text_classification_definition
+    Each country is assigned to one of four waste management system levels (1 to 4). Level 1 represents the most advanced systems (with more waste treated via recycling, composting, or incineration, and less waste mismanaged). Level 4 represents the least controlled systems (with less waste treated and more waste ending up in open dumps or burned in the open air).
```
spoonerf, line 42: I kind of prefer this description to the one you currently have on the chart, although the existing one is more detailed. I wonder if you could use this as a subtitle and then squeeze in more detail in a DoD? Up to you though!

spoonerf (approved):
Looks good, just a couple of minor comments! I find the colours on the chart a bit hard to distinguish, especially 1 and 2. You could check out these colour brewer options for diverging palettes, as an alternative: https://colorbrewer2.org/#type=diverging&scheme=RdYlGn&n=4

etl/steps/data/garden/papers/2026-01-20/anshassi_waste_management.meta.yml
veronikasamborska1994: yep, me too! Hannah made this chart so I'll suggest it to her :)

PR #5530, opened 2026-01-21
🎉 Make graphs a first-class citizen in ETL

pabloarosado:
## Summary

Create a new `graph` channel that allows metadata from charts and multidims to live in etl as first-class citizens.

## Context

Currently, etl allows you to define metadata for datasets, tables, and indicators. Charts' metadata has never had a place to live in etl, even though they often include some of our most important, public-facing metadata.

Since recently, we can also create metadata for multidims, which bridges that gap somewhat. But they have always lived in a bit of a mixed territory, as "export" steps, partially in ETL and partially in the database.

In this PR I have prototyped an alternative architecture, where charts and multidims can live fully in etl.

## Benefits
- All of our most important metadata can now live in the etl, instead of being spread between etl and database.
- After a dataset update, Indicator upgrader would not be needed. We would just need to update the DAG, and running the graph steps would automatically update the indicator ids in charts.
  - NOTE: I haven't tested this update workflow, but theoretically it should work.
- Charts and mdim steps are executed in the same way, namely: `etlr slug --graph`
- This opens the doors to streamlining the drafting and creation of charts and mdims from etl (with some wizard). Alternatively, if anyone creates charts (and eventually mdims) from the grapher admin, this also makes it easier to pull those charts into etl.
- For charts that require some manual work after an update, we can now have a .py script that would make those changes. See `graph://emissions/latest/per-capita-co2-vs-average`.

## Important notes
- All code has been a back and forth with Claude, so it would need to be reviewed thoroughly if we decide to go ahead.
- I've tried to touch as little as possible of the existing core code, to ensure nothing breaks.
- Because of the previous point, there's some logic duplication; if we like this prototype, we could refactor things a bit in collection modules to import some of that logic to `grapher/graph`.
- I'm not sure if `graph` is the best name, but I wanted to avoid, e.g. `chart`, or `grapher`, or other things that could cause future headaches given the status quo.

## How it works

Graph steps live in `etl/steps/graph/` and are defined in `dag/graph/*.yml` DAG files. They can be:

1. Metadata-only: Simple `.meta.yml` files for chart configs or multidim configs.
2. Metadata + code: A `.meta.yml` file and a `.py` file for complex chart or mdim steps.

I replicated some existing charts and mdims. Feel free to run them:
```
etlr banning-of-chick-culling --graph
```
Make changes in the yaml file, or in the admin, and see what happens.
You can check chart-diff for changes.

For mdims ([covid cases](http://staging-site-wip-prototype-graph-metadata/admin/grapher/covid%2Flatest%2Fcovid%23covid_cases) and [covid deaths](http://staging-site-wip-prototype-graph-metadata/admin/grapher/covid%2Flatest%2Fcovid%23covid_deaths/?period=weekly&metric=absolute)), you can run:
```
etlr covid-cases --graph
etlr covid-deaths --graph
```

## TODO
If we like the prototype, the following things (at least) would need to happen (they could happen in a separate PR):
- Clean up code and refactor to avoid logic duplication with collection modules.
- Check if all relevant fields in charts and multidims are properly handled.
- Adapt VSCode extension to make graph steps clickable in the DAG.
- Adapt step upgrader to update dependencies in graph steps.
- Neither Claude nor myself could easily figure out the logic of mdim slugs. So this needs to be decided.
- We could consider doing a big migration of charts and mdims to become graphs, or have some tooling to let users migrate individual charts and mdims easily in upcoming updates.


dag/graph/animal_welfare.yml
```diff
@@ -0,0 +1,8 @@
+steps:
+  # Replicate chart fur-farming-ban: https://admin.owid.io/admin/charts/7146/edit
+  graph://animal_welfare/latest/fur-farming-ban:
+    - data://grapher/animal_welfare/2024-12-17/fur_laws
```
lucasrodes, line 4: I have a strong preference for avoiding `graph://`. It's unnecessarily close to `grapher://`.

Before creating a new prefix, I'd re-consider what we already have and why we would or would not use them here. E.g. `data://`, `export://`, etc.

dag/main.yml
```diff
@@ -804,3 +804,6 @@ include:
   - dag/growth.yml
   - dag/labor.yml
   - dag/economics.yml
+  - dag/graph/animal_welfare.yml
+  - dag/graph/covid.yml
+  - dag/graph/emissions.yml
```
lucasrodes, line 881: This is a new strategy, where we would create new dag yaml files not only based on topic domain but also based on "step type".

I'm not sure now. I'd start small with just one yaml file like `dag/charts.yml` or so, before committing to multiple files.

etl/grapher/graph.py
```diff
@@ -0,0 +1,908 @@
...
+# and that we want to detect when checking for manual overrides
+USER_CONTENT_FIELDS = {
+    "title",
+    "subtitle",
+    "note",
+    "chartTypes",
+    "hasMapTab",
+    "tab",
+    "selectedEntityNames",
+}
```
lucasrodes, line 93: Most of these definitions I think should be outside of the python file, in an external yaml file that a dev or whoever can easily update. Or best case: programmatically.

etl/steps/graph/animal_welfare/latest/banning-of-chick-culling.meta.yml
```diff
@@ -0,0 +1,24 @@
+map:
```
lucasrodes, line 1: i think it can be confusing that there is no reference to the indicator being displayed here.

My understanding is that this would come from the DAG file?

I think this can be a bit confusing, as there would be like two (distant) places where one edits the chart config sort of.

lucasrodes (commented):
leaving some side comments. Most of them addressed in my comment before.

etl/steps/graph/animal_welfare/latest/banning-of-chick-culling.meta.yml
pabloarosado: My first idea was to include indicators in the DAG. Then I realised that this was problematic because (1) for some graphs we have many indicators, (2) it implies too many changes, given that the main unit in our DAG is the dataset. So I decided to just keep the DAG just for datasets.

The indicator is specified in the dimensions (the indicator name happens to be `status`), and follows the same logic as collections: you can name just the indicator, or the table_name#indicator, or the full catalog path. This way we avoid hardcoding versions in our steps.

```
dimensions:
  - property: y
    catalogPath: status
```

etl/grapher/graph.py
pabloarosado: Sure, don't worry about the code for now, it's currently a Claude monster, it would need to be properly cleaned and refactored.

dag/main.yml
pabloarosado: I started indeed with just a `graphs.yml` file. Then I realised that this would be unrealistic, since it would blow up very quickly. Breaking it out into namespaces felt much more scalable.
Then I also thought that adding charts to existing namespace dag files (mixing data steps and graph steps) would be inconvenient. That's why I created the separate dag files.
No strong opinion, whatever is most convenient, this is easy to change.

dag/graph/animal_welfare.yml
pabloarosado: I was unsure about "graph", but then I thought that:
* "view" is confusing; we use this word in analytics a lot for visits (it's already problematic to talk about "the number of views in explorer and mdim views").
* "chart" would be confusing, because it would sound as if mdims are not included.
* "viz" could be, though it's a bit informal; "visualization" is inconveniently long.

We could reuse "export" but I think that doesn't fulfil the promise of making graphs a first-class citizen. It sounds as if it's creating a thing outside of etl.

We can brainstorm some alternatives!

etl/steps/graph/animal_welfare/latest/banning-of-chick-culling.meta.yml
lucasrodes: I see. Nice!

Did you reuse some of the logic from collections for this? I remember breaking a bit my brain to resolve indicator names and convert them into full catalog paths given the step dependencies.

While the MDIM config was not showing the complete catalog path, the config submitted to the DB was. This was resolved on `Collection.save()`. Is this similar in some way here?

dag/graph/animal_welfare.yml
lucasrodes: **Using `export://`**
Sure. I'm then just confused with the purpose of `export://`. We have been using it just for explorers and MDIMs, which I think should live in the same space as charts.

In my opinion, explorers, MDIMs, and charts should probably live under the same prefix/channel/namespace.

If we agreed on that, then using a new prefix like `graph://` is effectively just a renaming of `export://`. 

**Using `graph://`**
I understand your opposition to `view`, `chart`, and `viz`. Especially against "view", which I also find confusing.

"chart," I think, is perfectly fine. For all intents and purposes, MDIMs, Explorers, and grapher charts are indeed just charts. Some have a single view (grapher charts), others have multiple views (Explorers/MDIMs). But all can be considered charts. I think we may be disagreeing on the terminology here, and maybe we need to better define it.

"viz" is fine, though maybe a bit too casual.

Anyway, I am happy to find another term and go with it.

**But I _really_  wouldn't go with `graph`.**

etl/steps/graph/animal_welfare/latest/banning-of-chick-culling.meta.yml
pabloarosado: For this I asked Claude to reuse the logic already existing in collections, but Claude said it wasn't possible without making changes to those modules (which I wanted to avoid for now). So Claude added some logic for this. I added a TODO to clarify that some of that logic could be reused from collections to avoid duplication.

dag/graph/animal_welfare.yml
pabloarosado: Regarding export steps, they are meant to be steps that **write** outside of etl (we also have `export://github` and `export://s3`). They came after the `external` steps, which are steps that are **read** from outside of etl.

I agree with having charts and mdims in the same place, that's part of the motivation of this prototype. My preferred choice would be to take mdims out of export, but we can discuss this further.

I like "chart" too, and it's probably the best word. I just decided against it to avoid confusion with the meaning of "chart" under the status quo. But if you think it's significantly better, I'm up for it. We can decide on Monday.

pabloarosado:
@codex review

pabloarosado:
@codex review

lucasrodes:
@pabloarosado shall we close in favor of #5969 ?

pabloarosado:
We can close this PR, and continue with the alternative approach: https://github.com/owid/etl/pull/5969

PR #5561, opened 2026-01-26
📊 Create labor force participation MDim

paarriagadap:


etl/steps/export/multidim/worldbank_wdi/latest/labor_force_participation_rate.py
```diff
@@ -0,0 +1,76 @@
...
+                # Extract the column name from the catalog path
+                indicator_col = first_indicator_path.split("#")[-1]
+
+            # Remove "Male " from beginning of title. Also capitalize first letter. We need to extract this from the table's metadata (Grapher step)
+            # It is only male because of how we grouped the views
+            view.config["title"] = (
+                tb[indicator_col].metadata.presentation.grapher_config["title"].replace("Male ", "").capitalize()
+                + ": males vs. females"
+            )
+            view.config["subtitle"] = (
```
paarriagadap: Hi @lucasrodes! I wanted to modify these grouped views in this MDim, and I found that the jinja instruction set is shown in the resulting subtitles and description_short. This is not the case for the non-combined views so I wonder if there is a fix to do here?

etl/steps/export/multidim/worldbank_wdi/latest/labor_force_participation_rate.py
```diff
@@ -0,0 +1,76 @@
...
+
+    #
+    # Edit views
+    #
+    for view in c.views:
+        # Initialize config and metadata if they're None
+        if view.config is None:
+            view.config = {}
+        if view.metadata is None:
+            view.metadata = {}
```
lucasrodes, line 53: i think there is no need for initializing these. The mdim should work fine without this

etl/steps/export/multidim/worldbank_wdi/latest/labor_force_participation_rate.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if view.dimensions["sex"] == "male_vs_female":
```
lucasrodes, line 55: you can use view.matches(sex="male_vs_female") too

etl/steps/export/multidim/worldbank_wdi/latest/labor_force_participation_rate.py
paarriagadap: Claude recommended that after calling an empty metadata field (it comes from each indicator)

PR #5562, opened 2026-01-26
📊 static viz: add code to generate world population growth static viz

veronikasamborska1994:


etl/steps/export/static_viz/2026-01-26/world_population_growth.py
```diff
@@ -0,0 +1,266 @@
+"""Create world population growth visualization (1700-2100)."""
```
Marigold: I'd rather put it to to `export/` folder. We could also save static vizs to R2 as the result.

veronikasamborska1994: nice! thanks for the tip! still work in progress but I'll move it there when I am done with the actual code for the chart ! :)

etl/steps/export/static_viz/population/2026-01-26/world_population_growth.py
```diff
@@ -0,0 +1,443 @@
...
+        )
+    tb.loc[~filter_mask, "note"] = "Growth rate omitted for smooth visualization"
+
+    tb = tb.format(["year", "country"], short_name=SHORT_NAME)
+
+    ds_export = create_dataset(OUTPUT_DIR, tables=[tb])
+    ds_export.save()
+
+    # Save charts AFTER dataset is saved (so they don't get deleted)
+    output_path = OUTPUT_DIR / "world_population_growth_1700_2100.svg"
```
Marigold: I'd rather save it `CURRENT_DIR` so that we can see commit it and see in github directly (especially side by side when updating).

etl/steps/export/static_viz/population/2026-01-26/world_population_growth.py
```diff
@@ -0,0 +1,443 @@
...
+        fontsize=9,
+        color=text_grey,
+        ha="left",
+        va="bottom",
+    )
+
+    # Layout: reserve space for header + footer (extra bottom space for projection label)
+    fig.subplots_adjust(top=0.84, left=0.06, right=0.97, bottom=0.16)
+
+    # Prepare table data
```
Marigold: Nit - having `fig` operations together (ideally in function) and `tb` before them would be nicer.

(In `export://` step you don't need to save the output dataset. Just letting you know, in this case it's clearly helpful.)

Marigold (approved):
lgtm!

PR #5588, opened 2026-01-30
🐝 archive steps and fix fasttrack dag

paarriagadap:
Hi @pabloarosado! I was archiving in the fasttrack dag and realized that there were several steps defined in this format:
```
data-private://grapher/fasttrack/latest/cleaned_traced_news_sentiment: !!set
    ? snapshot-private://fasttrack/latest/cleaned_traced_news_sentiment.csv
```

Is that OK? Not according to your "traffic lights" tool, so I fixed that. I wonder if we should fix something in the fasttrack tool as well


pabloarosado:
Thanks @paarriagadap that fix looks about right, feel free to commit it! And thanks for fixing the DAG (the traffic lights extension doesn't understand that notation).

PR #5611, opened 2026-02-03
🎉  static viz: population doubling times

veronikasamborska1994:


etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+def load_world_population() -> tuple[int, int, Table, Table]:
+    """Load historical + projection population for World (boundary years & origins).
+
+    Returns:
+        hist_last_year  – last year present in the historical series
+        proj_last_year  – last year present in the projections series
+        tb_hist_raw     – raw historical Table (for origins metadata)
+        tb_proj_raw     – raw projections Table (for origins metadata)
+    """
+    ds = Dataset(etl_paths.DATA_DIR / "garden/demography/2024-07-15/population")
```
lucasrodes: Use `paths.load_dataset`, as usual, to avoid hardcoding catalog paths.

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+
+    tb_hist_raw = ds["historical"]
+    tb_proj_raw = ds["projections"]
+
+    tb_hist = tb_hist_raw.reset_index()
+    tb_proj = tb_proj_raw.reset_index()
```
lucasrodes: Use `ds.read`, it's simpler

```suggestion
    tb_hist_raw = ds.read("historical")
    tb_proj_raw = ds.read("projections")
```

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+
+    world_hist = tb_hist[tb_hist["country"] == "World"]
+    world_proj = tb_proj[tb_proj["country"] == "World"]
+
+    # Boundary years – inferred from the data
+    hist_last_year = int(world_hist["year"].max())
+    proj_last_year = int(world_proj["year"].max())
+
+    return hist_last_year, proj_last_year, tb_hist_raw, tb_proj_raw
```
lucasrodes: I wouldn't estimate these boundary years here. This seems more like "plotting" logic. I'd have this function just return `tb_hist_raw` and `tb_proj_raw`, and estimate `hist_last_year` and `proj_last_year` in `create_visualization` function.

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+    """Load pre-computed doubling times from the garden dataset.
+
+    Returns a DataFrame with columns:
+        year_reached   – year the target population was first reached
+        doubling_years – number of years from half-target to target
+        from_b         – source population in billions (target / 2)
+        to_b           – target population in billions
+    """
+    ds = Dataset(etl_paths.DATA_DIR / "garden/demography/2024-07-18/population_doubling_times")
+    tb = ds["population_doubling_times"].reset_index()
```
lucasrodes: Use `paths.load_dataset` to avoid hardcoding catalog paths, and `ds.read`, as in usual data steps.


```suggestion
    ds = paths.load_dataset("population_doubling_times")
    tb = ds.read("population_doubling_times")
```

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+
+    # population_target is in absolute numbers; convert to billions
+    tb["to_b"] = tb["population_target"] / 1e9
+    tb["from_b"] = tb["to_b"] / 2
+
+    tb = tb.rename(columns={"year": "year_reached", "num_years_to_double": "doubling_years"})
+    tb = tb[["year_reached", "doubling_years", "from_b", "to_b"]].sort_values("year_reached").reset_index(drop=True)
+
+    return pd.DataFrame(tb)
```
lucasrodes: how come you switch to `pandas.DataFrame` here? I'd probably try to keep `Table` objects if possible.

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+# ---------------------------------------------------------------------------
+# Entry point
+# ---------------------------------------------------------------------------
+
+
+def run() -> None:
+    """Entry point: load data, render and save chart."""
+    hist_last_year, proj_last_year, tb_hist_raw, tb_proj_raw = load_world_population()
+
+    log.info(f"Data boundaries – historical last year: {hist_last_year}, " f"projections last year: {proj_last_year}")
```
lucasrodes: Use `paths.log.info` instead, then you don't need to import anything from `structlog`

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+"""
+
+import xml.etree.ElementTree as ET
+from pathlib import Path
+
+import matplotlib
+import matplotlib.pyplot as plt
+import pandas as pd
+import structlog
+from owid.catalog import Dataset, Table
```
lucasrodes: probably no need to import `Dataset` (see comments below)

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+inferred from the garden ``demography/2024-07-15/population`` dataset.
+"""
+
+import xml.etree.ElementTree as ET
+from pathlib import Path
+
+import matplotlib
+import matplotlib.pyplot as plt
+import pandas as pd
+import structlog
```
lucasrodes: Probably no need to import `structlog`, see comments below

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+from owid.catalog import Dataset, Table
+
+from etl import paths as etl_paths
+
+# Use non-path text so SVGs stay editable in Figma
+matplotlib.rcParams["svg.fonttype"] = "none"
+
+CURRENT_DIR = Path(__file__).parent
```
lucasrodes: To manage paths, and dependencies, please use `PathFinder` object as usual:

```suggestion
paths = PathFinder(__file__)
```

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+    fig = create_visualization(df, hist_last_year, proj_last_year, source_citation)
+
+    # --- save PNG ---
+    png_path = CURRENT_DIR / "pop_doubling.png"
+    fig.savefig(png_path, format="png", dpi=300, bbox_inches="tight")
+    log.info(f"Saved chart to {png_path}")
+
+    # --- save SVG ---
+    svg_path = CURRENT_DIR / "pop_doubling.svg"
+    fig.savefig(svg_path, format="svg", dpi=300, bbox_inches="tight", metadata={"Date": None})
```
lucasrodes: Instead of using `CURRENT_DIR`, just use `paths.directory`.

Also, I could see us having a method under `PathsFinder` to be re-used consider having a method like `export_fig`. I'm less sure, but maybe could be handy so you could run `paths.export_fig("pop_doubling", ["svg", "png"], dpi=300, bbox_inches="tight")`.

```suggestion
def export_fig(self, filename: str, extensions=list[str], **kwargs):
    for ext in extensions:
        path = paths.directory / f"{filename}.{ext}"
        kwargs = {
            "fname": path,
            "format": ext,
            **kwargs,
        }
        if ext == "svg":  # is this something applicable always?
            kwargs["metadata"] = {"Date": None}
        fig.savefig(**kwargs)
```

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+        key = (origin.attribution_short, origin.title, origin.date_published)
+        if key not in unique_origins:
+            unique_origins[key] = origin
+
+    source_parts = []
+    for (producer, _title, date_pub), _origin in sorted(unique_origins.items()):
+        year = date_pub.split("-")[0] if date_pub else ""
+        source_parts.append(f"{producer} ({year})")
+
+    return "Data sources: " + "; ".join(source_parts)
```
lucasrodes: i find this method a bit confusing. You could accomplish the same behavior without three separate loops.

something like

```suggestion
def build_source_citation(tb_historical: Table, tb_projection: Table) -> str:
    # Track seen origins to avoid duplicates
    seen_origins = set()
    source_parts = []
    # Iterate over variables and extract their origins, deduplicating by (producer, title, date)
    variables = [tb_historical["population_historical"], tb_projection["population_projection"]]
    for var in variables:
        for origin in var.metadata.origins:
            key = (origin.attribution_short, origin.title, origin.date_published)
            if key not in seen_origins:
                seen_origins.add(key)
                year = origin.date_published.split("-")[0] if origin.date_published else ""
                source_parts.append(f"{origin.attribution_short} ({year})")
    return "Data sources: " + "; ".join(sorted(source_parts))
```

(i haven't tested it)

etl/steps/export/static_viz/population/2026-02-03/pop_doubling.py
```diff
@@ -0,0 +1,564 @@
...
+        if len(groups[gid]) == 0:
+            container.remove(groups[gid])
+
+    # ── replace main_group children with the new flat structure ────────────
+    main_group.clear()
+    main_group.set("id", "figure_1")
+    for child in container:
+        main_group.append(child)
+
+    tree.write(svg_path, encoding="utf-8", xml_declaration=True)
```
lucasrodes: is this only applicable to this step? or would we want to always have some logic like this to export plots in a Figma-friendly way? If so, I'd consider moving this logic elsewhere so that other steps benefit from it too.

lucasrodes (commented):
I've left some comments! Thanks for working on this, the chart is turning out very cool.

etl/steps/export/static_viz/population/2026-01-26/world_population_growth_1700_2100.svg
lucasrodes, line 1: is this file expected to be here? I think it might have leaked from somewhere else.

Marigold: It looks like every rerun changes
```
<clipPath id="p8e7f060c00">
```
we should remove it from the SVG to ensure we don't get spurious diffs (I hope it's possible).

PR #5621, opened 2026-02-05
📊 ai: add stackoverflow survey on AI use

veronikasamborska1994:
@Marigold Claude is starting to really surprise me sometimes! It used to take me at least a couple of hours to get any AI assistant to get the right bits out of an html file but with this one I literally just gave it the links and it spit out the snapshot script without any additional prompts!!


Marigold (approved):
Yeah, it's getting smarter every year. By the way, you can use `@codex review` to get automatic review in addition to human one. The chart looks good.

Issue #5564, opened 2026-01-26
Migrate legacy OWID datasets to ETL


Marigold:
**Context**
We currently have approximately 485 legacy data sets (mostly manually uploaded CSVs) that lack catalog paths. This complicates our codebase (e.g., `init catalog`, `indicator upgrader`) because we have to maintain special handling for them.

The goal is to backport these to ETL to simplify the architecture, ensuring every chart relies on a valid catalog path.

**Action Items**
- [ ] **Kill datasets with just draft charts:** Analyze the legacy datasets and remove those that are only associated with draft charts to reduce volume.
- [ ] **Identify Econ datasets:** Isolate the ~100 datasets that belong to Core Econ to determine if they need specific handling.
- [ ] **Clean up old backports:** Ensure that no artifacts or datasets from previous backporting attempts exist (e.g., check for old `backport` namespace usage).
- [ ] **Create special DAG:** Backport the remaining valid datasets and place them into a designated DAG to keep them separate from standard pipelines.
- [ ] **Review Charts:** Utilize the housekeeper dashboard to identify low-traffic charts or further candidates for removal: http://metabase.owid.io/question/812-charts-to-review-housekeeper?num_days=365&is_published=

PR #5653, opened 2026-02-11
Migrate legacy OWID datasets to ETL

Marigold:
Closes https://github.com/owid/etl/issues/5564

## Summary

We have ~485 legacy datasets (manually uploaded CSVs) that lack `catalogPath` in MySQL. This complicates our codebase because we have to maintain special handling for them (e.g., in `init catalog`, `indicator upgrader`).

This PR adds tooling to migrate them into the ETL pipeline, giving every dataset a proper catalog path like:
```
grapher/plastic_pollution/2021-04-30/plastic_ocean_pollution__meijer_et_al__2021/plastic_ocean_pollution__meijer_et_al__2021#mismanaged_plastic_waste__metric_tons_year_1
```

Migrated datasets are added to `dag/migrated.yml` (we may reorganize them into topic DAGs later). Each dataset gets a meaningful namespace and version derived from the original source.

## What changed

- **`etl b migrate --run`**: New flag that automates the full migration pipeline in one command — backport, snapshot, ETL+grapher, indicator match, and chart upgrade
- **`etl b migrate --dry-run`**: Fixed to actually skip file generation (was broken before)
- **Claude skill** (`migrate-dataset`): Allows Claude Code to autonomously migrate a dataset by ID — looks up sources, picks namespace/version, runs the migration

## Example migration

This PR includes one example migration (dataset 5291 — Plastic ocean pollution) to demonstrate the approach. If the team is happy with how it looks, I'll run it automatically for the remaining ~300 datasets.

This will add a lot of files to git (snapshot .dvc, garden step, grapher step, metadata YAML per dataset), but it shouldn't be a big deal — and if we want to slim it down later, we can clean up by e.g. removing grapher steps.

## Notes

- After migration, we could get the raw data of all backported datasets from the snapshot directory if needed, but that's not urgent.
- There will still be datasets without `catalogPath` in the database (e.g. Core Econ datasets), but we can safely ignore them in our apps.

## Plan

[Full list of datasets to migrate](https://docs.google.com/spreadsheets/d/1T-rLx6acR1m_1HoNsNRnVlk7CwVT9wMzikllG1zKd2A/edit?usp=sharing)

1. Migrate all datasets with published charts and meaningful traffic using `etl b migrate --run`
2. Kill datasets that only have draft charts or zero views
3. Handle Core Econ datasets (~100) separately if needed
4. Clean up old backport namespace artifacts

### DIFF
--- a/.claude/skills/chart-editing/SKILL.md
+++ b/.claude/skills/chart-editing/SKILL.md
@@ -39,7 +39,7 @@ map:
       Banned: '#4881c6'
       No laws: '#b6a28c'
     customNumericColorsActive: true
-$schema: https://files.ourworldindata.org/schemas/grapher-schema.009.json
+$schema: https://files.ourworldindata.org/schemas/grapher-schema.010.json
 originUrl: /animal-welfare
 dimensions:
   - property: y
--- a/apps/chart_sync/admin_api.py
+++ b/apps/chart_sync/admin_api.py
@@ -64,9 +64,10 @@ class AdminAPI(object):
         return js
 
     def create_chart(self, chart_config: dict, user_id: Optional[int] = None) -> dict:
-        # Extract isInheritanceEnabled from config and prepare request params
+        # Extract chart-table fields; keep them out of chart_configs.full payload.
         config = chart_config.copy()
         is_inheritance_enabled = config.pop("isInheritanceEnabled", None)
+        config.pop("forceDatapage", None)
 
         # Build request parameters
         params = {}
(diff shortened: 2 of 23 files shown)


pabloarosado (approved):
Hey Mojmir, this looks like a solid plan. We can let it run in a staging server. If everything goes well, the charts should all be exactly identical.
I'm wondering a few things:
- Will chart diff work well, with these old charts? (I guess so).
- There's a file with LIS code in this PR, was that an accident?
- Wouldn't it be good to fix the auto archival process before running it? (If not, I guess we could also just run a one off script to archive the grapher datasets in one go, no big deal).
- Once this is done, it would be good to test some of our common tools on that staging server before merging, to see if they become significantly slower, due to the added steps in the dag. Hopefully not.

So, let's see how it goes, thanks for tackling this!

PR #5682, opened 2026-02-16
📜 Create technical publication on slavery data

pabloarosado:
## Summary
Add a technical publication on the slavery historical data to the ETL docs. It includes a long table, and Bastian suggesting making it expandable on demand. I got to a solution that I think works well (I haven't tried on mobile, but that's probably not very important).

[Here's the link to the publication](http://staging-site-docs-slaverydata-publication/etl/docs/analyses/slavery_historical_data/).

## Review

No need to go through the text, simply have a look and see if there's some small ways to tweak it (especially the table). I noticed that the footnotes don't show anything on hover, not sure if it's a zensical bug.

## TODO
Update links once article is published (both inside the index and in the technical publication itself).

### DIFF
--- a/docs/analyses/deadliest_animals/index.md
+++ b/docs/analyses/deadliest_animals/index.md
@@ -1,3 +1,6 @@
+---
+status: new
+---
 # What are the world's deadliest animals?
 
 !!! info ""
--- a/docs/guides/data-work/ai-workflow.md
+++ b/docs/guides/data-work/ai-workflow.md
@@ -3,7 +3,6 @@ tags:
   - 👷 Staff
   - Data Workflow
   - AI
-status: new
 icon: lucide/sparkles
 ---
 
--- a/docs/guides/demo.md
+++ b/docs/guides/demo.md
@@ -5,7 +5,6 @@ tags:
   - Documentation
   - Examples
 icon: lucide/popcorn
-status: new
 # hide:
 #   - toc  # Remove this line to show table of contents
 ---
--- a/docs/libraries/index.md
+++ b/docs/libraries/index.md
@@ -1,5 +1,4 @@
 ---
-status: new
 hide:
     - toc
 ---
--- a/etl/steps/data/garden/excess_mortality/latest/excess_mortality_economist/__init__.py
+++ b/etl/steps/data/garden/excess_mortality/latest/excess_mortality_economist/__init__.py
@@ -1,7 +1,6 @@
-from owid.catalog import Table
-
 from etl.helpers import PathFinder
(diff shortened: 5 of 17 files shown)


zensical.toml
```diff
@@ -126,7 +126,8 @@ nav = [
         { "Does the news reflect what we die from?" = "analyses/media_deaths/methodology.md" },
         { "Food supply of world regions" = "analyses/food_supply_of_world_regions/food_supply_of_world_regions.html" },
         { "Our COVID work" = "api/covid.md" },
-        # { "Land use for biofuels vs. solar" = "analyses/biofuels_land_use/index.md" },
+        { "Land use for biofuels vs. solar" = "analyses/biofuels_land_use/index.md" },
```
lucasrodes, line 129: is this expected?

docs/css/extra.css
```diff
@@ -140,3 +140,173 @@ div.highlight pre span.gp {
     transition: none;
   }
 }
+
+/* --- CSV Data Table ------------------------------------------------ */
```
lucasrodes, line 144: what was the motivation for customizing the table? was the default rendering not nice? The second table looks like its fonts are too large, doesnt it?

docs/analyses/slavery_historical_data/index.md
```diff
@@ -0,0 +1,82 @@
+# Tracking historical progress against slavery and forced labor: a long-run data view
```
lucasrodes, line 5: Id add the `status: new` to signal that the article is new. Like in [this entry](https://raw.githubusercontent.com/owid/etl/master/docs/guides/data-work/ai-workflow.md) (feel free to remove the new tag in the example page)

zensical.toml
pabloarosado: I don't know why it was commented out, I guess I forgot to uncomment it when we published it.

docs/css/extra.css
pabloarosado: The default table format was ok, but it's a big table. Bastian thought it could be convenient to make it collapsable. I tried a few things with Claued and came up with this solution.
I've fixed the font size in the table.

docs/analyses/slavery_historical_data/index.md
pabloarosado: Hmm, interesting. Do we have any way to track statuses? I feel like we may forget about what's new and leave those statuses as new for a while.

lucasrodes: no, we don't. it's a manual process.

PR #5683, opened 2026-02-16
📊 urbanization: test mdim

veronikasamborska1994:
@lucasrodes sorry for another longish review! I also had a quick question. Do you have an idea of what behaviour to expect if we combine estimates and projections in one view ? We can now show both projections and estimates without losing a map view but I wasn't sure if this would work here so figured I'd ask you first. I am not sure how complicated it would have to end up being given that I also have stacked area charts here but thought I'd ask what you think first.

### DIFF
--- a/etl/steps/data/garden/urbanization/2025-12-10/ghsl_countries.meta.yml
+++ b/etl/steps/data/garden/urbanization/2025-12-10/ghsl_countries.meta.yml
@@ -65,7 +65,7 @@ tables:
 
             - *text_small_territory_caveat
 
-          title: |-
+          title: &title_template |-
             <% if location_type == "rural_total" and metric == 'area' and data_type == 'estimates' %>
             Land covered by rural areas
             <% elif location_type == "rural_total" and metric == 'population' and data_type == 'estimates' %>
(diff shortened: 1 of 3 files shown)


etl/steps/data/garden/urbanization/2025-12-10/ghsl_countries.meta.yml
```diff
@@ -156,14 +215,176 @@ tables:
...
-            The share of total area or population for each urbanization level was calculated by dividing the area or population of each level (cities, towns, rural areas) by the overall total, providing a percentage representation for each category.
-           <% elif metric == 'density' %>
-            Population density was calculated by dividing the population of cities by the total area it covers, providing a measure of the number of people living in each km².
-           <%- endif -%>
+            Shares were calculated by dividing the area or population of each settlement type by the total, then multiplying by 100.
+            <% elif metric == 'density' %>
+            Population density was calculated by dividing the population by the total area of that settlement type.
+            <% elif metric == 'popshare_change' %>
+            The average exponential rate of change was calculated using the formula: ln(percPt/percP0) / 5 * 100, where percPt is the population share at time t and percP0 is the share 5 years earlier. This gives the constant annual growth rate (in percentage points) over the 5-year period.
+            <%- endif -%>
```
lucasrodes, line 228: just checking, is it always a 5-year period?

lucasrodes: i haven't checked 100%, but this seems to be a repetition of `title`. Shall we avoid duplicating it and just reference the already defined field? maybe with YAML anchors (`&` or `*`)

etl/steps/export/multidim/urbanization/latest/cities_towns_rural_areas.py
```diff
@@ -0,0 +1,267 @@
...
+        view
+        for view in c.views
+        if not (
+            (
+                view.dimensions.get("location_type") == "location_type_stacked"
+                or view.dimensions.get("location_type") == "urban_vs_rural"
+            )
+            and view.dimensions.get("metric") in ["popshare_change", "density"]
+        )
+    ]
```
lucasrodes: I'd suggest using `c.drop_views` instead. With that function, you can drop an arbitrary view (or views) that match a certain dimension filter

etl/steps/export/multidim/urbanization/latest/cities_towns_rural_areas.py
```diff
@@ -0,0 +1,267 @@
...
+            view.config["title"] = metadata["title"]
+            view.config["subtitle"] = metadata["description_short"]
+
+            # Only set stack mode if metric dimension exists
+            if metric:
+                # Use line charts for density and popshare_change (not stackable)
+                if metric in ["density", "popshare_change"]:
+                    view.config["chartTypes"] = ["LineChart"]
+                    view.config["hasMapTab"] = False
+                    view.config["tab"] = "chart"
```
lucasrodes: these views seem to be later dropped, don't they? my sense is that you don't need this if

PR #5694, opened 2026-02-19
📊 add international-$ conversion factors for owid-grapher

marcelgerber:
> [!IMPORTANT]  
> I don't really know what I'm doing in etl. Handle this with care!
> Also, feel free to just make changes directly to this branch!

This computes "exchange rates" between int-$ and local currencies.
We need these from the grapher side for a bunch of things:
- for poverty plots (to show the plot in LCUs)
- for How Rich are You (to convert the given income in LCUs into int-$)
- for "int-$ DoDs" (to convert int-$ to LCUs and show the conversion in a tooltip)

This is [the Notion doc behind this](https://www.notion.so/owid/2026-01-13-Converting-int-to-local-currency-2e774b71f92e802397dce3784a624d80?source=copy_link), for context.

Since I don't know how easy it is for you to get to the data here, I've also put it up in a Gsheet: https://docs.google.com/spreadsheets/d/1o-0AJnYRlSC-qEBi4GvB_pz1L91AOWxKK5p6_x1grfs/edit?usp=sharing

## Open questions
- [ ] Should all of this live in `garden` steps, and then there should just be a "simple" `external/grapher` step that pulls that data in and exports it to be available in owid-grapher
- [x] I'd also like to add ISO 4217 currency names & codes to the mix. I've established [an official source](https://www.six-group.com/en/products-services/financial-information/market-reference-data/data-standards.html) ("List One"; yes, this is the official data provider that ISO has elected!), but don't quite know how I would integrate this into the step. Snapshot?
- [x] Something seems to be going wrong with origins, as the meta file mentions UNWPP a bunch of times, but not any of the other origins. Also, I'm getting warnings for unspecified origins for several columns.
- [x] The JSON values we get out are at unreasonably high precision, can we somehow round them for real?

### DIFF
--- a/etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
+++ b/etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
@@ -129,7 +129,7 @@ def load_and_reconcile_ppp_data(ds_wdi: Dataset, ds_pip: Dataset) -> DataFrame:
     # Purchasing Power Parity (PPP) rates (2021 prices)
     tb_pip_ppp = (
         ds_pip["other_indicators"]
-        .rename(columns={"ppp__ppp_version_2021__welfare_type_income_or_consumption": "ppp"})
+        .query("ppp_version == @PPP_YEAR and ppp.notna()")
         .reset_index()
         .groupby("country")[
             "ppp"
--- a/dag/poverty_inequality.yml
+++ b/dag/poverty_inequality.yml
@@ -213,6 +213,6 @@ steps:
     - data://garden/regions/2023-01-01/regions
   data://external/owid_grapher/latest/int_dollar_conversions:
     - data://garden/worldbank_wdi/2026-02-27/wdi
-    - data://grapher/wb/2026-01-22/world_bank_pip
+    - data://garden/wb/2026-01-22/world_bank_pip
     - data://garden/regions/2023-01-01/regions
     - data://garden/worldbank_icp/2024-05-30/icp_2021_currencies


etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
```diff
@@ -0,0 +1,259 @@
...
+    UNCLEAR = "unclear"  # when it's not clear which source is better, due to large deviations or other issues, like recent currency changes or an unclear currency situation. drop these rows for now, but we might want to review them in the future and decide on a case-by-case basis which source to use.
+
+
+PPP_ADDITIONAL_COUNTRIES_TO_EXCLUDE = {
+    # Countries for which we want to drop the PPP value, even if the sources don't deviate much.
+    "Bulgaria": Source.UNCLEAR,  # Bulgaria has introduced the Euro in 2025. Both sources still seem to have data in the Bulgarian lev.
+}
+
+# For each case where the values from PIP and WDI don't line up (either because one has data but not the other, or because their values deviate by more than PPP_PIP_WDI_MAX_DEVIATION), we manually choose which source to use, or whether to exclude the country altogether.
+PPP_DEVIATIONS_SOURCE_TO_USE = {
```
marcelgerber: This is, roughly, what Joe told me to do:
When PIP and WDI line up in value (roughly, up to 3% deviation), then choose the PIP value.

When they don't line up, or one of the values is missing, then we need to decide on a case-by-case basis: If one source has a value but the other one hasn't, we usually just use the value.
If they both have values and we can see that WDI's is more reasonable (because there has been a currency change or redenomination, for example), then we choose WDI.
Otherwise, if we don't quite know what's going on or the situation seems complicated, then we don't use either of the two values and just drop this country.

dag/main.yml
```diff
@@ -14,6 +14,11 @@ steps:
     - data://grapher/demography/2024-07-15/population
   data://external/owid_grapher/latest/gdp:
     - data://grapher/worldbank_wdi/2025-09-08/wdi
+  # Create a step that computes conversion factors between local currencies and international dollars.
+  data://external/owid_grapher/latest/int_dollar_conversions:
+    - data://garden/worldbank_wdi/2026-01-29/wdi
+    - data://grapher/wb/2026-01-22/world_bank_pip
+    - data://garden/regions/2023-01-01/regions
```
marcelgerber: Should this go into the `poverty_inequality` dag? Not sure?

etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
marcelgerber, line 1: I'd like to add another source here that has info about the in-use currency for each country. The data is here: https://www.six-group.com/en/products-services/financial-information/market-reference-data/data-standards.html (this is ISO 4217).

But I wasn't quite sure which way to add that, probably as a snapshot?

snapshots/iso/2026-01-01/iso4217_currencies.py
marcelgerber, line 1: Hey @Marigold, thank you first of all!

There are two things in this XML file that we should watch out for also:
- There are 8 rows that have an extra `IsFund="true"`, e.g. `<CcyNm IsFund="true">WIR Euro</CcyNm>`. These are never useful (and I can't even imagine it being useful in the future), and we can just drop them.
- The country names are not harmonized at all, and contain names like `UNITED STATES OF AMERICA (THE)` or `VENEZUELA (BOLIVARIAN REPUBLIC OF)` that don't match ours. Should we run them through harmonization?

etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
```diff
@@ -227,20 +256,25 @@ def run() -> None:
     # Add OWID country code.
     tb["country_code"] = tb.index.map(country_name_to_code)
 
+    # Add ISO 4217 currency code for each country (informational, for consumers of this dataset).
+    df_iso = load_currency_codes(snap_iso4217)
+    tb["currency_code"] = tb.index.map(df_iso["currency_code"])
```
marcelgerber: Can we also add `currency_name`?

etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
```diff
@@ -0,0 +1,298 @@
...
+    return tb_wdi_cpi
+
+
+def load_currency_codes(snap: Snapshot) -> DataFrame:
+    """Load ISO 4217 currency codes, keeping one row per country (the first/primary currency listed)."""
+    records = json.loads(snap.path.read_text())
+    df = DataFrame(records).dropna(subset=["currency_code"])
+    # ISO 4217 country names are ALL CAPS — normalise to title case for joining.
+    df["country_name"] = df["country_name"].str.title()
+    # Some territories map to multiple currencies; keep the first occurrence (primary currency).
```
marcelgerber: After excluding funds, it might just be safer to drop any countries that still contain duplicates. We can't be entirely sure which currency the PIP/WDI data uses, then.

etl/steps/data/external/owid_grapher/latest/int_dollar_conversions.py
```diff
@@ -0,0 +1,299 @@
...
+        ds_wdi["wdi"]
+        .query("year == @PPP_YEAR and pa_nus_prvt_pp.notna()")
+        .reset_index()
+        .set_index("country")["pa_nus_prvt_pp"]
+    )
+
+    # Purchasing Power Parity (PPP) rates (2021 prices)
+    tb_pip_ppp = (
+        ds_pip["other_indicators"]
+        .rename(columns={"ppp__ppp_version_2021__welfare_type_income_or_consumption": "ppp"})
```
paarriagadap: I would prefer to use the garden step, but if using grapher step, maybe it's better to define this as an f-string, so we avoid hardcoding:
```
.rename(columns={f"ppp__ppp_version_{PPP_YEAR}__welfare_type_income_or_consumption": "ppp"})
```

dag/main.yml
paarriagadap: Yeah, I moved it, so I don't forget updating it with each PIP/WDI update

PR #5718, opened 2026-02-26
owid-catalog v1.0.1: ResponseSet ergonomics (.items, .to_dict(), .to_frame(all_fields=))

lucasrodes:
## Summary

- **Rename `ResponseSet.results` → `.items`** to eliminate the redundant `results.results` access pattern
- **Add `.to_dict()` method** that returns `[item.model_dump() for item in self.items]` — useful for serializing search results into AI/LLM context windows. Use it `results.to_dict()`
- **Add `all_fields` parameter to `.to_frame()`** to temporarily override display mode (basic vs all fields) without mutating instance state
- **Bump version to `v1.0.1`** with changelog entry

## Test plan

- [x] `make check` passes (format, lint, typecheck)
- [x] All 89 tests in `test_client.py` pass (including 2 new tests for `all_fields`)
- [x] Grep confirms no remaining `self.results` or `ResponseSet(results=` references

/schedule

### DIFF
--- a/lib/catalog/tests/test_client.py
+++ b/lib/catalog/tests/test_client.py
@@ -1057,7 +1057,6 @@ class TestResponseSetUI:
         assert rs._ui_advanced is True
 
 
-
 class TestIndicatorsSearchParams:
     """Test indicator search with sort_by and latest parameters."""
 


Marigold (approved):
nice!

lib/catalog/owid/catalog/api/models.py
```diff
@@ -22,44 +23,54 @@ class ResponseSet(BaseModel, Generic[T]):
...
     query: str = ""
     total_count: int = 0
     base_url: str = Field(frozen=True)
 
     # Tweak this to have a more advanced display
     _ui_advanced: bool = False
 
+    @property
+    def results(self) -> list[T]:
+        """Deprecated: use ``.items`` instead."""
```
Marigold: How about just removing it? I don't think we need to keep it backward compatible so early after release.

lucasrodes: technically it's a breaking change, but sure, we can drop it

PR #5719, opened 2026-02-26
🔨 (etl) replace hideLegend with hideSeriesLabels

sophiamersmann:
> [!IMPORTANT]
> Wait for PRs in owid-grapher to land before merging

Renames the chart config's `hideLegend` to `hideSeriesLabels`. Because this is a breaking change, the config schema version is bumped from 009 to 010.

For line charts, slope charts and stacked area charts, `hideLegend` is renamed to `hideSeriesLabels`. For all other chart types, `hideLegend` is dropped.

As far as I can see, all charts here that currently use the `hideLegend` are line charts, so I renamed them to `hideSeriesLabels`

The ETL unit tests are failing because there's no config schema yet at https://files.ourworldindata.org/schemas/grapher-schema.010.json. I guess, I'll wait for the owid-grapher PR to land, then run the test suite again?

I don't understand some of the differences reported by data-diff - seem to be unrelated.

### DIFF
--- a/schemas/explorer-schema.json
+++ b/schemas/explorer-schema.json
@@ -238,19 +238,19 @@
                         "description": "A subset of the grapher config.",
                         "properties": {
                             "$schema": {
-                                "$ref": "https://files.ourworldindata.org/schemas/grapher-schema.009.json#/properties/$schema"
+                                "$ref": "https://files.ourworldindata.org/schemas/grapher-schema.010.json#/properties/$schema"
                             },
                             "addCountryMode": {
-                                "$ref": "https://files.ourworldindata.org/schemas/grapher-schema.009.json#/properties/addCountryMode"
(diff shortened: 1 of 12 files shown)


schemas/explorer-schema.json
```diff
@@ -290,9 +290,6 @@
                             "matchingEntitiesOnly": {
                                 "$ref": "https://files.ourworldindata.org/schemas/grapher-schema.009.json#/properties/matchingEntitiesOnly"
                             },
-                            "hideLegend": {
-                                "$ref": "https://files.ourworldindata.org/schemas/grapher-schema.009.json#/properties/hideLegend"
-                            },
```
sophiamersmann, line 295: Dropped without replacement because explorers never supported the `hideLegend` field (and I also didn't add support for the new `hideSeriesLabels` field).

Marigold (approved):
It should be good based on [docs](https://docs.owid.io/projects/etl/guides/schema-management/?h=schema+mana#schema-management) I've created from the past version bump.

PR #5721, opened 2026-02-26
🔨🤖 update tags for datapage metadata

ikesau:
I gave the `tagless-datapages.csv` file to Claude and told it the issue and it wrote a script to update the metadata yml files where appropriate.

I haven't validated its work yet. Will do so once this staging site deploys.

### DIFF
--- a/etl/steps/data/garden/crime/2026-01-21/prison_rates.meta.yml
+++ b/etl/steps/data/garden/crime/2026-01-21/prison_rates.meta.yml
@@ -25,10 +25,7 @@ tables:
         unit: prisoners per 100,000 people
         display:
           numDecimalPlaces: 0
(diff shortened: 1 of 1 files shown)


lucasrodes:
@ikesau heads up, in case you are not aware, you can define topic tags at 'dataset level'. so, if the same tags apply to all indicators (maybe for covid datasets), you just define it once.

basically, need to edit `definitions.common` property. Here you have an example on ho to do it: https://docs.owid.io/projects/etl/architecture/metadata/structuring-yaml/?h=common#anchors-aliases

Optionally, you can also do this at table-level, just use `common` property in the desired table.

Leaving this option here, in case you want to feed this into claude. Up to you! Thanks for looking into this, btw

PR #5773, opened 2026-03-15
📊 ai: add MMLU

veronikasamborska1994:


veronikasamborska1994:
> @codex review

@Marigold I think I just wasn't logged in here for some reason. I don't really usually use codex so that's probably why :)

PR #5803, opened 2026-03-23
📊 Remove processing log

Marigold:
## 👨 

So long, processing log. It has been an honor, but times have changed, and you're no longer useful now that we can get AI summaries of steps.

## 🤖 

Remove the experimental processing log system that tracked data transformation provenance via a DAG of operations on each variable. The feature was opt-in (disabled by default via `PROCESSING_LOG` env var) and is no longer useful.

- Delete `processing_log.py` module and its tests
- Remove `processing_log` field from `VariableMeta`
- Remove all processing log tracking from Table/Indicator operations (save, load, rename, dropna, sort, arithmetic, groupby)
- Keep `update_log`/`amend_log` methods as no-ops for backwards compatibility
- Clean up imports and documentation references

### DIFF
--- a/lib/catalog/owid/catalog/__init__.py
+++ b/lib/catalog/owid/catalog/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.0.1"
+__version__ = "1.1.0"
 
 from owid.catalog import api
 from owid.catalog.api import Client, fetch, search


Marigold:
Yeah, it's sad. I was honestly surprised how far we got, but AI ruins everything...

Marigold:
@codex review

PR #5805, opened 2026-03-23
📊 Bump all dependencies to latest versions

Marigold:
@lucasrodes this also updates Streamlit to 1.55.0. I've only fixed compatibility, but it also has tons of [new features](https://docs.streamlit.io/develop/quick-reference/release-notes) like dynamic containers or `bind="query_params"` which could replace our `url_persist`. If you think we shouldn't rush it, I can keep it on the old version.


lucasrodes (approved):
Good cleaning. I think we can merge. The `bind` seems definitely worth looking into, so I've created issue #5817 for later look.

PR #5811, opened 2026-03-23
🎉🤖 Add air pollution multidim

HannahRitchie:
Hey @antea04,

This is our first try building MDims via Claude Code. Hopefully things go smoothly, but let me know if the process fails in some way.

This MDim will be useful for our Wellcome work, and brings together lots of disjointed charts we have on Air pollution deaths.

Could you review it for me, please?

### DIFF
--- a/.claude/skills/create-multidim/SKILL.md
+++ b/.claude/skills/create-multidim/SKILL.md
@@ -97,6 +97,8 @@ Place it right after the grapher step it depends on.
 
 ### Step 5: Run and verify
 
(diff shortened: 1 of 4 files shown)


HannahRitchie:
@Marigold just tagging you here in case there's anything useful in seeing how this process plays out.

HannahRitchie:
Thanks very much @antea04.

I actually didn't know I could incorporate "What you should know" bullets into this via Claude. It initially told me this wasn't possible, then changed its mind...

After some wrangling, I think I managed to get it to work and submit an updated PR request. Hopefully it works (and thanks for patience with it/me!)

If you're happy with it, you can then just merge and publish. But let me know if there are more issues.

PR #5815, opened 2026-03-23
🔨 Migrate add_population() to UN WPP 2024-07-12 schema

Marigold:
Stacked on top of #5814.

## What

Rewrites `etl.data_helpers.population.add_population()` to use the `2024-07-12/un_wpp` dataset instead of the old `2022-07-11` one.

## Why

The `2022-07-11` dataset had a custom `population_granular` table with single-year age resolution. The `2024-07-12` dataset restructures this into a standard `population` table with 5-year age buckets, separate columns per metric, and separate `estimates`/`medium` variants for historical vs projected years.

## How

**`population.py`**
- Reads `population` table (was `population_granular`)
- Filters to `variant in ['estimates', 'medium']` to cover all years (was just `'medium'`)
- New `_select_age_buckets()` helper replaces integer range arithmetic:
  1. First tries an **exact match** — handles pre-aggregated buckets already in the data: `'all'`, `'15+'`, `'18+'`, `'65+'`, `'0-4'`, `'1-4'`, etc.
  2. Falls back to selecting the **minimal non-overlapping atomic sub-buckets** that fit inside the requested range (e.g. `[5, 14]` → `['5-9', '10-14']`)

**DAG updates** — migrated all `add_population` callers from `2022-07-11` → `2024-07-12`:
- `ihme_gbd/2024-05-20/gbd_risk`
- `ihme_gbd/2026-02-07`: gbd_mental_health, gbd_cause_deaths, gbd_cause_dalys, gbd_prevalence, gbd_risk_deaths, gbd_risk_dalys
- `who/2024-07-30/ghe`
- `homicide/2025-08-06/who_mort_db`
- `tuberculosis/2026-02-05/burden_disaggregated`

Steps using `2022-07-11` for other purposes (terrorism, gapminder fertility, demography/population) are intentionally left untouched.


lucasrodes:
@Marigold i have this on my radar, will try to review this in the coming days. sorry for the delay!

Marigold:
@lucasrodes would you mind checking this out?

lucasrodes:
Really sorry for the delay, @Marigold! This LGTM, thanks for fixing it!

One minor concern about the new bucket-aware logic in `_select_age_buckets` (`etl/data_helpers/population.py`): the returned buckets are guaranteed to lie *inside* `[req_min, req_max]` but not to *cover* it. So requests that don't align with the UN WPP 2024 bucket grid can silently under-cover.

Two failure modes worth flagging:

1. **Silent NaN** — for a request with no bucket inside it (e.g. `[15, 17]`), `_select_age_buckets` returns `[]`, the loop in `add_population` only `log.warning`s and `continue`s, and the population becomes NaN for that group. IHME steps catch this via `tb_number["value"].notna().all()`, but other callers won't.

2. **Silent under-coverage** (worse) — request `[2, 9]`: `0-4` doesn't qualify (`b_min=0 < 2`), `5-9` does. The function returns `['5-9']` and the result is reported as the population of "ages 2-9" — silently missing ages 2–4. No warning fires because `buckets` is non-empty.

All currently migrated callers (IHME GBD, WHO GHE, WHO MortDB, TB burden) happen to use ranges that align with the bucket grid, so they're fine, but maybe an issue in the future.

Suggestion: inside `_select_age_buckets` (or right after the call in `add_population`, around the `if not buckets` check), assert that the union of selected buckets equals `[req_min, req_max]` and raise otherwise. Plus a docstring note that `age_group_mapping` ranges must align with the UN WPP bucket grid.

lucasrodes (approved):

Marigold:
Thanks for the careful read, @lucasrodes! Fixed in 0e3bab4f7e:

- `_select_age_buckets` now verifies the selected atomic buckets fully and contiguously cover `[req_min, req_max]`. Both failure modes you flagged — empty bucket list and silent under-coverage like `[2, 9]` → `['5-9']` — now raise `ValueError` naming the request range and the available buckets.
- Dropped the `log.warning` + `continue` block in `add_population` so the new error propagates instead of producing NaN.
- Docstrings on both functions note that ranges must align with the UN WPP bucket grid.
- Added `tests/data_helpers/test_population.py` covering exact-match, multi-bucket coverage, and the misalignment cases (`[2, 9]`, `[15, 17]`, `[0, 6]`).

lucasrodes:
Thanks, @Marigold! Let's go!

PR #5855, opened 2026-03-30
📊 Archive faostat/2025-03-17 steps

Marigold:
Archive all faostat/2025-03-17 steps (meadow, garden, grapher). These steps take almost an hour to run and are no longer used by any active steps, so archiving them to save CI time.

Note: `agriculture/2025-08-06/decoupling_food_production_and_land_use` is kept in the active DAG as it has active charts — its faostat dependencies are resolved from the archive DAG.


Marigold:
@pabloarosado would you mind checking it out?

pabloarosado:
> @pabloarosado would you mind checking it out?

Hey @Marigold thanks for looking into this. I'm still waiting for Max's feedback on some related issues. I'd prefer to keep the old steps for a bit longer, and I'll take care of the archiving once those issues are clarified and the old steps are safe to be archived.

Is this blocking anyone or causing any trouble (apart from the annoyance of the extra compute time)? If so, I can figure out some temporary solution.

Marigold:
Not a blocker at all, the ball is on your side then. Feel free to close this or merge when the time is right.

PR #5862, opened 2026-03-31
📊 Fix wizard page redirect flash

Marigold:
Move the `?page=` query parameter redirect logic to the top of `st_show_home()`, before any UI rendering happens.

Previously, the redirect check was at the very bottom of the function — after the entire home page (title, cards, sections) was rendered. This caused a ~2 second flash of the home page before `st.switch_page()` kicked in.

Now when visiting e.g. `/etl/wizard/?page=chart-diff`, the redirect fires immediately with no visible flash.


Marigold:
@codex review

PR #5909, opened 2026-04-10
📊 Fix WHO FluID meadow step crash

Marigold:
## Bug Fix: WHO FluID meadow step crash

The `data://meadow/who/latest/fluid` step was crashing with:

```
ValueError: Column iso_year is still object. Consider converting it to str.
```

### Root cause

Two issues combine to cause the crash:

1. **`pd.read_csv()` with default `low_memory=True`** reads the CSV in chunks, and when chunks disagree on column types, the column gets `object` dtype. Column 7 (`iso_year`) was affected.

2. **One malformed row in the source data** has shifted columns — a date value (`2024-09-01`) ends up in the `iso_year` field — which poisons type inference.

The `repack` step then rejects the `object`-typed `iso_year` column.

### Fix

- Add `low_memory=False` to `pd.read_csv()` to prevent chunked mixed-type inference
- Coerce `iso_year` to numeric and drop the one malformed row

@codex review

### DIFF
--- a/etl/steps/data/meadow/who/latest/fluid.py
+++ b/etl/steps/data/meadow/who/latest/fluid.py
@@ -23,7 +23,8 @@ def run(dest_dir: str) -> None:
     snap = paths.load_snapshot("fluid.csv")
 
     # Load data from snapshot.
-    df = pd.read_csv(snap.path)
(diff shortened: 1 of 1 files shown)


Marigold:
@codex review

PR #5912, opened 2026-04-10
📊 ai: update data robot density

veronikasamborska1994:


dag/artificial_intelligence.yml
```diff
@@ -268,6 +268,15 @@ steps:
   data://grapher/ifr/2026-01-14/robot_density:
     - data://garden/ifr/2026-01-14/robot_density
 
+  # IFR robot density 2024 data
+  data://meadow/ifr/2026-04-10/robot_density:
```
antea04, line 270: can we archive the previous steps (above this) or are they still being used for something?

antea04 (commented):
LGTM! 

I was wondering if you thought about enabling the line/ slope tab of the chart as well (now that you have two years of data)? 

It's not super insightful atm, but could be interesting for some people to explore - but I totally understand waiting for another year of updates as well!

antea04 (approved):

dag/artificial_intelligence.yml
veronikasamborska1994: I still need the garden step but I archived the grapher step :)

PR #5940, opened 2026-04-16
Mark all skills as internal to prevent external indexing

Marigold:
## Summary

- Added `metadata: { internal: true }` to all 14 skills in `.claude/skills/`
- Added a rule to CLAUDE.md requiring this flag on new skills by default

External skill marketplaces (skills.sh, skillsmp.com) were crawling our public repo and listing our internal skills. The `metadata.internal` flag is the standard mechanism from the [skills CLI](https://github.com/vercel-labs/skills) to hide skills from discovery — they'll only be installable when `INSTALL_INTERNAL_SKILLS=1` is set.

Our skills still work normally for anyone with the repo cloned. This only affects external index visibility.

## Test plan

- [ ] Verify skills still load in Claude Code (`claude --debug`)
- [ ] Confirm `npx skills add owid/etl --list` no longer shows skills (without `INSTALL_INTERNAL_SKILLS=1`)

### DIFF
--- a/.claude/skills/chart-editing/SKILL.md
+++ b/.claude/skills/chart-editing/SKILL.md
@@ -1,6 +1,8 @@
 ---
 name: chart-editing
 description: Edit and preview .chart.yml files for OWID graph steps. Use when user wants to edit chart config, preview charts, change chart appearance, or work with graph step chart files.
+metadata:
+  internal: true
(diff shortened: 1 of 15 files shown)


lucasrodes (approved):
lgtm!

PR #5946, opened 2026-04-17
:sparkles: Include snapshots in `etlr --watch` file watcher

paarriagadap:
## Summary

`etlr --watch` previously only observed `etl/steps/`, so snapshot changes were silently ignored. This PR extends watch mode to also track `.dvc` files under `snapshots/`.

### Design note

An earlier iteration of this PR also auto-ran `etls` on snapshot `.py` / `.dvc` edits. Per discussion with @Marigold, that approach pulled in a lot of edge cases (driver scripts, `--path-to-file` manual-upload snapshots, metadata-only redownloads of huge files, R2 object churn) — and it's unnecessary, because downstream steps already key off the `.dvc` checksum.

The workflow is therefore:

```
# Terminal A
etlr my_step --grapher --watch

# Terminal B (whenever you want fresh data)
etls my_snapshot
```

`etls` in terminal B rewrites the `.dvc` (md5 and/or metadata), which terminal A notices and rebuilds from automatically.

### Changes

1. **Watch `snapshots/` too.** `watch_folder` / `_mtime_mapping` in [etl/files.py](etl/files.py) are generalized to accept multiple paths; [etl/command.py](etl/command.py) passes both `paths.STEP_DIR` and `paths.SNAPSHOTS_DIR`.
2. **Under `snapshots/`, only `.dvc` files are watched** (not `.py`). A `.py` edit on its own doesn't propagate until `etls` rewrites the companion `.dvc`, so watching `.py` would cause spurious rebuilds off pre-`etls` script state. Files elsewhere (e.g. `etl/steps/`) are tracked as before.

## Test plan

- [x] `.venv/bin/etlr <step> --grapher --watch` in terminal A.
- [x] In terminal B, `etls <snapshot>` for a snapshot the step depends on → terminal A logs `--- File changed: <dvc>` and rebuilds.
- [x] Edit a `.py` under `snapshots/…` without running `etls` → no rebuild (expected: only the `.dvc` is watched).
- [x] Edit an `.meta.yml` / `.py` under `etl/steps/…` → still triggers a rebuild (regression check of pre-existing behavior).
- [x] `make check` passes.


paarriagadap:
@codex review

Marigold:
@paarriagadap neat, I've completely forgot about snapshots when adding `--watch` mode. Have you considered watching only `*.dvc` files with metadata (and not `.py` that create them)? When you change snapshot metadata (like origin or checksum by reuploading the snapshot), `etlr` will detect that and rerun meadow, etc. with proper snapshot metadata. So if the problem was only about metadata, this would be the simplest solution.

I'm a bit worried about running `etls` on every change. Snapshot scripts might use `--path-to-file`, can be very slow and each run uploads a file to R2 that stays there potentially forever. A better workflow would be to run
```
etlr abc --watch
```
then if you run `etls abc` in another terminal, it'll upload new snapshot to R2, change checksum in `.dvc` file and the `etlr` process above will run automatically.

What do you think? Would it be enough for your use case?

paarriagadap:
Hi @Marigold!

- I did consider watching the dvc files only, but I wanted to cover cases where the data is uploaded via the script like child_labor_report
- The case with path-to-file is covered by not running etls

But you're probably right (Claude agrees with you!), so I am simplifying the code to watch changes to the dvc only and if I find in the future that we need to add more we can go back to it

paarriagadap:
@codex review

Marigold (approved):
Thanks for a nice improvement! Approved, if we find that watching `.py` snapshot files is helpful, I'm open to making it more complex.

PR #5915, opened 2026-04-13
✨ improve faust of demography explorer

lucasrodes:
Resolves parts of owid/owid-issues#2084 — improving FAUST in the [Population & Demography Data Explorer](https://ourworldindata.org/explorers/population-and-demography).

### 1. Apply FAUST edits from `faust_suggestions_for_lucas.csv`
Source: [edomt's comment](https://github.com/owid/owid-issues/issues/2084#issuecomment-4224254010) with [faust_suggestions_for_lucas.csv](https://github.com/user-attachments/files/26632078/faust_suggestions_for_lucas.csv) covering 254 views.

- [x] Apply new **titles** (plain-English, e.g. "Number of children under 5" vs. "Population of children under the age of 5")
- [x] Apply new **subtitles** (add where missing: dependency ratios, age structure, sex ratios, …)
- [x] Apply new **footnotes** (remove noise like "Values as of 1 July of the indicated year.")
- [x] Apply **rounding** per indicator (≈4 sig figs for counts, 1–2 decimals for rates)
- [x] Rewrite **life expectancy** FAUST (incl. Max's fix: "Period life expectancy for individuals who have reached age 0" → "Period life expectancy at birth")
- [x] Add **DoD links** (fertility rate, age-specific fertility rate, period life expectancy)
- [x] Respect `<empty>` cells → leave field intentionally blank

### 2. Explorer changes beyond the CSV
Also from [edomt's comment](https://github.com/owid/owid-issues/issues/2084#issuecomment-4224254010):

- [x] **Indicator dropdown**: "Population by broad age group" → "Population by age group" (5668cb6e)
- [x] **Indicator dropdown**: "Number of child deaths" → "Child deaths"; "Number of infant deaths" → "Infant deaths" (95ae5e07)
- [x] **Age dropdown (Fertility rate & Births)**: "Mothers aged 15–19 years" → "Women aged 15–19 years" (denominator is all women, not just mothers) (0fd9331d, 8f2abafd)
- [x] **Age dropdown (Life expectancy)**: "Under 1 year" → "At birth" (also flagged by [Max](https://github.com/owid/owid-issues/issues/2084#issuecomment-3407437538))
- [x] **New age filter "15–49 years"** for Population & Annual population change (sum of UN 5-year groups 15-19 … 45-49; from [Max's suggestion](https://github.com/owid/owid-issues/issues/2084#issuecomment-3772250554))
[...]

PR #5950, opened 2026-04-17
:bug: Fix DoD hover in YAML files

paarriagadap:
## Summary

Fixes the DoD hover in YAML/Python files so it actually renders, and points it at the DB where DoDs are actually edited.

## 1 · Ship the on-demand hover fetch (v0.0.18 → v0.0.19)

The installed v0.0.18 only rendered a hover when the DoD was already in the in-memory cache — it relied entirely on the `preloadDodDefinitions` path, and the hover handler itself never called `fetchDodDefinition`. First-time hovers silently returned nothing.

That fix landed in #5915, but the version wasn't bumped — so `make vsce-sync` kept everyone on the broken v0.0.18. This PR ships it as v0.0.19.

## 2 · Read DoDs from staging first, not production

DoDs are edited on a staging server first and only synced to live_grapher later, so the hover while developing should read the developer's own staging DB. But the original script fell back to a hardcoded `owid@prod-db` config that didn't match any real machine, and a naive "just return `OWID_ENV`" fallback would silently query whatever DB the developer's env pointed at — flagged by Codex on the earlier iteration of this PR.

New design: walk a prioritized list of candidate envs and use the first one that accepts a connection:

1. `OWID_ENV` when `STAGING` is set — developer's staging server (current source of DoD edits).
2. `ENV_FILE_PROD` (explicit prod env file).
3. `.env.prod` in the repo root, by convention.
4. The repo's `.env`, read raw via `dotenv_values` — only if `DB_NAME=live_grapher`. Reading raw bypasses the STAGING override that `etl.config` would otherwise apply to `OWID_ENV`.
5. `OWID_ENV` as final fallback (local dev DB, etc.).

Each response carries a `source` label. The hover renders it unconditionally as *Source: `production`* or *Source: `staging (staging-site-foo)`*, so users can always tell where the content came from at a glance.

Falls through on `sqlalchemy.exc.OperationalError` — if the bore tunnel to live_grapher is down, or the staging server isn't reachable, the next candidate is tried instead of erroring.

Also fixed the repo-root path: the old `etl_root = parent.parent.parent` actually pointed at `vscode_extensions/`, which broke the `.env` / `.env.prod` lookups.

## Files changed

- `vscode_extensions/dod-syntax/package.json` + `package-lock.json`: 0.0.18 → 0.0.19
- `vscode_extensions/dod-syntax/install/dod-syntax-0.0.19.vsix`: freshly built with all of the above
- `vscode_extensions/dod-syntax/install/archived/dod-syntax-0.0.18.vsix`: old vsix archived
- `vscode_extensions/dod-syntax/scripts/fetch_dod.py`: staging-first candidate list + try/catch fall-through + `source` field in the response
- `vscode_extensions/dod-syntax/src/extension.ts`: always render *Source: `...`* in the hover

Built with `make vsce-compile EXT=dod-syntax` under Node 22.

Note: the Python script is spawned by the extension from the **workspace** path (not from the installed vsix), so the script fix takes effect immediately on next hover — no VSCode reload required. The version bump is only needed to re-deliver the compiled `extension.js`.

## Test plan

Verified on this branch:

- [x] `code --list-extensions --show-versions | grep dod-syntax` → `owid.dod-syntax@0.0.19`.
- [x] Standalone script returns the DoD from staging with `"source": "staging (staging-site-data-dodhover-yaml)"`:
      `.venv/bin/python3 vscode_extensions/dod-syntax/scripts/fetch_dod.py child-labor`
- [x] Fall-through on `OperationalError` works: with the bore tunnel down (connection refused on 127.0.0.1:3310), the script still returned content via the staging candidate instead of erroring.
- [x] `make check` passes.

Please reviewer-verify:

- [x] `make vsce-sync` upgrades from 0.0.18 → 0.0.19 on a machine that hadn't received this build yet.
- [x] After reloading VSCode, hovering over `[child labor](#dod:child-labor)` on line 128 of `etl/steps/data/garden/un/2026-04-08/child_labor_report.meta.yml` shows the DoD content and a `*Source: staging (staging-site-<branch>)*` line.
- [x] On a non-staging checkout with `.env` pointed at live_grapher (`STAGING` unset), the same hover shows the DoD content and `*Source: production*` instead.
- [x] Typing a new `[foo](#dod:some-dod)` and hovering immediately resolves — covering the on-demand fetch path (not just preload).


paarriagadap:
@codex review

paarriagadap:
@codex review

lucasrodes (approved):
Thanks, Pablo! LGTM

PR #5953, opened 2026-04-20
📊 Fix search API semantic search init

Marigold:
## Summary

Fixes `GET /indicators` on `search.owid.io` returning 503 for every request. Root cause was stale-pickle deserialization of `SentenceTransformer` from the joblib cache after the `sentence-transformers` major version bump (2.2.2 → >=5.1.0) in #4866 — the unpickled object was missing `tokenizer`, so `EmbeddingsModel.__init__` raised `'SentenceTransformer' object has no attribute 'tokenizer'`. The failure was swallowed by a background-thread `except Exception` and surfaced to callers as a misleading "still initializing" 503, with nothing reaching Sentry.

## Changes

**`apps/wizard/utils/embeddings.py`**
- Drop `@memory.cache` on `get_model`. Pickling a PyTorch model via joblib is fragile across dependency upgrades.
- Use `SentenceTransformer(model_name, local_files_only=True)` with an `OSError` fallback that re-fetches from HF. Measured: first load ~0.4s, subsequent ~0.08s — faster than the joblib hit (~0.1s) and no staleness hazard.
- Derive `EmbeddingsModel.model_name` from `model.tokenizer.name_or_path` (stable on 5.x). Callers may still override it to namespace cache files (e.g. `"sim_charts_title"`). Killed the outdated "hack / TODO" comment.

**`api_search/semantic_search.py`**
- Call `sentry_sdk.capture_exception(e)` in the startup background thread. `SentryAsgiMiddleware` only sees HTTP-handler exceptions, so startup crashes were invisible until now.
- Expose `is_initialization_complete()` / `get_initialization_error()` for endpoints to distinguish the two states.

**`api_search/v1/__init__.py`**
- Differentiate responses: 503 "still initializing" only while init is in progress; 500 with the real error once init has finished unsuccessfully. Previously both cases returned the same 503.

## Test plan

- [x] `get_model()` hot/cold timings measured locally
- [x] `EmbeddingsModel(model)` derives `all-MiniLM-L6-v2`; override still works
- [x] `make check` (ty) passes
- [ ] Verify on staging that `/indicators` returns real results and `/indicators/info` shows `ready: true`
- [ ] Confirm a forced init failure on staging now shows up in Sentry

### DIFF
--- a/api_search/semantic_search.py
+++ b/api_search/semantic_search.py
@@ -3,6 +3,8 @@
 import threading
 from typing import Any
 
+import sentry_sdk
+
 from apps.wizard.app_pages.indicator_search.data import Indicator, _get_data_indicators_from_db
 from apps.wizard.utils.embeddings import EmbeddingsModel, get_model
 from owid_mcp.data_utils import build_catalog_info
@@ -31,6 +33,7 @@ def _initialize_semantic_search():
 
         _initialization_complete = True
     except Exception as e:
+        sentry_sdk.capture_exception(e)
         _initialization_error = str(e)
         _initialization_complete = True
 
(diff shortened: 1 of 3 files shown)


Marigold:
@codex review

PR #5961, opened 2026-04-22
📜 Add chart-text-report skill

paarriagadap:
## Summary

Add a new skill `chart-text-report` that generates a compact Markdown audit of the user-facing text Grapher shows on charts — Title, Subtitle, Footnote (a subset of FAUST) plus indicator-level `description_short` and `description_key`. Each field is tagged `[override]` (set on the MDim view), `[inherited]` (pulled from the indicator's ETL metadata), or `[missing]` (absent in both).

Built by generalizing the iteration that produced the editorial FAUST reports for the four inequality MDims (`incomes_pip`, `gini_pip`, `gini_lis`, `incomes_wid`).

## Why this is useful

The output is designed to be **pasted straight into a Google Doc** to review and discuss changes to the FAUST and metadata with colleagues. All the text a reader actually sees on a chart shows up in one flat outline — with chart-title headings, human-readable dimension selections, preview links, and a deduplicated legend for description_key bullets — so non-technical collaborators can review copy without opening every chart or hunting through `.meta.yml` files.

## What's included

- `.claude/skills/chart-text-report/SKILL.md` — trigger guidance, inheritance rules, output-format spec, workflow, things to avoid.
- `scripts/_common.py` — shared helpers: grapher-channel metadata loader, inheritance resolvers, `BulletLibrary` (description_key dedup with auto-generated slugs), stopwords, preview-URL builder, `how_to_read_block`, `collect_used_tags`.
- `scripts/generate_mdim_text_report.py` — MDim view mode. Supports `collapse_dims` grouping with `{dim}` placeholder parametrization, human-readable selection subheaders, heading disambiguation when multiple groups share a title, and per-MDim JSON config.
- `scripts/grapher_dataset_mode.py` — whole-dataset or hand-picked indicator-list mode. Reuses all the shared helpers; every tag is `[inherited]`/`[missing]` (no view-level overrides exist).

## The critical rule captured

Chart-level FAUST (Title / Subtitle / Footnote) resolves **only** from `presentation.grapher_config.{title, subtitle, note}`. Falling back to `title`, `presentation.title_public`, `display.name`, or `description_short` produces text that does not match what Grapher actually renders — those are data-page fields. The skill documentation flags this explicitly to avoid rederiving the mistake.

## MDim rebuild workflow

The skill's rebuild recipe is `etlr <mdim> --export --private`. No DB-bypass helpers — if MySQL is unreachable, the skill reports the error and stops. Dropping `--grapher` from the default is deliberate: the FAUST report reads metadata directly from the local grapher-channel feather files, so the slow `grapher://<dataset>` MySQL upload that `--grapher` triggers isn't needed. Change detection still re-runs any garden or MDim step whose code/yaml changed, so editing metadata still takes effect. Only add `--grapher` when you need indicator data/metadata pushed to MySQL for live rendering.

Measured: forced single-MDim rebuild ~4 s without `--grapher` (vs ~60 s with it on WID); ~2 s when nothing changed.

## Codex review fixes (commit a2b1a8a05)

Addressed three flagged issues in `scripts/_common.py`:

- **Surface metadata load failures** — `get_indicator_meta` no longer swallows broad exceptions; only a missing column returns `None` (legit `[missing]` tag). Malformed paths, wrong channel, missing datasets, and unreadable feathers now propagate so broken state can't silently masquerade as `[missing]`.
- **Enforce grapher channel** — `get_indicator_meta` now raises on non-grapher catalog paths (previously a `garden/...` path would be silently resolved against grapher storage).
- **Preserve explicit empty-string overrides** — `resolve_field` no longer treats `note: ""` (MDim view intentionally clearing an inherited footnote) as "unset". Empty strings are now proper overrides, rendered as `_(empty)_` so the report matches what Grapher displays. Spot-check: 11 `incomes_wid` views + 2 `gini_pip` views now correctly show their empty-footnote override instead of the inherited note.

## Legend is scoped to tags that actually appear (commit 077f0ab21)

`how_to_read_block` now takes a `used_tags: set[str]` and only documents tags that occur in the rendered body. The body is built first, scanned via `collect_used_tags`, and the header is appended afterwards. This is a cleaner presentation — e.g. after the empty-string-override fix, no field in the four inequality MDims is tagged `[missing]`, so the `[missing]` bullet is now omitted from those files' legends.

## Why a skill

The inheritance rules, the output format decisions (dedup + auto-slug, human-readable selections, parametrized placeholders, heading disambiguation, tag legend scoped to what actually appears), and the rebuild recipe are all reusable well beyond the four MDims that first needed them. Keeping it as a skill (with runnable scripts) lets future agents produce these reports without rediscovering the pitfalls.

## Test plan

- [x] `generate_mdim_text_report.py` regenerates the four inequality MDim reports after rebuilding their `.config.json` via `etlr`.
- [x] `grapher_dataset_mode.py --dataset data/grapher/lis/2026-03-16/luxembourg_income_study` renders a 640-indicator report with `[inherited]`/`[missing]` tags and a bullet legend.
- [x] Heading disambiguation verified: in `incomes_pip`, previously-colliding headings like `Mean income or consumption per {period}` now render with a `(Breaks in data: Don't show)` / `(Breaks in data: Show)` suffix.
- [x] Slug cleanup verified: description_key slugs like `income-_pre-tax_-measured` now render as `income-pre-tax-measured` (underscores treated as word boundaries).
- [x] Empty-footnote override rendered as `_(empty)_` in incomes_wid / gini_pip; the skill's output now matches what Grapher renders for views that intentionally clear inherited text.
- [x] Legend pruning verified: `[missing]` bullet removed from all four files' legends now that no field is tagged `[missing]`.
- [x] `ruff check` passes on all scripts.
- [x] `etlr <mdim> --export --private` (no `--grapher`) rebuilds the config.json via the full ETL path on a staging server in ~2–4 s.
- [x] Paste one of the generated reports into a Google Doc and confirm the formatting (bullet nesting, bold field labels, preview links) survives the copy/paste.


paarriagadap:
@codex review

pabloarosado (approved):
Hey, I won't have the bandwidth to properly review (it's a lot of code!) or test the skills.
I think that having a lot of code across skills shouldn't affect token usage in other conversations (only the title and frontmatter description of each skill are always loaded, and the rest is pulled in on demand). So as long as these skills work well for you and the descriptions are concise and clear, it should be fine.

paarriagadap:
Thanks @pabloarosado! Yeah, I agree. I just wanted you to check any improvements to it, in the context of improving the workflow between data scientists and writers. Maybe I should have written that out.

Anyway, the skill will be available to play with it, without interfering other processes.

pabloarosado:
Hi @paarriagadap, feel free to merge, we can start using it and improve it further if needed later on. It may also inspire other alternative solutions to help in our workflow. Thanks a lot for the initiative!

PR #5962, opened 2026-04-22
✨ compact nested DAG syntax

### LINKED ISSUE #5881: 💡 ergonomics: simplify dag syntax for basic steps
Most of our steps follow the structure snapshot → meadow → garden → grapher, and are defined in a dag like:

```yaml
data://meadow/un/2022-07-11/un_wpp:
  - snapshot://un/2022-07-11/un_wpp.zip
data://garden/un/2022-07-11/un_wpp:
  - data://meadow/un/2022-07-11/un_wpp
```

I think we could allow for a simpler syntax for these cases, which is just equivalent:

```yaml
  - data://garden/un/2022-07-11/un_wpp:
    - data://meadow/un/2022-07-11/un_wpp:
      - snapshot://un/2022-07-11/un_wpp.zip
```

Marigold:
Closes #5881.

## Summary

Adds support for a compact nested DAG YAML syntax so a linear
`snapshot → meadow → garden → grapher` chain can be written once:

```yaml
  - data://garden/un/2022-07-11/un_wpp:
    - data://meadow/un/2022-07-11/un_wpp:
      - snapshot://un/2022-07-11/un_wpp.zip
```

The parser accepts both flat and nested forms side by side; the in-memory
graph shape is unchanged, so every caller of `load_dag()` keeps working.

## Writing to DAG files

The line-based writers in `etl/dag_helpers.py`
(`write_to_dag_file`, `_remove_step_from_dag_file`,
`create_dag_archive_file`) previously detected steps by matching lines
that end with `:` and do NOT start with `-`, which silently ignored
every nested step declaration. Each writer now calls `flatten_dag_file`
on its first line so the existing logic always operates on a flat file,
and callers from the Wizard / `etl update` / step_update archive flow
keep working without modification.

## Formatter

* `etl dag compact [FILE ...]` folds a dep `D` into parent `P` iff `D`
  has exactly one global consumer (`P`). Shared dependencies stay flat
  so every step is declared once. Chain-head comments migrate to the
  surviving outer step. Refuses to fold when inline dep comments would
  be orphaned. Idempotent in a single pass.
* `etl dag flatten [FILE ...]` is the inverse — useful for reviewing
  diffs and investigating a chain.
* `make format-dags` wraps `etl dag compact` but is not wired into
  `make format` by default (it loads the full DAG, which is heavier
  than ruff).

## Migration

Applied `etl dag compact` across every active `dag/*.yml`: 41 files
rewritten, net **−1577 lines**, `dag/archive/*.yml` untouched. The
full in-memory `load_dag()` result (2886 steps, full dep sets) is
byte-identical before and after the rewrite.

## Test plan

- [x] `.venv/bin/pytest tests/test_dag_utils.py -x` — 25 tests pass,
  including new coverage for nested parsing, mixed flat+nested,
  duplicate/multi-key errors, flatten no-op on flat files,
  compact idempotence, shared-dep behaviour, and writer operations
  on nested input.
- [x] `.venv/bin/pytest -m "not integration" tests -q` — 441 pass.
- [x] `ty check` + `ruff check` + `ruff format --check` clean.
- [x] `.venv/bin/etlr un/2022-07-11/un_wpp --dry-run` against the
  compacted DAG correctly resolves the full 3-step chain.
- [x] `etl dag compact` is idempotent — re-running on a compacted tree
  touches zero files.
- [x] `etl dag flatten dag/*.yml` + `etl dag compact dag/*.yml`
  round-trips to the same content.


Marigold:
@codex review

Marigold:
@lucasrodes do you think it's worth adding? I'm not sure if such syntactic sugar adds much value now that everything is autogenerated. But happy to take this over the finish line if you'd find it useful!

lucasrodes:
Hi, @Marigold! Sorry for not getting back on this. If resolving the conflicts is easy, it'd be nice to merge this. Otherwise, don't worry too much.

PR #5966, opened 2026-04-23
✨ Unify DAG YAML read/write helpers

Marigold:
## Summary

The repo had five slightly different ways to read or write a DAG YAML
file — line-based, ruamel-based, raw string concat, plus a duplicate
copy in the wizard. Consolidate them onto the canonical helpers in
``etl/dag_helpers.py`` so every caller gets the same comment-preserving
behaviour, with one place to audit when the DAG format evolves.

No behavioural change for existing callers.

### What moved

- **New**: ``etl.dag_helpers.load_single_dag_file(filename)`` — flat
  ``{step: deps}`` for one DAG file, without following ``include``.
  ``etl.version_tracker.load_steps_for_each_dag_file`` now uses it
  instead of doing its own ``yaml.safe_load``.
- **Delegated**: ``apps.utils.files.add_to_dag`` now calls
  ``write_to_dag_file``; ``remove_from_dag`` calls
  ``remove_steps_from_dag_file``. Callers keep the same public API
  (``backport.migrate``, ``fasttrack.fast_import``, wizard).
- **Dedup**: ``apps.wizard.utils.remove_from_dag`` was a verbatim copy
  of the ``apps.utils.files`` one — now a re-export.
- **VSCode extension** (``clickable-dag-steps``): extract the URI-line
  classifier as an exported ``classifyDagLine`` helper, seed a focused
  test suite (6 cases) in ``src/test/extension.test.ts``, bump to
  ``0.0.4``. No user-visible change.

### Follow-ups (out of scope)

- FAOSTAT's ``write_steps_to_dag_file`` and
  ``update_explorer_dependencies_in_dag`` still use raw
  ``file.write`` / full-file string replace. Worth migrating but the
  semantics differ subtly from ``write_to_dag_file`` (append-only, won't
  update existing steps) so I'd rather do it as its own PR.

## Test plan

- [x] ``.venv/bin/pytest -m "not integration" tests -q`` — 434 pass
  (includes new coverage for ``load_single_dag_file`` and for the
  delegating wrappers).
- [x] ``ty check`` + ``ruff check`` + ``ruff format --check`` clean.
- [x] ``cd vscode_extensions/clickable-dag-steps && npm run compile &&
  npm test`` — 6 tests pass.
- [x] Package + install updated ``.vsix``; open a DAG file to confirm
  status decorations still render.


Marigold:
@codex review

PR #5969, opened 2026-04-23
🎉 Charts as first-class ETL citizens via simple mdims

pabloarosado:
## Summary
Allow charts to be created and edited from ETL, as a special case of mdims (instead of creating a new `Graph` object, as in [this old PR](https://github.com/owid/etl/pull/5530)).

Additionally:
- Support export steps with just a config file (without a .py file).
- Adapt VSCode extensions.
  - Also adapt chart preview to be able to visualize draft charts.
- Adapt tests, version tracker, and step updater.
- Create an example chart config `etl/steps/export/multidim/animal_welfare/latest/banning_of_chick_culling.config.yml`.

## To be discussed
- The old PR had a way to pull and push grapher configs, but we weren't sure about this approach. Currently, we have linked and unlinked metadata fields in grapher charts. Those fields come from ETL. I'm wondering if we could extend that logic to other (or all?) grapher fields.


Marigold:
@codex review

etl/collection/chart_upsert.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    return chart_id
+
+
+def _resolve_chart_slug(collection: "Collection", view: "View") -> str:
+    """Derive the chart slug from the collection's short_name.
+
+    Grapher chart slugs are conventionally dash-separated; the mdim short_name is snake_case.
+    """
+    del view  # unused for now; kept in the signature for future explicit-slug overrides
```
Marigold, line 84: What's the point of `del view`?

Marigold (approved):
Neat, I'm surprised how elegant this is compared to the previous PR. After you merge it, I'll make it work with Chart Preview in VSCode.

etl/collection/chart_upsert.py
```diff
@@ -0,0 +1,121 @@
...
+            existing = Chart.load_chart(session, slug=slug)
+        except NoResultFound:
+            existing = None
+
+    if existing is not None:
+        # Preserve publication state unless the user explicitly overrode it.
+        config.setdefault("isPublished", existing.config.get("isPublished", False))
+        log.info("collection.chart.update", slug=slug, chart_id=existing.id)
+        admin_api.update_chart(chart_id=existing.id, chart_config=config, user_id=user_id)
+        chart_id = existing.id
```
lucasrodes, line 62: So, if a chart already exists in DB, we can just overwrite it via ETL? Am I understanding it correctly?

Couldn't this cause unwanted overwrites? Say the chart exists in DB and is live (i.e., this `if` evaluates to `True`). If the config in ETL differs from that in the DB (because someone has edited it in either ETL or the DB), wouldn't the DB version always be overwritten?

I see that on staging, one can reject changes in the Chart diff (I suppose), but the ETL config still appears and may have been merged into `master`. Then, in `master`, if `etl` runs, the DB chart would be overwritten, no?

I'm just not sure I understand this logic. At the moment, it feels a bit dangerous, but I could be misunderstanding it.

etl/collection/chart_upsert.py
```diff
@@ -0,0 +1,121 @@
...
+    Expects `len(collection.dimensions) == 0` and `len(collection.views) == 1`.
+    """
+    if len(collection.dimensions) != 0:
+        raise ValueError("upsert_collection_as_chart called on a collection with dimensions.")
+    if len(collection.views) != 1:
+        raise ValueError(f"Chart mode (no dimensions) requires exactly one view; got {len(collection.views)}.")
+
+    view = collection.views[0]
+    slug = _resolve_chart_slug(collection, view)
+    config = _build_chart_config(view, slug)
```
lucasrodes, line 46: I think it'd be nice to have a function like `mdim_to_chart` that just runs all this code and returns some kind of Chart object. Could probably use some of the classes from `etl.grapher.model`, maybe `ChartConfig`? This is conceptually nicer and clearer. Also, it would be cool if we could use the existing data model (and tweak it if needed).

I haven't tested the code below, and it's probably not correct, but this is an estimate of what I'm thinking here.

```python
from etl.grapher.model import ChartConfig
from etl.collection import Collection
def mdim_to_chart(mdim: Collection) -> ChartConfig:
    if len(mdim.dimensions) != 0:
        raise ValueError("upsert_collection_as_chart called on a collection with dimensions.")
    if len(mdim.views) != 1:
        raise ValueError(f"Chart mode (no dimensions) requires exactly one view; got {len(mdim.views)}.")

    view = mdim.views[0]
    slug = _resolve_chart_slug(mdim, view)
    config = _build_chart_config(view, slug)

    chart_config = ChartConfig(
        slug=slug,
        full=config,
    )
    return chart_config
```

lucasrodes (commented):
Awesome, Pablo! Thanks for working and pushing for this. I've left two comments. One more on the "code design" side of things; the other is a minor concern about the source of truth/conflicts with chart configs.

pabloarosado:
> Neat, I'm surprised how elegant this is compared to the previous PR. After you merge it, I'll make it work with Chart Preview in VSCode.

Thanks Mojmir! In principle Chart Preview should already work in this PR. I tweaked a few things and checked on a couple of examples, but I haven't checked thoroughly.

etl/collection/chart_upsert.py
pabloarosado: Not much point, I'll remove that.

etl/collection/chart_upsert.py
pabloarosado: I see your point, we can refactor in a separate PR.

lucasrodes: I think it could be done in this PR; it's a small addition. It is unlikely that we will remember to do this (I already had forgotten about it)

pabloarosado:
Closing this PR without merging; the good thing of the current approach is that no changes were required in owid-grapher. However, after various discussions, it seems that having a proper layered approach (making some changes in owid-grapher) is preferrable.
So, the new approach should be: [this PR](https://github.com/owid/etl/pull/6174)

PR #5970, opened 2026-04-23
📊 animal_welfare: Create animal welfare mdim

pabloarosado:


pabloarosado:
I think we should rethink a bit the structure of the explorer/mdim. I'll close this for now.

PR #5989, opened 2026-04-28
Create skill to draft data-updates-comms message after an update

pabloarosado:
Thanks @edomt for the initiative around skills!

I've created this one that generates a draft response to the data-update-comms template.
It should check what datasets have been updated in your current branch, and creates a temporary .md file with the answers to the slack prompts.

At least for the mechanical fields it seems to be working well (e.g. publication dates, sources, data coverage), so it can save some time. The user is then prompted to manually refine the non-mechanical fields.

NOTE: There's no need to review, it turns out that we already have a similar one in the code (see conversation below), so I'll either combine them both or delete it.


Marigold:
There's an overlap with [this skill](https://github.com/owid/etl/blob/master/.claude/skills/update-dataset/slack-announcement-template.md) that gets triggered at the end of `update-dataset`. A nice thing about doing it at the end of updating the dataset is that most of things needed for an announcement are already in the context. Would it make sense to merge them together? I'm okay with replacing it by your version.

pabloarosado:
> There's an overlap with [this skill](https://github.com/owid/etl/blob/master/.claude/skills/update-dataset/slack-announcement-template.md) that gets triggered at the end of `update-dataset`. A nice thing about doing it at the end of updating the dataset is that most of things needed for an announcement are already in the context. Would it make sense to merge them together? I'm okay with replacing it by your version.

Oh, thanks @Marigold I didn't know this already existed. Maybe it makes sense to keep the skill of creating a data-update-comms as a separate one (in case someone does a manual update and doesn't use the full update-dataset skill), and then reference data-update-comms in the existing skill. What do you think? If you think it's not worth it, feel free to close this PR.

Marigold:
I just wanted you to know about the skill. Actually, maybe the best way would be to merge this as is and DRY it when we work on a real dataset update. I don't want to be deciding which one is "better" now.

(Interestingly, Claude already knows about the [duplication](https://github.com/owid/etl/pull/5989/changes#diff-397969891f7e91a33b55e6e2c40c1d1346399ed5e2b56828f5a60405ae400fcdR273-R274) :) )

edomt:
Thank you @pabloarosado! I like your approach here 🙂 

I see you've already merged, so I just made some additions in f3ddaa7ad528b73d5083f5f6f478a63a3423c384 (mainly giving Claude better context for the task, and including small prompts in the draft for the data scientist to write from a reader-centric POV).

pabloarosado:
> Thank you @pabloarosado! I like your approach here 🙂
> 
> I see you've already merged, so I just made some additions in [f3ddaa7](https://github.com/owid/etl/commit/f3ddaa7ad528b73d5083f5f6f478a63a3423c384) (mainly giving Claude better context for the task, and including small prompts in the draft for the data scientist to write from a reader-centric POV).

Thank you @edomt, good points!

PR #5994, opened 2026-04-29
✨ meta-yaml-hover: find references from declarations

paarriagadap:
## Summary

Adds a reverse-lookup feature to the `meta-yaml-hover` VSCode extension: Cmd+Click on a YAML key declaration (or an `&anchor` token) now peeks every reference to it in the same file — the inverse of the existing reference→declaration jump.

## What changed

- New `ReferenceProvider`: powers `Shift+F12` / right-click → **Find All References** for both placeholders (`{<dotted.path>}`) and YAML anchors (`*<name>`). Honors `context.includeDeclaration`. Works from either side — on a declaration it returns usages; on a usage it returns sibling usages.
- Extended `DefinitionProvider`: when the cursor sits on a *declaration* (a YAML key in left-of-colon position, or an `&anchor` token), it returns all references. VSCode's multi-definition peek then shows the list, so Cmd+Click on a declaration works as the inverse of the existing forward jump.
- New helpers: `getDeclAtCursor` (walks up YAML indent to build the dotted path), `findPlaceholderRefs`, `findAnchorRefs`. Reuses the existing `PLACEHOLDER_RE`, `ALIAS_RE`, `findHoverMatch`, `findKeyDeclLine`, `findAnchorDeclLine`, `isMetaYamlFile`.
- Module docstring + README updated to describe the new UX. Version bumped `0.0.2 → 0.0.3`; old VSIX archived under `install/archived/`.

Same-file scope only — no cross-file or `shared.meta.yml` traversal, matching the existing extension behavior.

## Test plan

- [x] `npm run check-types` and `npm run lint` clean inside `vscode_extensions/meta-yaml-hover/`
- [x] `npx @vscode/vsce package` produces `install/meta-yaml-hover-0.0.3.vsix`; `code --install-extension` installs without errors
- [x] `make check` clean at repo root
- [x] After **Developer: Reload Window**, on a real `*.meta.yml` file:
  - [x] Cmd+Click on a key declaration under `definitions:` peeks every `{definitions.<key>}` usage
  - [x] Cmd+Click on an `&anchor` token peeks every `*anchor` usage
  - [x] Shift+F12 from the same positions matches the peek results in the References panel
  - [x] Shift+F12 from a usage site lists sibling usages plus the declaration (since `includeDeclaration` defaults to true)
  - [x] Forward Cmd+Click (reference → declaration) still works (regression check)
  - [x] Cmd+Click on a key whose path is never referenced falls through with no spurious peek


paarriagadap:
@codex review

PR #6001, opened 2026-04-29
🎉 Let `etl pr` create worktrees

pabloarosado:
## Summary

Let `etl pr` create git worktrees, and optionally share the same `data` folder.

## Context

Given that it's becoming more common to work with multiple ETL branches in parallel (especially when running AI agents), it may be convenient to use git worktrees.

This PR allows "etl pr" to create worktrees easily, e.g.
```
etl pr "Testing things" data --worktree
```
This will create a new worktree and a PR from there. It will also copy the `.env` file to the new folder.

One downside of creating new ETL worktrees is that one may need to re-run many ETL steps (e.g. regions, population, WDI...) on every new worktree, which can take a long time. For this reason, this PR also gives the option to `--share-data` between worktrees, e.g.
```
etl pr "Testing things" data --worktree --share-data
```
This creates a symlink of the original data folder in the new worktree, so that one can work in the new worktree using the same local ETL catalog.

## Notes
* I have tested that it works locally with a few simple examples. I haven't yet stress-tested what happens with grapher uploads from two worktrees. In theory each worktree pushes to its own staging DB without any conflict. Of course, if one was working on the same files from the two worktrees, there could be complications; but for such cases, one shouldn't use `--share-data`.

TODO: Consider adding the following notes to the documentation, if we agree on this approach:
* When working with worktrees, I think it's convenient to automatically load the virtual environment of a given folder as soon as you enter that folder. This can be achieved by adding this snippet to the `.zshrc`:
```
# To automatically load a virtual environment when entering a directory, as long as it's called either .venv or env
autoload -U add-zsh-hook
load-py-venv() {
    if [ -f .venv/bin/activate ]; then
        # enter a virtual environment that's here
        source .venv/bin/activate
    elif [ -f env/bin/activate ]; then
        source env/bin/activate
    elif [ -f venv/bin/activate ]; then
        source venv/bin/activate
    elif [ ! -z "$VIRTUAL_ENV" ] && [ -f poetry.toml -o -f requirements.txt ]; then
        # exit a virtual environment when you enter a new project folder
        deactivate
    fi
}
add-zsh-hook chpwd load-py-venv
load-py-venv
```
* There are different ways to work with git worktrees and LLMs. I think the most convenient one is to have two separate VSCode windows, one in each folder.

* I noticed that, when working in a new worktree with a shared `data` folder as a symlink, git notices the changes in that folder (because it doesn't find `data/.gitfolder`). This is annoying, so I created [a separate PR](https://github.com/owid/etl/pull/6005) to fix this, and let git always ignore the `data` folder.

### DIFF
--- a/apps/pr/cli.py
+++ b/apps/pr/cli.py
@@ -449,6 +449,9 @@ def print_worktree_hint(worktree_path: Path, shared_data: bool = False) -> None:
     print()
     print("Then set up the env (one-time):")
     print("  uv sync")
+    print()
+    print("When you're done with this worktree, remove it with:")
+    print(f"  git worktree remove {display_path}")
     if shared_data:
         print()
         print("WARNING: data/ is a symlink to the original repo's data/, so:")
--- a/apps/pr/cli.py
+++ b/apps/pr/cli.py
@@ -40,12 +40,26 @@ etl pr "some title for the PR" --direct --base-branch "develop" --work-branch "t
 # Shorter
 etl pr "some title for the PR" --direct -b "develop" -w "this-temporary-branch"
 ```
+
+**Custom use case (4)**: Create the new branch in a sibling git worktree so you can keep editing your current branch in parallel.
+
+```shell
+etl pr "some title for the PR" --worktree
+# Shorter
+etl pr "some title for the PR" -t
+# With a custom path
+etl pr "some title for the PR" -t --worktree-path /tmp/etl-mybranch
+```
+
+The new working directory is printed at the end (default: `../etl-BRANCH`); `cd` into it to start working there.
 """
 
 import hashlib
 import os
 import re
+import shutil
 import uuid
+from pathlib import Path
 from typing import cast
 
 import click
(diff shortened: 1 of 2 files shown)


lucasrodes (approved):
nice, thanks for doing this! I'm not that familiar with worktrees; I had to look them up. Looks interesting, and I can totally see why they can be useful for agents.
                                                                                                            
a few things that came to mind while reviewing this PR:                                                          
                                                                                                            
- on `--share-data`: the `rm -rf data/` footgun is well documented, but nothing prevents two concurrent ETL runs writing to the same `data/` dir. You mention this in the description, but it might be worth surfacing in the printed hint as well, not just the docstring.                                                               
- the new branch is created off the local master (after a `fetch --prune` in init_repo, but no rebase). If the local master is stale, you'd start the worktree from an old commit. worth a `git pull` first, or branching from origin/master directly?                                                                              
- (nit) No easy way to clean up afterward. could be nice to add an `etl pr --remove-worktree` or at least mention `git worktree remove ../etl-foo` somewhere in the printed hint.                                            
- worth a mention in `AGENTS.md` / `CLAUDE.md` under the Git Workflow section so agents know about this flag when running in parallel. I wouldn't bake it into the individual skills (update-dataset, fix-dependabot, etc.) since worktree-vs-not is a per-run decision, not a per-task one.                                  

Re: the docs, TODO — +1 on adding both the .zshrc snippet and the "two VSCode windows" tip. For me, the latter is exactly how I'd want to use this.

PR #6002, opened 2026-04-29
📊 natural_disasters: Create natural disasters mdims

pabloarosado:
## Summary
Create 3 mdims about natural disasters.

## Context
More details in https://github.com/owid/owid-issues/issues/1545
In this PR, I created the following mdims:
* [Deaths from natural disasters](http://staging-site-data-naturaldisasters-mdims/admin/grapher/natural_disasters%2Flatest%2Fdeaths%23deaths)
* [People affected by natural disasters](http://staging-site-data-naturaldisasters-mdims/admin/grapher/natural_disasters%2Flatest%2Faffected%23affected)
* [Economic damages from natural disasters](http://staging-site-data-naturaldisasters-mdims/admin/grapher/natural_disasters%2Flatest%2Feconomic_damages%23economic_damages)

## Additional changes:
- A fix and model bump of etl inspector.
- Some improvements on natural disasters metadata.
- The collections module used to expand indicator names into full catalog paths only at the point when the collection was created. But this was problematic if one wanted to add indicators in view configs programmatically, after the collection is already created. In this PR, that operation of expanding names into paths happens twice: when creating the collection, and when saving it. I thought of doing it only on saving; but just in case, to ensure nothing breaks, I decided to let it do it twice (it's not an expensive operation).

## TODO (Pablo):
* [x] Currently, all indicators show the same "What you should know", mentioning all disasters; I'll try to fix it so that only the relevant disasters appear.
* [x] Group wet and dry mass movement into a "landslide" category.
* [x] Include glacial lake outburst flood (currently ignored) under "flood".
* [x] The disaster type "fog" only has one event, the Great Smog of London in 1952. I'd suggest removing "fog" altogether.

## Review
@lucasrodes no need to go through the code in detail. I'd appreciate if you could:
* [x] Play around with the mdims a bit in case you find issues or have suggestions.
* [x] Let me know if you agree with the change in collections (expanding catalog paths).
* [x] (we briefly discussed it during triage) To be able to show a multi-indicator stacked-bar view with a map tab showing the sum, I hacked the system a bit: I included the sum (which already exists as a separate indicator) as a "color" property, so we can call it afterwards and use it in the map. I think that ideally (both for charts and mdims) grapher could allow for a "map" indicator. If you agree with this approach, I can create an owid-grapher issue. Otherwise let me know if you think this is not a good move.
Thanks!


lucasrodes:
Nice work, Pablo! These are great MDIMs. I'll lay out my thoughts below. I have some suggestions for how the MDIMs are presented, then just some follow-up questions on your changes to `etl.collection` and the map hack. Have a nice weekend!

### Exploring MDIMs
Some comments on the MDIMs, I’ve started writing this down as I checked the MDIM “Deaths from natural disasters”, but I think that all my points hold for the other MDIMs.

- **Non-standard colors**: I noticed that you use non-standard colors (a mustard yellow and a striking orange). It would be better to adhere to our color palette and choose from our 24 available colors. If you think that the palette could be improved, I’d definitely suggest bringing this up to design/product.
- **Description key is too verbose**: I see that for the views with all disasters, you’ve defined the different natural disasters in the description key. Super cool that there are descriptions in hand. I think, though, that because of them, the description key section feels quite heavy and hard to process with human eyes. I see two options to improve this situation: either (a) use DoDs for the terms in the subtitle and the second bullet point in the description key, or (b) maybe nest bullet points describing specific events one level more to create a visual hearchy (also, I would bold the terms to make it even easier to scan).
- **Open brackets**: In map views, maybe worth considering having open brackets for the upper limit?
- **Year range**: I see that you are using 2000-2025 by default. Is there a way for different views to use a different range? Mostly looks fine, but I thought maybe for the views on decadal average it’d be nicer if we could have a longer period of years shown by default. But maybe this is just not possible.


In the "People affected by natural disasters" MDIM, in the “Impact” dropdown, I’m wondering if you need the description of the choices at all. At least for the "Individual" group. What about replacing the choice names for the descriptions?  "Injuries" → "People injured", "Homeless" → "People left homeless", etc.
  image

Also, more broadly, maybe reconsider if we need descriptions at all sometimes? Like, do we need "Damages caused by drought events?" under "Droughts" in the Economic damage MDIM?


### Changes to `etl.collection.`
I am fine with it! Out of curiosity, is this needed again in case we add incomplete paths during processing and after creating the MDIM object?

### Showing sum in map
So, if I understand correctly, in your case, this works bc the `color` property doesn't affect the presentation of maps or stacked bar charts, right? Like, if you had line charts, the hack wouldn't be free, right? Just so I understand.

I guess this hack is tolerable. And yes, in general, I think that it'd be nice if we could allow the map to use an "arbitrary" indicator that is not part of the chart. I think that it'd be even nicer if we could condition this "arbitraryness" and actually only allow for indicators that are related to the ones in the chart (e.g., the "sum", etc.). This would mean that we need to somehow register this relationship in ETL or Grapher, which I think would be nice, actually.

Another possibility we discussed was to just show the sum, which is computed by grapher, just like it is computed and shown in the column chart. The benefit of this approach is that we wouldn't need to have an indicator existing in ETL. But, I suppose it comes at the cost of not having proper metadata for the summation indicator (although not sure this is being shown anywhere).

lucasrodes (approved):

pabloarosado:
Thanks @lucasrodes for the detailed feedback!
I've implemented most of your suggestions (some are tricky, e.g. keeping map brackets open). I'll wait for Hannah's feedback before merging.
Have a good weekend!

PR #6005, opened 2026-04-30
🔨 Remove unnecessary code related to the old data/garden/reference

pabloarosado:
## Summary
Clean up code related to the old `data/garden/reference` (where we used to commit regions data).

## Context
In the early days of ETL, we used to commit some data to the repos, namely in `data/garden/reference` (I think mostly or only regions data). But we stopped doing that almost 3 years ago (see [data folder history](https://github.com/owid/etl/commits/master/data)). I don't think we are moving back to committing any data to the repos.

In this PR I've simplified `.gitignore` to ignore the full `data` folder. I've also removed a couple of other mentions of `garden/reference`, including an old data step that was creating data inside `garden/reference` (and is no longer in the DAG).

The motivation for doing this, apart from cleaning up old code, is to be able to work with various ETL worktrees in parallel, reading and writing to a shared `data` folder (living in the main repos folder). When doing that, under the status quo, git complains because it detects changes in the `data` folder.

This PR should let git simply always ignore the `data` folder.

Related to https://github.com/owid/etl/pull/6001


Marigold (approved):
Nice

PR #6016, opened 2026-05-04
💄 Update metadata for CSET investment indicators

edomt:
Hey @veronikasamborska1994,

Thanks for updating the CSET data! I was equally surprised to see the high 2026 figures on the `private-investment-in-artificial-intelligence-cset` chart, and spent some time understanding how CSET arrives at them. That investigation made me think it was a good moment to revisit the chart's metadata more broadly — we now have more years of data, fewer open questions about the 2021–2023 years, and a few wording improvements I'd like to suggest.

## The main change

I'm proposing we drop CSET's partial-year data from this chart. Their underlying method (assigning median values to undisclosed deals from comparable Crunchbase deals) seems fairly robust for full years — but for 2026 specifically, around 80% of the value is imputed, vs. ~40% in normal years, simply because there are very few disclosed deals so far to anchor the medians. That makes the 2026 figure feel too preliminary to display.

To do this with minimal disruption:

- In case we want to change this later, both `estimated_investment` and `estimated_investment_projected` indicators stay in the dataset, with the existing splitting logic untouched.
- I've just removed `estimated_investment_projected__field_all` from the chart in Grapher.

Another important reason is that using two indicators drops the data page UI, which means readers don't get to see any of the WYSK. For a tricky indicator like that, I think there are more benefits to showing all the metadata.

## What's in this PR

- `cset.py`: small tweak — the subtitle's `{month}` / `{year}` placeholders now read from `last_date_published` rather than `last_date_accessed`, so they reflect CSET's data cutoff (March 2026) rather than our snapshot date (April 2026).
- `cset.meta.yml`: rewritten title, description_short, and description_key for both indicators, plus a second description_short anchor (`investment_description_short_partial`) so the projected indicator can keep the "Data for {year} is incomplete..." clause without it bleeding into the actuals.

## What's intentionally preserved

- All processing code in `cset.py` (the meadow split into actuals/projections, the last-actual-year merge, etc.).
- The `_projected` indicator itself — it remains in the catalog and could be used later.

### DIFF
--- a/etl/steps/data/garden/artificial_intelligence/2026-04-27/cset.meta.yml
+++ b/etl/steps/data/garden/artificial_intelligence/2026-04-27/cset.meta.yml
@@ -25,12 +25,11 @@ definitions:
     - Patent data has a lag because some offices take years to publish new filings. Recent years are excluded here.
 
   description_key_investment: &description_key_investment
-    - The data covers external private-market investment in AI companies, including venture capital deals, private equity deals, and mergers and acquisitions.
-    - It excludes internal corporate R&D, capital spending, and government funding. Publicly traded companies, including large technology firms, are also excluded.
-    - Because this is only one type of funding, the data understates total global spending on AI.
-    - A small number of very large deals can drive big year-on-year swings. Broader economic conditions, such as interest rates and investor sentiment, also affect totals in ways that are not specific to AI.
-    - CSET's figure for {year} is preliminary and is likely to rise as more deals are reported. We show it to provide a more current view, but it may change as CSET updates its estimates.
-    - The "World" total only includes countries that are covered by the source, so it understates global activity.
+    - This data tracks the flow of external money into private companies. It excludes internal spending on R&D and infrastructure.
+    - On the recipient side, this only includes privately held AI companies. It excludes publicly traded companies like "Big Tech".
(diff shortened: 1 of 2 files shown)


edomt:
@codex review

veronikasamborska1994 (approved):
Thanks Ed. I agree it's worth dropping the partial-year indicator from the chart and your changes to the metadata look good.

If helpful I can add a note to the relevant scheduled data update issue to check what `estimated_investment_projected__field_all` looks like in case you want to revisit adding it later in the year when the share of values that are imputed should be more comparable to earlier years. Though I also see your point that it's nice to have a data page for an indicator this nuanced. 

I'm not sure if there are any plans for extending what multi-indicator charts we can show as data pages, but I think charts with 2 indicators where 1 is a projection is something the P&D team could consider. Not that this would change what we do with this chart now of course.

edomt:
> If helpful I can add a note to the relevant scheduled data update issue to check what estimated_investment_projected__field_all looks like in case you want to revisit adding it later in the year when the share of values that are imputed should be more comparable to earlier years. Though I also see your point that it's nice to have a data page for an indicator this nuanced.

Thanks, that would be nice! 🙂 

> I'm not sure if there are any plans for extending what multi-indicator charts we can show as data pages, but I think charts with 2 indicators where 1 is a projection is something the P&D team could consider. Not that this would change what we do with this chart now of course.

I think it's on the list of things P&E are considering for data pages in the coming cycle. So we might indeed return to this in a few weeks/months if we can show all the nuanced metadata _and_ the projected data.

PR #6023, opened 2026-05-05
💄 ai: add IEA regional aggregates for data center electricity demand

veronikasamborska1994:
## What this PR does

This PR extends the IEA data center energy dataset to include **regional breakdowns** for the share of electricity demand indicator, and improves metadata documentation.

### Code changes (`energy_ai.py`)

The `create_share_of_electricity_demand` function previously only computed shares for World, US, China, and World excl. US and China. It now also computes shares for IEA's six regional aggregates: **Africa, Asia Pacific, Central and South America, Europe, Middle East, and North America**.

For each region, the electricity demand denominator is built by summing Ember's country-level demand across the IEA member countries in that region. For the global estimate, **Ember's world total is used directly** rather than summing across IEA regions. This is because the IEA does not publish a data center electricity figure for its Eurasia regional aggregate (which includes Russia), so a regional sum would undercount global demand. Using Ember's world total gives a global share of ~1.5% in 2025, consistent with the IEA's own reporting.

### Metadata changes (`energy_ai.meta.yml`)

- Added a shared `eurasia_note` anchor applied to all five indicators, noting that the IEA does not publish a regional aggregate for Eurasia (which includes Russia).
- Expanded `description_processing` for the share indicator to explain the two-source methodology (IEA numerator, Ember denominator), the regional vs. global denominator approach, and why Ember's world total is preferred for the global figure.


pabloarosado (approved):
Looks good!
I double-checked IEA's [regions](https://www.iea.org/regions/) and they are correct. Ideally, we could include these definitions in our `regions.yml` file (so that grapher would recognise those regions and show their members), but it's up to you; you don't need to do that for this PR.

veronikasamborska1994:
> Looks good! I double-checked IEA's [regions](https://www.iea.org/regions/) and they are correct. Ideally, we could include these definitions in our `regions.yml` file (so that grapher would recognise those regions and show their members), but it's up to you; you don't need to do that for this PR.

thanks @pabloarosado! I've added the regional mapping to our `regions.yml` file but the regional DoDs aren't showing up on the chart yet. Is there anything else that I am missing here?

pabloarosado:
> > Looks good! I double-checked IEA's [regions](https://www.iea.org/regions/) and they are correct. Ideally, we could include these definitions in our `regions.yml` file (so that grapher would recognise those regions and show their members), but it's up to you; you don't need to do that for this PR.
> 
> thanks @pabloarosado! I've added the regional mapping to our `regions.yml` file but the regional DoDs aren't showing up on the chart yet. Is there anything else that I am missing here?

Thanks for doing that! Simply merge the current PR (ideally, by the end of the day, to not block other runs) and write a message in #developers to point out the change in regions, so that someone can pull the changes into owid-grapher.

(According to [our documentation](https://www.notion.so/owid/Regions-and-income-groups-1fc74b71f92e8064b02dd8c01361856b?source=copy_link#25674b71f92e809d8f19d16467f4c1bc), there's an automatic action to do that, but I think it doesn't happen regularly (maybe once a week). So it's better to ask someone to do it manually, and it should be quick).

PR #6029, opened 2026-05-05
🔨 explorers: migrate `food-prices`

lucasrodes:
| PROD | STAGING |
|--------|--------|
| [link](https://ourworldindata.org/explorers/food-prices) | [link](http://staging-site-refactor-migrate-explorers-e/admin/explorers/preview/food-prices) |

@pabloarosado could you review the explorer and verify it looks good? I've been checking and it looks fine, I think.

The explorer was grapher-based, now it is indicator-based. Most of the metadata has been moved to the garden step so that charts relying on these indicators also benefit from this metadata. I've also adapted these charts to be in sync with this metadata. I'm actually not sure how relevant the stand-alone charts are, or if they were there just for the explorer. But that's up to you, totally fine to keep them of course.

Also, there seems to be an old version of the "[Food prices for nutrition](https://admin.owid.io/admin/datasets/6719)" dataset. Could you confirm that it is archivable?

/schedule


pabloarosado (approved):
Looks good! Just a minor issue: I noticed that some views have a line tab, even though the data has only a single year covered ([example](http://staging-site-refactor-migrate-explorers-e/admin/explorers/preview/food-prices?tab=line&time=latest&Diet=Calorie+sufficient+diet&Cost+or+Affordability=Cost&Cost+metric=%25+of+average+food+expenditure&country=NGA~BGD~IND~ETH~MEX~USA~BRA~GBR)). For those cases, it would be better to drop the line tab and keep only map and bar.

PR #6037, opened 2026-05-06
📊 oil spills: update ITOPF

veronikasamborska1994:
Hi @antea04! I am going to be handing over this dataset to you when I leave OWID so I've assigned this PR so you can familiarise yourself with the data. It was a tricky update to begin with because we need to extract the numbers from the pdf but now with claude code it should be relatively straightforward.


antea04 (approved):
Thanks Veronika! 

LGTM! Thank you so much for your great work at OWID over the last few years and I'm wishing you all the best for your next adventure ❤️

PR #6040, opened 2026-05-06
🔨 explorers: migrate `food-footprints`

lucasrodes:
| PROD | STAGING |
|--------|--------|
| [link](https://ourworldindata.org/explorers/food-footprints) | [link](http://staging-site-refactor-migrate-foodfootpri/admin/explorers/preview/food-footprints) |

@pabloarosado I've migrated the `food-footprints` explorer into ETL. I realized this explorer was relying on both grapher charts and CSV files! Because of this, I had to import a dataset (Clark et al.) to ETL, which just existed in owid-datasets. Please review that the explorer looks fine. I sanity-checked it, and it mostly looks fine. I simplified some subtitles.

As I created the explorer, I realised that there were some inconsistencies in titles and subtitles. I think this comes from the fact that `commodity` views came from charts and `specific product` views came from an external CSV.

Summary of changes:
- Stop using `owid-distinct` color scheme for bar chart. The current explorer in PROD uses that scheme inconsistently. Let's just use default blue, which is the standard now.
- Some titles were inconsistent across `Commodity` and `Specific food products`. 
  - "Eutrophication" and "Eutrophying emissions" used in titles depending on commodity vs specific product. Chose "Eutrophying emissions".
  - "Scarcity-weighted water use" and "Water-scarce withdrawals" used in titles depending on commodity vs specific product. Went with "Scarcity-weighted water use".

/schedule


pabloarosado (approved):
Looks good, thanks for this work!
My only concern is that, being a popular explorer, maybe it would be worth it to ask on Slack if others (especially Hannah) agree on changing the colors. I have seen this explorer often shared in social media, and I identified the charts by its colors (which I agree, are arbitrary).

One easy thing that needs to be fixed: origin URL is missing.

And some other minor issues (feel free to ignore):
* I've noticed different num decimal places here and there (but they were not necessarily well chosen in the original anyway).
* The title of Clark et al. (2022) in the snapshot should be "Estimating the environmental impacts of 57,000 food products".
* I think that the snapshot description of Clark et al. doesn't need to be repeated in the description key of its indicators.
* The short name of data step of Clark et al is a bit long and messy.

lucasrodes:
@pabloarosado I'll leave the color choice to you, then. I've rolled back to restore the colors to what they were in PROD, so this isn't a blocker. FWIW, I don't think they matter that much. I say this because the color seems random and not very stable: it changes as the user adds more entities (see recording below). I think we should revisit this after this merge; it makes it look buggy, imo.

https://github.com/user-attachments/assets/86e37f20-171d-42fa-94a5-4597cd7deaec



---

> The title of Clark et al. (2022) in the snapshot should be "Estimating the environmental impacts of 57,000 food products".

Done!

> I think that the snapshot description of Clark et al. doesn't need to be repeated in the description key of its indicators.

Thanks for catching. Removed these.

> The short name of data step of Clark et al is a bit long and messy.

Replaced it with `clark_et_al_2022`, inspired by other steps in agriculture:
- `snapshot://agriculture/2024-05-23/floud_et_al_2011_daily_calories_europe.csv`
- `data://garden/papers/2024-03-26/broadberry_et_al_2015`
- `data://garden/papers/2024-03-26/mueller_et_al_2012`

PR #5358, opened 2025-12-15
📊 Migrate minerals explorer to multidim format

Marigold:
## Summary

Original explorer: https://ourworldindata.org/explorers/minerals
New mdim: http://staging-site-data-migrate-minerals-multid/admin/grapher/minerals%2Flatest%2Fminerals%23minerals

Migrates the minerals explorer from the legacy `create_explorer_legacy` format to the new multi-dimensional (multidim) format using `create_collection`.

### Changes
- **New multidim explorer**: `export://multidim/minerals/latest/minerals`
- **Declarative YAML config**: `minerals.config.yml` with dimensions and default settings
- **Python step**: Parses column metadata and sets dimensions dynamically
- **DAG updated**: Both legacy and multidim explorers available during transition

### Features Preserved
- All dimension choices (Mineral, Metric, Type, Measure) with proper capitalized labels
- Map tab disabled for specific columns with sparse geographic data (14 columns)
- Map tab disabled for "Unit value" metric
- Dynamic `minTime` based on data availability (bar chart for sparse data with <5 years range)
- Default view: Copper, Mine, Production, Absolute
- Default selection: World, Australia, Chile, China, United States
- Y-axis starting at 0
- Line chart + Bar chart options

### Generated Output
- **303 views**
- **65 minerals** (Copper, Gold, Lithium, etc.)
- **36 types** (Mine, Refinery, Processing, etc.)
- **3 metrics** (Production, Reserves, Unit value)
- **2 measures** (Absolute, Share of global)

## Test plan
- [x] Run `etlr "multidim/minerals/latest/minerals" --only --export` locally
- [x] Verify on staging that views match expected dimensions
- [x] Compare with legacy explorer output

## TODO before merging
- [ ] Remove old minerals explorer

## TODO after merging
- [ ] Set up a redirect from the old explorer (is that a thing? where to do that?)
- [ ] Migrate `Minerals Supply and Demand Prospects` explorer

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6042, opened 2026-05-06
🔨 explorers: migrate `minerals`

lucasrodes:
| PROD | STAGING |
|--------|--------|
| [link](https://ourworldindata.org/explorers/minerals) | [link](http://staging-site-refactor-migrate-minerals/admin/explorers/preview/minerals) |

@pabloarosado The explorer looks good! I see that there was prior work on porting this explorer to an MDIM in https://github.com/owid/etl/pull/5358. So basically here I'm borrowing most of the code from the MDIM.

Also, I've realized that there is an explorer `minerals_supply_and_demand_prospects`. I suppose that's fine to keep?

### DIFF
--- a/etl/steps/export/explorers/minerals/latest/minerals.config.yml
+++ b/etl/steps/export/explorers/minerals/latest/minerals.config.yml
@@ -1,0 +1,49 @@
+config:
+  explorerTitle: Minerals
+  explorerSubtitle: Explore the production, reserves and value of minerals.
+  isPublished: true
+  selection:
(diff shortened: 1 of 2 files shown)


pabloarosado (approved):
Thanks for this.
I checked a few random views and it all seems to work fine.
The `minerals_supply_and_demand_prospects` lives in the archive DAG. There's still a `.py` file; if you think it causes confusion, we could delete that (and remove the step from the archive DAG).

PR #6065, opened 2026-05-11
🚧 Attempt to avoid chart syncing issues between etl and owid-grapher

pabloarosado:
In this branch I figured out a way to avoid syncing issues between etl and owid-grapher charts. But it's more of a hack rather than a proper solution.


pabloarosado:
The implementation in this branch seems to work, but it's hacky, not the right way forward. I also implemented a chart-pull tool, which may be useful. But we can rethink it differently in a separate PR. So I'll close without merging.

PR #6080, opened 2026-05-11
📊 Activate make check pre-commit hook by default

Marigold:
## Summary

Make the `make check` pre-commit hook active by default for everyone, so it doesn't depend on a manual `make install-hooks` step (follow-up from #6059).

## Changes

- **`install-hooks` switches from copy → `core.hooksPath`.** Previously it copied `scripts/hooks/pre-commit` into `.git/hooks/pre-commit`. Now it runs `git config core.hooksPath scripts/hooks`, which is idempotent and stays in sync if the hook is ever edited (no stale copy in `.git/hooks/`).
- **`install-hooks` is a dep of `.venv`.** Anyone running `make .venv` (directly or via `make test`, `make check`, etc.) gets the hook activated on the first run.
- **Docs**: working-environment.md now says the hook is auto-activated; `make install-hooks` is kept as a manual fallback.

## Notes / tradeoffs

- `core.hooksPath` is per-clone (lives in `.git/config`), so existing checkouts pick it up the next time anyone runs `make .venv`.
- If a user has a personal hook in `.git/hooks/`, `core.hooksPath` overrides it. Unlikely here.
- Guarded with `if [ -d .git ]` so it's a no-op in environments without a `.git/` (e.g. some CI containers).

## Test plan

- [x] `make install-hooks` sets `core.hooksPath` and prints the confirmation.
- [x] `make -n .venv` shows both `.venv-default`'s `uv sync` recipe **and** the install-hooks block.
- [x] `make check` passes.
- [x] Commit on this branch was blocked by the hook (it ran `make check` automatically) before landing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ help:
 	@echo '  make fasttrack 	Start Fast-track on port 8082'
 	@echo '  make chart-sync 	Start Chart-sync on port 8083'
 	@echo '  make query SQL="..." Run SQL query on staging MySQL for current branch'
-	@echo '  make install-hooks	Install git hooks (runs make check before commit)'
+	@echo '  make install-hooks	Activate pre-commit hook (auto-runs with make .venv)'
 	@echo '  make test      	Run all linting and unit tests'
 	@echo '  make test-all  	Run all linting and unit tests (including for modules in lib/)'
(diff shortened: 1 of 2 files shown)


Marigold:
@codex review

PR #5978, opened 2026-04-27
Migrate Source → Origin for dag/migrated.yml snapshots

Marigold:
## Summary

Migrates legacy `meta.source` to modern `meta.origin` for every snapshot DVC referenced from `dag/migrated.yml`, plus DEFRA air pollution which kicked this off.

Origins follow OWID's documented style: producer is a clean institution / surnames-only label, no year/`&`/`;`/OWID; `title_snapshot` and `description_snapshot` are reserved for genuine slices of named multi-product databases (FAO, V-Dem, OECD reports, etc.); paper/book sources collapse into `title` + `description` only; URLs and dates pulled verbatim from the legacy source — no fabrication.

Also adds **`.claude/skills/migrate-source-to-origin/`**, an instruction-only Claude Code skill that drives the same migration interactively for ad-hoc fixes (no API key, no helper script).

After this PR: 0 legacy `meta.source` blocks remain in `dag/migrated.yml` snapshot DVCs.

## Out of scope (tracked in #5979)

- 7 garden/grapher `.meta.yml` files with legacy `sources:` keys.
- 73 active legacy snapshot DVCs outside `dag/migrated.yml`.

## Testing

- All modified snapshot DVCs parse via `SnapshotMeta.load_from_yaml`.
- Sample garden + grapher runs for Fouquin/Hugot, plastic-waste, etc.
- Skill validated across 26 fresh subagent test runs spanning papers, OWID compilations, and multi-product DB slices.

PR #6081, opened 2026-05-12
🔨 Remove Source metadata object (origins only)

Marigold:
## Summary

Removes the legacy `Source` metadata object from `owid-catalog` and every active ETL code path that still wrote to it. `Origin` is now the only producer-metadata shape that survives end-to-end into Grapher.

### What changes

- **`owid-catalog`**: `Source` is dropped from `owid.catalog.core.meta` and the public `owid.catalog` namespace. `DatasetMeta.sources`, `VariableMeta.sources`, and `TableMeta` source plumbing are gone. `VariableMeta` now requires at least one origin.
- **`etl/snapshot.py`**: the DVC loader rejects snapshots without an `origin:` block (hard `AssertionError`, surfaced at dirty-step detection so missing migrations fail fast in CI rather than partway through a step).
- **`etl/grapher/to_db.py`**: drops `cleanup_ghost_sources`, the `_upsert_source_to_db` path, and every per-variable source upsert. Variables now upsert origins only. The `sources` MySQL table and `gm.Source` ORM model are intentionally left in place — ETL just stops writing to them so existing rows can be cleaned up separately on the DB side.
- **`apps/backport/`**: the full CLI pipeline (`backport`, `fasttrack_backport`, `migrate`, the three cookiecutter templates) is deleted. It was the last consumer of `Source` and only worked against the legacy backport flow.
- **Fasttrack wizard**: source fields removed from the form; submission now emits a single `origin:` block.

### Compatibility shim

A slim `etl/backport_helpers.py` keeps `long_to_wide` importable so the ~370 archived snapshot scripts that re-build cached feathers still load. This is a pure import-compatibility shim — no new behavior.

### Depended on

- #6115 (📊 Migrate 8 flat-legacy snapshot DVCs to origin) — merged. Those snapshots used the pre-`source:` flat metadata shape and would otherwise trip the new origin-required assertion during dirty-step detection.

## Test plan

- [x] Full unit-test suite (`make test`) green
- [x] `lib/catalog` unit tests green (354 passed)
- [x] Format + lint clean on both root and `lib/catalog`
- [x] Buildkite `etl-unit-tests` green
- [x] Buildkite `etl-automated-staging-environment` green — full pipeline runs against fresh staging DB
- [x] Active DAG scan: no remaining `from owid.catalog import Source` in any live step
- [x] Active DAG scan: every active snapshot DVC has `meta.origin:`


lucasrodes (commented):
GREAT! Thanks for this great cleanup!! So much pleasure in merging this PR.

Maybe already on your radar, but aren't there some snapshot steps that still rely on `Source`?

The new `assert meta.get("origin")` in `etl/snapshot.py:846` runs inside `SnapshotMeta.load_from_yaml`, which `SnapshotStep.is_dirty()` calls during dep resolution. A quick scan shows ~128 snapshots in `dag/migrated.yml` still use legacy `meta.source:` with no `meta.origin:`, so `etlr` runs touching them (or their downstream steps) would hit the assertion. #6115 covered 8 similar ones — not sure if these 128 are intentionally out of scope or planned for a follow-up.

Repro:

```bash
.venv/bin/python -c "from etl.snapshot import SnapshotMeta; \
  SnapshotMeta.load_from_yaml('snapshots/agriculture/2014-01-01/nitrogen_efficiency__lassaletta_et_al__2014.feather.dvc')"
# AssertionError: "origin" must be set
```


---

Also, could you bump `owid-catalog` to 1.2.0? Also, if you could add a changelog section in its README.md for the new version, that'd be great.

Marigold:

Thanks @lucasrodes! Two quick replies:

**On the `assert meta.get("origin")`:** I think you may have been running against a stale local checkout. The repro file (`snapshots/agriculture/2014-01-01/nitrogen_efficiency__lassaletta_et_al__2014.feather.dvc`) already has `origin:` set at the branch tip (`a447ff52`) — it was migrated by #5978 and #6082 which are both on this branch. I verified all 255 snapshots referenced in `dag/migrated.yml` have `origin`, and ran the load on your exact file path — it loads fine.

I also scanned all `.dvc` files on disk: 6319 are still missing `origin`, but **0 of them are referenced by any active DAG file** — they only appear in `dag/archive/*.yml` (170 refs). So `etlr` runs never load them. Happy to migrate the archived ones too in a follow-up, but they're effectively dead.

**On `owid-catalog`:** bumped to `v1.2.0` and added a changelog entry in `lib/catalog/README.md`. Also merged master in.

lucasrodes (approved):
ha! my bad. Then let's proceed with the cleaning!

PR #6082, opened 2026-05-12
📊 Migrate active snapshots from meta.source to meta.origin

Marigold:
## Summary

Migrate the legacy `meta.source:` block to modern `meta.origin:` for **all 110 unarchived snapshot DVC files** still using the legacy schema. Each file was migrated by an LLM subagent following the project's [`migrate-source-to-origin`](.claude/skills/migrate-source-to-origin/SKILL.md) skill, then validated against `SnapshotMeta.load_from_yaml`.

The migration is **restructure-not-rewrite**: every legacy fact survives into the new origin, just placed in the right field per OWID's documented Origin style.

## Scope

- **110 snapshot DVC files modified** (find them via `git diff --name-only origin/master`)
- **No downstream sweep** — only the snapshot `.dvc` files were touched. Downstream `meta.yml`, `.py` steps, and `dataset.sources:` blocks that duplicate this source info were intentionally left in place for a follow-up sweep.
- **Active DAG only** — these are the 110 snapshots referenced in `dag/*.yml`. Archived snapshots (referenced only from `dag/archive/`) were not touched.

## Per-file decisions

For each file, the subagent ran:

1. **STEP 1 (coincidence test)** — does the snapshot coincide with the producer's data product (default), or is it a producer-defined slice (rare)? This decides whether `title_snapshot` / `description_snapshot` are set.
2. **STEP 2 (field rules)** — fill `producer`, `title`, `citation_full`, `date_published`, `date_accessed`, `url_main`, `attribution_short`, `version_producer`, `license` per OWID style.
3. **Self-check** against lint rules (no `&`/`;`/year in `producer`, no trailing period in `title`, `attribution_short` ≠ `producer`, etc.).
4. **Validate** via `SnapshotMeta.load_from_yaml` — all 110 files parse cleanly.

## Common patterns

- The legacy `meta.name` is OWID's internal label — it was discarded when it embedded the producer name, year, or `(et al.)` suffix; the producer-issued title from `published_by` was used instead.
- Single journal papers → STEP 1 says they coincide → no `title_snapshot`.
- Multi-product producer databases (e.g. WDPA, GID-DB, BoE Millennium, BIMTS, DOTS, OECD Gender Data, GGDC, Eurostat Comext) → STEP 1 says they differ → `title_snapshot` is set, slice-specific prose lives in `description_snapshot`.
- OWID compilations of multiple raw sources → producers enumerated when ≤3 are named, else `Various sources`; legacy "Our World in Data based on …" prefix stripped from `producer` and preserved in `citation_full`.
- HTML in legacy descriptions (`<a href>`, `<ul>`, `<br>`) converted to Markdown — grapher renders origin descriptions through a markdown-only path.
- `is_public: false` is a `SnapshotMeta` sibling field and was preserved verbatim alongside `meta.origin`.

## Out of scope (deliberately deferred)

- **Downstream `meta.yml` sweep** — many of these snapshots have garden/grapher steps with `dataset.sources:`, `all_sources:`, or variable-level `descriptions:` blocks that duplicate the snapshot's source info. Migrating those is a separate per-chain task (workflow step 9 in the skill).
- **Variable-level `description:` → `description_short:` rename** — out of scope for the snapshot migration.

## Test plan

- [x] All 110 files parse cleanly with `SnapshotMeta.load_from_yaml`
- [x] No residual `meta:\n  source:` blocks in the modified files
- [x] All modified files contain a `meta.origin:` block
- [ ] Spot-check a few representative migrations against the legacy diff (single paper, OWID compilation, producer-defined slice)

### DIFF
--- a/snapshots/agriculture/2014-01-01/nitrogen_efficiency__lassaletta_et_al__2014.feather.dvc
+++ b/snapshots/agriculture/2014-01-01/nitrogen_efficiency__lassaletta_et_al__2014.feather.dvc
@@ -1,7 +1,7 @@
 meta:
-  source:
-    name: |-
-      Lassaletta, Billen, Grizzetti, Anglade & Garnier (2014). 50 year trends in nitrogen use efficiency of world cropping systems: the relationship between yield and nitrogen input to cropland. Environmental Research Letters.
+  origin:
+    producer: Lassaletta et al.
+    title: '50 year trends in nitrogen use efficiency of world cropping systems: the relationship between yield and nitrogen input to cropland'
     description: |-
       This dataset measures the nitrogen balance of agricultural inputs and outputs.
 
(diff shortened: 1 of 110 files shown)


Marigold:
@codex review

PR #6084, opened 2026-05-12
🔨 explorers: migrate `conflict-data`

lucasrodes:


lucasrodes:
Restarting under conflict-data-source — that's the modular per-source explorer; conflict-data will be derived from it later.

PR #6128, opened 2026-05-15
📊 Add income distribution S3 export for bespoke viz

marcelgerber:
Our other bespoke viz also upload data to the owid-public S3 bucket (see e.g. https://github.com/owid/etl/pull/6067 and https://github.com/owid/etl/blob/6e3a93b6d1afcbc99771e00556162131eb3f9c59/etl/steps/export/s3/ihme_gbd/latest/gbd_treemap_json.py), so let's do that for the income distribution plot, too. We also transform the data into a JSON format that's easily consumable in JS.

@paarriagadap If you want, you could now get rid of the `data://external/poverty_inequality/latest/thousand_bins_distribution` step. I believe that specifying the latest version in this step's DAG should be enough.

## Summary
- Add an S3 export step for the 2026 World Bank PIP thousand-bin income distribution data.
- Register `export://s3/poverty_inequality/latest/income-distribution` in the poverty and inequality DAG.
- Emit compact country-keyed JSON at `data/poverty-inequality/income-distribution.2026.json`.

## Validation
- `DRY_RUN=1 .venv/bin/etlr export://s3/poverty_inequality/latest/income-distribution --export --private`


etl/steps/export/s3/poverty_inequality/latest/income-distribution.py
```diff
@@ -0,0 +1,116 @@
...
+
+from etl.config import DRY_RUN
+from etl.data_helpers.misc import round_to_sig_figs
+from etl.helpers import PathFinder
+from etl.paths import EXPORT_DIR
+
+# S3 bucket name and folder where dataset files will be stored.
+S3_BUCKET_NAME = "owid-public"
+S3_DATA_DIR = Path("data/poverty-inequality")
+EXPORT_YEAR = 2026
```
paarriagadap: I would prefer this to be the latest year and not hardcode to 2026

etl/steps/export/s3/poverty_inequality/latest/income-distribution.py
```diff
@@ -0,0 +1,116 @@
...
+OUTPUT_FILE = f"income-distribution.{EXPORT_YEAR}.json"
```
paarriagadap: Same here, hardcoding the year can be a problem for future versions of the data

paarriagadap (approved):
Looks good!

My only comment is about hardcoding the year, which is not so future-proof. I would prefer to assign the latest year instead.

It's true that the previous step looks redundant now, but I am keeping it for now, because I might use it for static charts.

marcelgerber:
I've now changed it so that it always creates a data file for each year. Then we don't need to hardcode anything here, and can just flexibly use whichever data file we need.
It's not like it costs us much in terms of time or space, anyhow.

PR #6150, opened 2026-05-19
✨ Surface fast-track errors and fix unnamed-column edge case

Marigold:

## Summary

Three fast-track UX fixes that came out of importing a dataset on prod, plus a couple of region aliases needed for harmonization. The dataset itself will be added in a separate PR.

### Wizard fixes

- **Surface forked-step errors in the UI.** `etl_main(...)` runs the step via `os.fork()`; the child prints the traceback to `stderr` then exits non-zero, the parent calls `sys.exit(exit_code)`. Streamlit treats `SystemExit` as normal termination, so the wizard silently stopped and the traceback was only visible in `pm2`'s `wizard-error.log`. Wrapped the call in try/except for `SystemExit`/`Exception` and tee'd stderr into a buffer so the captured traceback is rendered via `st.code(..., language=\"python-traceback\")`. The original stderr stream still reaches the log file. (`apps/wizard/app_pages/fasttrack/app.py`)
- **Drop fully-empty columns from sheet imports.** A stray trailing comma in the published sheet header makes pandas synthesize `Unnamed: 9`, which becomes `unnamed__9` after `underscore()` and fails validation. Added `data_df.dropna(axis=1, how=\"all\")` in `parse_data_from_sheets` so empty columns disappear before validation. For the local-CSV path, `parse_metadata_from_csv` now infers from the *parsed* data columns rather than the raw `csv_df.columns`, keeping data and metadata aligned. (`apps/wizard/app_pages/fasttrack/load.py`)
- **Better error when an unnamed column still slips through** (e.g. it has a stray value somewhere). Validation now tells the user to delete empty columns in the `data` tab and wait ~5 min for the published CSV to refresh, instead of just \`Variable unnamed__9 is not defined in metadata\`. (\`apps/wizard/app_pages/fasttrack/process.py\`)

### Region aliases

- \`México\` → Mexico
- \`Democratic Congo\` → Democratic Republic of Congo

## Test plan

- [ ] Trigger an ETL failure inside fast-track (e.g. by re-importing a sheet with duplicate \`(country, year)\` rows) and confirm the traceback now renders in the wizard UI.
- [ ] Verify a sheet with a trailing-comma header still imports cleanly (no \`unnamed__N\` error).
- [ ] Upload a local CSV with a trailing empty column and confirm neither \"Variable unnamed__N is not defined in metadata\" nor \"Variable unnamed__N in metadata is not in the data\" appears.
- [ ] Confirm \`México\` and \`Democratic Congo\` resolve to the expected ISO codes through \`paths.regions.harmonize_names\`.


Marigold:
@codex review

PR #6172, opened 2026-05-28
🎉 Add review-explorer-mdim-mapping skill

pabloarosado:

## Review

Hey @lucasrodes there's no need to review the code; we'll probably use the skill just once for the remaining csv-based explorer. But feel free to try it out with natural disasters and see the resulting HTML tool.

## Summary

Adds a new internal skill `review-explorer-mdim-mapping` that turns the proposal produced by [`map-explorer-to-mdim`](.claude/skills/map-explorer-to-mdim/SKILL.md) into a **single self-contained HTML** for human side-by-side review of an explorer → MDIM view mapping.

The natural-disasters migration was the seed case: we needed a way to share 204 (explorer view → proposed MDIM view) pairs with a reviewer so they could approve or flag each one. The HTML works out of the box (zero deps, no server, decisions auto-save in the reviewer's browser), and the skill prints a pre-build coverage summary that surfaces gaps a human can't see by eye — distinct MDIM targets, many-to-one collapses, unresolved rows, MDIM views never targeted.

## How it fits in

`migrate-explorer-*` produces an MDIM → `map-explorer-to-mdim` proposes the redirect mapping → **this skill** lets a human verify each pair → the verified mapping (export CSV/JSON) drives the redirects. Two skills, two concerns: proposal vs. review.

## How to use it

Once `map-explorer-to-mdim` has produced `ai/<slug>-mdim-mapping/`, just ask Claude:

> "build the review HTML for the &lt;slug&gt; explorer→MDIM mapping"

(or invoke `/review-explorer-mdim-mapping`). Claude reads `SKILL.md`, asks for the **published MDIM Grapher slugs** if you haven't given them (the one thing it can't auto-derive), runs the underlying script, prints the coverage report, and tells you where the HTML landed. You then send the single `.html` file to your reviewer.

Under the hood it runs:

```bash
.venv/bin/python .claude/skills/review-explorer-mdim-mapping/scripts/build_review.py \
    --mapping-dir ai/<slug>-mdim-mapping \
    --explorer-slug <slug> \
    --mdim-slug <short>=<published-grapher-slug> [--mdim-slug ...]
```

— exposed as a normal CLI so power users (or CI) can call it directly if they want.

## What's in the HTML

- One pair at a time: old explorer (left) vs proposed MDIM (right), both rendered live from the configured host (default production); selection chips + "open ↗" full URL under each chart.
- **Approve ✓ / Flag ⚠ / Clear**, per-row note, prev/next + jump dropdown, filters (All / To review / Approved / Flagged), live counts, keyboard shortcuts (`←`/`→`, `a`, `f`, `c`).
- Persistence: every action saves to `localStorage` (refresh-safe). Optional **Auto-save to file** mirrors changes to a JSON on disk (Chrome/Edge File System Access). **Import** restores/merges from any exported JSON — the recovery path and the way two reviewers combine decisions.
- Settings panel lets a reviewer swap the host to compare against a `staging-site-<branch>` instead of production.

## Test plan

- [x] `make check` passes (lint + types).
- [x] Generated a `mapping_proposal.csv` + `mapping_rules.py` + `multidim_*_views.csv` fixture from the natural-disasters hand-made mapping and ran the skill end-to-end — it produced an 80 KB self-contained HTML embedding 204 view pairs.
- [x] Coverage report fires before HTML generation and reports distinct targets, many-to-one rows, and unresolved rows.
- [x] Embedded JavaScript syntax-checks clean (`node --check`).
- [x] HTML opens locally; both panes load from production, both sides default to World, navigation + approve/flag persistence + Settings + Import all work.
- [ ] Try on the next real explorer→MDIM migration that goes through `map-explorer-to-mdim` end-to-end (the natural-disasters validation used a retro fixture in `ai/`, not the real upstream output, because that mapping was hand-made before `map-explorer-to-mdim` existed).


lucasrodes (commented):
Hi @pabloarosado! Thanks for drafting this. I think the idea sounds great overall.

The only thing I'd maybe change is to add some context on how the resulting `index.html` can be shared with the team: via vibe.owid.io internal site. To create apps/entries in there, we have a [dedicated skill](https://github.com/owid/owid-claude-plugins-staff/blob/main/plugins/owid-vibe-apps/skills/vibe-app/SKILL.md). So I'd suggest making your skill aware of that skill. Note that this skill should be installed beforehand (use `\plugin` in claude shell).

pabloarosado:
@codex review

pabloarosado:

Thanks @chatgpt-codex-connector — both findings landed in 7e2d1a949:

- **Filtered-decide skip**: `decide()` now remembers the next record's id *before* re-applying the filter, so approve/flag under the *To review* filter lands on the row that was next in the queue (not the one after it).
- **MDIM prefix shadowing**: `parse_mapping()` now tries the longest MDIM short as the column prefix first, so e.g. `deaths_by_age_type` won't be misattributed to `deaths`.

Validated by re-running the skill against the natural-disasters fixture; coverage and JS syntax check unchanged.

Issue #6069, opened 2026-05-11
tracking: Bringing charts and mdims closer together in etl and grapher


pabloarosado:
> _Written by Claude Fable 5 — @pabloarosado at the wheel._

## Summary

This issue will track progress on the roadmap to bring charts and MDIMs closer together in both etl and grapher.

For more context, see [this Notion page](https://www.notion.so/owid/2026-05-Merging-charts-and-MDIMs-from-in-etl-and-grapher-36b74b71f92e80eea99ecc0dc3dafbae?source=copy_link).

### Old
* https://github.com/owid/etl/pull/5530
* https://github.com/owid/etl/pull/5969

### Phase 1
* https://github.com/owid/owid-grapher/pull/6553 (superseded by owid/owid-grapher#6825 and owid/owid-grapher#6826)
* https://github.com/owid/etl/pull/6174 (superseded by owid/etl#6511)
* https://github.com/owid/owid-grapher/pull/6825
* https://github.com/owid/owid-grapher/pull/6826
* https://github.com/owid/etl/pull/6511

### Chart-configs refactor (groundwork)
* https://github.com/owid/owid-grapher/pull/6971
* https://github.com/owid/owid-grapher/pull/6979
* https://github.com/owid/owid-grapher/pull/6980
* https://github.com/owid/owid-grapher/pull/6988
* https://github.com/owid/owid-grapher/pull/6987
* https://github.com/owid/owid-grapher/pull/7002
* https://github.com/owid/owid-grapher/pull/7004

### Related
* https://github.com/owid/owid-grapher/pull/6922

PR #6174, opened 2026-05-28
🎉 Charts as first-class citizens in ETL: layered approach with grapher

pabloarosado:
## Human summary

Lets ETL create and edit charts as zero-dimension mdim collections, using a layered approach, so that changes made from the grapher chart admin supersede the ETL config.

Depends on **[owid/owid-grapher#6553](https://github.com/owid/owid-grapher/pull/6553)**.
Tracking issue: https://github.com/owid/etl/issues/6069
This PR is the etl-side of Phase 1 in the [roadmap document](https://app.notion.com/p/owid/2026-05-Bringing-charts-and-MDIMs-closer-together-in-etl-and-grapher-36b74b71f92e80eea99ecc0dc3dafbae).

---

## Core implementation idea

A chart is defined in etl as an `export://multidim` step (with a `.config.yml` file) with `dimensions: []` (i.e. no dropdown selectors; not to be confused with the grapher config's `dimensions`, which lists the plotted indicators) and exactly one view. This step gets routed through `Collection.upsert_to_db` to the new `PUT /admin/api/charts/:id/etlConfig` endpoint, which stores the chart's grapher config as a separate `chart_configs` row, pointed to by a new `charts.configIdETL` column (mirroring how `variables.grapherConfigIdETL` already works).

The rendered config is `merge(variableETL, etlConfig, patch)`: ETL writes the `etlConfig` row, admins write `patch`, so the two never collide and an ETL re-push can't clobber an admin edit. We also set `charts.catalogPath` as the chart's stable ETL identity (like multi-dims already have).

## Additional changes

- Export multidim steps can now be executed just with the config file (even without a .py file).
- Adapted the chart-preview skill to let Claude preview draft charts (this is important for newly created charts).
- Adapted chart-editing skill to the new approach.
- Adapted VSCode extensions (clickable dag steps and chart preview).
- Adapted version tracker.
- Created step to replicate [the chick-culling laws chart](http://staging-site-etl-chart-config/admin/charts/7118/edit) as an example.

## How to test

- Fetch this branch and (pointing to the corresponding staging server) ask Claude to create a new chart (on whatever topic). Use the chart preview extension to visualize it.
- Make changes in the chart admin (e.g. edit the title).
- Run the export step of the chart again, and check that the title is still the one manually set from the admin.
- From the chart admin, link again the title field to etl, and check that the chart uses the etl values again.
- (After doing the test, remove the chart and the step).


pabloarosado:
@codex review

schemas/multidim-schema.json
```diff
@@ -328,11 +328,11 @@
         "dimensions": {
             "type": "array",
             "title": "Collection dimensions",
-            "description": "Array of dimension objects that define the filter dropdowns available to users.",
+            "description": "Array of dimension objects that define the filter dropdowns available to users. An empty array (and a single view) means the collection is pushed as a regular Grapher chart instead of a multi-dim page.",
             "requirement_level": "required",
             "guidelines": [
                 [
-                    "Must contain at least one dimension."
+                    "Can be empty when the collection is a single chart (no dimension dropdowns)."
```
lucasrodes, line 340: note for self: check that step fails whenever there is no dimensions and >1 views.

etl/steps/export/multidim/animal_welfare/latest/banning_of_chick_culling.config.yml
```diff
@@ -0,0 +1,38 @@
...
+  - "Animal Welfare"
+dimensions: []
+views:
+  - dimensions: {}
+    indicators:
+      y:
+        - "chick_culling_laws#status"
+    config:
+      $schema: "https://files.ourworldindata.org/schemas/grapher-schema.009.json"
+      title: "Which countries have banned chick culling?"
```
lucasrodes, line 10: redundancy between `views[].dimension.config.title` and `title.title`?

schemas/multidim-schema.json
lucasrodes: this is actually accounted for in `etl.collection.model.core`, so nvm!

etl/collection/model/core.py
```diff
@@ -251,6 +251,22 @@ def save(  # ty: ignore[invalid-method-override]
         self.upsert_to_db(owid_env)
 
     def upsert_to_db(self, owid_env: OWIDEnv):
+        # Degenerate "single chart" case: no dimensions means no multi-dim page —
+        # push to the charts table via the chart admin endpoint instead of
+        # `/multi-dims/`. A zero-dimension collection with multiple views is
+        # not a chart and not a valid MDIM either, so we refuse it explicitly.
```
lucasrodes: for our future selves, maybe we can add a note that the ideal scenario, is that both charts & mdims are pushed to the same table and grapher handles that? As in, there's no need for having mdims and charts tables separate other than technical/legacy reasons

etl/collection/chart_upsert.py
```diff
@@ -0,0 +1,121 @@
...
+        result = admin_api.create_chart(chart_config=bootstrap)
+        chart_id = result["chartId"]
+    else:
+        chart_id = existing.id
+        log.info("collection.chart.update", slug=slug, chart_id=chart_id)
+
+    # Write the chart's ETL-authored config. This recomputes `full` server-side
+    # as merge(variableETL, etlConfig, existing patch); any admin patches
+    # already in chart_configs.patch are preserved.
+    admin_api.put_chart_etl_config(chart_id=chart_id, grapher_config=chart_config)
```
lucasrodes: not sure if needed. Or if here is a good place (or maybe inside `put_chart_etl_config`), but I remember using a method to validate the chart config before adding it to the DB.

The method is `validate_chart_config`, in `etl/indicator_upgrader/schema.py` (odd place now that i think of it)

https://github.com/owid/etl/blob/2342205b7cd2d6f947f14663f62ee36d166f7650/etl/indicator_upgrade/schema.py#L13-L17

again, not sure if it is needed or if we already run checks on the side of Grapher API.

etl/steps/export/multidim/animal_welfare/latest/banning_of_chick_culling.config.yml
pabloarosado: Thanks, good catch! The redundancy was due to the fact that mdims used to require a `title` (and a `title_variant`), which is meaningless for charts.
I've changed things a bit in the last commit:
- `title.title` is now optional in the schema; but an error is raised when running an mdim (with dimensions) if there's no top-level title.
- `title.title_variant` is now optional (I don't understand why it was mandatory in the first place).
- `default_selection` is also optional; but a warning is raised when a non-map-only view (e.g. a view with a line tab) has neither its own `selectedEntityNames` nor a top-level `default_selection` to fall back to.

Do you agree?

Marigold:
Very nice! Worked a charm for me. A couple of ideas, but nothing prevents merging.

1. We risk overriding hand authored charts with the same slug. If the chart doesn't have `etlConfig`, we should be cautious about overwriting it.
2. Would it help to have some tests for e.g. `_build_chart_config`?
3. If we start using this intensively, having a tool to migrate hand authored chart to ETL would be cool
4. We should highlight ETL-authored charts in Admin

lucasrodes (commented):
Awesome, @pabloarosado! This is a great addition. I quickly scanned the code, and left some minor suggestions. But overall, nothing critical! Excited to try out this tool in PROD.

I've tried the workflow with Claude and it mostly worked fine. Below are some comments on the different stages and a final one on removing charts


### Creating the chart
> Fetch this branch and (pointing to the corresponding staging server) ask Claude to create a new chart (on whatever topic).
This worked fine! I prompted Claude with the following instruction

```
i want to test the work from this PR. I've been suggested
  to create a chart to begin with, can you assist me w that?
  I'd like a scatterplot  that shows life expectancy vs
  population.
```

And it generated a new YAML file in `etl/steps/export/multidim/demography/latest/life_expectancy_vs_population.config.yml`. It looked good.

On the flip side, I felt it was a bit slow (took ~4 minutes). I am not sure if this is expected, or maybe there is something we can improve in the SKILL to make this faster. From my understanding, the skill presumably in use (`chart-editing`) is focussed on editing charts. Should we adapt it to also have the appropriate context for *creating* charts?

<details>
<summary>Claude's full transcript</summary>

```
● Skill(chart-editing)
  ⎿  Successfully loaded skill

● Let me look at the current state of the graph steps and PR
  work to understand the setup.

  Searched for 1 pattern (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl log --oneline master..HEAD
      | head -20; echo ---; git -C /home/lucas/repos/etl diff
      master...HEAD --stat | tail -30)
  ⎿  8c6a5d037 Merge remote-tracking branch 'origin/maste
     r' into refactor-migrate-co2
     6b640f214 📊 Guinea Worm data update (#6178)
     … +45 lines (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl branch --show-current;
      echo ---; git -C /home/lucas/repos/etl log --oneline -8;
      echo ---; rg --files /home/lucas/repos/etl -g "*.ch…)
  ⎿  refactor-migrate-co2
  work to understand the setup.

  Searched for 1 pattern (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl log --oneline master..HEAD | head -20; echo ---; git -C /home/lucas/repos/etl diff
      master...HEAD --stat | tail -30)
  ⎿  8c6a5d037 Merge remote-tracking branch 'origin/master' into refactor-migrate-co2
     6b640f214 📊 Guinea Worm data update (#6178)
     d9e42c81b Add women's rights, human rights, crime, media, social connection and technology indicators to Int
     … +36 lines (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl branch --show-current; echo ---; git -C /home/lucas/repos/etl log --oneline -8;
      echo ---; rg --files /home/lucas/repos/etl -g "*.ch…)
  ⎿  refactor-migrate-co2
     ---
     8c6a5d037 Merge remote-tracking branch 'origin/master' into refactor-migrate-co2
     … +45 lines (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl branch -a | rg -i "chart" ; echo ---; git -C /home/lucas/repos/etl log --oneline
      -5 etl-chart-config 2>/dev/null; echo ---; git -C…)
  ⎿    bug-chart-diff-failure
       bug-chart-diff-query
       bug-remove-frame-chart
     … +57 lines (ctrl+o to expand)

● The branch exists but I need to see what's actually in it — the paths the skill mentions didn't match. Let me check
  what the PR changed:

● Bash(git -C /home/lucas/repos/etl diff master...etl-chart-config --stat | tail -40)
  ⎿   .claude/skills/chart-editing/SKILL.md              | 230 +++++++++++----------
      .claude/skills/check-chart-preview/SKILL.md        |  40 ++--
      .../check-chart-preview/get_chart_png_url.py       |  78 +++++++
     … +18 lines (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl show
      etl-chart-config:etl/steps/export/charts/animal_welfare/latest/banning_of_chick_culling.config.yml; echo
      ===DAG===; git -C /ho…)
  ⎿  Error: Exit code 128
     fatal: path 'etl/steps/export/charts/animal_welfare/latest/banning_of_chick_culling.config.yml' does not exist in
     'etl-chart-config'
     (eval):1: ==DAG=== not found

● Bash(git -C /home/lucas/repos/etl show etl-chart-config:.claude/skills/chart-editing/SKILL.md)
  ⎿  ---
     name: chart-editing
     description: Create or edit an ETL-authored Grapher chart — a single-chart `.config.yml` in `etl/steps/expor
     … +161 lines (ctrl+o to expand)

● Bash(git -C /home/lucas/repos/etl checkout etl-chart-config && git -C /home/lucas/repos/etl status --short | head)
  ⎿  Switched to branch 'etl-chart-config'
     Your branch is up to date with 'origin/etl-chart-config'.
     ?? scripts/apply_softhyphen_prefixes.py
     … +4 lines (ctrl+o to expand)

  Listed 1 directory (ctrl+o to expand)

● Bash(rg "data://grapher/demography/2024-07-18/population|data://grapher/demography/2025-10-22/life_expectancy"
      /home/lucas/repos/etl/dag/*.yml -l | head -5; echo ---…)
  ⎿  /home/lucas/repos/etl/dag/demography.yml
     /home/lucas/repos/etl/dag/static_viz.yml
     ---LE-COLS---
     … +19 lines (ctrl+o to expand)

  Searched for 1 pattern (ctrl+o to expand)

● Bash(rg -A5 "grapher/demography/2024-07-18/population" /home/lucas/repos/etl/dag/demography.yml | head -10; echo
      ---; ls /home/lucas/repos/etl/etl/steps/data/grapher…)
  ⎿  Error: Exit code 2
       data://grapher/demography/2024-07-18/population_doubling_times:
         - data://garden/demography/2024-07-18/population_doubling_times

       ## Population comparison
       data://grapher/demography/2023-07-03/world_population_comparison:
         - data://garden/demography/2023-06-27/world_population_comparison:
     ---
     population_doubling_times.py
     __pycache__

● Branch switched — the skill list now reflects the PR version. Now let me pin down the population grapher step and
  the exact indicator paths:

  Listed 1 directory (ctrl+o to expand)

● Bash(.venv/bin/python -c "
      from owid.catalog import Dataset…)
  ⎿  ['historical', 'population', 'population_density', 'population_growth_rate', 'projections']
     ['population', 'source', 'world_pop_share']
     ---
     … +40 lines (ctrl+o to expand)

● Indicators identified. Let me check how existing configs handle scatterplots (especially the continents color
  dimension):

  Searched for 4 patterns (ctrl+o to expand)

● I have everything I need. Let me verify how config-only export steps work (whether a .py is needed) and check the
  chick-culling DAG entry:

● Bash(ls /home/lucas/repos/etl/etl/steps/export/multidim/animal_welfare/latest/; echo ---DAG---; rg -n -B3 -A4
      "multidim/animal_welfare" /home/lucas/repos/etl/dag/ani…)
  ⎿  banning_of_chick_culling.config.yml
     ---DAG---
     112-  #
     … +55 lines (ctrl+o to expand)

● YAML-only config steps confirmed. Now let me look at where to add the DAG entry and check the multidim schema for
  the scatter indicator shape:

● Bash(rg -n "life_expectancy" /home/lucas/repos/etl/dag/demography.yml | head; echo ---; tail -20
      /home/lucas/repos/etl/dag/demography.yml; echo ---SCHEMA---; .venv/b…)
  ⎿  184:  data://grapher/demography/2025-10-22/life_expectancy:
     185:    - data://garden/demography/2025-10-22/life_expectancy:
     208:  data://grapher/oecd/2023-10-11/life_expectancy_birth:
     … +89 lines (ctrl+o to expand)

  Read 1 file (ctrl+o to expand)

● Now I'll create the chart config and DAG entry:

● Write(etl/steps/export/multidim/demography/latest/life_expectancy_vs_population.config.yml)
  ⎿  Wrote 31 lines to etl/steps/export/multidim/demography/latest/life_expectancy_vs_population.config.yml
       1 topic_tags:
       2   - "Life Expectancy"
       3 dimensions: []
       4 views:
       5   - dimensions: {}
       6     indicators:
       7       y:
       8         - catalogPath: "life_expectancy_at_birth#life_expectancy_0"
       9           display:
      10             name: "Life expectancy at birth"
     … +21 lines (ctrl+o to expand)

● Update(dag/demography.yml)
  ⎿  Added 8 lines
      188          - data://garden/demography/2025-10-22/life_tables
      189          - data://garden/un/2024-07-12/un_wpp
      190
      191 +  #
      192 +  # Life expectancy vs. population — chart authored in ETL.
      193 +  #
      194 +  export://multidim/demography/latest/life_expectancy_vs_population:
      195 +    - data://grapher/demography/2025-10-22/life_expectancy
      196 +    - data://grapher/demography/2024-07-15/population
      197 +    - data://grapher/regions/2023-01-01/regions
      198 +
      199    # Broken limits of Life Expectancy
      200    data://grapher/demography/2025-10-22/broken_limits_le:
      201      - data://garden/demography/2025-10-22/broken_limits_le:

● Both files are in place. Let me validate the DAG wiring with a quick dry-run (no execution):

● Bash(.venv/bin/etlr export://multidim/demography/latest/life_expectancy_vs_population --export --dry-run 2>&1 | tail
      -15)
  ⎿  --- Detecting which steps need rebuilding...
     OK (1.1s)
     --- Would run 9 steps (at least 3m 51.1s):
     … +9 lines (ctrl+o to expand)

● Everything resolves. The chart is ready to test. Here's what I set up:

  What was created

  1. Chart config — etl/steps/export/multidim/demography/latest/life_expectancy_vs_population.config.yml

  A zero-dimension mdim (the PR's single-chart format) with one scatter view:
  - y: life_expectancy_at_birth#life_expectancy_0 (OMM life expectancy, demography/2025-10-22/life_expectancy)
  - x: population#population (demography/2024-07-15/population), log scale since population spans ~5 orders of
  magnitude
  - color: regions#owid_region — continent coloring, same pattern the COVID scatters use

  2. DAG entry — dag/demography.yml:191 with the three grapher deps.

  The dry-run validates fine — public slug will be life-expectancy-vs-population.

  Next step (for you to run)

  Some upstream deps aren't built locally (it wants to rebuild the hmd → life_tables chain, ~4 min), so fetch them
  from the catalog instead:

  PREFER_DOWNLOAD=1 .venv/bin/etlr export://multidim/demography/latest/life_expectancy_vs_population --export
  --private

  On success it prints admin_url=http://staging-site-etl-chart-config/admin/charts/<id>/edit.

  Then for iterating on the YAML afterwards, the fast loop is:

  .venv/bin/etlr export://multidim/demography/latest/life_expectancy_vs_population --export --force --only

  Once it's pushed, I can fetch the PNG preview (via get_chart_png_url.py from the PR) so we can check the scatter
  visually — let me know how the run goes.

✻ Cooked for 4m 14s
```

</details>

### Previewing the chart
My understanding from Claude's output is that I need to execute the `export` step. So  the preview is only possible after the chart exists in Admin. Just to confirm, did I understand this correctly?

After running the `etlr` command, I went ahead and typed `/check-chart-preview`.

I like the previewing chart a lot, but I think it'd be helpful at some point to get the URL to the chart. Maybe when `etlr` runs the `export://` step and upserts a chart to the DB, so that it is easier to go to the live chart.

E.g. some message after the `--- Finished export://multidim/demography/latest/life_expectancy_vs_population (2.8s)`, like `chart available at staging-.../admin/charts/<chart_id>/edit`.

### Updating & syncing with Admin
> Run the export step of the chart again, and check that the title is still the one manually set from the admin.

works perfectly! Also, if then i go and edit the title in ETL to match Admin's, then Admin recognizes that it is synced with ETL.

> From the chart admin, link again the title field to etl, and check that the chart uses the etl values again.

Works!


### Removing charts
One thing that came to mind is what should be the behavior when someone removes a chart in Admin. If that happens, and we don't do the same in ETL (e.g. archiving the chart step), the chart will appear again in PROD. Do you have an intuition on how we can solve this to avoid human errors? Somehow we should make ETL aware of the list of catalog paths of the "removed charts".

Maybe we need to add a new table in Grapher (e.g. `charts_removed`) that lists the IDs and chart paths of the removed charts (along with other fields like `removedAt`, `removedBy`, etc.). This is a bit related to the idea of having `catalogPath` in the charts table, which I shared in https://github.com/owid/owid-grapher/pull/6553#issuecomment-4612342600.

etl/steps/export/multidim/animal_welfare/latest/banning_of_chick_culling.config.yml
lucasrodes: sounds good.

I only fear we might end up with a too flexible schema file for multidims because of the "degenerate" chart version. It works fine here bc we have a Python method checking, but otherwise the schema is starting to be less strict. Maybe we end up requiring two schemas in the end... (with some common fields defined in one place or so)

pabloarosado:
> Very nice! Worked a charm for me. A couple of ideas, but nothing prevents merging.
> 
> 1. We risk overriding hand authored charts with the same slug. If the chart doesn't have `etlConfig`, we should be cautious about overwriting it.
> 2. Would it help to have some tests for e.g. `_build_chart_config`?
> 3. If we start using this intensively, having a tool to migrate hand authored chart to ETL would be cool
> 4. We should highlight ETL-authored charts in Admin

Hi Mojmir, thanks for the review! I agree with all your points:
1. I can't think of any dangerous cases. Chart diff would catch differences between full chart configs.
2. Done.
3. This is already in [TODO for Phase 1](https://app.notion.com/p/owid/2026-05-Bringing-charts-and-MDIMs-closer-together-in-etl-and-grapher-36b74b71f92e80eea99ecc0dc3dafbae?source=copy_link#37374b71f92e80598545dd6da0b63731).
4. Added to [TODO for Phase 1](https://app.notion.com/p/owid/2026-05-Bringing-charts-and-MDIMs-closer-together-in-etl-and-grapher-36b74b71f92e80eea99ecc0dc3dafbae?source=copy_link#37374b71f92e80598545dd6da0b63731).

etl/steps/export/multidim/animal_welfare/latest/banning_of_chick_culling.config.yml
pabloarosado: Ideally, we'll end up fully unifying charts and mdims ([Phase 4 of the roadmap](https://app.notion.com/p/owid/2026-05-Bringing-charts-and-MDIMs-closer-together-in-etl-and-grapher-36b74b71f92e80eea99ecc0dc3dafbae?source=copy_link#37374b71f92e80f2ac5af0db2a0bcae1)). So the current situation is an imperfect but working solution. We could split the schemas, but then we might be moving away from the place where we want to be in the near future.

pabloarosado:
Hi @lucasrodes, thanks for trying it out and for the feedback! Let me address your comments:
- From looking at the transcript, I think there are multiple things taking Claude time (e.g. changing branches, getting more context, etc.). We could figure out some way to improve the skill to avoid some of those steps and reduce the time. Let's do that in a separate PR.
- On previewing: The chart preview extension already executes the export step on opening (even before the chart exists on grapher). And chart preview already has a link to the corresponding chart preview at the top.
- The `/check-chart-preview` skill is mostly for Claude to be able to inspect the outputs. The user should be inspecting the chart preview directly (or go to the admin).
- On removing charts: good point. Your `charts_removed` idea make sense. More broadly we currently don't have a good way to handle the removal of charts on staging (e.g. when you delete a chart on staging, chart-sync doesn't delete the chart in prod; you just get a warning message). We should handle the issue of how we handle chart removals in a separate issue/PR.

I'll have a look at the rest of your comments, thanks for the thorough review!

etl/collection/model/core.py
pabloarosado: Done.

etl/collection/chart_upsert.py
pabloarosado: Done (without reusing the old validate_chart_config).

pabloarosado:
@lucasrodes @Marigold I just realised you never approved this PR 🧐
Any further concerns from your side?

pabloarosado:

## End-to-end testing on staging

Since this branch had been sitting for a while and the merge picked up a fair amount of upstream change, I exercised Phase 1 end-to-end against `staging-site-etl-chart-config` (driving the real export step, the grapher `etlConfig` endpoints, and admin edits, and inspecting the resulting `charts` / `chart_configs` / `charts_x_parents` rows after each step). Every throwaway chart and step was cleaned up afterwards, so staging is back to baseline.

### Verified working

| Behavior | Result |
|---|---|
| Create chart from ETL + correct 4-layer structure (`etlConfig` row, tiny bootstrap `patch`, rendered `full`, `catalogPath` stamped), with the `dimensions` residue correctly stripped from `patch` | ✅ |
| ETL edit propagates on re-push | ✅ |
| **Admin edit survives an ETL re-push** (the core guarantee) | ✅ |
| Re-linking a field back to ETL falls back to the ETL value | ✅ |
| **No version churn on re-runs**: a scheduled (unchanged) run skips the step entirely; a forced run is detected as a no-op (no version bump, no revision, no R2 re-upload, no rebuild) | ✅ |
| **Indicator re-versioning**: chart re-points to the new variable and the inheritance parent follows it, no stale `dimensions` in `patch` | ✅ |
| **Admin-hand-edited dimensions win** after a subsequent ETL re-version | ✅ |
| Full three-layer inheritance `variableETL → etlConfig → patch` (a chart inherits an omitted field from the indicator's own grapher config) | ✅ |
| Adopt-by-slug of an existing hand-made chart, without clobbering its content, plus the guard that refuses to hijack a chart owned by another step | ✅ |
| Detaching from ETL is render-neutral | ✅ |
| Publishing survives an ETL re-push, and the published-by-slug config is served from R2 | ✅ |
| Malformed config rejected pre-push with a clear, field-located error | ✅ |
| Multi-indicator chart dimensions | ✅ (the working-parent/link behavior is client-side and covered by the grapher PR's vitest tests) |
| Unit suite for the changed modules (`chart_upsert`, `chart_diff`, `geo`, `grapher_model`) | ✅ |

### Known soft gap (already noted in the roadmap)

A deleted ETL chart is recreated by the next run **as an unpublished draft** (`isPublished=false`), and only when the step actually executes again (a routine unchanged run skips it). So a deliberately-deleted chart reappears as a draft rather than silently going back live. Fine to handle in a follow-up.

### Not covered here

The rendered pixels (I validated the config instead; the rendering path itself is unchanged by this PR) and a live staging-vs-prod Chart Diff run (needs prod; the logic is covered by unit tests).

Net: the behaviors that would actually hurt in production, clobbering human edits, churning every chart every run, breaking on data updates, or hijacking/resurrecting charts, are all either verified correct or characterized as soft and deferred. Nothing surfaced that should block Phase 1.

Marigold (approved):
Looks good on my end!

pabloarosado:
This PR has been superseeded by https://github.com/owid/etl/pull/6511
but let's keep it open for now as a reference, while the other one is not merged.

PR #6181, opened 2026-06-01
🐝 Discontinue monkeypox update — Global.health mpox-2024 source dead

Marigold:

## What

Disables the automated monkeypox update because the **Global.health mpox-2024 line list has been discontinued**.

## Why

- Global.health's last mpox-2024 report is dated **2024-12-20**, and its download endpoint (`https://mpox-2024.s3.eu-central-1.amazonaws.com/latest.csv`) now returns **`403 Access Denied`**.
- There is **no replacement endpoint** — the official mpox-2024 report still points at the same (now-dead) URL, and the data portal/API gateway alternatives don't serve the same country×date series.
- The Global.health data was only ever the supplementary `suspected_cases_cumulative` column and was already **frozen** (nothing past 2023-12-01). It remains available from our catalog/R2, so downstream steps are unaffected.
- The WHO source (the primary one) is still live, but the update was failing in `scripts/update-monkeypox.sh` on the Global.health snapshot step, blocking the whole run.

## Changes

- **`scripts/update-monkeypox.sh`** — disabled. It now logs an informative line and `exit 0`s cleanly so the scheduler sees a clear reason; the original commands are kept commented out for reference/re-enabling.
- **`snapshots/health/latest/global_health_mpox.csv.dvc`** — added a note that the source is discontinued and `url_download` returns 403; data stays available from R2.

The snapshot script and downstream ETL steps are left untouched.

### DIFF
--- a/scripts/update-monkeypox.sh
+++ b/scripts/update-monkeypox.sh
@@ -1,24 +1,34 @@
 #!/bin/bash
 
-set -e
+# This update has been discontinued.
+# Global.health discontinued the mpox-2024 line list (last report 2024-12-20) and its download
+# endpoint (https://mpox-2024.s3.eu-central-1.amazonaws.com/latest.csv) now returns 403 Access Denied.
+# There is no replacement endpoint, so we no longer refresh this snapshot — the frozen historical data
+# (supplementary suspected_cases_cumulative) stays cached in R2 and downstream steps keep using it.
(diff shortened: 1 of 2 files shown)


Marigold:
cc @spoonerf, the link stopped working. We can revert this if we find a new reliable source.

PR #6190, opened 2026-06-03
✨🤖 Modernize add_population_to_table → paths.regions.add_population

Marigold:

## What

Modernizes the deprecated standalone `geo.add_population_to_table(...)` to the recommended `paths.regions.add_population(...)` across **54 active call sites in 41 steps**, in two commits:

1. **run() tier (23 sites / 21 files)** — calls directly in `run()`.
2. **helper-param tier (33 sites / 33 files)** — calls inside helpers that received `ds_population` as an argument; the param is dropped from the helper signature and from its caller(s), and the orphaned `load_dataset("population")` line removed.

## Why this is a no-op

`PathFinder.regions` loads `ds_population` via the step's **own `load_dataset("population")` dependency** (`etl/helpers.py:375`, `auto_load_datasets=False`) and delegates to the same underlying `add_population_to_table`. So the migration preserves the pinned population vintage and produces byte-identical output — it does **not** force the latest population version.

Only sites passing the standard `population` dataset were touched; population version is irrelevant to safety.

## Out of scope

- `countries/{cow_ssm,gleditsch}` — wrap the call in a *local* helper also named `add_population_to_table` (bespoke year-extension logic).
- A few sites with non-standard datasets / unresolved arg tracing.
- Archived/orphaned step files.
- `geo.add_region_aggregates` (85 sites) — separately deprecated; same safe-subset logic applies, possible future PR.

## Verification

- `ruff` + `ruff format` clean; `make check` passes.
- The helper-param edits were applied via a reconstruction-based AST codemod (position/keyword-aware signature + caller rewrites) and **every changed file's diff was reviewed**, including order-sensitive arg removal (`igme`), multi-call helpers (`famines`, `un_members`, `gcp`), and shared population-var cases.
- Rebuilt **11 representative steps** end-to-end (exit 0, each using its declared population dep), spanning all structural shapes: `meijer_2021`, `corruption_barometer`, `un/2025-07-03/refugee_data`, `igme`, `worldbank_wdi/2026-02-27/wdi`, `tuberculosis/burden_estimates`, `maternal_mortality`, `un_members`, `deaths_karlinsky`, `cigarette_sales`, `famines_by_regime_gdp_population`.

### DIFF
--- a/etl/steps/data/garden/animal_welfare/2024-12-12/farmed_finfishes_used_for_food.py
+++ b/etl/steps/data/garden/animal_welfare/2024-12-12/farmed_finfishes_used_for_food.py
@@ -1,8 +1,7 @@
 """Load a meadow dataset and create a garden dataset."""
 
-from owid.catalog import Dataset, Table
+from owid.catalog import Table
 
-from etl.data_helpers import geo
 from etl.helpers import PathFinder, create_dataset
 
 # Get paths and naming conventions for current step.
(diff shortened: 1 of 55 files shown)


Marigold:
@codex review

PR #6192, opened 2026-06-03
📊 Add dumbbell plots to inequality MDims

paarriagadap:

Adds a **Dumbbell** chart tab to the "Before vs. after taxes and benefits" view of the LIS and WID inequality MDims, so users can see each country's before- and after-tax value side by side (two dots connected by an arrow) alongside the existing line chart.

## MDims affected

| Source | MDims |
|---|---|
| **LIS** | `gini_lis`, `incomes_lis`, `relative_poverty_lis` |
| **WID** | `gini_wid`, `incomes_wid`, `palma_ratio_wid` |

WB PIP is **excluded** — its welfare types are income vs. consumption, not before/after tax, so a before-vs-after dumbbell doesn't apply there.

## What changed

- **Dumbbell tab.** `"Dumbbell"` added to `chartTypes` on each `before_vs_after` view (kept as a secondary tab; the line chart stays the default). With two y-indicators, Grapher's dumbbell renders in TwoColumn mode at the latest year — exactly the before/after-tax comparison.
- **Arrow direction (before → after).** The arrow points from `yColumns[0]` → `yColumns[1]`, i.e. the indicator order. `group_views` orders indicators by choice order, so the LIS choices were flipped to `["mi", "dhi"]` (before-tax first); WID was already before-first. The LIS metadata helpers now key off the before-tax indicator explicitly so the combined title/subtitle/description_key are unaffected by the reorder.
- **Entity sort.** Each view sets `sortBy: column` / `sortColumnSlug: <after-tax indicator>` / `sortOrder: asc` so the dumbbells (and table) order entities by after-tax value, lowest first. (Dumbbell's TwoColumn mode excludes the `startValue`/`endValue` sort keys, so sorting by the after-tax *column* is the way to do this.)
- **Schema sync.** `schemas/dataset-schema.json` gains the upstream `dumbbell` grapher_config object and the `Dumbbell` chartTypes enum value, so `test_grapher_config_schema_sync` passes.
- **Skill doc.** Documents a before/after FAUST regression-diff workflow in the `faust-metadata-audit` skill (used to verify this PR — see below).

## Verification

Diffed the generated MDim configs (pre-PR master vs. branch): the only changes are the intended ones (Dumbbell tab + LIS indicator reorder). FAUST reports (Title/Subtitle/Footnote/description_short/description_key) are **byte-identical** before/after for all three LIS MDims, confirming the dhi→mi inheritance-base shift changes no user-facing text. One accepted side effect: the LIS line chart's two series swap their auto-assigned colors (no explicit colors pre-PR, so palette follows series order).

### DIFF
--- a/schemas/dataset-schema.json
+++ b/schemas/dataset-schema.json
@@ -1643,9 +1643,52 @@
                               "SlopeChart",
                               "StackedBar",
                               "WorldMap",
-                              "Marimekko"
+                              "Marimekko",
+                              "Dumbbell"
                             ]
                           }
+                        },
+                        "dumbbell": {
+                          "type": "object",
+                          "description": "Configuration of the dumbbell chart",
+                          "properties": {
+                            "connectorStyle": {
+                              "type": "string",
+                              "default": "arrow",
+                              "description": "How the connector between dumbbell heads is drawn.\n- arrow: draw an arrow pointing from start to end\n- line: draw a simple line (only respected when two indicators are plotted)\n",
+                              "enum": [
+                                "arrow",
+                                "line"
+                              ]
+                            },
+                            "valueLabelMode": {
+                              "type": "string",
+                              "default": "absolute",
+                              "description": "What to display as value labels next to each dumbbell.\n- absolute: show the raw values at the start and end\n- change: show the absolute change (e.g. +50)\n- percentChange: show the percentage change (e.g. +33%)\n- none: hide value labels\n",
+                              "enum": [
+                                "absolute",
+                                "change",
+                                "percentChange",
+                                "none"
+                              ]
+                            },
+                            "trendColorMap": {
+                              "type": "object",
+                              "description": "Custom colors for the time-range encoding",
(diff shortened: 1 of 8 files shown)


schemas/dataset-schema.json
```diff
@@ -1643,10 +1643,53 @@
                               "SlopeChart",
                               "StackedBar",
                               "WorldMap",
-                              "Marimekko"
+                              "Marimekko",
+                              "Dumbbell"
                             ]
                           }
                         },
+                        "dumbbell": {
```
paarriagadap, line 1651: Hi @Marigold!

Let me know if this is the right thing to do: I wanted to try [Dumbbell plots](https://owid.slack.com/archives/C014SGT6W9X/p1780492799498209) but the unit test failed. Claude recommended this.

paarriagadap: Disregard this, it looks it's right and Lucas is adding more changes here https://github.com/owid/etl/pull/6196

PR #6222, opened 2026-06-08
📜🤖 Clarify snapshot vs garden boundaries and update-dataset guidance

edomt:

## Summary

Two doc improvements that came out of a frustrating session where Claude (a) put aggregation/dedup/date-conversion logic in the snapshot script before being told to read the docs, and (b) wrote too much detail in the PR description.

### CLAUDE.md

- **"Snapshot is raw passthrough only"** note after the Pipeline Overview table, with concrete examples of what belongs in garden instead (summing rows, dedup across files, label→date conversion, source-category renaming).
- **"Description fields: .dvc vs garden description_processing"** subsection — producer-side description in .dvc, OWID-side processing in garden, no overlap.

### update-dataset/SKILL.md

- Rewrote the `etl update` URI guidance. The old version told you to run from the garden URI, which only bumps garden+grapher; the new version: snapshot URI + --include-usages if source data changes, garden URI if only garden/metadata changes.
- Tightened PR description guidance: collapsed per-step sections only for steps that changed in a non-obvious way. Skip pure `etl update` boilerplate.


edomt:
@paarriagadap This is the PR I was telling you about today! Feel free to edit/merge.

paarriagadap (approved):
Thanks Ed! I think they make sense.

Some people might prefer not to set `--include-usages` at least at the initial update, but I think it's fine. Note that the skills asks to archive the dataset after all steps are done, so the run would fail if not all usages are updated.

PR #6225, opened 2026-06-08
🐛 Require pr-clean to run from the main repo

pabloarosado:

## What

Make `etl pr-clean` refuse to run from a secondary worktree. It now errors out unless invoked from the main repo, pointing you to `cd` there.

## Why

While working inside a worktree, I switched it to `master` and ran `etl pr-clean`. It deleted the branch but left the worktree behind — and a second run couldn't see the orphan either.

Root cause: `list_worktrees()` maps **currently-checked-out branch → path**. Once the worktree was switched off its branch, that branch had no worktree in the map, so `worktrees.get(branch)` returned `None`. In `clean_branch()` the worktree guards and the session-copy + `worktree remove` block are all gated on `worktree_path is not None`, so they were skipped and the code went straight to `git branch -D` — deleting the branch and orphaning the worktree. The orphan (now on `master`) is invisible to pr-clean because pr-clean only iterates over branches.

## Fix

A blanket "must run from the main repo" guard is simpler and more robust than trying to repair the branch↔worktree linkage:

- It can't pull the rug out from under you regardless of what branch any worktree is currently on.
- No path-matching or broken-link edge cases.
- The only thing it blocks is cleaning from inside worktree A instead of main — and the workaround is a single `cd` to main, which can clean *every* worktree anyway.

Changes:
- Add a global guard at the top of `clean_cli()` that bails when the current working tree isn't the main repo.
- Collapse the now-redundant per-branch current-worktree guard in `clean_branch()` (with the global guard, current always equals main).
- Note the requirement in the `pr-clean` help text.

## Test plan

- `make check` passes (lint, format, types).
- Manual: running `etl pr-clean` from a worktree now raises a clear `ClickException`; from the main repo it behaves as before.

### DIFF
--- a/apps/pr/cli.py
+++ b/apps/pr/cli.py
@@ -81,7 +81,7 @@ from typing import cast
 import click
 import questionary
 import requests
-from git import GitCommandError, Repo
+from git import GitCommandError, InvalidGitRepositoryError, Repo
 from rich_click.rich_command import RichCommand
 from structlog import get_logger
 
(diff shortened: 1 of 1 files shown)


pabloarosado:
@codex review

pabloarosado:
@codex review

PR #6223, opened 2026-06-08
📊🤖 Update NVIDIA quarterly revenue (2026-06-08)

edomt:
> _Written by Claude Code — @edomt at the wheel._

## Summary

- Bumped `artificial_intelligence/nvidia_revenue` from `2026-04-02` to `2026-06-08`, picking up NVIDIA's Q1 FY27 release (May 20, 2026; period ending Apr 26, 2026; revenue $81.6B).
- **Adopted NVIDIA's new market-segment presentation** introduced in Q1 FY27: the four previous non-data-center segments ("Gaming", "Professional Visualization", "Auto", "OEM & Other") were rolled into a single combined bucket (NVIDIA's term: "Edge Computing"; our label: **"Gaming, devices, automotive"**).
- **Fixed the fiscal-quarter → calendar-date logic**: each data point is now dated to NVIDIA's reported fiscal-quarter end (e.g., Q4 FY26 = `2026-01-25`), replacing the previous mapping that placed each point at the start of a roughly-aligned calendar quarter (e.g., Q4 FY26 was dated `2025-10-01`).

## Data-shape diff (verified)

| | OLD (2026-04-02) | NEW (2026-06-08) |
|---|---|---|
| quarters | 48 | 49 |
| indicators | 4 (Data centers and AI, Gaming, Other, Total) | 3 (Data centers and AI, Gaming/devices/automotive, Total) |
| date range | 2014-01-01 → 2025-10-01 | 2014-04-27 → 2026-04-26 |

Numeric reconciliation across the 8 overlapping quarters (Q1 FY25 .. Q4 FY26): NVIDIA's new "Edge Computing" equals the sum of the four older non-data-center categories to the dollar; data center total unchanged; overall total unchanged.

## DAG

Old `2026-04-02` entries moved to `dag/archive/artificial_intelligence.yml`; new `2026-06-08` entries placed in the original slot under the `# NVIDIA quarterly revenue dataset` section header.

PR #6242, opened 2026-06-11
🐛 Fix row-shift bug in NVIDIA revenue snapshot

edomt:

## Summary

Follow-up bugfix on [#6223](https://github.com/owid/etl/pull/6223). Esteban (with Claude) spotted that the four FY2020 quarters (Apr 2019 → Jan 2020) had corrupted segment values in the just-merged dataset.

The "Data centers and AI" line was understated by ~2× for those four bars, but the values were small enough relative to the recent post-2022 boom that the visual anomaly didn't jump out either.

## What was wrong

The Q4 FY21 PDF prints "Professional Visualization" with the row label wrapping across two lines (the row's *values* sit between the words). My fallback for unlabelled rows did a substring check (`"Professional Visualization" in text`), which fails on wrapped labels. So Professional Visualization was silently dropped from the canonical-segment assignment list, and every row beneath it shifted up by one label.

Concretely, for Q1 FY20:

| Row in PDF | Real value | Label the buggy extractor gave it |
|---|---|---|
| Gaming | $1,055M | Gaming ✓ |
| Professional Visualization | $266M | Data Center ❌ |
| Data Center | $634M | Auto ❌ |
| Auto | $166M | OEM & Other ❌ |
| OEM & Other | $99M | **TOTAL** ❌ |
| TOTAL | **$2,220M** | (unlabelled → dropped) ❌ |

Why FY20 only: the Q4 FY21 PDF is the most-recent source for FY20 in the dedup step. Older PDFs win for older quarters; newer PDFs win for newer quarters.

## Fix

1. Match each word of the segment label separately, so a wrapped label is recognised when "Professional" *and* "Visualization" both appear somewhere in the page text.
2. Add two snapshot-time integrity checks that would have caught this:
   1. Sum of top-level segments must equal the published TOTAL for every `(source_pdf, quarter)`.
   2. Hyperscale + ACIE must equal Data Center in the new (Q1 FY27+) presentation.

Both invariants pass on the corrected snapshot.

## Verified

- All 13 source PDFs spot-checked one random `(quarter, segment)` cell each against the rendered PDF in Firefox; all match.
- FY20 quarters now report correct values (e.g. Q1 FY20 Total = $2,220M, Data Center = $634M).

### DIFF
--- a/snapshots/artificial_intelligence/2026-06-08/nvidia_revenue.csv.dvc
+++ b/snapshots/artificial_intelligence/2026-06-08/nvidia_revenue.csv.dvc
@@ -16,6 +16,6 @@ meta:
       name: ©2026 NVIDIA Corporation
       url: https://www.nvidia.com/en-us/about-nvidia/terms-of-service/
 outs:
-  - md5: ef7de151621248d0c631605d277a82fa
-    size: 20239
+  - md5: c109a9519747ae14568973299cd04962
+    size: 20631
     path: nvidia_revenue.csv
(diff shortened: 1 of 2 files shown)


edomt:
@codex review

PR #6251, opened 2026-06-11
📊 wildfires: add historical data to the main script

veronikasamborska1994:
hey @Marigold ! how feasible would it be to run all of the API requests starting from 2003 in the automated update? the data seems to change retrospectively for earlier years too and I am just wondering what the best solution would be for this.

### DIFF
--- a/dag/climate.yml
+++ b/dag/climate.yml
@@ -32,7 +32,6 @@ steps:
   #
   data://garden/climate/latest/weekly_wildfires:
     - data://meadow/climate/latest/weekly_wildfires:
-      - snapshot://climate/2024-02-22/weekly_wildfires_2003_2023.csv
       - snapshot://climate/latest/weekly_wildfires.csv
     - data://garden/regions/2023-01-01/regions
     - data://garden/faostat/2026-02-25/faostat_rl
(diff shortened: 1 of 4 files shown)


Marigold:
Hey @veronikasamborska1994! I benchmarked this to see how bad it would be — and it turns out it's very doable. The EFFIS API is happy to serve ~12k requests (252 countries × 24 years × both endpoints) in a few
  minutes when queried in parallel, with zero rate-limiting.

  The only bottleneck was the script itself: it fetched sequentially with a 0.5 s sleep between requests, which would have meant ~3 hours for the full 2003+ fetch — too slow for the daily autoupdate, which runs
  all snapshots one after another. So I pushed a commit to this branch (e0f2f6f4) that parallelizes the fetch with 8 worker threads. The full fetch now takes **~3 minutes**, which is actually faster than the old
  2024-only sequential version.

  To validate it, I ran the parallelized script end-to-end: it produced a byte-identical file to your run (same md5, since the fetch order is preserved), so the `.dvc` didn't even change.

  So: 👍 to fetching everything from 2003 in the automated update, no separate historical snapshot needed. Old years that get retroactively revised by GWIS will now just flow in with each daily update.

Marigold (approved):

PR #6262, opened 2026-06-12
📊 Add Maddison, WID and ILO regions to regions dataset

paarriagadap:
> _Written by Claude Code — @paarriagadap at the wheel._

## Summary

Adds the regional groupings of three external providers — the [Maddison Project Database](https://www.rug.nl/ggdc/historicaldevelopment/maddison/), the [World Inequality Database](https://wid.world/), and [ILOSTAT](https://ilostat.ilo.org/) — to OWID's regions system, exposes them as map indicators on the [region-definitions page](https://ourworldindata.org/world-region-map-definitions), and captures the whole workflow as a reusable skill.

Membership for every region was taken from the **source** (each provider's own dataset / classification) and verified by set-equality — not hand-typed.

### Maddison Project Database — 8 regions (`defined_by: maddison`)
`MADDISON_EA/EE/LA/MENA/SSEA/SSA/WE/WO`. Membership extracted from the `region` column of `garden/ggdc/2024-04-26/maddison_project_database` (169 countries; includes historical entities USSR, Czechoslovakia, Yugoslavia, Sudan (former) where the source assigns them).

### World Inequality Database — 9 regions (`defined_by: wid`)
`WID_EA/EUR/LA/MENA/NA/OC/RCA/SSEA/SSA`. Membership from WID's official country table; the combined "North America & Oceania" is split into North America and Oceania (216 countries).

### ILOSTAT — 15 regions across two tiers
Mirrors ILOSTAT's own classification (its `ilo_region` / `ilo_subregion_broad` columns), kept at the two coarser tiers:
- **`ilo_1`** (5 broad): Africa, Americas, Arab States, Asia and the Pacific, Europe and Central Asia.
- **`ilo_2`** (11 intermediate): Northern Africa, Sub-Saharan Africa, LAC, Northern America, Arab States, Eastern Asia, South-Eastern Asia and the Pacific, Southern Asia, Northern/Southern/Western Europe, Eastern Europe, Central and Western Asia.

The 14 most-detailed subregions are **not** kept — both retained tiers independently partition the same 233 countries. Arab States belongs to both tiers, so it's tagged `ilo_1` and back-filled into the `ilo_2` indicator in the grapher step (mirroring the existing `un_region` pattern). Channel Islands handling is documented with a `# NOTE` in the YAML.

## ILOSTAT dataset (`garden/un/2026-02-03/ilostat`)

[...]

PR #6263, opened 2026-06-12
📊 Add World Inequality Database regions to regions dataset

paarriagadap:


paarriagadap:
Closing — WID regions will go into #6262 instead.

PR #6276, opened 2026-06-15
📊 Add Ebola outbreak datasets (2026 tracker + historical)

Marigold:

> **Status (not merging for now).** After review with @edomt, we're not pursuing Ebola charts at this point: the data isn't well-suited to charting, modeled GBD estimates are unreliable at the scale of small regional outbreaks, and covering an active outbreak is a high-commitment, one-way decision (and not reproducible in Grapher — no DRC health-zone shapefile). This PR is parked as a working reference: the historical CDC table is *reported* (non-modeled) per-outbreak counts that could serve as a factual backbone **if** we ever write about Ebola. Likely to be closed.

Adds two new datasets under a new `ebola` namespace, both built end-to-end (snapshot → meadow → garden → grapher) and wired into `dag/health.yml`.

## ⚠️ Licensing — read before publishing

Neither dataset is cleared for republication yet:

- **2026 tracker** is built on the DRC **Institut National de Santé Publique (INSP)** SitRep series, compiled by the Kraemer Lab (Oxford). The upstream metadata flags: *"reuse with attribution to INSP and citation of the specific report number and date; confirm distribution terms with INSP before external republication."* The `.dvc` records this.
- **Historical** data is from the **CDC** chronology (US Government work, public domain) — lower risk, but worth a confirm.

**We should reach out to INSP / the Kraemer Lab directly before making any of the 2026 charts public.** Until then this is a build only — no charts created. (Snapshots are uploaded to R2 so the pipeline builds on staging; that's internal storage, not reader-facing republication.)

## A. Live 2026 outbreak tracker — `ebola/2026-06-15/ebola_drc_2026`

The 2026 Bundibugyo Ebola outbreak (DRC + Uganda), declared a PHEIC by WHO on 17 May 2026.

- **Source:** [`INRB-UMIE/Ebola_DRC_2026`](https://github.com/INRB-UMIE/Ebola_DRC_2026) (Kraemer Lab), which transcribes the INSP SitRep PDFs into tidy per-metric CSVs.
- **Snapshot** stacks the 11 core case/death CSVs into one long table, keeping the source `ND`/`NA` sentinels verbatim. Mobility/conflict/demographic layers in the upstream repo are deliberately excluded (restrictive third-party licenses; not needed for a cases/deaths tracker).
- **Garden** produces two tables: **national** (cumulative confirmed/suspected cases & deaths) and **by health zone** (38 zones; cumulative + newly reported). Health-zone names are kept as subnational entities (not country-harmonized).
- **Cross-check:** the 38 health zones sum to 782 confirmed cases on 13 June — exactly the national figure (and 676 on 10 June matches WHO/press reports).

## B. Historical outbreak chronology — `ebola/2026-06-15/ebola_outbreaks`

Every known Ebola outbreak since 1976.

- **Source:** [CDC Ebola outbreak history](https://www.cdc.gov/ebola/outbreaks/index.html). The snapshot archives the raw HTML; the **meadow** step parses the year/country accordion structure into one row per outbreak, with integrity checks that fail loudly if the page is restructured.
- **Garden** harmonizes 21 country labels (the 2014–2016 West African epidemic and the 2026 DRC+Uganda outbreak are kept as combined entities), aggregates the 5 country-years that had two distinct outbreaks, and recomputes the case fatality rate from the totals.
- **Result:** 49 country-years, 1976–2026, totalling 34,982 cases / 15,443 deaths — consistent with the published ~34,936 / 15,409 for 1976–2022 plus recent years.

## Notes for reviewers

- **Snapshots are uploaded to R2** (public-domain CDC HTML + the INSP-derived CSV), so the pipeline builds on staging. This is internal data storage; reader-facing publication (charts) stays gated on the licensing question above.
- **Provincial level** is omitted from the tracker for v1 (the source's provincial linelist uses a different date format and an onset basis); national + health-zone only.
- The tracker covers an **active outbreak** — the upstream schema may shift as it evolves, and it needs a refresh cadence once we decide to maintain it.

### DIFF
--- a/dag/health.yml
+++ b/dag/health.yml
@@ -4,6 +4,22 @@ steps:
   data://garden/owid/latest/covid:
     - etag://raw.githubusercontent.com/owid/covid-19-data/master/public/data/owid-covid-data.csv
 
+  # 2026 Bundibugyo Ebola outbreak (DRC + Uganda), via Kraemer Lab / INSP SitReps
+  data://meadow/ebola/2026-06-15/ebola_drc_2026:
+    - snapshot://ebola/2026-06-15/ebola_drc_2026.csv
+  data://garden/ebola/2026-06-15/ebola_drc_2026:
+    - data://meadow/ebola/2026-06-15/ebola_drc_2026
+  data://grapher/ebola/2026-06-15/ebola_drc_2026:
+    - data://garden/ebola/2026-06-15/ebola_drc_2026
+
+  # Historical Ebola outbreak chronology (CDC), 1976-present
+  data://meadow/ebola/2026-06-15/ebola_outbreaks:
+    - snapshot://ebola/2026-06-15/ebola_outbreaks.html
+  data://garden/ebola/2026-06-15/ebola_outbreaks:
+    - data://meadow/ebola/2026-06-15/ebola_outbreaks
+  data://grapher/ebola/2026-06-15/ebola_outbreaks:
+    - data://garden/ebola/2026-06-15/ebola_outbreaks
+
   # Death completion (Karlinsky, 2024)
   data://grapher/health/2025-04-18/deaths_karlinsky:
     - data://garden/health/2025-04-18/deaths_karlinsky:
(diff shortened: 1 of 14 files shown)


Marigold:
Parking this for now.

PR #6277, opened 2026-06-15
📊 emissions: New approach to estimate emissions by human activity

pabloarosado:

Reworks "emissions by custom sector" into a clearer set of steps that group greenhouse-gas emissions by the **human activity** they serve, rather than by source sector. The old `emissions_by_custom_sector` steps (2025-09-24, 2026-02-11, 2026-03-04) are retired in favor of a single `2026-06-15/emissions_by_human_activity` family.

## What it does

Climate Watch (WRI) subsectors are folded into five everyday activities:

- **Growing food** — agriculture, land use, other fuel combustion
- **Getting around** — transport
- **Keeping warm and cool** — buildings
- **Electricity** — electricity and heat
- **Making things** — manufacturing & construction, industry, fugitive emissions, waste

Each step also produces an `_including_other` table variant that breaks the miscellaneous emissions (other fuel combustion, waste, fugitive) out into their own "Other" activity instead of folding them in. Tables are country-level and report both CO₂ and total GHG.

## The four grapher outputs

1. **`emissions_by_human_activity`** — electricity kept as its own activity (Climate Watch only).
2. **`..._splitting_electricity`** — electricity emissions redistributed into the activities that consume the electricity, using UNdata's Energy Statistics Database (final electricity consumption by sector). Reports direct, indirect, and total emissions per activity.
3. **`..._including_gross_lulucf`** — same as (1) but "Growing food" uses *gross* agriculture-and-land-use emissions from Jones et al. (national contributions), converted with IPCC AR4 GWPs. Climate Watch's land-use emissions are net (≈0 globally), which understates land's climate impact; the gross figure corrects that.
4. **`..._including_gross_lulucf_splitting_electricity`** — combines (2) and (3).


pabloarosado:
@codex review

pabloarosado:

Addressed the Copilot review (commit 0ad53a690):

- **Conservation check** (`splitting_electricity` comment): the stale comment is now a real assertion. Per country-year, the indirect emissions must sum back to the full electricity-generation total (shares sum to 1, checked within 0.1%). Added to both splitting-electricity steps.
- **Archive DAG**: regenerated `dag/archive/emissions.yml` with `etl archive-dag`, so the retired `emissions_by_custom_sector` entries now carry the standard `# archived; last active in <sha>` recovery markers.

Dismissed the 7 `pr.merge`-vs-`DataFrame.merge` comments: in these steps the left operand is always a `Table`, so `tb.merge(...)` dispatches to `owid.catalog`'s metadata-preserving `Table.merge` override (verified it's literally the same function `pr.merge` points to). No origins are dropped, so there's no provenance bug to fix.

PR #6289, opened 2026-06-16
Verify Tailscale after waking a staging server

Marigold:

## Problem

A colleague pressed **Wake up** on a staging server ([#6054](https://github.com/owid/etl/pull/6054)) and got a success toast, but `http://staging-site-…/admin` stayed unreachable.

Diagnosis: `owid-lxc start` (what Wake runs) does two things — start the container **and** re-authenticate it to Tailscale. The container started fine, but the **Tailscale login failed silently** (it's only a warning inside `owid-lxc`, which still exits 0). The result: the container is `Running` but `Logged out` of Tailscale, so its hostname doesn't resolve via MagicDNS and admin/SSH are unreachable — while the dashboard happily reported success.

## Fix

After `owid-lxc start`, confirm the container actually rejoined the tailnet by checking it has a `100.x` Tailscale IP (`owid-lxc exec … tailscale ip -4`, which works even when Tailscale is down). If it didn't, return a clear, actionable error instead of a misleading success:

> Server … started, but it is NOT on the Tailscale network, so it's unreachable by hostname (admin/SSH won't work). Tailscale failed to re-authenticate on wake. Try waking it again; …

`restart_server` inherits the check (it calls `start_server`).

## Testing

Exercised against live containers:
- Waking a real idle container → `ok=True`, *"is awake and on the tailnet"* (only after Tailscale genuinely reconnected).
- The check returns `False` for a container that's stopped / not on the tailnet, so the failure path reports honestly.

## Follow-up (not in this PR)

This makes the failure **visible**; it doesn't change *why* Tailscale sometimes fails to re-auth on wake (likely the host running `owid-lxc` can't read the Tailscale auth key from the vault). Worth confirming the prod-wizard host has vault access so wakes reliably rejoin the tailnet. Could also make `owid-lxc start` itself exit non-zero on Tailscale-login failure, but that's left alone to avoid affecting the deploy pipeline that tolerates Tailscale warnings.

### DIFF
--- a/apps/wizard/app_pages/servers_dashboard/app.py
+++ b/apps/wizard/app_pages/servers_dashboard/app.py
@@ -128,6 +128,16 @@ with st.container(horizontal=True, vertical_alignment="bottom", horizontal_align
         **Branch Names**: The 'staging-site-' prefix is automatically removed for cleaner display.
         """)
 
+# Show the outcome of the last server action (carried across the post-action rerun, so it's
+# displayed alongside the freshly-refreshed server list).
+_action_result = st.session_state.pop("servers_action_result", None)
+if _action_result is not None:
+    _ok, _label, _msg = _action_result
+    if _ok:
+        st.success(f"✅ {_label}: {_msg}")
+    else:
+        st.error(f"❌ {_label} failed: {_msg}")
+
 # Fetch server data and host memory stats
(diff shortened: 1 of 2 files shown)


Marigold:
@codex review

PR #6291, opened 2026-06-16
🐛 Fix cloud SessionStart hook so .venv builds reliably

Marigold:

## Problem
In Claude Code on the web (cloud sessions), sessions started **without a working `.venv`** — Claude had to build it inline on every session. Two bugs in `scripts/remote_setup.sh` (our `SessionStart` hook):

1. **It never `cd`s to the repo root.** SessionStart hooks don't guarantee the working directory is the checkout, so `make .venv` ran outside the repo and failed with `make: *** No rule to make target '.venv'`.
2. **`set -e` + a non-critical step abort the run.** The hook installs the `gh` CLI (a github.com download) *before* building the venv; under `set -e`, any failure there aborts the script before `uv`/`make` ever runs.

## Fix
- `cd "$CLAUDE_PROJECT_DIR"` and assert `pyproject.toml` is present before doing anything.
- Install deps with **`uv sync --all-extras --group dev`** directly — exactly what `make .venv` runs underneath, minus the `include` / `install-hooks` / sanity-check layers that add fragility in an ephemeral sandbox.
- Make the `gh` install **best-effort and non-fatal** (cloud sessions already ship built-in GitHub tools that authenticate as the user).
- Drop `set -e`; **always `exit 0`** so a SessionStart hook hiccup never blocks the session.
- Add the `startup|resume` matcher to the hook in `.claude/settings.json`.

## Notes
- Dependency install belongs in a `SessionStart` hook, **not** the cloud environment "setup script" — the setup script runs outside the repo, so `uv sync` can't find `pyproject.toml` there. This is the documented split (setup script = repo-independent system tools; SessionStart hook = project deps).
- Local runs are unaffected: the hook still early-exits when `CLAUDE_CODE_REMOTE != true`.

## Testing
- `bash -n` clean; local gate verified (skips when not remote); `.claude/settings.json` validated as JSON.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -8,6 +8,7 @@
   "hooks": {
     "SessionStart": [
       {
+        "matcher": "startup|resume",
         "hooks": [
           {
             "type": "command",
--- a/scripts/remote_setup.sh
+++ b/scripts/remote_setup.sh
@@ -25,8 +25,14 @@ if [ "$CLAUDE_CODE_REMOTE" != "true" ]; then
 fi
 
 # Always operate from the repo root (where pyproject.toml / uv.lock live).
+# Prefer $CLAUDE_PROJECT_DIR, but it can be empty depending on how the hook is
+# invoked — fall back to deriving the root from this script's own location
+# (scripts/remote_setup.sh → repo root is one level up).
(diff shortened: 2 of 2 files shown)


Marigold:
@codex review

PR #6305, opened 2026-06-17
📊 Fix Dependabot security vulnerabilities

Marigold:
> _Written by Claude Code — @Marigold at the wheel._

Resolves open Dependabot security alerts by upgrading vulnerable dependencies across the monorepo (root + `lib/` subpackages + VS Code extensions).

## pip (uv.lock)

| Package | From → To | Severity fixed | Manifests |
|---|---|---|---|
| `bleach` | 6.3.0 → 6.4.0 | medium, low | root |
| `starlette` | 1.2.1 → 1.3.1 | high, low | root |
| `tornado` | 6.5.5 → 6.5.7 | high ×2, medium | root, catalog, datautils, owl |
| `cryptography` | 48.0.0 → 48.0.1 (root) / 49.0.0 (libs) | high | root, catalog, datautils, owl |
| `python-multipart` | 0.0.28 → 0.0.32 | high, low ×3 | root |
| `pyjwt` | 2.12.1 → 2.13.0 | high, medium ×3 | root |

`lib/datautils` pins `cryptography` and `tornado` directly — bumped the version floors to the patched releases and bumped the package `version` to `0.6.4` (the publish workflow rejects republishing an existing version). Re-locked root + catalog + owl to pick up `owid-datautils==0.6.4`.

## npm (js-yaml → 4.2.0)

- **meta-yaml-hover**: real direct dependency (`^4.1.0 → ^4.2.0`).
- **find-latest-etl-step**, **dod-syntax**: js-yaml is only a dev-tooling transitive dep (eslint/mocha) — forced to `^4.2.0` via `overrides` rather than adding a phantom runtime dependency.

## Not fixed

- **#559** (bleach `linkify(parse_email=True)` CPU exhaustion, medium): no patched version published yet — nothing to upgrade to.

`make check-all` passes (root + all `lib/` packages).

PR #6307, opened 2026-06-17
📊 Clear remaining Dependabot version-update PRs

Marigold:

Follow-up to #6305. That PR cleared every Dependabot **security alert**; this one clears the remaining routine **version-update PRs** (no associated alert) so the Dependabot PR list ends up empty.

## esbuild 0.27.x → 0.28.1 (devDependency)

Bumped in 8 VS Code extensions: `chart-preview`, `clickable-dag-steps`, `compare-previous-version`, `detect-outdated-practices`, `dod-syntax`, `find-latest-etl-step`, `meta-yaml-hover`, `run-until-cursor`. esbuild is build-only (devDependency); all 8 extensions compile on 0.28.1.

## js-yaml → 4.2.0 (in the 2 extensions #6305 didn't cover)

- **clickable-dag-steps**: js-yaml is imported in `src/extension.ts` but was only declared as `@types/js-yaml` (the runtime package resolved transitively). Added `js-yaml@^4.2.0` as a proper `dependencies` entry.
- **compare-previous-version**: js-yaml isn't imported (dev-tooling transitive only) — forced to `^4.2.0` via `overrides`, matching the approach used in #6305.

## Skill update

Extended the `fix-dependabot` skill (`.claude/skills/fix-dependabot/SKILL.md`) so future runs also sweep non-alert version-update PRs (new Step 4c) and end with an empty Dependabot PR list — plus guidance on direct-vs-transitive npm deps (when to add to `dependencies` vs. `overrides`).

## Supersedes
Closes the 10 open Dependabot version-update PRs: #6306, #6281, #6273, #6272, #6271, #6270, #6269, #6268, #6265, #6264.

### DIFF
--- a/.claude/skills/fix-dependabot/SKILL.md
+++ b/.claude/skills/fix-dependabot/SKILL.md
@@ -9,6 +9,8 @@ metadata:
 
 Resolve open Dependabot security alerts by upgrading vulnerable dependencies across the ETL monorepo — including pip packages (`pyproject.toml` + `uv.lock`), lib/ subdirectories, and npm packages in `vscode_extensions/`.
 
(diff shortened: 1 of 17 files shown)


Marigold:
@codex review

PR #6318, opened 2026-06-19
📊 Add .corrections.yml mechanism for known upstream data errors

Marigold:

A declarative **`.corrections.yml`** mechanism for recording known **upstream data errors** and applying the local fix, replacing scattered inline `.loc[...]` / `.drop(...)` exceptions with one auditable record per error (carrying its reason, provider, and reporting status) — plus a CLI that inventories and visualises every correction in the repo.

## 🔎 Visual review

**[Open the corrections dashboard →](https://htmlpreview.github.io/?https://gist.githubusercontent.com/Marigold/3e46907c92259d6d03caf2f4a98723c2/raw/a79cd61f093683abe73c741b5dc967575e6b5d71/corrections.html)** — every correction in the repo, grouped by provider/status. Expand a row for a before→after table and a time-series chart showing the *problematic* values (red) and the fix (green). Generated by `etl corrections -o out.html --charts`.

## What's here

- **`etl/data_corrections.py`** — loader (`load_corrections`) + `apply_corrections(tb, corrections)`, plus an audit (`build_audit` / `write_audit` / `read_audit`) that records each correction's affected points for visualisation.
- **`PathFinder.corrections_path`** + **`PathFinder.apply_corrections(tb)`** in `etl/helpers.py`, mirroring the `.excluded_countries.json` ergonomics — one line in a step. It also writes the audit on each run.
- **`etl corrections`** CLI (`apps/corrections/cli.py`) — scans all `.corrections.yml`, validates them (repo-wide lint), prints a terminal summary, and renders a self-contained HTML dashboard.
- **Six real cases migrated** from inline code, each verified **value-identical** end-to-end.
- **JSON Schema** (`schemas/corrections-schema.json`, wired in `.vscode/settings.json`) so `.corrections.yml` files are linted in the editor like `.meta.yml`; a test enforces every repo corrections file against it.
- Unit tests in `tests/test_data_corrections.py`.

## The format

A `<short_name>.corrections.yml` next to the step, a list of entries:

```yaml
- indicator: consumption_emissions          # the COLUMN to correct
  entity: Panama                            # country/entity
  years: [2006, 2008, 2016]                 # list | all | latest | {after/before/from/to: Y}
  action: drop                              # drop | override | scale | flag
  reason: Negative consumption-based CO2 emissions (physically impossible) in the source data.
  provider: Global Carbon Project
  reported: 2024-11-13                       # optional — when we told the provider
  status: reported                           # open | reported | acknowledged | fixed_upstream
```

- Rows are located by `entity` + `years`, or by their current value: `match: {value: 4.5}`.
- **Actions**: **`drop`** removes the matched rows (long/tidy tables); **`override`** replaces the indicator value with `value` (use `value: null` to blank one bad cell in a wide table without dropping the row); **`scale`** multiplies by `factor` (e.g. a unit mismatch — `factor: 0.001`); **`flag`** logs the known error without changing data.
- `years` grammar: a literal list, `all` (every year for the entity), `latest` (the table's most recent year), or `{after/before/from/to}` ranges.
- **`expect:` guard** (optional): a mapping of `eq`/`ne`/`gt`/`ge`/`lt`/`le` → threshold, asserted over all matched rows *before* the fix. `drop` is self-validating (the row vanishing trips the no-match assert), but `override`/`scale` keep the row — `expect` is how they **fail loudly when the source fixes the value**. Cannot combine with `flag`.
- **No `id` field** — a stable label is derived from `provider / indicator / target` for logs and error messages.

## Migrated cases

| Step | Correction | Action |
|---|---|---|
| `gcp/2025-11-13/global_carbon_budget` | Panama negative consumption emissions (2006/2008/2016) | `drop`, year list |
| `ifr/2026-01-14/robot_density` | China 2023 value revised by the source | `override` |
| `usgs/2025-12-15/mineral_commodity_summaries` | Australia zinc reserves 2022 + Kazakhstan asbestos 2020 | `override: value: null` (blank cell) |
| `wb/2025-07-01/income_groups` | Ethiopia's unclassified latest-year row | `drop`, `years: latest` |
| `literacy/2026-05-12/historic_literacy_omm` | Cote d'Ivoire 2019 likely-error literacy rate | `drop`, year list |
| `tourism/2026-01-21/unwto` | Taiwan domestic hotel guests reported in units, not thousands | `scale: factor: 0.001` + `expect: {gt: 1000000}` |

These exercise every action (`drop`, `override`, value-blanking, `scale`) and the `entity`+list / `entity`+`latest` / `entity`+`all` locators, plus the `expect` guard.

## Visualisation & audit

`PathFinder.apply_corrections` records, at apply time, each affected entity's full pre-correction series and the exact before/after of every affected point, under `data/corrections_audit/` (gitignored). `etl corrections --charts` reads it to render — per correction — a before→after table plus a **Plotly** time-series that highlights the problematic values in red and the fix in green (with a Linear/Log toggle). Charts are exact and immune to mid-pipeline column renames (the audit is captured where the correction runs, not re-read from the published dataset).

> Note: the audit reflects the value **at the correction's apply point**. For corrections that run before a downstream transform (e.g. UNWTO, applied before the garden ×1000), the "After" is the intermediate value, not the final published number.

## Verification

For each migrated step I built the garden output with the old inline code and with the new corrections file and asserted `assert_frame_equal(check_exact=True)` **plus** per-column metadata equality. All are value-identical — a behaviour-preserving refactor, not a data change. `make check` (ruff + ty) and the unit tests pass.

## Notes for review

- **`override` changes a published number** (IFR China 2023). The corrected value and the source's note are in the correction's `reason`/`status` — worth a glance.
- **The old "lost guard" concern is now addressed** by `expect:` — `override`/`scale` corrections can assert the rows are still anomalous before acting (e.g. UNWTO's `expect: {gt: 1000000}`), failing loudly if the source fixes the issue. Richer cross-row guards (e.g. animal-welfare's `barn == free_range`) remain inline for now.
- **WB `EXPECTED_MISSING_COUNTRIES_IN_LATEST_RELEASE`** keeps its `Ethiopia` entry on purpose — dropping Ethiopia's latest row is *why* it shows as missing, and a downstream assertion checks that. I only updated the comments to point at the corrections file.

## Surveyed but intentionally NOT migrated

I swept `garden/`, `meadow/` and `snapshots/` for inline point-corrections (multiple angles). Good candidates remain for follow-ups — deferred here with reasons, so the next pass is just picking from this list:

**Mechanism supports them, deferred for verification care:**
- **`gfs/.../gfs_wave_two` drinks 4.5→5** and **WHO typo fixes** (`avian_influenza` "225"→"2025", `vaccine_stock_out` "Inaccurtae"→"Inaccurate") — these use `.astype().replace(...)` rather than a `.loc` assignment, so a value-identical migration needs care, not a copy.
- **`animal_welfare/.../laying_hens_keeping_eu` Belgium 2025 + Greece 2013** — clean drops, but the inline code carries *cross-row* value-guards (`barn == free_range`, `total < 100k`) that the per-row `expect:` clause can't express yet.
- **`gapminder/.../maternal_mortality` year typos** (meadow) and **`health/.../polio_free_countries` Somalia 2000** — in-scope point corrections, lower priority.

**Out of scope (not provider point-errors):**
- **`democracy/.../lexical_index` `regime_lied==7 → regime_redux_lied=6`** — looks like a point fix but rewrites **4,234 rows**: a systematic scale-recoding *rule*, not a point correction. Left inline.
- **`emissions/.../gdp_and_co2_decoupling` Honduras 2008/2013** — reads more like an analytical decision (removing distorting peaks).
- **GCP Kuwait-1991 Oil Fires merge**, **`health/.../wgm_2018` answer-code merge** — aggregation / cross-wave harmonization.
- **`faostat/.../detected_anomalies`** — already a dedicated anomaly-class framework with value-based `check()`, multi-key (item/element) locators, and a replace-from-another-source fix that the current mechanism can't express; consolidation is a separate effort.
- **NOAA missing-country fix** (categorical handling), **waste-management UAE** (×100 rescale via `scale` is now possible — candidate for a follow-up).
- **BGS USSR-after-1991, Energy Institute zero-fills, agriculture spurious zeros, WB-PIP snapshot validation, `ggdc` source-flagged outliers, OECD pre-1990 marriage rates** — geopolitical/temporal/harmonization/systematic *rules*, not enumerated provider errors.

> Note: the search also turned up corrections living in **archived (superseded) step versions** — I verified active-DAG membership before touching anything and skipped archive-only files.

## Explicitly NOT in this PR (follow-ups)

Per-provider report generator, auto-expiry on re-ingestion, Anomalist-seeded patch drafting. (The cross-dataset dashboard — originally a follow-up — is now included via `etl corrections`.)

### DIFF
--- a/apps/corrections/cli.py
+++ b/apps/corrections/cli.py
@@ -65,8 +65,13 @@ def collect_corrections(step_dir: Path) -> list[dict[str, Any]]:
 
 
 def _fmt(v: float | None) -> str:
-    """Compact human-readable number (e.g. -1.23, 6.4e+07)."""
-    return "—" if v is None else f"{v:.4g}"
+    """Human-readable number with thousands separators (e.g. 56,180,148 · -742,811 · 14.2)."""
+    if v is None:
+        return "—"
+    a = abs(v)
+    if a != 0 and (a < 0.01 or a >= 1e15):
+        return f"{v:.3g}"  # fall back to scientific only for extreme magnitudes
+    return f"{v:,.2f}".rstrip("0").rstrip(".") or "0"
 
 
 def _affected_table(affected: list, action: str) -> str:
--- a/apps/corrections/cli.py
+++ b/apps/corrections/cli.py
@@ -11,6 +11,7 @@ the cross-dataset view of known upstream data errors and what we've reported to 
 
 import datetime as dt
 import html
+import json
 import webbrowser
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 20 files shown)


Marigold:
@codex review

etl/data_corrections.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _row_mask(tb: Table, correction: dict[str, Any], country_col: str, year_col: str) -> np.ndarray:
+    """Return a positional boolean mask over `tb` selecting the rows a correction targets."""
+    index_names = [name for name in tb.index.names if name is not None]
+    label = _label(correction)
+
+    if "match" in correction:
+        mask = np.ones(len(tb), dtype=bool)
+        for key, value in correction["match"].items():
```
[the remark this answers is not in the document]
Marigold: > _Written by Claude Code — @Marigold at the wheel._

Good catch — fixed in d73ed339e3. `_validate_correction` now rejects an empty (or non-dict) `match`, so a `match: {}` entry fails loudly at load time instead of building an all-true mask and wiping the table (drop) or overwriting the whole indicator (override). The existing `n_matched > 0` guard couldn't catch this, since an empty match matches every row. Also rejected combining `match` with `entity`/`years` (those would have been silently ignored). Added two tests covering both.

Marigold:
Here's another example that could be migrated to corrections. It should also include the sanity check https://github.com/owid/etl/pull/6354.

Marigold:
@pabloarosado I've decided not to turn FAO data issues into corrections. It uses long format and is quite specific (and works well already), but I borrowed a couple of ideas from you.

etl/steps/data/garden/ifr/2026-01-14/robot_density.corrections.yml
```diff
@@ -0,0 +1,10 @@
+# Known upstream data errors in IFR robot density, corrected locally until fixed at the source.
+# See etl/data_corrections.py for the format. Values are in robots per 10,000 employees (post the /10 rescale).
+- indicator: robot_density
+  entity: China
+  years: [2023]
+  action: override
+  value: 14.2
```
lucasrodes, line 7: What do you think about adding a field like `expected` (maybe with some tolerance)? to confirm that the original value continues to be "wrong"?

etl/steps/data/garden/literacy/2026-05-12/historic_literacy_omm.py
```diff
@@ -66,8 +66,9 @@ def run() -> None:
     tb_unesco["source"] = "UNESCO Institute for Statistics"
     tb_unesco["source_url"] = "https://databrowser.uis.unesco.org/resources/bulk"
 
-    # Remove Cote d'Ivoire for 2019 due to likely error
-    tb_unesco = tb_unesco[~((tb_unesco["country"] == "Cote d'Ivoire") & (tb_unesco["year"] == 2019))]
+    # Drop known upstream data errors (e.g. Cote d'Ivoire 2019), declared in
+    # historic_literacy_omm.corrections.yml.
+    tb_unesco = paths.apply_corrections(tb_unesco)
```
lucasrodes, line 71: probably fine to keep this as this for now, it's flexible enough.

But as I saw it, I was wondering about detecting the corrections YML automatically, and applying it on `.format` maybe. But not sure if enforcing a specific place where we apply it would be too opinionated for now.

apps/corrections/cli.py
lucasrodes, line 1: Could we have this service running per-staging too? So we have a preview without launching the command. Also would be possible to share accross the team.

I thought of vibe.owid.io, but seems like maybe per-staging would be better?

lucasrodes (approved):
LGTM! Left minor suggestions

also: this is great!!! Step codes looks cleaner now 😌

apps/corrections/cli.py
Marigold: Good idea, but only if the corrections get some traction.

etl/steps/data/garden/ifr/2026-01-14/robot_density.corrections.yml
Marigold: Added!

etl/steps/data/garden/literacy/2026-05-12/historic_literacy_omm.py
Marigold: I did it, but in `create_dataset`. One can still call `apply_corrections` helper manually.

PR #6325, opened 2026-06-22
📜 add-provider-regions skill: suffix-slug rule + hover article links + Step 10

paarriagadap:

Docs-only improvements to the `/add-provider-regions` skill, from the ILO frontend work and Codex feedback on #6321. Three things:

## 1. Keep the bare suffix slug for tiered providers (addresses Codex on #6321)

`parseLabel` resolves an entity to a provider by its `(Provider)` name suffix (lowercased, spaces stripped), **not** by `defined_by`. For a provider whose suffix doesn't match its per-level keys (e.g. `(ILO)` → `ilo`, not `ilo_1`), the bare slug must stay registered in `ADDITIONAL_REGION_DATA_PROVIDERS` + labels, or entity grouping and the hover silently break. (Real bug — fixed in owid/owid-grapher#6629.)

## 2. Embed the article link in hover descriptions

The Step 9 `descriptions` instruction now shows the markdown-link format pointing at the `world-region-map-definitions` anchor (matching the WB/WHO/UN entries).

## 3. New Step 10 — add the provider to the world-region-map-definitions article

The editorial step of adding the provider's section to the article (a gdoc): the heading → anchor slug rule, and that multi-tier providers share one section.

Docs-only.


paarriagadap:
@codex review

PR #6326, opened 2026-06-22
📜 Regions-workflow skill updates: drift check (update-dataset) + hover colors (add-provider-regions)

paarriagadap:

Two related documentation updates to the region-workflow skills, bundled together.

## `update-dataset` skill — region-provider drift check
- Adds a check during the harmonization step to flag when a dataset's provider regions aren't yet in the regions dataset, so they get added where needed.
- Step 5d falls back to the [world-region-map-definitions](https://ourworldindata.org/world-region-map-definitions) article links to re-derive an *existing* provider's regions when the dataset itself doesn't encode them (updates only — a brand-new provider's regions won't be in the article yet).

## `add-provider-regions` skill — hover color matching
- Documents the optional step of pinning a provider's region **hover colors** to the exact colors its map chart renders — the default `CategoricalMapPalette10` in the indicator's `sort`/display order — by adding name-keyed entries to `MapContinentColors` in owid-grapher's `CustomSchemes.ts`.
- Notes the placement requirement: the entries go in `MapContinentColors`, **not** `ContinentColors` (which is declared before `OwidMapColors` and can't reference it → a "used before declaration" typecheck error).
- Explains when to **skip** pinning: a region shared across tiers (e.g. ILO's `Arab States (ILO)`) sits at a different palette position in each tier, so one name-keyed color can't match both charts without a clash — leave such a provider on the position-based fallback.
- Corrects the shared-region grey-hole back-fill guidance to the centralized `PROVIDER_REGION_BACKFILLS` in `getAggregatesByProvider` (covers the hover **and** the admin entity presets), rather than a `getRegionsForKey`-only special-case.

The corresponding owid-grapher implementation (pinning Maddison/WID hover colors to their charts) ships separately in [owid/owid-grapher#6629](https://github.com/owid/owid-grapher/pull/6629).


paarriagadap:
@codex review

PR #6331, opened 2026-06-23
📊 Retry transient Metabase 502s in read_metabase

Marigold:

## Problem

owidbot chart-diff runs (and any other caller of the analytics → Metabase path) intermittently hard-fail with:

```
RuntimeError: Metabase API request failed with status code 502: <html>
... nginx/1.24.0 (Ubuntu) ...
```

The 502 page is generated by **nginx**, not Metabase — the upstream Metabase JVM is briefly unreachable. This is expected and benign: the analytics pipeline restarts Metabase every time it rebuilds the DuckDB mirrors so it picks up the fresh files (`ops/.buildkite/analytics/shared_duckdb_steps.yml` → `make metabase.restart` + `pm2 restart metabase`). That happens **daily**, plus on every push to the `analytics` / `owid-grapher` main branches, so the restart window is hit fairly often. Metabase reads the semantic layer straight from on-disk DuckDB files, whose inode is rewritten on rebuild (documented at `analytics/Makefile`), which is why a restart is needed in the first place.

The bug on our side: `read_metabase` fired a single request with **no retry**, so a few-second restart window failed the entire run.

## Fix

Wrap the request in a `tenacity.Retrying` loop (same pattern already used in `etl/snapshot.py` / `etl/grapher/io.py`; `tenacity` is already a dependency):

- **5 attempts, exponential backoff 2→60s** — covers a normal JVM reboot window (~30s of total wait across the gaps).
- **Retried:** HTTP 502/503/504 (`RETRYABLE_STATUS_CODES`) plus `ConnectionError` / `Timeout`.
- **Not retried:** 4xx and CSV parse failures reraise immediately (`reraise=True` preserves the original exception type rather than wrapping in `RetryError`).
- Each retry logs a `metabase.request_retry` warning with the attempt number, so restarts stay visible rather than silently swallowed.

## Testing

Verified with mocked responses: transient 502s retry and recover, a 400 fails fast with no retry, and a persistent 502 exhausts all attempts then reraises `MetabaseTransientError`. `make check` passes.

### DIFF
--- a/etl/analytics/metabase.py
+++ b/etl/analytics/metabase.py
@@ -9,6 +9,8 @@ from io import BytesIO
 import pandas as pd
 import requests
 from metabase_api import Metabase_API
+from structlog import get_logger
+from tenacity import Retrying, retry_if_exception_type, stop_after_attempt, wait_exponential
 
 from etl.config import (
     METABASE_API_KEY,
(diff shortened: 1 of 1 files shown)


Marigold:
@codex review

PR #6340, opened 2026-06-25
✨ ETL cli tool PR - add assignee

antea04:
Added an option to automatically assign a new PR to yourself. 

Currently added it with a new flag `--assign`, so nothing breaks, but this could also be default behaviour. Tested it with this PR and it works :)

/schedule

### DIFF
--- a/apps/pr/cli.py
+++ b/apps/pr/cli.py
@@ -546,16 +546,25 @@ def create_pr(repo, work_branch, base_branch, pr_title, assign: bool = False):
         "body": "",
         "draft": True,
     }
+    assignee: str | None = None
     if assign:
         user_response = requests.get("https://api.github.com/user", headers=headers)
         if user_response.status_code == 200:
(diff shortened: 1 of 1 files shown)


apps/pr/cli.py
```diff
@@ -192,6 +192,11 @@
     is_flag=True,
     help="Symlink the new worktree's data/ to the original repo's data/ (only with --worktree). Avoids recomputing upstream ETL steps. Don't run heavy ETL ops in both worktrees concurrently, and never `rm -rf data/` in the worktree.",
 )
+@click.option(
+    "--assign",
```
lucasrodes: very brief, can we change this to `auto-assign`? I feel that reflects better what the flag does. Possibly change also other references from `assign` to `auto_assign`.

lucasrodes (approved):
LGTM! Thanks, Tuna

apps/pr/cli.py
antea04: sure np

PR #6345, opened 2026-06-26
📊 Repoint poverty and inequality steps to new PIP and WID datasets

paarriagadap:

## What this does

Repoints the poverty/inequality steps off the wide-flat **legacy** datasets onto the new **dimensional** ones:

- `data://garden/poverty_inequality/2025-01-22/poverty_inequality_file`
- `data://garden/poverty_inequality/2025-01-22/inequality_comparison`
- `data://garden/covid/latest/compact`

all now read `wb/2026-03-24/world_bank_pip` and `wid/2026-02-10/world_inequality_database` instead of `*_legacy`.

## Why

`world_bank_pip_legacy` and `world_inequality_database_legacy` are kept only to feed the CSV-based explorers we plan to deprecate. These three steps were the remaining non-explorer consumers blocking that. After this PR the legacy datasets are used **only** by the two CSV explorers (`explorers/wb/latest/world_bank_pip`, `explorers/wid/latest/world_inequality_database`), so they can be removed cleanly once the explorers are sunset.

## How

A new `shared.py` rebuilds, from the dimensional long-form tables, exactly the key-indicator columns the `keyvars` transforms need:

| Adapter | Source | Notes |
|---|---|---|
| `build_pip_unsmoothed` | `world_bank_pip/complete_series` | all 6 indicators; `gini`/`palma_ratio` sit on the PPP-less inequality rows; `reporting_level` from the country-name suffix |
| `build_wid_main` | `world_inequality_database` `inequality`/`incomes`/`relative_poverty` | welfare→code map, `extrapolated` dim → `_extrapolated` suffix |

`inequality_comparison` reads only `keyvars` (plus the same adapters for indicator metadata). COVID's `extreme_poverty` is rebuilt from the dimensional `poverty` table (consolidated income-or-consumption series at the $3/day line).

No version bump — variable IDs are preserved, so existing charts are unaffected.

### Simplification: keyvars-only output

`poverty_inequality_file` historically also emitted `percentiles` and `wdi` tables, but nothing consumes them (only `keyvars` feeds `inequality_comparison`). They're dropped, which also removes the legacy long→wide→long roundtrip in the percentiles path. The step now reads the dimensional tables directly and emits a single `keyvars` table.

**Build time: ~167s → ~3s**, and the `wdi` dependency is removed from the DAG.

## Parity review (value-preserving)

Feeding the unchanged `keyvars` transforms with the **current** legacy table vs. the dimensional reconstruction yields identical values (max abs diff = float32 rounding ≈ 1e-6 for PIP; WID large-magnitude values match to 1e-4 relative):

| Output | Result |
|---|---|
| PIP keyvars | ✅ 0 value mismatches |
| WID keyvars (both extrapolation states) | ✅ 0 value mismatches |
| COVID `extreme_poverty` input | ✅ 1272 = 1272 rows, 0 mismatches |
| `inequality_comparison` (grapher output) | ✅ unchanged — reads the byte-identical `keyvars` |

One documented, harmless difference: the dimensional PIP dataset restricts regional aggregates to ≥1990 upstream (`regional_data_from_1990`), so the keyvars `keyvars` loses 243 pre-1990 WB-region values (9 regions × 1981–1989 × mean/median/relative-poverty). Those years are outside `inequality_comparison`'s 1993/2019 matching windows, so the chart output is unaffected.

Verified with `ai/parity_check_pip_wid_migration.py`. `make check` passes.

### DIFF
--- a/dag/covid.yml
+++ b/dag/covid.yml
@@ -47,7 +47,7 @@ steps:
       - data://garden/un/2024-07-12/un_wpp
     - data://garden/un/2024-07-12/un_wpp
     # Econ
-    - data://garden/wb/2026-03-24/world_bank_pip_legacy
+    - data://garden/wb/2026-03-24/world_bank_pip
     - data://garden/un/2025-05-07/undp_hdr
     # Health
     - data://garden/wash/2025-12-08/household
--- a/dag/poverty_inequality.yml
+++ b/dag/poverty_inequality.yml
@@ -95,7 +95,6 @@ steps:
       - data://garden/poverty_inequality/2025-01-22/poverty_inequality_file:
         - data://garden/wb/2026-03-24/world_bank_pip
         - data://garden/wid/2026-02-10/world_inequality_database
-        - data://garden/worldbank_wdi/2026-02-27/wdi
       - data://garden/demography/2024-07-15/population
       - data://garden/regions/2023-01-01/regions
       - data://garden/wb/2026-03-24/world_bank_pip
--- a/dag/poverty_inequality.yml
+++ b/dag/poverty_inequality.yml
@@ -93,13 +93,13 @@ steps:
     - data://garden/poverty_inequality/2025-01-22/inequality_comparison:
       # Poverty and inequality file for inequality comparisons at OWID
       - data://garden/poverty_inequality/2025-01-22/poverty_inequality_file:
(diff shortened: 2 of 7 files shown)


paarriagadap:
@codex review

PR #6354, opened 2026-06-29
🐛 fix Taiwan unit mismatch in UNWTO domestic accommodation guests

veronikasamborska1994:
### Problem

Taiwan's `dom_tour_accom_short_term_hotels_and_similar_isic_5510_guests` was **~1000× too large**. The UNWTO source reports Taiwan's domestic hotel guests in **units** while all other 98 countries report in **thousands**. After the garden step's blanket ×1000 conversion, Taiwan showed ~60 billion guests instead of ~60 million.

### Fix

- **`UNIT_CORRECTIONS`**: Divides Taiwan's raw value by 1000 before the ×1000 conversion. The correction is **self-validating** — it asserts that the raw value is still >100× the median of other countries. If the source fixes the inconsistency in a future release, this assert fails with a clear message telling the developer to remove the entry.

### Sanity checks

- **`sanity_check_population_ratios()`**: After unit conversion, checks that no person-count indicator (guests, arrivals, departures, trips) exceeds a plausible multiple of the country's population. Uses domain-appropriate thresholds:
  - **Domestic** (`dom_*`): max **50×** population
  - **Inbound** (`in_*`): max **200×** population (micro-states like Andorra genuinely reach ~122×)

This catches any **new** unit mismatches automatically, without depending on what other countries report — it's anchored to population, a hard physical bound.

### Future scenarios

| Scenario | What happens |
|---|---|
| Source fixes Taiwan | The `ratio > 100` pre-check fails → tells you to remove the `UNIT_CORRECTIONS` entry |
| New country has same bug | Population check catches it with a clear error |
| Source changes units globally | Population check catches all affected countries |

### DIFF
--- a/etl/steps/data/garden/tourism/2026-01-21/unwto.py
+++ b/etl/steps/data/garden/tourism/2026-01-21/unwto.py
@@ -280,7 +280,11 @@ def sanity_check_population_ratios(tb: Table) -> None:
       genuinely receive >100× their population in visitors.
     """
     # Columns that represent counts of people (guests, arrivals, departures, trips).
-    PERSON_COUNT_COLUMNS = [col for col in tb.columns if any(k in col for k in ["guest", "arrival", "departure", "trip", "vis_tourist", "vis_excur"])]
+    PERSON_COUNT_COLUMNS = [
+        col
+        for col in tb.columns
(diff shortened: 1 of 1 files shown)


veronikasamborska1994:
@Marigold  is this ok to merge?

Marigold (approved):
Yes, merge it. The error is unrelated, I'm working on a fix.

PR #6356, opened 2026-06-29
📜 Explain shares above 100% in food trade notebook

pabloarosado:

## What

Adds a short **"Shares above 100%"** subsection to the food-trade technical companion notebook ([docs/analyses/food_trade/food_trade.ipynb](https://github.com/owid/etl/blob/master/docs/analyses/food_trade/food_trade.ipynb)), explaining why a country can show exports above 100% of its production, or imports above 100% of its domestic supply.

## Why

The notebook already documents the TM/SCL consistency checks and the import-coverage gate, but never explains the >100% shares a reader will actually see in the visualization (e.g. roasted coffee for Brazil). This closes that gap.

## What it says

Two causes:

- **Measurement artefact** — trade comes from TM while production and supply come from SCL, and where the two datasets disagree a share can exceed 100% on its own (links back to the *Consistency between TM and SCL trade reports* section).
- **Real effect** — some countries export more than they produce by re-exporting imported goods. Since `domestic supply = Production + Imports − Exports − Stock variation`, when `Exports > Production` it follows that `Imports > Supply + Stock variation`, so (for small stock variations) both shares exceed 100% together.

## Notes

- Only the canonical, git-tracked `docs/` notebook is touched; the untracked `site/` copy is a build artefact.
- The in-page link uses the case-preserved anchor that `docs/ignore/post-build/convert_notebooks.py` generates, so it resolves on the rendered docs site.
- The full notebook was not re-executed: the additions render text / are a collapsed silent-assert cell, so a full run would only churn every chart output. The assertion logic was validated separately.

### DIFF
--- a/docs/analyses/food_trade/food_trade.ipynb
+++ b/docs/analyses/food_trade/food_trade.ipynb
@@ -8,7 +8,7 @@
     "# Global food trade data\n",
     "\n",
     "!!! info \"\"\n",
-    "    :octicons-person-16: **[Pablo Rosado](https://ourworldindata.org/team/pablo-rosado)** • :octicons-calendar-16: June 24, 2026 *(last edit)* • [**:octicons-mail-16: Feedback**](mailto:info@ourworldindata.org?subject=Feedback%20on%20technical%20publication%20-%20Analysing%20global%20food%20trade%20data)"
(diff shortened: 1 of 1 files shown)


pabloarosado:
@codex review

PR #6362, opened 2026-06-30
📊 Use AR5 GWPs for land emissions in emissions by human activity

pabloarosado:

Follow-up to [owid-issues#2431](https://github.com/owid/owid-issues/issues/2431), on the `emissions_by_human_activity_including_gross_lulucf` step (the chart chosen for the Energy chapter of Max's book, option 2).

### What changed

**1. GWP convention: AR4 → AR5.** The step replaces "Growing food" with Jones et al. gross agriculture-and-land-use emissions, converting CH₄/N₂O to CO₂e. It was using IPCC **AR4** GWPs (25 / 298), but Climate Watch (the source for every other activity in the chart) switched to **AR5** in its September 2025 update, and our `climate_watch/2026-02-10` snapshot uses AR5. The land bar was therefore on a different basis than the rest of the chart. Now it uses AR5 (28 / 265) to match. Net effect on the World land bar is small (~1.6%), since the CH₄ and N₂O changes partly cancel.

**2. Removed the silent Climate-Watch fallback.** Where Jones lacked complete gross figures, the step backfilled Climate Watch's *net* land emissions, mixing conventions within a single stacked chart. It now drops those entity-years instead (and the matching rows of the other activities, so no stacked total mixes gross and net land). A warning logs anything dropped.

Only two tiny entities are affected: **Marshall Islands** (all years) and **Palau** (1990–2011) report land-use CO₂ but no agricultural CH₄/N₂O. Both are negligible globally and unaffected at World level.

### Verification

- Pipeline runs clean through the dependent `..._splitting_electricity` step; `make check` passes.
- World 2023: total **54.9 Gt CO₂e**; land-use CO₂ within "Growing food" ≈ **4.8 Gt**, consistent with the Global Carbon Budget figure in the chapter's opening chart. No NaN in the output.

### DIFF
--- a/dag/emissions.yml
+++ b/dag/emissions.yml
@@ -156,7 +156,7 @@ steps:
     - data://garden/emissions/2026-06-15/emissions_by_human_activity_splitting_electricity:
       - data://garden/emissions/2026-06-15/emissions_by_human_activity
       - data://garden/un/2026-02-11/energy_statistics_database
-  # As above, but "Growing food" uses gross agriculture-and-land-use emissions from Jones et al. (AR4).
+  # As above, but "Growing food" uses gross agriculture-and-land-use emissions from Jones et al. (AR5).
   data://grapher/emissions/2026-06-15/emissions_by_human_activity_including_gross_lulucf:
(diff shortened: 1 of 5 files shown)


pabloarosado:
@codex review

PR #6364, opened 2026-07-01
🐛 Decouple metadata tabs from data columns in export_table_to_gsheet

### LINKED ISSUE #6343: export_table_to_gsheet: metadata tabs are coupled to the data columns (can't document an indicator without dumping its values)

## Summary
`export_table_to_gsheet` (in `etl/data_helpers/misc.py`) couples two things that should be independent: **which columns get written to the data tab** and **which indicators get a documentation ("metadata") tab**. As a result, you can't generate a metadata tab for an indicator without also dumping that indicator's values into the data sheet.

## Root cause
- `_add_metadata_tabs(sheet, table, metadata_variables)` extracts metadata via `table[column_name].metadata`, so a variable in `metadata_variables` is only documented **if it's a column of the `table` you pass in**.
- But the same `table` is what `_prepare_dataframe` + `write_dataframe` write to the data tab.
- So the "what data to show" list and the "what to document" list are the same list.

`_get_variables_to_process` confirms it — variables not present in the table are silently skipped with just a warning:
```
variables_to_process = [var for var in metadata_variables if var in table.columns]
...
log.warning(f"Warning: Variables {missing_vars} not found in table")
```

## The either/or this forces
- Pass `literacy_rate` → metadata tab is generated ✅, but its **values land in the data tab** ❌.
- Omit `literacy_rate` → data tab is clean ✅, but **no metadata tab is produced** (silent warning) ❌.

There's no way to say "document indicator X but don't show its values."

## Real-world impact (how this surfaced)
`grapher/literacy/2026-05-12/historic_literacy_omm.py` exports a **sources-only** data tab (`source_url`, `source`) but requests `metadata_variables=["literacy_rate"]`. Because `literacy_rate` isn't in the exported table, **no metadata tab is generated** — the only reason the live sheet appears to have one is a stale leftover tab from an older code version (`update_existing=True` overwrites the data tab but never deletes old tabs). Any freshly-created sheet gets no metadata tab at all.

## Proposed fix
Decouple the two inputs. Options, roughly in order of preference:
1. Add a separate `metadata_table` (or let `metadata_variables` resolve against a table other than the one written to the data tab), so callers can document indicators that aren't shown as data.
2. At minimum, **fail loudly** instead of silently warning when a requested `metadata_variables` entry isn't found — today this is a silent no-op that hides the problem.

## Related / follow-on
- Worth pairing with the broader gsheet-export rework so refreshes aren't tied to one person's `client_secret.json` (a shared **service account** + **fixed sheet ID** so any maintainer — and CI — can update the same sheet). The current `drive.file` scope means each person's run spawns a new sheet, which is its own maintenance problem.

Marigold:

Fixes #6343.

## Problem

`export_table_to_gsheet` resolved `metadata_variables` only against the same table that gets written to the data tab (`table[column_name].metadata`). So a variable could only get a documentation tab if its values were also dumped into the data sheet — there was no way to document an indicator without showing it. Missing variables were also silently dropped with just a `log.warning`.

This bit `historic_literacy_omm.py`: it exports a sources-only data tab (`source_url`, `source`) but requests a metadata tab for `literacy_rate`, which isn't a column of that table — so no metadata tab is generated on a fresh sheet.

## Fix

- Added an optional `metadata_table` param to `export_table_to_gsheet`. When given, `metadata_variables` resolves against it instead of `table`, so the data tab and metadata tab can be sourced from different tables. Defaults to `None` → falls back to `table`, so existing callers are unaffected.
- `_get_variables_to_process` now raises `ValueError` on a missing variable instead of silently dropping it with a warning.
- Updated `historic_literacy_omm.py` to pass `metadata_table=tb` (the full table, which has `literacy_rate`) alongside `table=tb_sel` (the sources-only data tab).

The other three callers (`long_run_child_mortality.py` ×2, `life_expectancy.py`) already include their indicator column in the table they pass, so they're unaffected by this change.

## Unrelated CI fix bundled in

`buildkite/etl-automated-staging-environment` was failing on `test_app_fasttrack`, unrelated to the above. That test reimports whichever `snapshots/fasttrack/**/*.dvc` entry has the most recent `date_accessed` — currently `france_wealth_share_top1` (a DRAFT/private dataset merged to master separately), whose Google Sheet isn't published as CSV yet. The test already tolerates one flavor of "the picked sheet isn't in the expected state" (a missing-template error); extended that same allowlist to also cover "not published as CSV," since both are expected states of real, in-progress fasttrack data rather than app bugs.

## Test plan

- [ ] `make check` (lint/format/typecheck) — done locally, all green
- [ ] Run `historic_literacy_omm` grapher step and confirm the exported sheet has both the sources-only data tab and a `metadata_literacy_rate` tab
- [x] CI (`buildkite/etl-automated-staging-environment`) passes with the draft fasttrack sheet present

### DIFF
--- a/etl/data_helpers/misc.py
+++ b/etl/data_helpers/misc.py
@@ -945,12 +945,11 @@ def _get_variables_to_process(table: Table, metadata_variables: list[str] | None
     if metadata_variables is None:
         return list(table.columns)
 
-    variables_to_process = [var for var in metadata_variables if var in table.columns]
     missing_vars = [var for var in metadata_variables if var not in table.columns]
     if missing_vars:
-        log.warning(f"Warning: Variables {missing_vars} not found in table")
+        raise ValueError(f"metadata_variables not found in table: {missing_vars}")
 
-    return variables_to_process
+    return metadata_variables
 
 
 def _extract_metadata_rows(metadata_obj, included_fields: set) -> list[list[str]]:
@@ -1097,8 +1096,13 @@ def export_table_to_gsheet(
     general_access: Literal["anyone", "domain", "user"] = "anyone",
     include_metadata: bool = True,
     metadata_variables: list[str] | None = None,
+    metadata_table: Table | None = None,
 ) -> tuple[str, str] | None:
     """Export a Table to Google Sheets with improved error handling and performance.
+
+    `metadata_variables` is resolved against `metadata_table` (falling back to `table`
+    if not given), so a variable can be documented in a metadata tab even if it isn't
+    one of the columns written to the data tab.
 
     Returns
     -------
@@ -1128,7 +1132,7 @@ def export_table_to_gsheet(
 
         # Add metadata if requested
         if include_metadata:
-            _add_metadata_tabs(sheet, table, metadata_variables)
+            _add_metadata_tabs(sheet, metadata_table if metadata_table is not None else table, metadata_variables)
 
         # Set permissions
         _set_permissions(sheet.sheet_id, role, general_access)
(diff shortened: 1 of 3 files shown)


Marigold:
@codex review

Marigold:
@codex review

PR #6366, opened 2026-07-01
📊 agriculture: Update food trade pipeline and technical publication

pabloarosado:

## What

Removes the percentage shares (share of a country's production that it exports, share of its domestic supply that it imports) from the food-trade Sankey visualization pipeline. Per the discussion in `#proj-food-trade-viz`, these shares occasionally gave useful context but more often caused confusion, so we drop them. This also removes the SCL dependency, which was only there to compute those shares.

## Changes

- **Garden `food_trade`**: drops the SCL load, the Food Balance Sheet supply computation, the import-coverage gate, and the `exporter_production` / `importer_supply` columns. The step now builds only the viz-ready bilateral trade flows (importer-first reconciliation and combined-item logic are unchanged).
- **`food_trade.meta.yml`**: removes the two now-absent variable definitions.
- **`dag/faostat.yml`**: drops the `faostat_scl` dependency of the `food_trade` garden step.
- **S3 export** (`export://s3/faostat/latest/food_trade`): each per-product JSON now carries `flows` only; the `production` and `supply` blocks are gone.
- **Technical publication** (`docs/analyses/food_trade/food_trade.ipynb`): the "Estimating production and supply" section now explains that we explored deriving these shares from SCL and why we decided not to publish them (TM and SCL are compiled independently and often disagree; the shares frequently exceed 100% from re-exports or dataset mismatch; SCL omits major traders such as Japan and Singapore). SCL/QCL are also now read from the published feather inside that section, rather than up front, so the notebook no longer exceeds Colab's memory limit.

## Verification

- The bilateral `flows` are **byte-identical** to production across all 149 products (confirmed against the live JSONs); only the unused `production`/`supply` blocks were removed.
- The notebook re-executes cleanly end to end, with all assertions passing.
- The viz on `owid-grapher` `master` already stopped reading `production`/`supply`, so publishing flow-only data does not break the live visualization.


pabloarosado:
@codex review

PR #6369, opened 2026-07-01
📊 Update World Bank income groups (2026-07-01)

paarriagadap:

Tracks: [owid/owid-issues#2437](https://github.com/owid/owid-issues/issues/2437)

## Summary

Annual update of the **World Bank income classification** (`wb/2025-07-01/income_groups` → `wb/2026-07-01/income_groups`). The new World Bank release (OGHIST, July 1, 2026) is the **FY2027** classification, based on **2025 GNI per capita** (Atlas method).

**Notable data changes (all upward or newly classified — no downgrades):**

| Country | 2024 | 2025 |
|---|---|---|
| Ethiopia | *(unclassified)* | **Low-income** |
| Venezuela | *(unclassified)* | **Lower-middle-income** |
| Togo | Low-income | **Lower-middle-income** |
| Jordan | Lower-middle-income | **Upper-middle-income** |
| Micronesia | Lower-middle-income | **Upper-middle-income** |
| Philippines | Lower-middle-income | **Upper-middle-income** |
| Sri Lanka | Lower-middle-income | **Upper-middle-income** |
| Vietnam | Lower-middle-income | **Upper-middle-income** |

Ethiopia and Venezuela return to the classification after years of unclassification, so the Ethiopia manual-unclassification workaround was removed. All historical (pre-2025) values are byte-identical to the previous version.

**FY2027 GNI thresholds (2025 GNI):** Low ≤ $1,175 · Lower-middle $1,176–$4,635 · Upper-middle $4,636–$14,375 · High > $14,375.

## Downstream dependencies (deferred to #6416)

All **85 references** to `wb/2025-07-01/income_groups` across **15 DAG files** are repointed to `2026-07-01` in the follow-up PR #6416 (stacked on this one), which also removes/archives the old chain and resyncs `electricity_mix`'s EI-vs-Ember discrepancy list (only correct against the 2026 groups). This mirrors the 2025 two-PR pattern (#4673 → #4692) and lets the downstream impact be reviewed via the data-diff report on that PR. Until #6416 merges, both chains stay active: downstream datasets keep consuming `2025-07-01`.

The `countries_that_must_have_data` fixes in `population`/`unaids` (Philippines/Vietnam moved to upper-middle income) also live in #6416 — they only become necessary once consumers point at the 2026 groups, so this PR stays purely additive (the new chain only).
## Charts

26 published charts use income_groups indicators (3 dedicated: `world-bank-income-groups` map, `countries-by-income-classification`, `population-by-income-classification`; 23 use the classification as a scatter dimension). All remapped to the new variables on staging, plus 3 narrative charts (including two that pinned stale 2024-07-29 / 2025-07-01 variables the auto-upgrader left behind).

<details>
<summary>Source file layout change (meadow sanity checks)</summary>

The WB dropped the legend block ("Low income (L)", "Lower middle income (LM)", "Upper middle income (UM)") that used to sit between the year header and the first country. So the raw row count fell 239 → 234 and the first country (Afghanistan) moved from row 10 to row 5. Data extraction is unaffected (those rows have no country code and are dropped), but the meadow input sanity checks — which key off absolute row positions — were updated accordingly, with a `# NOTE` for the next maintainer.

Also: the WB relabeled Somalia's source name `Somalia` → `Somalia, Fed. Rep.`; updated `income_groups.countries.json`.
</details>

<details>
<summary>Modernized outdated practices (+ A/B test)</summary>

- `geo.harmonize_countries` → `paths.regions.harmonize_names` (garden, mechanical).
- Deleted the snapshot `.py` script entirely — the `.dvc` has a stable `url_download`, so the snapshot runs script-less via `etls` / `etl snapshot`.
- `geo.add_regions_to_table` → `paths.regions.add_aggregates` in `income_groups_aggregations`. This changes the aggregation core, so it was validated with a **controlled A/B test**: built both ways against the same catalog and diffed — **0 columns differ, 0 max abs diff (identical output)**.
</details>

<details>
<summary>New sanity checks</summary>

Added a grounded `sanity_check_outputs` to the aggregations step (it previously had none): no fully-NaN column, non-negative country counts, non-negative status populations. Each assertion was negative-tested to confirm it fires.

`classification_missing_pop` (the region-total-minus-summed-classified-population residual) can compute slightly negative when the summed classified population marginally exceeds the region total — only **Europe 1990** (−66M). Such residuals are **clamped to 0** (missing population can't be negative), guarded by an `EXPECTED_NEGATIVE_MISSING_POP = {("Europe", 1990)}` whitelist so any *new* `(region, year)` negative **fails for review** rather than being silently zeroed. With the clamp in place, `missing_pop` is now included in the non-negative output check.
</details>

<details>
<summary>Metadata updates</summary>

- `description_from_producer` + `description_key` thresholds bumped to FY2027 / 2025 GNI.
- Reworked the Venezuela/Ethiopia `description_key` bullet (both are now classified).
- `classification` processing level set to `minor` (it is a relabeling of the WB's own published classification); the derived aggregations remain `major` with `description_processing`.
- Set `presentation.attribution_short: World Bank` (does not inherit from origin).
- Clarity: `income_groups_latest` description_short "latest year informed" → "latest available year".
- `.dvc`: `url_download` switched to the stable, date-less WB DDH OpenAPI endpoint (`.../resources/DR0095334/download`); `date_published` and `citation_full` → 2026.
</details>

## Pre-existing (not a regression)

The aggregation step logs an "Unknown overlaps: Netherlands Antilles / Aruba (1987–2009)" warning (both were separate reporting entities in those years). Present in the prior version too — not introduced here.


paarriagadap:
@codex review

snapshots/wb/2026-07-01/income_groups.xlsx.dvc
```diff
@@ -4,15 +4,15 @@ meta:
     title: Income Classifications
     description: |-
       Every year, the World Bank Group classifies the world’s economies into four income groups: low, lower-middle, upper-middle, and high. These classifications, updated each year on July 1, are based on the previous year’s Gross National Income (GNI) per capita, expressed in U.S. dollars using the [Atlas method](https://datahelpdesk.worldbank.org/knowledgebase/articles/77933-what-is-the-world-bank-atlas-method).
-    citation_full: World Bank (2025). World Bank Country and Lending Groups.
+    citation_full: World Bank (2026). World Bank Country and Lending Groups.
     url_main: https://datahelpdesk.worldbank.org/knowledgebase/articles/906519-world-bank-country-and-lending-groups
-    url_download: https://datacatalogfiles.worldbank.org/ddh-published/0037712/DR0095334/OGHIST_2025_07_01.xlsx
-    date_accessed: '2025-07-01'
-    date_published: '2025-07-01'
+    url_download: https://ddh-openapi.worldbank.org/resources/DR0095334/download
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Code — @paarriagadap at the wheel._

Intentional — keeping the date-less DDH OpenAPI endpoint (`.../resources/DR0095334/download`). We already use stable, always-current download endpoints for other snapshots (e.g. OECD) and it has not been a problem in practice.

Downstream consumption is pinned to the `outs` md5 (`a330bbe7…`) — etlr pulls the file from S3 by checksum and never re-hits `url_download` — so builds are fully reproducible regardless of the endpoint. The mutability only matters on a deliberate *re-creation* of this exact version (rare), and the convenience (next year's updater does not have to reverse-engineer the dated filename) outweighs that narrow risk. Not changing.

paarriagadap:
Hi @pabloarosado!

Can you take a look to see if this migration makes sense and if I covered all necessary changes? [Here](http://staging-site-data-worldbank-incomegroups/etl/wizard/chart-diff?show_all=&chart-diff-sort-by-results=relevance&show-narrative-charts=False&show-article-citations=False) you can see all affected charts. I don't think there should be a problem.

The update includes bumping versions where `income_groups` is used as a dependency, so we can archive the previous version.

I don't know why `etl-automated-staging-environment` is failing, it makes it look like `fur_laws fails`, but I don't see it's the case.

pabloarosado:
Hey @paarriagadap, I pushed a fix. Philippines and Vietnam were moved from lower-middle to upper-middle income, but both were still hardcoded as `countries_that_must_have_data` for the lower-middle aggregate in `population.py` (and Philippines in `unaids.py`). The aggregates were hence missing in some charts (e.g. TB death rate):
tuberculosis-death-rate

(In a separate PR, we should decide how to avoid this kind of silent issues in the future)

paarriagadap:
Ahh, you're right @pabloarosado. Thanks. I remember changing this in the past but not this time. I will also think about options to tackle this.

pabloarosado (approved):
Thanks @paarriagadap, I did some checks and I think now this PR is ready to merge!
The staging server is having strange issues; I think we'd need to create a new one or reboot it to fix those issues, but I don't think that's necessary. So, feel free to merge later today.

etl/steps/data/garden/energy/2026-04-24/electricity_mix.py
```diff
@@ -463,6 +459,11 @@ def fix_discrepancies_in_aggregate_regions(tb_review: Table, tb_ember: Table, co
             "South America": [],
         }
     )
+    # Collect any region whose actual EI-vs-Ember discrepancies no longer match the expected list.
```
paarriagadap: @pabloarosado Hey, I checked the migration last year https://github.com/owid/etl/pull/4692 and there was a change in electricity_mix. Claude had a look and suggested changes, but I am not sure if this is what you intend. Please take a look and we can modify/revert if necessary

paarriagadap:
Hi @pabloarosado!

I decided to do what we did last year and separate the PRs: this one will be for the data update and this one https://github.com/owid/etl/pull/6416 for the downstream changes. I'm doing this because I want to try the new data-diff to check the largest changes.

etl/steps/data/garden/energy/2026-04-24/electricity_mix.py
paarriagadap: See the same change at https://github.com/owid/etl/pull/6416

PR #5996, opened 2026-04-29
Add catalog dataset JSON-LD artifacts

Marigold:
> _Written by Claude Code — @Marigold at the wheel._

## Summary

Adds an ETL-owned path for generating Schema.org Dataset JSON-LD for high-quality public garden catalog datasets. The implementation writes per-dataset `dataset.jsonld` files next to catalog `index.json`, generates a sitemap for canonical dataset pages, and records skipped/warning datasets in `jsonld_quality_report.json` so metadata gaps can be fixed upstream.

This PR only creates and publishes the JSON-LD/sitemap artifacts; it does not yet make Dataset Search results useful on its own. Follow-up work in the catalog Cloudflare Worker is still needed to publish the sitemap and serve human-readable dataset landing pages that embed these JSON-LD files.

The first cut keeps JSON-LD generation behind `etl d publish --jsonld`, with publishing synced independently from normal dataset checksums so mapper changes can update structured metadata without rerunning data steps. `--jsonld-base-url` lets staging generate self-consistent links for `catalog-staging.ourworldindata.org`.

## v1 canary allowlist (`--jsonld-only`)

The catalog-discovery project scoped a deliberately small **v1 canary**: expose JSON-LD for only the three datasets OWID already publishes openly on GitHub (`co2-data`, `energy-data`, `poverty-data`) — a discovery surface for already-public data, not a new disclosure — rather than the whole garden catalog. Exposing all garden datasets at once was pushed back on by leadership.

To support this without locking the feature to those three, `--jsonld-only '<namespace>/<dataset>'` (repeatable) restricts generation to an allowlist. The match is on `namespace` + `dataset` (**not** the dated path), so it survives data re-versioning; the latest public version is always chosen. Default (no `--jsonld-only`) stays "all eligible garden datasets" so the PR remains useful for a later catalog-wide rollout. Allowlist entries that match no dataset are logged as a warning (`catalog_jsonld.allowlist_entry_unmatched`) so a typo or renamed dataset is visible rather than failing silently. The allowlist and the quality gate compose: the allowlist *selects*, the quality gate *protects*.

Publish the trio with:

```bash
.venv/bin/etl d publish --jsonld --dry-run \
  --jsonld-only emissions/owid_co2 \
  --jsonld-only energy_data/owid_energy \
  --jsonld-only wb/world_bank_pip
```

### Where the allowlist lives

[...]

PR #6375, opened 2026-07-02
✨ Fill catalog JSON-LD table descriptions from existing metadata

Marigold:

## What & why

After the catalog-discovery v1 canary (#5996) shipped, the JSON-LD quality report flagged `missing_table_description` on every table of all three curated datasets (CO2, energy, and the 7 World Bank PIP poverty tables).

The original plan was to backfill a table-level `description` for each. But `table.description` is a mostly-internal field (its own schema doc: *"Description of the table (mostly for internal purposes, or for users of our data catalog)"*), we focus on **dataset**-level descriptions, and authoring new per-table prose would need dataset-owner sign-off. So this fixes the generation/check logic to **reuse descriptions that already exist in the metadata** instead of requiring new content.

## Changes

- **Generator** (`lib/catalog/owid/catalog/schema_org.py`): new `table_description(table, dataset_meta)` helper. For the multi-table `hasPart` representation, a table node's description now falls back to the producer (origin) description, then the dataset description, instead of being left blank. Single-table datasets are unchanged (their table already flattens into the top-level `Dataset`, which carries `dataset.description`). Nothing is synthesized — if no description exists anywhere, the node simply has none.
- **Quality check** (`etl/catalog_jsonld/quality.py`): `missing_table_description` now fires only when no description resolves from the table, its origins, **or** the dataset — matching what the JSON-LD actually emits. This turns a near-universal false positive into a real signal.
- **Tests**: fallback ordering (explicit → origin → dataset → none) in the catalog lib, and the check's warn/no-warn behaviour in `tests/catalog_jsonld/test_quality.py`.

## Verification

Ran the artifact builder in dry-run against locally-built CO2, energy, and PIP. `missing_table_description` clears for all three (only energy's out-of-scope `wide_table_variableMeasured_truncated` remains). Confirmed the fallback yields real content on genuinely blank tables: CO2/energy → their OWID origin description; the PIP tables → the World Bank PIP dataset description (with `percentiles` keeping its more specific origin text).

## Note — reverses the brief's guardrail

The original task brief said "fix the data, not the check." That assumed we'd author descriptions; since that's off the table (owner approval), the fix moved into the generator/check. The `missing_table_description` change is a calibration, not a removal — it still fires when a table has no description available from any source.

## Out of scope / follow-ups

- **CO2 dataset-level description bug**: the built `garden/emissions/2025-12-04/owid_co2` dataset has inherited junk dataset metadata (`title: "Population"`, `description:` = the Maddison/GDP origin text) via `default_metadata` from an auxiliary input. Its JSON-LD description is only good in prod because `latest` seems to resolve elsewhere. Worth its own fix (likely needs owner-approved text).
- **Issue 1 (Downloads section)**: the visible `<h2>Downloads</h2>` on multi-table dataset pages is a rendering gap in `owid/cloudflare-workers` (`datasetPage.ts` reads only top-level `distribution`, not `hasPart[].distribution`) — a separate PR in that repo, not addressed here.

### DIFF
--- a/etl/catalog_jsonld/quality.py
+++ b/etl/catalog_jsonld/quality.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 
 from owid.catalog.core.meta import DatasetMeta
-from owid.catalog.schema_org import TableSchemaInput, license_to_url
+from owid.catalog.schema_org import TableSchemaInput, license_to_url, table_description
 
 
 @dataclass
@@ -51,7 +51,10 @@ def assess_dataset_quality(
 
     for table in tables:
         warnings = []
-        if not table.metadata.description:
+        # A table rarely sets its own description (a mostly-internal field); the emitted
+        # JSON-LD falls back to the producer (origin) or dataset description, so only warn
+        # when no description resolves from any of those existing sources.
+        if not table_description(table, dataset_meta):
             warnings.append("missing_table_description")
         variable_count = len(
             [name for name in table.variables if name not in set(table.primary_key or table.metadata.primary_key)]
--- a/lib/catalog/owid/catalog/schema_org.py
+++ b/lib/catalog/owid/catalog/schema_org.py
@@ -121,12 +121,32 @@ def dataset_to_schema_org(
         if distributions:
             result["distribution"] = distributions
     elif tables:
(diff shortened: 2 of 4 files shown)


Marigold:
@codex review

PR #6385, opened 2026-07-02
📊 Require an explicit dataset description for catalog JSON-LD

Marigold:

## What

`owid_energy`'s catalog page description was silently the population dataset's blurb ("Our World in Data builds and maintains a long-run dataset on population..."), even though `owid_energy` has nothing to do with population.

`owid_energy.meta.yml` has no `dataset.description` (removed in #6378 for being unhelpful boilerplate). With no explicit description, `owid.catalog.schema_org._dataset_description()` fell back to scanning every column's indicator origins and used the description of whichever one it hit first — which happened to be the `population` column's origin, an auxiliary indicator pulled in only for per-capita calculations.

## Fix

- Removed the origin-scanning fallback from `_dataset_description()` entirely. A dataset must now have an explicit `dataset.description` — otherwise `dataset_to_schema_org()` raises instead of guessing.
- `TableMeta.description` is also never used as a substitute, even for single-table datasets — it's a mostly-internal field authors rarely set with a public-facing description in mind. Only `dataset_meta.description` counts.
- Tightened `assess_dataset_quality`'s `missing_description` check to match: neither an origin's description nor `TableMeta.description` counts as the dataset "having" a description anymore. This means datasets lacking a real `dataset.description` are blocked from JSON-LD generation (skipped, reported in the quality report) *before* they'd ever reach the new raise — consistent with how `missing_title`/`missing_license_url`/`missing_provenance` already work.
- Added a real `dataset.description` to `owid_energy.meta.yml`, written from the language actually published in the [energy-data repo's README](https://github.com/owid/energy-data) (the public-facing repo this dataset is exported to), rather than an invented summary.
- `table_description()` (used only for individual table nodes in multi-table `hasPart` datasets) still falls back to an origin's or the dataset's description when a table lacks one of its own — that part is pre-existing, deliberate, and already covered by its own tests. It just no longer considers `TableMeta.description` first.

## Note

`emissions/2025-12-04/owid_co2.meta.yml` still has the same kind of boilerplate description that was removed from `owid_energy` in #6378 ("this dataset will be loaded by the co2-data repository..."). Left untouched here since it's out of scope and already has *an* explicit description, but worth cleaning up separately.

## Test plan

- [x] `pytest lib/catalog/tests/test_schema_org.py` — tests for the raise, and confirming `TableMeta.description` is never used
- [x] `pytest tests/catalog_jsonld/` — a test confirming `missing_description` blocks a dataset that only has an origin-level description
- [x] `pytest lib/catalog` (full suite) — 368 passed
- [x] `make check`

### DIFF
--- a/etl/catalog_jsonld/quality.py
+++ b/etl/catalog_jsonld/quality.py
@@ -117,14 +117,10 @@ def _has_title(dataset_meta: DatasetMeta, tables: list[TableSchemaInput]) -> boo
 
 
 def _has_description(dataset_meta: DatasetMeta, tables: list[TableSchemaInput]) -> bool:
-    if dataset_meta.description or any(table.metadata.description for table in tables):
-        return True
-    for table in tables:
-        for variable in table.variables.values():
-            for origin in variable.origins:
-                if origin.description or origin.description_snapshot:
-                    return True
-    return False
+    # Matches the contract in owid.catalog.schema_org._dataset_description: an indicator
+    # origin's description (e.g. an auxiliary population/GDP column) does not count as the
+    # dataset having a description of its own — only an explicit dataset- or table-level one does.
+    return bool(dataset_meta.description or any(table.metadata.description for table in tables))
 
 
 def _has_license_url(dataset_meta: DatasetMeta, tables: list[TableSchemaInput]) -> bool:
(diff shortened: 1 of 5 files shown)


Marigold:
@codex review

PR #6404, opened 2026-07-07
✨ Improve data-diff PR comment and add HTML report

Marigold:

## Why

The data-diff comment is posted on every PR but rarely used, while chart-diff is popular. The gap is structural: chart-diff's comment is a compact scoreboard linking to an interactive page, whereas data-diff's comment *is* the whole deliverable — a wall of monospace text that leads with `= Dataset` noise, buries the actual signal, truncates at 64k chars, and whose call-to-action is "re-run this locally with more flags".

## What

Two-part redesign, deliberately lightweight (no new Wizard app):

**1. Signal-first PR comment** — only changed/new/removed datasets appear, with per-column counts and up to 3 sample rows; identical datasets collapse to a single count line:

```
data-diff: ❌ 1 changed · 3 identical · 16 skipped — full report ↗

~ garden/who/2025-01-17/vaccine_stock_out
    ~ vaccine_stock_out.value (changed data)
        ~ Changed values: 4 / 119,537 (0.00%)
            Austria 2019 …: Inaccurtae forecasts → Inaccurate forecasts
            (+3 more)
```

**2. Sidecar HTML report** — a self-contained page with the full diff (old→new value tables with up to 100 sample rows, metadata diffs, "show identical" toggle, live filter), uploaded to `s3://owid-catalog/diffs/<branch>/data-diff.html` and linked from the comment at `https://catalog.ourworldindata.org/diffs/<branch>/data-diff.html`. Cloudflare edge-caches for ~5 min, so the link carries a `?v=<timestamp>` param to always point at the latest run.

## How

- `etl/datadiff_report.py` (new): structured diff model + JSON serialization + HTML renderer.
- `etl/datadiff.py`: `DatasetDiff` fills a structured `result` while printing; new `--output-json` / `--output-html` CLI flags. **The printed text output is byte-identical to before** (existing exact-string tests pass unchanged).
- `apps/owidbot/data_diff.py`: builds the comment from the structured JSON instead of ANSI-decoding stdout; uploads the HTML report (degrades gracefully to a link-less comment if the upload fails).
- `lib/catalog/owid/catalog/s3_utils.py`: `upload()` gains `content_type` / `cache_control` params so the report renders inline in the browser.

## Verification

- 7 tests pass (`tests/test_datadiff.py`, new `tests/apps/test_owidbot_data_diff.py`); `make check` green here and in `lib/catalog`.
- End-to-end: tweaked a local dataset value, ran `etl diff REMOTE data/ --output-html/--output-json`, verified text/JSON/HTML all show the change; browser-checked the report incl. toggle and filter.
- Confirmed `catalog.ourworldindata.org` serves uploaded reports with `content-type: text/html` (test objects deleted afterwards).
- Real output demo: see the refreshed owidbot comment on #6397.

### DIFF
--- a/apps/owidbot/cli.py
+++ b/apps/owidbot/cli.py
@@ -82,7 +82,7 @@ def cli(
     services_body = {}
     for service in services:
         if service == "data-diff":
-            services_body["data-diff"] = data_diff.run(include)
+            services_body["data-diff"] = data_diff.run(include, branch=branch)
 
         elif service == "chart-diff":
             charts_df = chart_diff.call_chart_diff(branch)
--- a/apps/owidbot/data_diff.py
+++ b/apps/owidbot/data_diff.py
@@ -48,7 +48,9 @@ def run(include: str, branch: str | None = None) -> str:
         call_etl_diff(include, output_json=json_path, output_html=html_path)
 
         report = DiffReport.from_json(json_path.read_text())
-        report_url = upload_html_report(html_path, branch) if branch else None
+        # A report that just says "no differences" isn't worth hosting — upload only when
+        # there's something to show.
+        report_url = upload_html_report(html_path, branch) if branch and report.status != "clean" else None
 
     return format_comment(report, report_url)
 
--- a/apps/owidbot/data_diff.py
+++ b/apps/owidbot/data_diff.py
@@ -1,77 +1,204 @@
+"""data-diff service for owidbot.
+
+Runs `etl diff` against the remote catalog, uploads the full self-contained HTML report to R2
+(publicly served at https://catalog.ourworldindata.org/) and posts a compact, signal-first
+summary to the PR: only changed datasets with a few sample rows, everything deeper lives in
+the linked report.
+"""
+
 import os
 import re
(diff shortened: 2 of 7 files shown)


Marigold:
@codex review

Marigold:

Valid finding — thanks. Note it predates this PR (the fast path came in with #6154), but this PR made the consequence worse since a "clean" diff now skips the report entirely.

Fixed in 1e0d8c0660: the skip now additionally requires (a) dataset-level metadata to compare equal in-memory (`_dataset_metadata_dict`, which already excludes `source_checksum` — no extra fetch, the remote `index.json` is already loaded), and (b) every table's `.meta.json` to be byte-identical via the same local-MD5-vs-ETag check as the feathers. Any mismatch falls through to the normal full diff, so a false negative only costs speed, never correctness.

The fix immediately proved itself locally: two datasets previously skipped as "identical" now correctly surface a dataset-metadata diff (an `owners` field added remotely after the local build). Covered by a new test (`test_dataset_files_match_covers_metadata`).

PR #6419, opened 2026-07-09
✨ data-diff report: same format for small and large reports

paarriagadap:

Follow-up to #6415.

## Why

The triage aids (tier strip + "Top changes" watch lists) only rendered on reports with ≥ 3 changed datasets. Two problems: reviewers get two different report formats depending on diff size, and the gate keyed on *dataset count* — so a **single-dataset diff with hundreds of changed indicators** (WDI-style, exactly where a ranked Indicators watch list helps most) got no watch list at all. Conversely, a fully uniform format would add pure duplication to a one-indicator diff.

## What

Replace the size gate (`TRIAGE_MIN_DATASETS`) with **per-element content gates** — each element renders only when it can discriminate, the same principle already used for the controls:

- **Tier strip**: ≥ 2 differing datasets ("Of the 1 dataset with differences…" restates the headline).
- **Datasets watch sub-list**: ≥ 2 entries.
- **Indicators watch sub-list**: ≥ 2 entries (losses + changes) — a ranking of one thing ranks nothing.

Resulting behavior: a one-indicator diff gets a clean minimal report (chips only); a single-dataset/many-indicators diff keeps the ranked **Indicators** list without the singleton Datasets list or the strip; big reports keep everything. The control gating (tier dropdowns ≥ 2 categories, identical-toggle only when identical datasets exist, show-more only when truncated) is unchanged.

## Verification

22 + 4 tests pass (`test_triage_aids_gate_on_content` covers all three shapes); verified on a real single-dataset diff (`electricity_mix` vs production): no tier strip, no Datasets sub-list, ranked Indicators list present with anchors.

### DIFF
--- a/etl/datadiff_report.py
+++ b/etl/datadiff_report.py
@@ -684,8 +684,9 @@ def render_html(report: DiffReport) -> str:
     else:
         ind_tier_select = ""
 
-    # Triage aids — one format for every report size; content-based gating on the controls
-    # (dropdowns, toggles, show-more) already prevents dead UI on small diffs.
+    # Triage aids gate on content, not report size: each element renders only when it can
+    # discriminate — the strip needs >= 2 differing datasets ("Of the 1 dataset…" restates the
+    # headline), and a watch sub-list needs >= 2 entries (a ranking of one thing ranks nothing).
     tier_strip = ""
     top_block = ""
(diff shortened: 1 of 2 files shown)


paarriagadap:
@codex review

PR #6420, opened 2026-07-10
🐛 Post data-diff comment even when some datasets error

paarriagadap:

## Problem

When any dataset fails to compare, `etl diff` prints `⚠ Found errors, create an issue please` **instead of** `❌ Found differences` (they're `elif` branches) and exits 1. owidbot's gate in `call_etl_diff` only tolerates exit 1 when stdout contains the literal `"Found differences"`, so it raises `RuntimeError`, the owidbot process dies before posting, and the ops wrapper's `|| true` swallows the crash. Net effect: the PR comment silently keeps whatever stale content it had (typically the pre-bake `✅ no differences`).

This is exactly what happened on #6416: the bake finished at 16:34 UTC, every subsequent owidbot run crashed on this gate, and the comment stayed frozen at the 15:57 pre-bake state. Reproduced manually on the staging box:

```
RuntimeError: etl diff failed (exit 1)
...
⚠ Found errors, create an issue please
```

## Fix

Accept exit 1 when stdout contains either summary line. This is safe because `etl diff` calls `_write_reports()` **before** exiting, so the JSON/HTML reports are always complete — and the report pipeline already models error datasets explicitly (`⚠️` status icon, `! path — error: ...` lines in the comment, severity override in the HTML report). The gate was just never updated for that.

Runs that exit 1 *without* either summary line (a genuinely broken run) still raise, as do exit codes ≥ 2.

## Test

`test_call_etl_diff_accepts_completed_runs` covers all four gate outcomes (errors line, differences line, exit 2, exit 1 with no summary line) with a stubbed `Popen`.

### DIFF
--- a/apps/owidbot/data_diff.py
+++ b/apps/owidbot/data_diff.py
@@ -248,8 +248,12 @@ def call_etl_diff(include: str, output_json: Path, output_html: Path) -> None:
     # Remove all warnings from stderr
     stderr = re.sub(r"^.*WARNING.*", "", stderr, flags=re.MULTILINE).strip()
 
-    if result.returncode == 1 and "Found differences" in stdout:
-        log.info("etl diff found differences", returncode=result.returncode)
+    # `etl diff` exits 1 for any completed run with findings; when a dataset fails to compare,
+    # the summary line is "⚠ Found errors, create an issue please" *instead of* "Found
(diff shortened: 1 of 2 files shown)


paarriagadap:
@codex review

PR #6423, opened 2026-07-10
Attribute AI-written content by model name, not "Claude Code"

Marigold:
> _Written by Sonnet 5 — @Marigold at the wheel._

## Summary
- The disclosure attribution line (`CLAUDE.md`'s "Team" section, plus two skills that echoed it) said `Written by Claude Code`, which doesn't identify which underlying model actually produced the content.
- Changed the template to `Written by <model name>` (e.g. "Sonnet 5", "Opus 4.8", "Fable 5", "Haiku 4.5") — substituted with whichever model is generating the content at the time — and added a note clarifying that.
- Updated the two skill files (`add-provider-regions`, `review-data-pr`) that duplicated the old wording, so all three stay consistent.


pabloarosado (approved):
Good idea! Maybe it could be "`Claude Code <model name>`" or "`Claude <model name>`" instead, given that the name of the models (Haiku, Fable...) can be a bit obscure to some people, but mentioning Claude makes sense to most. But up to you, your proposal is also good.

Marigold:
Good idea! Implemented.

PR #6431, opened 2026-07-10
📊 🐛 Correct India universal suffrage years in Lexical Index

lucasrodes:
> _Written by Claude Code — @lucasrodes at the wheel._

Corrects a confirmed error in the Lexical Index source data (Skaaning et al.), requested in [owid-issues#2301](https://github.com/owid/owid-issues/issues/2301#issuecomment-4924264618).

India's universal right to vote (men **and** women) is coded as beginning in **1947** (independence) but actually began in **1950**, when the Constitution came into force. Bastian confirmed this with Svend-Erik Skaaning; the upstream fix is due in the next annual release (2027).

**Approach:** a `lexical_index.corrections.yml` (the standard mechanism for known upstream errors) overrides `male_suffrage_lied` / `female_suffrage_lied` to `0` for India, 1947–1949. It's applied on the raw suffrage columns right after `preprocess`, so the fix propagates automatically to every derived indicator — universal right to vote in practice, the three-tier classification, and all the `num_*` region counts used on the new democracy topic page.

An `expect: {eq: 1}` guard makes the step fail loudly once the 2027 release flips the source values, so the correction can't silently re-mangle fixed data.

### DIFF
--- a/etl/steps/data/garden/democracy/2026-04-02/lexical_index.corrections.yml
+++ b/etl/steps/data/garden/democracy/2026-04-02/lexical_index.corrections.yml
@@ -16,7 +16,7 @@
     Universal male suffrage in India is coded as beginning in 1947 (independence), but it began in
     1950 (Constitution in force). Confirmed with Svend-Erik Skaaning; fix due in the 2027 release.
   provider: Lexical Index
-  reported: 2026-07-09
+  reported: "2026-07-09"
   status: acknowledged
 - indicator: female_suffrage_lied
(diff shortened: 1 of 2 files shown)


lucasrodes:
@Marigold i tried your recently added corrections module for this case. This was [the brief](https://github.com/owid/owid-issues/issues/2301#issuecomment-4924264618). Letting you know for visibility on how I used the tool. Will likely merge by the end of Monday

PR #6439, opened 2026-07-14
📊 Update worldbank_wdi/2026-02-27/wdi to 2026-07-14

Marigold:
## Summary

Tracks: [owid/owid-issues#2449](https://github.com/owid/owid-issues/issues/2449)

Updates the World Bank World Development Indicators (WDI) dataset from `2026-02-27` to `2026-07-14`, per the [July 2026 WDI release](https://datatopics.worldbank.org/world-development-indicators/release-note/jul-2026.html), which highlights updates to national accounts data and new indicators on governance.

Key changes:
- **Found and fixed three indicator-upgrader bugs** (repo-wide, not WDI-specific) while investigating why this update's chart diffs looked unexpectedly noisy — split out into #6457 since none of that code is WDI-specific. One (a type mismatch in `map.columnSlug` handling) has silently left the map tab of upgraded charts pinned to the old variable since the function was introduced in 2022; confirmed 155 of this PR's 173 map-tab charts were affected before the fix.
- **Worldwide Governance Indicators (WGI) renamed and restructured** by the World Bank: the legacy 6-metric-per-dimension structure (estimate, percentile rank ± CI, number of sources, standard error) is replaced by a simplified 2-metric structure (raw estimate + 0–100 score) under new `GOV_WGI_*` codes. Per the [release note](https://datatopics.worldbank.org/world-development-indicators/release-note/jul-2026.html), the 2025 WGI edition also involved "enhanced data source screening, better indicator mapping, and a refined aggregation model," and the new 0–100 score uses fixed benchmark countries for consistency (replacing the old percentile-rank approach, which was always relative to the current set of countries reporting). 24 legacy codes retired, 12 new/renamed governance indicators added. No published chart used any of the retired codes (verified against production Datasette), so this is a clean retirement with no chart-remap concern.
- Several other minor indicator renames/additions from the source: 3 new Women, Business and the Law (WBL) indices, a GDP/consumption "price level index" pair replacing the old PPP-conversion-factor ratio (recomputed this cycle using GDP deflator and IMF CPI data, with Eurostat/OECD annual PPP estimates for those regions and ICP-extrapolated estimates elsewhere), 2 new consumption-per-capita PPP indicators, and 1 new COVID cause-of-death indicator. `SG.LAW.INDX` was dropped by the source entirely.
- Fixed a persistent flakiness in the snapshot's legacy per-indicator metadata fetch (api.worldbank.org 502s under high concurrency) — see the snapshot section below.
- Widespread value-level revisions across national-accounts, trade, and mortality indicators — all spot-checked as small, plausible revisions or previously-missing years now filled in (see garden section).

<details>
<summary>Snapshot</summary>

- `date_published` moved from `2026-01-28` to `2026-07-01` (from the WB metadata API's own `last_updated_date`); `citation_full` year bumped to match.
- The generic-download-URL workaround (WB's metadata API returns a versioned URL that lags the actual latest release) is still needed this cycle — verified the API still returns a stale versioned URL.
- **Fixed flaky legacy-metadata fetch**: the per-indicator legacy API call (`api.worldbank.org/v2/indicator/<code>`, ~1,498 calls) hit persistent 502s under the script's original 20-way parallel load — three consecutive runs each failed on a different indicator, even after adding HTTP-adapter-level retry/backoff. Lowered `MAX_WORKERS` to 6 and increased retry `total`/`backoff_factor`; the fetch then completed cleanly.
- One indicator (`SE.LPV.PRIM.SD`, "Primary school age children out-of-school (%)", which has real non-null data) disappeared from the legacy API's response this cycle (1516 → 1497 indicators) while still present in `WDISeries.csv`. Added a meadow-level fallback so indicators missing from the legacy API but present in `WDISeries.csv` get their core metadata (name/source/topic/unit) backfilled from the CSV instead of being silently dropped.
</details>

<details>
<summary>Garden</summary>

- Ran `update_wdi_metadata.py` (copied forward from the previous version folder — `etl update` doesn't carry over non-standard helper scripts) to reconcile `wdi.meta.yml` against the new indicator set: 39 deleted (36 legacy WGI + `pa_nus_pppc_rf` + `sg_law_indx`, plus a stale `population` entry that was never actually a final output column), 5 titles updated, 20 new indicators added.
- Extended the tool's `KEEP_PATTERNS` to cover garden-computed columns (`labor_force`, `number_employed`, etc., `ny_gdp_*_adjusted`, `*_ilo_modeling_comparison_absolute`) that don't exist in the raw meadow table and were being incorrectly flagged for metadata deletion every cycle (this recurred from the previous cycle's PR #5723, which fixed it by hand each time).
- Resolved 53 source citations missing from `wdi.sources.json` (mostly date-string churn in the raw source name; a handful of genuinely new DHS/GBD/ICP/WGI sources got hand-written citations following the existing naming convention).
- Adapted the curated `wdi.meta.override.yml` entry for the retired `pa_nus_pppc_rf` ("Price level ratio…") to its renamed successor `pa_nus_gdp_pli` ("Price level index (GDP)"), updating the ratio-based wording to the new base-100 index framing.
- Removed an obsolete typo-fix patch (`"threated"` → `"threatened"` in three species-threat indicators' producer descriptions) — the World Bank fixed the typo upstream this cycle; verified against the new `WDISeries.csv`.
- **Diffed against the published 2026-02-27 garden dataset** (not a local rebuild, to avoid conflating this update with upstream population/regions/income_groups drift): schema changes are fully explained by the known indicator churn above; value-level changes (1,091 of 1,503 common columns) are widespread but every spot check landed on small, plausible revisions or previously-missing years now filled in, consistent with the release note's "national accounts data" update. Full write-up in `workbench/wdi/garden_diff.md` (not included in this PR — internal working notes).
- Harmonization audit: clean. No missing/unused country mappings, one already-excluded WB aggregate (`Fragile and conflict affected situations`) went completely blank in the source this cycle (harmless — already outside our canonical set), and the only non-canonical garden output entity is OWID's own `World (excluding China)` custom aggregate (intentional). WB's `defined_by: wb` regions in `regions.yml` are unchanged (verified against `WDICountry.csv`'s `Region` column).

**Not covered in this PR**: `geo.add_regions_to_table` is still used at two call sites in the garden step (remittance and GDP-adjustment aggregations) — deprecated in favor of `paths.regions.add_aggregates`, but modernizing it needs a controlled A/B test given WDI's chart footprint, and this is a routine data refresh. Left for a follow-up cleanup PR.
</details>

## Indicator-upgrader bugs found and fixed

While reviewing why upgraded charts showed unexpectedly large config diffs, found and fixed three bugs in `etl/indicator_upgrade/` (repo-wide, not WDI-specific). **Moved to #6457** since none of that code touches WDI — summary here, full write-up there:

1. **Stale `map.columnSlug` on upgraded map charts (the significant one).** `update_chart_config_map` compared `map.columnSlug` (stored as a *string*) directly against `indicator_mapping`'s *int* keys, so the membership check always failed silently — the map tab kept pointing at the old variable even though `dimensions` (and every other view) upgraded correctly. This function dates to 2022 (#2558), so it has likely affected every past dataset upgrade with map-tab charts, not just this one. Confirmed on this PR: **155 of 173 WDI charts with a map tab** had a stale `columnSlug` after the (buggy) upgrade ran. Fixed by casting to `int` before the lookup.
2. **Inheritance-enabled charts keep every schema default forever (bloat, not a correctness bug).** #5911 fixed a real bug (schema-default overrides silently reverted on inheritance-enabled charts) by skipping the defaults-removal step entirely for those charts. That over-corrects: every schema default gets baked in permanently, even fields the chart never touched. Replaced the blanket skip with a real diff against the indicator's own resolved config, so only genuine overrides survive. A Codex review pass caught that the first version tracked "explicitly set" fields from the *resolved* config instead of the chart's actual stored `.patch`; fixed by threading the real `chart_configs.patch` through explicitly.
3. **Same bloat still slipped through for charts the fix above couldn't build a baseline for** (multi-dimension charts, or indicators with no `grapherConfigETL` at all). Fixed by using the `y`-dimension's own indicator config specifically, and by treating "no `grapherConfigETL`" as an explicit empty baseline rather than "unknown."

All three are covered by new regression tests, and verified end-to-end on staging: reset the affected charts to their clean production config, re-ran the (fixed) upgrade, and confirmed `chart_configs.patch` came out minimal in each case, with no functional/visual change.

**Remediation status**: staging was rebuilt from scratch partway through this PR's review, which gave a clean slate to re-run the fixed upgrade against — all ~480 WDI-affected charts (and their narrative-chart children) are now in a correct, verified state. The one gap: production Datasette only reliably covers ~80% of chart IDs for cross-checking, so about 90 charts on this branch couldn't be individually diffed against production; spot checks of the reachable ones found no other stale/bloated cases beyond the ones already described and fixed above.

## Downstream dependencies (follow-up PR)

The following still depend on `worldbank_wdi/2026-02-27` and are intentionally **not** repointed in this PR (mirrors the historical two-PR pattern for foundational-dataset updates — chain update now, bulk-bump separately):

- `data://grapher/health/2024-04-22/gpei_funding`
- `export://multidim/worldbank_wdi/latest/ilostat_comparison`
- `export://multidim/worldbank_wdi/latest/labor_force_participation_rate`
- `data://grapher/labor/2026-02-17/female_labor_force_participation_omm`
- `data://grapher/covid/latest/compact`
- `data://grapher/sipri/2026-04-27/military_expenditure`
- `data://garden/growth/2025-01-16/gdppc_vs_living_conditions`
- `data://garden/emissions/2026-03-05/gdp_and_co2_decoupling`
- `data://external/owid_grapher/latest/int_dollar_conversions`
- `data://garden/emdat/2026-04-30/natural_disasters` (+ its 4 grapher variants, explorer, and 3 multidim exports)

None of these need chart remapping — they keep their own version and variable IDs; only their *dependency* on WDI changes, so their recomputed aggregates will show up in Chart Diff once they're bumped.

## Harmonization audit

<details>
<summary>Details</summary>

- **Missing in mapping**: none.
- **Unused mappings**: none.
- **Unknown excluded entries**: `Fragile and conflict affected situations` — a WB operational-classification aggregate that went from 27,712 non-null values (Feb 2026) to zero across all indicators and years in this release. Already excluded from OWID's WDI output, so no impact; flagged only as a heads-up in case it's a transient WB-side gap.
- **Targets not in canonical regions/income groups**: none.
- **Excluded entries matching canonical regions**: none.
- **Garden output entities not in canonical regions/income groups**: `World (excluding China)` — an OWID-computed custom aggregate for poverty comparisons, not a bug.
</details>

## Sanity-check findings

The garden step has no dedicated `sanity_check_inputs`/`sanity_check_outputs` functions (pre-existing — this is a long-standing legacy step). The snapshot's own inline citation/license asserts passed on the new data.

## Not covered in this PR: known follow-ups

The indicator upgrade + a catch-all scan turned up items unrelated to (or only tangentially caused by) this update — flagging here so a future update doesn't have to rediscover them:

- **9 narrative charts (12 stale pins)** still reference WDI variables from versions several cycles old (`2024-05-20`, `2025-01-24`, `2025-09-08` — none from the `2026-02-27` this PR bumps from). Each needs remapping to its **parent chart's current equivalent indicator** (not a straight short_name match — some are legacy pre-dimensional variables with no modern twin). Affected: `men-smoke-more-than-women` (narrative chart 1), `gdp-per-capita-in-international-and-market-dollars-2022` (17), `richer-countries-tend-to-have-higher-self-reported-life-satisfaction` (19), `change-maternal-mortality-gdp-maldives` (22), `south-korea-stands-out-in-research-funding-and-patents` (94), `unemployment-rate-spain-di` (96), `electricity-reaches-less-than-one-in-four-people-in-some-sub-saharan-african-countries` (112), `anemia-particularly-low-income-countries-di` (120, 3 pins). Fix via `push_new_narrative_charts_cli` per parent chart (see the `update-dataset` skill's step 7 narrative-chart guidance).
- **Several narrative charts have stale FAUST overrides** (subtitle/note text that's a near-identical but outdated copy of the parent's current text, e.g. "living costs" vs "cost of living" wording drift, a tolerance override that no longer matches the parent) that the upgrader flagged but — correctly — did not auto-fix, since FAUST changes need editorial sign-off. Left as-is per the same reasoning; a content reviewer should confirm each reset before it's applied.
- **`pa_nus_pppc_rf` → `pa_nus_gdp_pli` and `pa_nus_prvt_pp` → `pa_nus_prvt_pli` are not a like-for-like rename** — despite matching titles suggesting otherwise, the Bank replaced a currency-conversion ratio (LCU per international $, centered around 1) with a base-100 price-level index (a different quantity, different scale). The indicator mapping initially treated this as a rename, which would have silently changed 4 charts' displayed metric while keeping their old ratio-framed title/subtitle/units/map-bins (`gdp-price-levels-relative-to-the-us`, `price-levels-relative-to-us-vs-gdp-per-capita`, `higher-labor-productivity-and-price-levels-in-richer-and-poorer-countries`, `ppp-conversion-factor-for-private-consumption`). Rejected in Chart Diff instead — those 4 charts stay on their current `2026-02-27` config for now. They'll need either a proper price-level-index rewrite (title/subtitle/unit/map-bins) or retirement once someone has the editorial bandwidth; until then they're frozen (no future data) and add another reason `2026-02-27` can't yet be archived, on top of the downstream-dependency bump above.

All three predate or are tangential to this update and don't block it.


Marigold:
@codex review

etl/steps/data/garden/worldbank_wdi/2026-07-14/wdi.py
```diff
@@ -0,0 +1,1049 @@
...
+    tax_rev_share_gdp_code = "gc_tax_totl_gd_zs"
+    gdp_percap_code = "ny_gdp_pcap_pp_cd"
+    omm_tax_rev_code = "omm_tax_rev_percap"
+    assert (
+        df[tax_rev_share_gdp_code].min() >= 0.0
+        and df[tax_rev_share_gdp_code].quantile(0.995)
+        <= 100.0  # timor has > 100 tax revenue as a % of GDP for 2010-2012
+        and df[tax_rev_share_gdp_code].max() > 1.0
+    )
+    tb_omm[omm_tax_rev_code] = df[gdp_percap_code].multiply(df[tax_rev_share_gdp_code].divide(100)).round(2)
```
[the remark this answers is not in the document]
Marigold: > _Written by Claude Sonnet 5 — @Marigold at the wheel._

Good catch on flagging this, but the underlying data is clean in this release — verified directly:

- `gc_tax_totl_gd_zs` min value across all 5,688 rows is 0.0407 (no rows below 0.001, none in the 1e-8/1e-9 range the note described).
- The specific example the note cited (US ~1e-08 instead of ~11%) doesn't reproduce: US shows 8.6–12.6% across 2010–2024 in the actual downloaded data.
- Checked the *previous* published version (2026-02-27) too, for comparison — same clean values, no corruption there either.

So the note (introduced 2026-02-02 during the 2025-09-08 update cycle, per `git log -S`) was stale — whatever caused the corruption back then was already fixed upstream, and it got carried forward unverified through three more update cycles. Removed it in ad8c266ce9 rather than adding quarantine logic for a problem that isn't present in the current data.

Marigold:
@codex review

etl/indicator_upgrade/indicator_update.py
```diff
@@ -146,22 +158,50 @@ def __init__(
...
+            metadata, used to strip the same kind of redundant defaults from
+            `dimensions[*].display` -- a separate inheritance path from `indicator_config`
+            above (see `prune_dimension_displays`). Applied regardless of
+            `is_inheritance_enabled`, since dimension display always inherits from the
+            variable's own display metadata.
+        """
         # Fix errors in schema
         config_new = fix_errors_in_schema(config)
+        # Remember which paths were explicitly present before defaults get filled in below.
+        original_config = config_new
```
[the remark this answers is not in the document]
Marigold: > _Written by Claude Sonnet 5 — @Marigold at the wheel._

Confirmed, good catch — `chart.config` is built from `chart_configs.full` (the resolved config, inherited values merged in), not `.patch`. Tracking "explicit" fields from it meant an inherited-but-untouched field would look explicit and get pinned as a fake override if it happened to differ from the new indicator's value.

Fixed by threading the chart's actual `chart_configs.patch` through as a new `original_patch` parameter (`ChartIndicatorUpdater.run`, `update_chart_config`, `_update_single_chart`), used for explicit-field tracking instead of the resolved full config. Added a regression test (`test_inheritance_pruning_uses_patch_not_full_config`) that reproduces exactly this scenario — an inherited title the chart never set, differing between old and new indicator — and verified end-to-end again on the real chart (755) on staging. Fixed in b304840b22.

Marigold:
@paarriagadap could you take a quick look? I rejected some charts from this update because WDI changed their definitions (I'll address them in the subsequent PR). I've also added a couple of improvements/ bugfixes to data-diff and indicator upgrader (in different PRs).

snapshots/worldbank_wdi/2026-07-14/wdi.zip.dvc
```diff
@@ -0,0 +1,19 @@
...
+    citation_full: World Development Indicators (WDI), The World Bank (2026).
+    version_producer: '125'
```
paarriagadap: > _Written by Claude Fable 5 — @paarriagadap at the wheel._

`version_producer: '125'` is stale — it was carried over verbatim from the February `.dvc`, but the WB DDH API reports the current release as **129** (`reference_system/version`, production date 2026-07-01).

Since `update_snapshot_metadata()` already parses that same API response, it could set this field automatically (from `meta_orig["reference_system"]["version"]`) so it can't go stale on future updates.

snapshots/worldbank_wdi/2026-07-14/wdi.py
```diff
@@ -0,0 +1,278 @@
...
+    )
+
+    # Sanity checks.
+    error = "Citation has changed. Review the new citation and adjust the code."
+    assert meta_orig["identification"]["citation"] == "World Development Indicators, The World Bank", error
+    error = "License has changed. Review the new license and adjust the code."
+    assert meta_orig["constraints"]["license"]["license_id"] == "Creative Commons Attribution 4.0", error
+
+
+if __name__ == "__main__":
```
paarriagadap: > _Written by Claude Fable 5 — @paarriagadap at the wheel._

Outdated `__main__` block — `etls` invokes the module's click command directly (`run_snapshot_script` in `etl/snapshot_command.py` checks `hasattr(module, "main")` before falling back to `run`), so these two lines are dead code and can be deleted.

etl/steps/data/garden/worldbank_wdi/2026-07-14/wdi.meta.yml
```diff
@@ -0,0 +1,5257 @@
+# WARNING: Do NOT use this file to manually curate WDI indicators (since this file is generated automatically). Use the wdi.meta.override.yml file instead.
+dataset:
+  update_period_days: 365
```
paarriagadap: I think it's more like 180

paarriagadap (changes requested):

Hi @Marigold! I want to request changes — none is critical. Two small fixes are flagged inline (stale `version_producer`, dead `__main__` block).

## Empty entities

Audited every surface using these indicators (charts, MDim views, explorers, narrative charts, article embeds with `?country=` overrides) against the entities that actually have data, on staging and production. **No case was introduced by this PR.** I fixed the empty selections on the affected published charts (pending approval in [chart-diff](http://staging-site-Marigold-wdi-update/etl/wizard/chart-diff)):

| Chart | Fix |
|---|---|
| [air-passengers-carried](http://staging-site-Marigold-wdi-update/admin/charts/1111/edit) | selected the WB regions |
| [healthcare-access-quality-un](http://staging-site-Marigold-wdi-update/admin/charts/4544/edit) | selected the WB regions |
| [community-health-workers](http://staging-site-Marigold-wdi-update/admin/charts/4545/edit) | selected a country set |
| [oil-rents-as-a-share-of-gdp-draft-8498](http://staging-site-Marigold-wdi-update/admin/charts/8498/edit) | left empty (unpublished draft) |

One **article citation opens a fully empty view** — pre-existing on production, so content follow-up, not this PR: [childhood-stunting-fell-dramatically-over-the-20th-century](<https://ourworldindata.org/childhood-stunting-fell-dramatically-over-the-20th-century#:~:text=One%20in%20four%20children%20in%20the%20world%20today%20suffers%20from%20stunting.%20That%E2%80%99s%20150%20million%20children%20under%20five%20years%20old.>) links [share-of-children-younger-than-5-who-suffer-from-stunting](http://staging-site-Marigold-wdi-update/grapher/share-of-children-younger-than-5-who-suffer-from-stunting) with only `World` selected (×2 links), but the WB publishes no World aggregate for this indicator — the chart opens empty. A handful of other article links pin an entity without data (mostly unsuffixed WB region names vs the `(WB)`-suffixed entities), but their remaining selection still renders, so I've left those out.

## Stale pins (archival blockers, not for this PR)

- **Narrative charts**: 13 pins are on the new `2026-07-14` (remediation worked); 8 remain on `2024-05-20` / `2025-01-24`, matching the follow-up in the PR body. `/update-dataset` step 7 already documents the fix (`push_new_narrative_charts_cli` per parent chart).
- **Migration explorer**: pins 3 remittance indicators (`bx_trf_pwkr_dt_gd_zs`, `si_rmt_cost_ob_zs`, `si_rmt_cost_ib_zs`) to `worldbank_wdi/2025-09-08` — four cycles old, invisible to DAG greps. Add to the "can't archive old WDI versions yet" list.

## Suggestions

1. The "Downstream dependencies" list misses 5 consumers still on `2026-02-27`: `growth/2024-05-16/gdp_historical`, `growth/2026-07-02/structural_transformation_omm`, `malnutrition/2024-12-16/malnutrition`, `technology/2024-12-23/internet`, `aviation_safety_network/2024-06-05/aviation_statistics`.
2. Track the deferred cleanup: `geo.add_regions_to_table` ×2, plus `pd.concat` (wdi.py:78) and `pd.DataFrame(table)` (wdi.py:249) → `pr.*` equivalents.
3. The `.dvc` `description` is missing spaces after sentence ends ("decades.WDI serves…"). It's regenerated by the snapshot script via `BeautifulSoup.get_text()`, so fix it there (e.g. `get_text(separator=" ")` + whitespace normalization) or it re-breaks every update.

snapshots/worldbank_wdi/2026-07-14/wdi.zip.dvc
Marigold: > _Written by Claude Sonnet 5 — @Marigold at the wheel._

Fixed in b9207c9bb1 — `update_snapshot_metadata()` now sets `version_producer` from `meta_orig["reference_system"]["version"]`, same as the other origin fields it already updates each run, and the `.dvc` reflects the current `129`. Verified against the live API.

snapshots/worldbank_wdi/2026-07-14/wdi.py
Marigold: > _Written by Claude Sonnet 5 — @Marigold at the wheel._

Removed in b9207c9bb1.

etl/steps/data/garden/worldbank_wdi/2026-07-14/wdi.meta.yml
Marigold: > _Written by Claude Sonnet 5 — @Marigold at the wheel._

Agreed — checked `date_published` across the last 4 releases (2024-05-20, 2025-01-15, 2025-07-03, 2026-01-28, 2026-07-01), gaps average ~190 days. Changed to 180 in b9207c9bb1.

Marigold:
@paarriagadap thanks for review! I'll check if we have more empty charts like that and address your suggestions and downstream deps in the next PR.

PR #6450, opened 2026-07-15
📜 Clone owid-issues when no local checkout exists for the scheduled-issue check

paarriagadap:

## Summary

Follow-up to #6448, addressing a Codex comment that landed just after the merge: the scheduled-issue workflow check (step 6d of `/update-dataset`) told the reader to fall back to `gh api repos/owid/owid-issues/contents/.github/workflows --jq '.[].name'` when no local `~/owid-issues` checkout exists — but that only lists **filenames**, while every subsequent check needs file **contents** (content grep for group workflows whose filename doesn't mention the dataset, plus `cron`/`body`/`assignees` parsing), and the commit step needs a working tree regardless.

The fallback is now to clone the repo (`gh repo clone owid/owid-issues ~/owid-issues`), which makes all later instructions work identically in a fresh environment. The reviewer-side mirror in `/review-data-pr` §13 had the same flaw and is fixed the same way.


paarriagadap:
@codex review

PR #6459, opened 2026-07-16
🐛 Preserve nested-only DAG definitions when updating parent steps

### LINKED ISSUE #6165: Nested-only step definitions can be silently wiped when a parent step is updated

## Problem

The nested DAG format lets a step be defined only as a nested dependency under a parent (no top-level entry). When the parent's dependency list is later rewritten — for example by \`etl update --include-usages\`, or by a manual edit — the nested child's definition is lost along with it. The version tracker never gets to flag the old step as \`UNUSED → ARCHIVABLE\`, and \`etl archive\` never gets a chance to move it to the archive DAG.

## Example

Starting DAG (step defined only nested):

```yaml
steps:
  export://multidim/foo/latest/bar:
    - data://grapher/foo/2025-01-01/bar:    # only definition (nested)
      - data://garden/foo/2025-01-01/bar    # only definition (nested)
```

Run:

```
etl update data://grapher/foo/2025-01-01/bar --include-usages
```

Result:

```yaml
steps:
  data://garden/foo/2026-01-01/bar: [...]   # new top-level
  data://grapher/foo/2026-01-01/bar:        # new top-level
    - data://garden/foo/2026-01-01/bar
  export://multidim/foo/latest/bar:
    - data://grapher/foo/2026-01-01/bar     # dep pointer replaced
```

The two \`2025-01-01\` entries are gone — wiped, not archived.

## Why it happens

[\`_write_to_nested_dag_file\`](etl/dag_helpers.py) replaces the entire dependency list at the found location: \`mapping[key] = _build_dependency_sequence(...)\`. When the export step's deps get rewritten, anything nested inside that subtree is overwritten — including step definitions that lived nowhere else.

pabloarosado:

Fixes #6165

## Problem

The nested DAG format lets a step be defined only as a nested dependency under a parent (with no top-level entry). When the parent's dependency list was rewritten (for example by `etl update --include-usages`), `_write_to_nested_dag_file` replaced the entire dependency subtree, so any step definitions that lived only inside it were silently wiped. The version tracker never got a chance to flag the old steps as unused/archivable.

## Fix

Before replacing a step's dependency list, the writer now collects the nested definitions inside it. After all updates are applied, any collected definition that is no longer declared anywhere in the document is re-declared as a top-level entry (with its own nested subtree intact).

Two related refinements in the same code path:

- The wiped definitions are removed from the `declared_steps` set before building the replacement sequence, so a child updated in the same call keeps its nested place under the parent instead of being flattened.
- If a displaced step was itself part of the update (its declaration was wiped before it was looked up), its **new** dependency list is written rather than the stale one.

## Example

Updating the DAG from the issue now produces:

```yaml
steps:
  export://multidim/foo/latest/bar:
    - data://grapher/foo/2026-01-01/bar:
      - data://garden/foo/2026-01-01/bar:
        - snapshot://foo/2026-01-01/bar.csv
  data://grapher/foo/2025-01-01/bar:
    - data://garden/foo/2025-01-01/bar:
      - snapshot://foo/2025-01-01/bar.csv
```

The old `2025-01-01` chain survives as a top-level entry, so it can later be flagged `UNUSED → ARCHIVABLE` and archived through the normal procedure.

## Tests

Three new tests in `tests/test_dag_utils.py` cover the issue scenario, a parent update that keeps an unchanged nested child, and the no-duplication case where the child is re-declared in place.

### DIFF
--- a/etl/dag_helpers.py
+++ b/etl/dag_helpers.py
@@ -422,11 +422,21 @@ def _write_to_nested_dag_file(dag_file: Path, dag_part: dict[str, Any], comments
     original_steps = set(_parse_dag_yaml(doc))
     declared_steps = set(original_steps)
 
+    # Replacing a dependency list below wipes any nested definitions that lived inside it. Those
+    # nested definitions may be the only place a step is declared, so collect them here and
+    # re-declare the lost ones at the top level after all updates (see issue #6165).
+    displaced: list[tuple[str, Any]] = []
+
     for step, dependencies in dag_part.items():
         node = _find_step_mapping(steps, step)
         if node is None:
             continue
         mapping, key = node
+        displaced_here = list(_iter_nested_dep_items(mapping[key]))
+        displaced.extend(displaced_here)
+        # The wiped definitions are no longer declared, so the new sequence is free to re-nest them
+        # (e.g. a child that is updated in this same call keeps its nested place under this step).
+        declared_steps -= {sub_step for sub_step, _ in displaced_here}
         mapping[key] = _build_dependency_sequence(dependencies, dag_part, declared_steps, {step})
         declared_steps.update(_collect_declared_steps(mapping[key]))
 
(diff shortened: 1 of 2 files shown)


pabloarosado:
@codex review

PR #5869, opened 2026-03-31
📊 Expand indicator paths for x, size, color dimensions

Marigold:
## Closes #5665

Expand indicator paths for `x`, `size`, and `color` dimensions — previously only `y` was supported.

### What changed

**1. `expand_config()` / `create_collection()` now accept `additional_indicators`**

Adds `x`/`size`/`color` indicators to every auto-generated view during expansion:

```python
config = expand_config(
    tb=tb,
    indicator_names="deaths",
    additional_indicators={
        "size": "historical#population_historical",
        "color": "regions#owid_region",
    },
)
```

Short paths are expanded using the same logic as `y` indicators; paths with `#` or full catalog paths are kept as-is.

**2. `Collection.save()` now auto-expands short paths added after creation**

Previously, path expansion only ran once during `create_collection()`. If you added `x`/`size`/`color` indicators *after* that (e.g. after `group_views`), you had to hardcode full catalog paths. Now `save()` re-runs `expand_paths()` before validation — already-complete paths are safely skipped.

This enables the pattern from the issue:
```python
# Before — hardcoded full paths:
view.indicators.size = Indicator(
    catalogPath="grapher/demography/2024-07-15/population/historical#population_historical"
)
view.indicators.color = Indicator(
    catalogPath="grapher/regions/2023-01-01/regions/regions#owid_region"
)

# After — short paths, auto-expanded at save():
view.indicators.set_indicator(
    size="historical#population_historical",
    color="regions#owid_region",
)
```

**3. Updated real-world steps to use short paths**

- `incomes_wid.py` — the exact example from the issue
- `dummy.py` — also used a full path for `color`

### Testing

- 6 new tests covering full paths, table#indicator paths, short slugs, all `expand_path_mode` values, invalid axis validation, and combined with `common_view_config`
- All 137 collection tests pass (2 pre-existing integration failures unrelated to this change)

PR #6461, opened 2026-07-16
✨ Support x, size, and color indicators in expand_config

### LINKED ISSUE #5665: MDIMs: Automatically expand indicator paths of dimensions other than `y`
## One-liner
Expand the indicator paths of dimensions `x`, `size`, and `color` (currently only `y` is supported)

This would prevent us hardcoding complete ETL paths as in:
https://github.com/owid/etl/blob/c5e25da669df0ea3e6a9064cfe1d0ec0231d45f6/etl/steps/export/multidim/wid/latest/incomes_wid.py#L233-L236

I _think_ that what needs to change is the logic in `build_views` and the calls to `_expand_indicator_path`.

https://github.com/owid/etl/blob/811f439ddd8e35e959d0e2a7e28f3c04db125e0f/etl/collection/core/expand.py#L308-L315

_ranked as 4 in importance/annoyance (1-5 scale), from work in https://github.com/owid/etl/pull/5593_

pabloarosado:

Fixes #5665

## What this does

`expand_config` could only generate views with a `y` indicator, so building scatter views meant hand-editing views afterwards and hardcoding complete ETL paths for `x`, `size`, and `color` (see the example in the issue, from `incomes_wid.py`).

This PR adds three new parameters to `expand_config` (and `CollectionConfigExpander.build_views`): `indicator_x`, `indicator_size`, and `indicator_color`. Each one is applied to every generated view and accepts either:

- **a column of the table** — its path is expanded exactly like the `y` indicator paths, following `expand_path_mode` (this removes the hardcoding), or
- **a catalog path** (any value containing `#`) — used as given; short forms like `table#indicator` are still resolved against the collection dependencies at save time, since `ViewIndicators.expand_paths` already supports all chart dimensions.

A value that is neither (no `#` and not a column of the table) fails with a clear assertion message.

## Example

```python
config = expand_config(
    tb=tb,
    indicator_names="disposable_income",
    indicator_x="market_income",             # column of tb -> expanded automatically
    indicator_size="population#population",  # short catalog path -> resolved at save time
    common_view_config={"chartTypes": ["ScatterPlot"]},
)
```

## Notes

- The same indicator is applied to all generated views. Per-view-varying `x`/`size`/`color` (e.g. an x-indicator that follows each view's dimensions) is left as a future improvement, noted in the docstring.
- Not plumbed through `create_collection` / `paths.create_collection` in this PR: those handle lists of tables with per-table parameter distribution, which would grow the API surface considerably. Happy to add it as a follow-up if it turns out to be useful there too.

## Tests

Four new tests in `tests/collections/test_expand_config.py`: expansion from table columns, catalog paths used as given, `expand_path_mode` being honored, and the error case.

### DIFF
--- a/etl/collection/core/expand.py
+++ b/etl/collection/core/expand.py
@@ -15,6 +15,9 @@ def expand_config(
     indicator_as_dimension: bool = False,
     indicators_slug: str | None = None,
     expand_path_mode: Literal["table", "dataset", "full"] = "table",
+    indicator_x: str | None = None,
+    indicator_size: str | None = None,
+    indicator_color: str | None = None,
 ) -> dict[str, Any]:
     """Create partial config (dimensions and views) from multi-dimensional indicator in table `tb`.
 
@@ -48,7 +51,8 @@ def expand_config(
         - Out-of-the box sorting for dimension values.
             - Example: This could be alphabetically ascending or descending, or numerically ascending or descending.
             - IDEA: We could pass strings as values directly to the keys in dimensions dictionary, e.g. `dimensions={"sex": "alph_desc", "age": "numerical_desc", "cause": ["aids", "cancer"]}`. To some extent, we already support the function "*" (i.e. show all values without sorting).
-        - Support using charts with 'x', 'size' and 'color' indicators. Also support display settings for each indicator.
+        - Support display settings for each indicator.
+        - Support 'x', 'size' and 'color' indicators that vary per view (currently the same indicator is applied to all views).
 
     Parameters:
     -----------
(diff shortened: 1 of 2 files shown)


pabloarosado:
@codex review

pabloarosado:
@codex review

pabloarosado:

Closing without merging.

The three new parameters only exist on `expand_config`, which just two steps in the repo call directly; everything else builds collections through `paths.create_collection`, where they were deliberately not plumbed. And the case they cover (the same `x`/`size`/`color` on every generated view) is not a case any collection has, while the case #5665 was actually about (a subset of views, set after creation) already works with short-form paths: `Collection.save()` expands them for every chart dimension, so no version is ever hardcoded. Details in https://github.com/owid/etl/issues/5665#issuecomment-5147529874.

The last place in the repo that still spelled out full paths for these dimensions was `incomes_pip.py`, fixed in #6584 with a verified no-op on the generated config.

If we do want expand-time defaults later, #5869 is the better starting point: it plumbs `create_collection` and migrated its call sites.

PR #6462, opened 2026-07-16
✨ Add helper to filter a table to a single dimension value and drop the dimension

### LINKED ISSUE #5670: MDIMs: Keeping single dimension value, and dropping dimension in table metadata
## One-liner
In some cases, we want to just use a subset of columns from a given table `tb` that correspond to a certain dimension (e.g., `sex="female"`). Provide tools/functions to easily filter table columns and drop dimension attributes.

## Context
Sometimes we might only be interested in a specific dimension value (e.g., `sex="female"`), so we want to use only the indicators in the table that concern that dimension. In such cases, we also want to "drop the dimension entirely" so that it is not rendered in the MDIM. That is, in our example, we wouldn't want a dropdown `sex` that only has a single option "female".

At the moment, filtering and dropping these dimensions has to be done manually before calling `paths.create_collection`:

https://github.com/owid/etl/blob/b9abfcd6303e6ce4b87610ec4b3b94fef21d28b4/etl/steps/export/multidim/lis/latest/incomes_lis.py#L30-L40

If this is not done, the code would fail in the example above because a dimension is not defined in the YAML. A quick fix would then be to actually define the dimension in the YAML (which already contains some manually created views)... But that seems like a bad idea, as we'd have to define the dimension in multiple places only to be dropped.

## Possible solutions
1. We could have this logic in `paths.create_collection`, where if for a dimension only one choice is given, that dimension is completely dropped from the MDIM config. That would mean that there is no need of "dummy-defining" the dimension in the YAML. I'm unsure if this is easily implementable, or if there is some technical challenge like a strong assumption in the code that dimensions given should match those from YAML. Let's review.
    - **Good**: User doesn't need to run any other method, and just uses out of the box `paths.create_collection`, as usual.
    - **Bad**: A bit riskier, in that it might affect other MDIMs config. We would need to define some comparison tools to compare pre and after change MDIM configs.
3. Implement a function `drop_dimension_keeping_single_value(tb: Table, dimension_name: str, dimension_choice: str)`, which for a given table `tb` it only keeps columns (indicators) that have dimension `dimension_name` with value `dimension_choice`. Then, it drops the `dimension` attribute from indicators' metadata. This function is imported and manually called before `paths.create_collection`
    - **Good**: No need to dig into `paths.create_colelction`, so low risk (low change of corrupting other MDIMs)
    - **Bad**: User needs to know where the function is and import it. Might be tricky to remember it exists.

pabloarosado:

Fixes #5670

## Why it is needed

Several collections use only one choice of a dimension (`equivalence_scale="square root"`, `ppp_version=2021`) and must drop that dimension, otherwise the MDIM renders a dropdown with a single option. Today each step does it with a hand-rolled loop over the table's columns.

`Collection.save` does prune single-choice dimensions, but that is too late here: the table has several choices of the dimension, so `expand_config` builds a view per choice and the dimension never becomes single-choice in the first place.

Filtering through `dimensions={"equivalence_scale": ["square root"]}` instead does not work when the collection's YAML also has hand-written views: those views do not carry the dimension, so creation fails before the single-choice dimension is ever pruned.

```
AssertionError: Dimension equivalence_scale not found in dimensions! View:
dimensions:
  indicator: share
  decile: '10_40_50'
  period: nan
  welfare_type: dhi
```

## What this does

Adds `filter_columns_by_dimension_choices(tb, dimension_choices, drop_single_choice_dimensions=True)` in `etl/collection/utils.py`, exported from `etl.collection`:

```python
tb = filter_columns_by_dimension_choices(tb, {"equivalence_scale": "square root"})
tb = filter_columns_by_dimension_choices(tb, {"sex": ["female", "male"], "age": "adults"})
```

It keeps the indicators whose dimensions match every given `{dimension: choice | [choices]}` entry, removes from the metadata (column-level and table-level) every dimension left with a single choice, works on a copy so the caller's table is untouched, and raises a clear error when a dimension or a choice is missing, listing the available choices.

`drop_single_choice_dimensions=False` filters the columns without touching the dimensions.

This is option 2 from the issue. Option 1 (pruning single-choice dimensions inside `create_collection`) was not taken: it would change what steps see between `create_collection` and `save`, where a lot of code keys on dimensions (`view.matches`, `group_views`, `get_choice_names`), so a dimension disappearing early would silently match nothing instead of failing.

## Steps migrated

`incomes_lis.py` (the example in the issue), plus `incomes_pip.py` and `poverty_pip.py`, which filter to a single `ppp_version` with the same hand-rolled loop.

Note for future migrations: the helper keeps columns that do not carry dimensional metadata at all, including `country`/`year`, while the loops it replaces dropped them. Loops that assume every column has dimension metadata need a `None` guard, as done here for the `survey_comparability` loops in the two PIP steps.

## Verification

Ran the three export steps before and after and compared the generated collection configs (`export/multidim/.../<short_name>.config.json`), normalizing two pre-existing sources of non-determinism (the y-indicator order in grouped views and the order of the `dependencies` list):

| collection | views | result |
|---|---|---|
| `incomes_lis` | 91 | identical |
| `incomes_pip` | 57 | identical |
| `poverty_pip` | 44 | identical |

Dimensions are unchanged in all three. `pytest tests/collections` passes.

`drop_single_choice_dimensions` drops every dimension left with a single choice, not only the filtered one, so its blast radius was checked against the three real tables' published metadata: after filtering, the only single-choice dimension in each is the one being filtered (`equivalence_scale` in LIS, `ppp_version` in both PIP tables), so nothing else is dropped and the configs above still hold.

## Still open

- **Proposed** — @lucasrodes suggested moving the helper into `owid-catalog` as a `Table` method. Deferred: every caller today is a collection export step and `lib/catalog` has a slower release loop. Happy to move it in this PR if preferred.
- **Proposed** — @lucasrodes also suggested consolidating with `expand_config`, whose `dimensions` argument filters by dimension choices too. Deferred to a follow-up: it filters at view level in the emitted config rather than at table level, never drops the dimension, and raises if any dimension is missing from the dict, so sharing the logic means refactoring `CollectionConfigExpander`.
- **Declined** — Codex asked the helper to reject a table whose indicators end up with no dimensions at all (filtering to one choice of *every* dimension). Not taken: narrowing a table to a single indicator is deliberate and tested behaviour, and `drop_single_choice_dimensions=False` covers callers who need the dimensions kept. The `expand_config` consequence is documented in the docstring instead.
- **Unverified** — the three export steps were not re-run after the merge from master, so the config-identity table above still dates from July. `Collection.save` validates indicators against this branch's staging DB, which no longer exists, so the steps cannot complete locally. Verified instead at the level the helper actually controls: for all three real tables the filtered output (columns, per-column dimensions, table-level dimensions) is identical to the pre-fix helper, so the July comparison carries over. A staging run is still needed to confirm end to end.
- **Not migrated** — `gini_pip.py` and `palma_ratio_pip.py` still hand-roll the same loop. Left out to keep this PR scoped; worth a follow-up so all five call sites agree.


pabloarosado:
@codex review

etl/collection/utils.py
```diff
@@ -152,6 +152,58 @@ def get_complete_dimensions_filter(
...
+
+        >>> tb = drop_dimension_keeping_single_value(tb, dimension="equivalence_scale", value="square root")
+
+    """
+    columns_keep = []
+    values_found = set()
+    for column in tb.columns:
+        dims = tb[column].m.dimensions
+        if not dims or dimension not in dims:
+            columns_keep.append(column)
```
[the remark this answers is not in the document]
pabloarosado: > _Written by Claude Opus 5 — @pabloarosado at the wheel._

Valid, fixed in eda4d7244. The helper now separates the two cases: columns with no dimensional metadata at all (`country`, `year`) are still kept, while an indicator that has dimensions but not the one being filtered raises, listing the offending columns, since it cannot be filtered and keeping it would add views for values the caller excluded.

None of the three migrated tables has such a column (LIS incomes, PIP incomes, PIP poverty: 0 each), so their generated configs are unchanged, verified by re-running the export steps.

etl/collection/utils.py
```diff
@@ -152,6 +152,70 @@ def get_complete_dimensions_filter(
     return expand_combinations(dimensions_filter_complete)
 
 
+# steps (public API, exported from etl.collection)
+def drop_dimension_keeping_single_value(tb: Table, dimension: str, value: Any) -> Table:
```
lucasrodes: I'd suggest using `dimension_choice` over `value`. Also, if we can define a bit better the typing for this argument instead of using `Any`, that'd be great. I suppose it's `str`?

etl/collection/utils.py
```diff
@@ -152,6 +152,70 @@ def get_complete_dimensions_filter(
...
+
+    An indicator that has dimensions but not this one cannot be filtered, and keeping it would add
+    views that the caller did not ask for, so it raises instead. Drop those columns beforehand if
+    they are not wanted.
+
+    The input table is not modified; a copy is returned.
+
+    Example:
+
+        >>> tb = drop_dimension_keeping_single_value(tb, dimension="equivalence_scale", value="square root")
```
lucasrodes: I find the description a bit confusing. Maybe an example on input vs output would be useful to help understand what's going on?

I think my struggle also comes from the function doing two things at once: (i) filtering columns and keeping only those that match `dimension='value'`, and (ii) then dropping that dimension because it's 'useless'.

lucasrodes (commented):
Thanks for implementing this, Pablo! I've left some minor comments.

It's been some time since I last looked into this, and I'm not actually sure why this was needed when there is single-dimension pruning already happening on `Collection.save`. I suppose it's good to filter earlier as it makes data wrangling easier? Anyway, I suppose the motivation makes sense.

Two other things I wonder is about function signature and its discoverability.

### Function signature ー name
`drop_dimension_keeping_single_value` feels a bit cryptic (although I recognize that it was me that suggested something like that in #5670). I would suggest maybe `filter_columns_by_dimension_choice`.

### Function signature ー arguments
I would probably change what its arguments mean. And possibly make the function a bit more generic.

I would allow for filtering by multiple dimension choices (e.g. `"sex" = ["female", "male"]), and add a flag to tell the function to drop dimensions with single choices. So, it could be something like:

```python
def filter_columns_by_dimension_choice(
    dimension_choices: dict[str, str | List[str],
    drop_single_choice_dimensions: bool = True
)
```

Examples for `dimension_choices`:

```python
{
  "sex": "female",
}
```
or
```python
{
  "sex": "female",
  "age": "adults",
}
```
or
```python
{
  "sex": ["female", "male"],
}
```

### Function discoverability
I can see how this method transcends `etl.collection`, and can be placed elsewhere as it is more of 'table'-utils. 

I would even argue that this method could be part of `owid-catalog`, so we could then call it as `tb.filter_by_dim_choices(...)` or sth like that.

### Final comment
I realize that with my suggestion, we would slightly converge to what [expand_config](https://github.com/owid/etl/blob/04a105276e26560034828df1146bb6fa76fa0ef7/etl/collection/core/expand.py#L10) does, where we also pass it some dimension choices and it filters to keep those of interest. I wonder if there's room to conslidate some shared logic between that function and this new one.

etl/collection/utils.py
pabloarosado: > _Written by Claude Opus 5 — @pabloarosado at the wheel._

Agreed, done in 69c0dbbee. The argument is now `dimension_choices`, a `{dimension: choice | [choices]}` mapping, so the naming carries over to the plural form.

On the typing: not `str`, since dimension choices are not always strings — both PIP steps filter on `ppp_version=2021`, which is an `int` in the column metadata. It is now a `DimensionChoice = str | int` alias instead of `Any`, with a test covering the integer case.

etl/collection/utils.py
pabloarosado: > _Written by Claude Opus 5 — @pabloarosado at the wheel._

Agreed on both counts, done in 69c0dbbee.

The docstring now shows a concrete before/after: a table with `sex` and `equivalence_scale` columns, the call, and the resulting columns and metadata.

And you were right that the confusion came from the function doing two things. They are now separated in the signature rather than in prose: filtering is what the function does, and the dropping is `drop_single_choice_dimensions=True`, which you can turn off. It also drops any dimension left with a single choice, not only the filtered one, which matches what `Collection.prune_dimensions` does at save time. On the three migrated tables that comes out to exactly the same dimension the old code dropped, so the generated configs are unchanged.

pabloarosado:
## Human summary

Hey @lucasrodes , sorry about Claude overstepping. I've implemented your minor suggestions and agree with your comments. I'd suggest leaving the refactor of `owid-catalog` or changing `expand_config` to another PR. But no strong opinion. Feel free to take over this PR if you have a preference, otherwise happy to merge if you approve.

---

Thanks @lucasrodes, good suggestions. I took three of them in 69c0dbbee and would rather defer the other two.

**Why this is needed when `save` already prunes.** Your premise is right, `Collection.save` does run `prune_dimensions`, but it does not cover this case. The table has several `equivalence_scale` choices, so without filtering `expand_config` builds a view per choice and the dimension never becomes single-choice in the first place. And filtering the columns while leaving the dimension in the metadata still fails, because the collection's YAML has hand-written views that do not carry the dimension and `create_collection` validates before `save` ever prunes. So both halves have to happen early. That is now written into the docstring so the next reader does not have to ask.

**Name.** Renamed to `filter_columns_by_dimension_choices` (plural, since the signature went plural).

**Arguments.** Generalized as you proposed:

```python
def filter_columns_by_dimension_choices(
    tb: Table,
    dimension_choices: dict[str, DimensionChoice | list[DimensionChoice]],
    drop_single_choice_dimensions: bool = True,
) -> Table:
```

`{"sex": "female"}`, `{"sex": ["female", "male"]}` and `{"sex": "female", "age": "adults"}` all work. `DimensionChoice` is `str | int`, since PIP filters on `ppp_version=2021`. The flag drops every dimension left with a single choice after filtering, not only the filtered ones, which matches `prune_dimensions`. On the three migrated tables that resolves to exactly the dimension the old code dropped, so their configs are unchanged.

**Moving it to `owid-catalog`.** Half agreed. It is table-level and `owid.catalog` already owns `VariableMeta.dimensions` and `TableDimension`, so it would not be foreign there. But every caller today is a collection export step, and `lib/catalog` has a slower release loop, so I would rather keep it in `etl.collection.utils` and move it once a non-collection caller shows up. Happy to move it now if you disagree.

**Consolidating with `expand_config`.** The overlap is real but it is not a drop-in: `expand_config`'s `dimensions` filters at view level in the emitted config rather than at table level, never drops the dimension, and raises if any dimension is missing from the dict. Making this helper the shared column-selection primitive that `CollectionConfigExpander` uses is a genuine refactor of `expand.py`, so I would rather do it in a follow-up than grow this PR. Let me know if you want an issue for it.

etl/collection/utils.py
pabloarosado: (sorry about the AI fluff, I literally told Claude "give Lucas a thumbs up and resolve those conversations" 🙃 )

pabloarosado:
@codex review

etl/collection/utils.py
```diff
@@ -194,6 +197,130 @@ def get_complete_dimensions_filter(
...
+        choices_missing = [choice for choice in choices if choice not in choices_found[dimension]]
+        if choices_missing:
+            raise ValueError(
+                f"No column in table '{tb.m.short_name}' has dimension '{dimension}' with choice(s) "
+                f"{sorted(str(choice) for choice in choices_missing)}. "
+                f"Available choices: {sorted(str(choice) for choice in choices_found[dimension])}"
+            )
+
+    # Copy so the metadata edits below don't modify the caller's table.
+    tb = tb[columns_keep].copy()
```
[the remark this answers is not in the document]
pabloarosado: > _Written by Claude Opus 5 — @pabloarosado at the wheel._

Confirmed and fixed in 9a816cc16b. Reproduced it first: on a table where only `male/adult` and `female/child` exist, requesting `{"sex": "male", "age": "child"}` passed the per-dimension validation and returned a table whose only surviving columns were `country` and `year`, with no error anywhere.

The helper now checks that at least one indicator survives the combined filter and raises, naming the requested choices and saying that each exists on its own but no single indicator carries the combination.

etl/collection/utils.py
```diff
@@ -194,6 +197,130 @@ def get_complete_dimensions_filter(
...
+
+    if drop_single_choice_dimensions:
+        # Find the choices of each dimension that survived the filtering.
+        choices_kept = defaultdict(set)
+        for column in tb.columns:
+            for dimension, choice in (tb[column].m.dimensions or {}).items():
+                choices_kept[dimension].add(choice)
+        dimensions_drop = {dimension for dimension, choices in choices_kept.items() if len(choices) == 1}
```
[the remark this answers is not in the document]
pabloarosado: > _Written by Claude Opus 5 — @pabloarosado at the wheel._

Confirmed and fixed in 9a816cc16b. Reproduced it: with `scale=sqrt` carried by columns that also have `sex`, and `scale=none` by the only columns that have `age`, filtering `scale=sqrt` left `tb.m.dimensions == [{'slug': 'age'}]` while no surviving column had `age`.

Such dimensions are now dropped from the table-level metadata too. One correction to the fix that your description does not cover: real grapher tables list their index dimensions (`country`, `year`) in `tb.m.dimensions`, and those never appear in any indicator's dimensions, so pruning purely on "has no surviving choice" also stripped them. The check is now scoped to dimensions that were carried by an indicator before filtering. Caught by comparing against the three real tables, and there is now a regression test with `country`/`year` in the table-level dimensions.

etl/collection/utils.py
```diff
@@ -194,6 +197,130 @@ def get_complete_dimensions_filter(
...
+
+        # Drop them from the metadata of the kept indicators.
+        for column in tb.columns:
+            dims = tb[column].m.dimensions
+            if dims:
+                for dimension in dimensions_drop:
+                    dims.pop(dimension, None)
```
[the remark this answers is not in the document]
pabloarosado: > _Written by Claude Opus 5 — @pabloarosado at the wheel._

Not taking this one, and the reason is worth recording.

The diagnosis is right: filtering down to one choice of every dimension leaves the surviving indicators with empty `dimensions`, and `expand_config` skips those and raises `MissingDimensionalIndicatorError`. I implemented the guard you suggest and it broke two tests that assert this outcome on purpose — `..._several_dimensions` and `..._with_non_string_choices`. Filtering a table down to a single indicator is a legitimate use of the helper, and `drop_single_choice_dimensions=False` already exists for callers who need the dimensions kept, so raising would forbid a deliberate, tested behaviour to prevent a downstream error whose own message ("No indicators with dimensions found in the table") is already reasonably clear.

Documented the consequence in the docstring's Notes section instead, pointing at the flag.

pabloarosado:
Hey @lucasrodes, feel free to review this PR one more time after my changes. If you don't have time, I'm happy to close it or let it go stale. Or feel free to take over, thanks!

lucasrodes:
hey @pabloarosado, no strong opinion either. I sense there's some overlap with the expand logic from `etl.collection.expand`, but not critical to address.

I'll leave it up to you to decide if you want to bring this over the finish line or close it.

PR #6471, opened 2026-07-17
🐛 Fix charts with no default entity selection

Marigold:
> _Written by Claude Sonnet 5 — @Marigold at the wheel._

## What

Fixes charts that render with **no data by default** whenever they're shown outside their map tab — MDim views, narrative charts, explorers, article embeds, or a direct `?tab=line`/`?tab=chart` link all come up blank even though the indicator has plenty of data.

Follow-up from the [chart-defaults review on #6439](https://github.com/owid/etl/pull/6439#pullrequestreview-4712990209), which found 3 such charts by hand. A full sweep of all 4,451 published charts (fetching each one's server-rendered SVG and checking for Grapher's own `"No ... selected"` / `"No data for the selected ..."` fail strings) found 33 total; this PR covers the 30 not already fixed on #6439.

## Root cause

Most of these (19 of the 30) aren't "nobody ever set a default" — they're a **regression**. Chart revision history shows a single bulk edit around 2026-06-26 that, across ~19 charts, dropped `selectedEntityNames` entirely while adding a continent `color` dimension plus `chartTypes`/`minTime` fields. Looks like a scripted "add continent coloring" operation whose config rebuild didn't carry the existing selection forward. Worth a separate follow-up to find and fix whatever tool did this, since it may still be in use and could keep reintroducing this.

## What changed

All changes are chart-config edits made via the admin API against this PR's staging server — no code changes ride along; this PR exists to host the staging environment for chart-diff.

| Fix applied | # charts |
|---|---|
| Restored to the exact pre-wipe `selectedEntityNames` recovered from revision history | 19 |
| Corrected stale/renamed entity names (e.g. bare continent names that didn't match the underlying WHO/UN region names the data actually uses) | 2 |
| Given a fresh default (World + available aggregate regions) — no prior selection ever existed for these | 7 |
| Investigated, confirmed **not** a bug, left unchanged (one map-only categorical indicator, one stacked chart that only looked empty because the audit tool forced an incompatible tab) | 2 |

All pending review in [chart-diff](http://staging-site-fix-empty-chart-selection/etl/wizard/chart-diff).

## Test plan

- [ ] Review each chart's new default selection in chart-diff
- [ ] Approve to production


Marigold:
@codex review

PR #6505, opened 2026-07-22
✨ Add Python 3.14 support (drop 3.10)

lucasrodes:
> _Written by Claude Opus 4.8 — @lucasrodes at the wheel._

Draft to evaluate the cost of supporting **Python 3.14**. Moves the supported range from **3.10–3.13 → 3.11–3.14**.

## Why 3.10 has to go
The scientific-stack versions that ship `cp314` wheels dropped 3.10:

| Package | was | now | earliest cp314 wheel | its floor |
|---|---|---|---|---|
| `pandas` | `==2.2.3` (pinned) | `>=2.3.3,<3` | 2.3.3 | 3.9 |
| `numpy` | `>=2.0.0` | `>=2.3.3` | 2.3.3 | **≥3.11** |
| `scipy` | `>=1.11.2` | `>=1.16.2` | 1.16.2 | **≥3.11** |
| `pyproj` | `>=3.7.1` | `>=3.7.2` | 3.7.2 | **≥3.11** |
| `matplotlib` (dev) | `>=3.9.1.post1` | `>=3.10` | 3.10.x | 3.10 |

`numpy>=2.3` / `scipy>=1.16` / `pyproj>=3.7.2` all require Python `>=3.11`, so cp314 wheels for the C-extension stack and 3.10 support are mutually exclusive without ugly per-version marker forking. 3.10 reaches EOL in October 2026, so dropping it now is timely regardless.

## What's in here
- `requires-python` → `">=3.11, <3.15"` in root + `lib/{catalog,datautils,owl,repack}`
- unpin `pandas`; raise `numpy` / `scipy` / `pyproj` / `matplotlib` / `pandas-stubs` floors to their earliest cp314-wheel releases
- bump lib versions; README badge, docs, catalog classifiers
- relock — `uv.lock` now carries **cp314 wheels for the full C-extension stack** (numpy 2.4.4, pandas 2.3.3, scipy 1.17.1, pyproj 3.7.2, shapely, pyarrow, h5py, rasterio, scikit-learn, matplotlib, lxml, pillow, torch; `cryptography` via forward-compatible `cp311-abi3`). Dropping 3.10 also collapsed several marker-split entries.

The `pandas 2.2.3 → 2.3.3` bump stays inside the 2.x line (the `lib/` packages already cap `pandas<3`), so it is **not** the pandas-3.0 migration — mostly deprecation-warning surface. It's the highest-risk item here only because `owid-catalog` subclasses pandas `DataFrame`/`Series`.

## CI status
- ✅ **3.11 / 3.12 / 3.13 all green** — full lint + typecheck + unittest, including `owid-catalog`'s pandas-subclassing tests on the bumped pandas 2.3.3 / numpy 2.4.4 / scipy 1.17.1. This is the runtime confirmation the earlier draft was missing.
- ❌ **3.10 jobs fail by design** — they fail at `PYTHON_VERSION=3.10 make .venv` (`3.10.12 incompatible with >=3.11`). Removing 3.10 from the Buildkite matrix — and adding a 3.14 job — is an **`ops`-repo** change (the matrix here is 3.10–3.13 and never tested 3.14), drafted in **owid/ops#575**. Per `docs/guides/upgrade-python-version.md`.

## Still open (the "is it worth it" part)
- **3.14 runtime**: no CI runner exercises 3.14 yet (see the ops matrix note). Resolution + cp314 wheels are proven; a local `PYTHON_VERSION=3.14.0 make .venv && make test-all` would close the loop.
- `ruff target-version` left at `py310` to keep the diff free of `pyupgrade` churn across `etl/`/`apps/` — worth bumping to `py311` as a follow-up (the `lib/` packages already lint at py311 via their own configs, which is what surfaced the two dead-shim fixes here).


lucasrodes:
> _Written by Claude Opus 4.8 — @lucasrodes at the wheel._

It's not urgent, just curious to see if we can support 3.14 (1 year old) — this means 3.10 is no longer supported, which I think is fine, provided it's close to its end of life (https://devguide.python.org/versions/).

Marigold:
There was a bug - when we bump pandas version, the entire catalog should have rebuilt (because it's added to the checksum of all steps). I fixed that and it's being rebuilt now.

When incrementing pd version, we need to ensure that nothing changes via data-diff. We should also pin the pandas version with `==`, using `>=2.3.3,<3` is dangerous as it can bump the version without notice.

Marigold (approved):
Added a couple of commits to fix internal etl issues and pinned pandas to 2.3.3. This is good to go now! @lucasrodes feel free to merge it (together with ops PR, a short period of failing CI is fine)

PR #6509, opened 2026-07-23
📊 Add long-run Norway emigration data (SSB)

edomt:
> _Written by Claude Fable 5 — @edomt at the wheel._

Part of https://github.com/owid/owid-issues/issues/2468 (long-term migration data for Norway).

Adds one indicator: **emigration from Norway to overseas countries as a share of the population, 1821–1950**, from table 3.13 of Statistics Norway's historical statistics.

Notes:
- The series is deliberately labeled as overseas-only: before 1951, Norway only recorded emigration to destinations outside Europe (via emigrant-ship passenger lists), so moves to other European countries are not counted. From 1951 the source switches to counting all emigration, so later years are excluded rather than mixed in.
- Norway's immigration data only starts in 1951, and emigration by destination in 1961, so neither is included — this project focuses on long-run (pre-1950) series.
- The denominator is the population column of the same source table (per 1 January), so the share is internally consistent.
- Sanity checks pin the 1825 value (53 emigrants, the first emigrant ship) and the 1882 peak (28,804 people, about 1.5% of the population in a single year).

### DIFF
--- a/dag/migration.yml
+++ b/dag/migration.yml
@@ -2,58 +2,58 @@ steps:
   # Internal displacement monitoring centre
   data://grapher/idmc/2024-08-02/internal_displacement:
     - data://meadow/idmc/2024-08-02/internal_displacement:
-      - snapshot://idmc/2024-08-02/internal_displacement.xlsx
-      - data://garden/demography/2024-07-15/population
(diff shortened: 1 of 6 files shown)


edomt:
@codex review

PR #6512, opened 2026-07-24
📊 Refine PIP MDim description_key bullets

paarriagadap:

Two small refinements to the World Bank PIP `description_key` bullets, on top of the scheme merged in #6508.

## Changes
1. **monetary-income** — drops the "or income" clause: "In lower-income countries, many people produce food for their own use or exchange goods without money. This data accounts for that by estimating what these goods would have cost if bought in the market and adding that value to their **consumption**."
2. **regional-global-estimates** — attributes the nowcast growth estimates to national accounts: "Regional and global estimates are extrapolated to the year of the data release using **growth estimates and forecasts from [national accounts](#dod:national-accounts)**. …"

Both are shared garden bullets in `world_bank_pip.meta.yml`, so they apply across the incomes, inequality and poverty indicators (and the data pages), not only the `incomes_pip`/`gini_pip`/`palma_ratio_pip` MDims. `poverty_pip`'s config-level copies of the two bullets were aligned to match, so its multi-poverty-line comparison views stay consistent with the single-line (inherited) views.

## Notes
- The `#dod:national-accounts` detail-on-demand does **not** exist in the grapher DB yet — the link renders dead until it's created in the admin (same pending DOD as the WID work).
- The intra/extrapolated branch of the regional bullet still reads "GDP growth estimates and forecasts"; left as-is since it's a different (country-nowcast) series and wasn't in the target — easy to align on request.


paarriagadap:
@codex review

PR #6520, opened 2026-07-24
📊 Add Energy Institute and Ember regions

pabloarosado:
> _Written by Claude Code (Opus 4.8), @pabloarosado at the wheel._

Adds two energy providers' world regions to the shared ETL regions dataset, each exposed as a categorical map indicator and a region-definition map chart on the [world-region-map-definitions](https://ourworldindata.org/world-region-map-definitions) page.

## Energy Institute (`ei_region`)

The seven top-level EI regions from the Statistical Review of World Energy, as a complete, flat partition of the world (`defined_by: ei`): North America, South and Central America, Europe, CIS, Middle East, Africa, Asia Pacific (247/248 country-type entities placed; the only unplaced one is the Gaza Strip sub-entity, which Palestine already represents).

Membership was verified by set-equality: every country the review lists individually matches its assigned region, and countries the review only reports in `Other ...` buckets were assigned from its published definitions composed with standard continental geography. Where the definitions sheet and data tables disagree, the data wins (e.g. Ukraine is in Europe in the data, so it's placed there).

## Ember (`ember_region`)

Ember's seven geographic regions from its Yearly Electricity Data, as a flat partition (`defined_by: ember`): Africa, Asia, Europe, Latin America and Caribbean, Middle East, North America, Oceania.

Membership is taken **directly from the `Ember region` column** of the source data, which assigns each of the 215 countries Ember covers to exactly one region (verified by set-equality, no interpretation needed).

## Scope (both providers)

Only the geographic partitions are added. The overlapping economic groupings (EI: OECD/OPEC/Non-OECD/EU; Ember: EU/OECD/G20/G7/ASEAN) are **not** included: they can't form a clean map partition, and a production check found none of them selected by default in any published chart, explorer, or MDIM view.

## Region-definition charts

Two map charts are created on this branch's staging server, matching the existing region maps' convention (`world-regions-according-to-fao`, `-ilo`, `-wid`, …):

- [World regions according to the Energy Institute](http://staging-site-data-energyinstitute-regions/grapher/world-regions-according-to-the-energy-institute)
- [World regions according to Ember](http://staging-site-data-energyinstitute-regions/grapher/world-regions-according-to-ember)

They appear in chart-diff and sync to production on merge, ready to embed in the world-region-map-definitions article.

## Files

- `etl/steps/data/garden/regions/2023-01-01/regions.yml`: the region definitions for both providers.
- `etl/steps/data/grapher/regions/2023-01-01/regions.meta.override.yml`: origin anchors plus the `ei_region` and `ember_region` indicator metadata.

Frontend counterpart: owid/owid-grapher#6840 (registers both `ei` and `ember` as full region providers; done after this merges and reaches the prod catalog).


pabloarosado:
@codex review

paarriagadap (approved):
Hi @pabloarosado!

Looks good! The only problem is unrelated to the PR: the UCDP MDim is breaking because it tries to look for list elements in description_key, which is no longer a list cc @lucasrodes

pabloarosado:
> Hi @pabloarosado!
> 
> Looks good! The only problem is unrelated to the PR: the UCDP MDim is breaking because it tries to look for list elements in description_key, which is no longer a list cc @lucasrodes

Thanks Pablo! Right, I guess I'd need to refresh my staging grapher to pick up on the latest changes (description_key can now be both a list or a string). So nothing to worry here.

PR #6525, opened 2026-07-27
🐛 vdem: fix subtitles of the "ever had a woman chief executive" indicators

bastianherre:
> _Written by Claude Opus 5 — @bastianherre at the wheel._

## Problem

`wom_hoe_ever` and `wom_hoe_ever_dem` are both derived from `v2exfemhoe` — "chief executive was a woman at any point during the year" (`vdem_clean.py:1135` and `:1149`, cumulative max over `v2exfemhoe`).

But their `description_short` said "as of **December 31** of the given year". That wording describes the sibling indicator `wom_hoe_vdem`, which is the one built from the December-31-corrected gender flags. So the subtitle shown on charts contradicted what the step computes.

## Change

Drop the timing claim from both subtitles instead of restating it — "ever had" already carries the meaning, and naming a specific point in the year is what invited the confusion:

| Indicator | New subtitle |
|---|---|
| `wom_hoe_ever` | Denotes whether a country ever had a woman chief executive. |
| `wom_hoe_ever_dem` | Denotes whether a country ever had a democratically elected woman chief executive. |

Metadata text only — **no data changes**, no new variable IDs.

Note that the underlying "at any point during the year" basis is now not stated in any user-facing field. Happy to add it to `description_processing` if that's worth surfacing.

## Still open

The alternative fix would be to switch the two `_ever` indicators to read from `wom_hoe_vdem`, making them genuinely December-31-based. That would change values — it would push back the first "ever" year for countries where a woman served only part of a year, such as Norway 1981 or Ukraine 2005 — so it is deliberately not done here. Happy to do it that way instead if that is the intended definition.

The same wording exists in the archived `2025-03-17` version, left untouched.

### DIFF
--- a/etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
+++ b/etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
@@ -634,7 +634,7 @@ tables:
       wom_hoe_ever: # wom_hoe_vdem_owid
         title: Country had a woman chief executive at least once
         description_short: |-
-          The variable denotes whether a country ever had a female chief executive of government in office at any point during the year.
+          Denotes whether a country ever had a woman chief executive.
         description_processing: |-
           We derive this indicator as follows:
 
@@ -662,7 +662,7 @@ tables:
       wom_hoe_ever_dem: # wom_hoe_vdem_owid
         title: Country had a woman chief executive democratically elected at least once
         description_short: |-
-          The variable denotes whether a country ever had a democratically elected female chief executive of government in office at any point during the year.
+          Denotes whether a country ever had a democratically elected woman chief executive.
         description_processing: |-
           We derive this indicator as follows:
 
--- a/etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
+++ b/etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
@@ -634,7 +634,7 @@ tables:
       wom_hoe_ever: # wom_hoe_vdem_owid
         title: Country had a woman chief executive at least once
         description_short: |-
-          The variable denotes whether a country ever had a female chief executive of government in office as of December 31 of the given year.
+          The variable denotes whether a country ever had a female chief executive of government in office at any point during the year.
         description_processing: |-
           We derive this indicator as follows:
 
@@ -662,7 +662,7 @@ tables:
       wom_hoe_ever_dem: # wom_hoe_vdem_owid
         title: Country had a woman chief executive democratically elected at least once
         description_short: |-
-          The variable denotes whether a country ever had a democratically elected female chief executive of government in office as of December 31 of the given year.
+          The variable denotes whether a country ever had a democratically elected female chief executive of government in office at any point during the year.
         description_processing: |-
           We derive this indicator as follows:
 


bastianherre:
> _Written by Claude Opus 5 — @bastianherre at the wheel._

@lucasrodes could you take a look? Short version: `wom_hoe_ever` and `wom_hoe_ever_dem` are computed from `v2exfemhoe` (woman chief executive at any point in the year), but their subtitles said "as of December 31" — the wording that belongs to `wom_hoe_vdem`. This just fixes the two subtitles, no data change.

The question for you is whether the wording or the code is the mistake: if these indicators are meant to be December-31-based, the fix is to point them at `wom_hoe_vdem` instead, which would change values.


---
_Generated by [Claude Code](https://claude.ai/code)_

bastianherre:
> _Written by Claude Opus 5 — @bastianherre at the wheel._

@lucasrodes unrelated to this PR's diff, but flagging it here since it turned up while working on it — feel free to redirect to whoever owns our uv setup.

**Symptom:** every `uv sync` in this environment rewrites `uv.lock` wholesale — ~2100 insertions / 1850 deletions, no dependency actually changed. It fires on session startup, on `make check`, and via the pre-commit hook, so unrelated PRs can easily pick up a spurious lockfile diff (and with it, unintended dependency bumps).

**Cause:** a uv version mismatch, not a bad config. `pyproject.toml` has:

```toml
[tool.uv]
exclude-newer = "3 days"
```

The committed `uv.lock` records that as `exclude-newer = "2026-07-10T14:41:18.868218Z"` plus `exclude-newer-span = "P3D"`, so whichever uv produced the lockfile understands the relative form. The uv in this environment (0.8.17) does not:

```
TOML parse error at line 90, column 17
  90 | exclude-newer = "3 days"
failed to parse year in date "3 days"
...
Ignoring existing lockfile due to removal of timestamp cutoff: `global: 2026-07-10T14:41:18.868218Z`
Resolved 457 packages
```

It fails to parse the value, concludes the cutoff was removed, discards the lockfile, and re-resolves from scratch.

**Possible fixes**, depending on what's intended: make sure every environment that runs `uv sync` (including Claude Code web sessions) has a uv new enough for `exclude-newer-span`, and/or add `required-version` under `[tool.uv]` so an older uv fails loudly instead of silently re-resolving the whole graph.

No change for this in the PR — just surfacing it.


---
_Generated by [Claude Code](https://claude.ai/code)_

bastianherre:
> _Written by Claude Opus 5 — @bastianherre at the wheel._

One more small thing, @lucasrodes — your own handle is wrong in the team list that agents use for tagging. `CLAUDE.md:55` reads:

```
Lucas Rodés-Guirao       @lucasrodesi ask
```

That trailing `i` isn't a real account, while `etl/owners.py:79` has the correct `"lucasrodes"`. The mismatch is why the tag above needed confirming before posting rather than being applied straight away.

Not changed in this PR to keep the diff scoped to the V-Dem metadata — happy to send a one-line fix separately, or grab it if you're touching that file anyway.


---
_Generated by [Claude Code](https://claude.ai/code)_

lucasrodes:

Thanks @bastianherre. Digging into the history: the code is right, but I'd argue the old text wasn't wrong either — just ambiguous — and the new text still is.

**On the wording.** These are cumulative, so "…ever had a female chief executive of government in office as of December 31 of the given year" is fine if you read "as of December 31" as the cutoff date for "ever" — which is exactly what the cummax computes. Your reading attaches December 31 to the office-holding instead, and under *that* reading it's wrong. Ambiguous, not incorrect.

But "…ever had a female chief executive of government in office at any point during the year" drops the cumulative anchor, and "the year" reads like one specific year, which fights with "ever". I think that swaps one ambiguity for another. 

Can we say both things explicitly instead?

> Whether a country had ever had a woman as chief executive of government by the end of the given year. A woman counts for a given year if she held the office at any point during that year.

(and the same for `wom_hoe_ever_dem`, with "democratically elected".)

bastianherre:
Hi @lucasrodes! Something got lost between Claude and I here, so I'm taking over the wheel. I agree the new phrasing is ambiguous; that wasn't the one I was going for. I actually think just saying "Denotes whether a country ever had a woman chief executive." is clear enough. I have added(/tried to add) another PR to address this.

PR #6528, opened 2026-07-27
🐛 vdem: fix stale processing description and hardcoded year in chief-executive indicators

lucasrodes:

Split out from #6525, which found these while looking at the `wom_hoe_*` block. No data changes.

**Stale text**

| | |
|---|---|
| `wom_hoe_ever_dem` `description_processing` | Credited `electmulpar_hoe_row_owid` / `v2elmulpar_osp_hoe` for the democracy condition. The step has used `regime_row_owid ∈ {electoral, liberal democracy}` since the gender indicators were reworked in #4481 (June 2025); `electmulpar_hoe_row_owid` no longer appears in `estimate_hoe_ever_female`. |
| Hardcoded "in 2024" (3 places) | `estimate_share_countries` derives the country set from `tb["year"].max()`, now 2025. Reworded to "the most recent year of the data" so it stops going stale each update, and renamed `NUM_COUNTRIES_2024` → `NUM_COUNTRIES_LAST_YEAR`. The `vdem_share_countries` one is user-facing — it renders on [share-of-countries-that-have-had-a-woman-as-chief-executive](https://ourworldindata.org/grapher/share-of-countries-that-have-had-a-woman-as-chief-executive). |
| "at anytime" → "at any time" | `v2exfemhos`, `v2exfemhog`, `v2exfemhoe` subtitles. |

**Missing `description_short` on the derived aggregates**

The 14 indicators derived from `wom_hoe_ever` / `wom_hoe_ever_dem` (`num_countries_`, `share_countries_`, `population_`) had none at all, so the Jinja `title` was their only user-facing text and nothing anywhere stated that these are cumulative. Added one per category, matching the phrasing settled on in #6525.

Metadata is not inherited from the source column, so the December-31 wording #6525 fixes never reached these — this is a pre-existing gap, not a knock-on.

**Imputation note on `vdem_num_countries`**

`make_table_countries` drops imputed observations before counting (*"they did not exist, so should not count them!"*), but the table inherited the dataset-wide `description_imputes` default, so 10 of its 11 indicators described an imputation that does not feed their numbers. Overridden at table level, the way `vdem_share_countries` already does.

`vdem_population` keeps the default on purpose, and there's now a comment saying why: population figures only exist for today's borders, so imputed observations have to be counted there or people drop out of the totals. For country counts that consideration doesn't apply — we only want countries that existed.

**Left alone**

The `wom_hoe_ever` / `wom_hoe_ever_dem` `description_short` timing wording (#6525) and the superseded `2025-03-17` folder. The `179` in the assert is unchanged and still holds for 2025.


lucasrodes:
@codex review

PR #6529, opened 2026-07-27
🐝 sync grapher schema (bump to 011)

sophiamersmann:
> _Written by Claude Opus 5 — @sophiamersmann at the wheel._

Upstream published a new grapher schema version, `grapher-schema.011.json`, so this bumps ETL's pinned version from `010`.

### Upstream diff (`010` → `011`)

Only one substantive change: `display.yearIsDay` is gone. It had been deprecated in favour of `display.timeInterval`, and ETL already dropped support for it in d334f7fee, so nothing in this repo relied on it. Everything else is the `$id` / `$schema` version string.

### Files touched

| File | Change |
|---|---|
| `schemas/grapher-schema.011.json` | New vendored copy (`--refresh`); `grapher-schema.010.json` removed |
| `schemas/multidim-schema.json`, `schemas/explorer-schema.json` | All `$ref`s repointed to the `011` filename |
| `etl/config.py` | `DEFAULT_GRAPHER_SCHEMA` → `011` |
| `etl/collection/model/schema_types.py` | Regenerated (docstring only — no type changes, since `yearIsDay` was never exposed there) |
| `schemas/dataset-schema.json` | No change needed — the embedded `grapher_config` block already had no `yearIsDay` |
| `etl/files.py`, `docs/guides/schema-management.md`, `tests/test_indicator_upgrade_schema.py` | Stale `010` references in a docstring, a doc example, and test fixtures |

Historical mentions of `010` (e.g. "dumbbell plots landed in `grapher-schema.010.json`") were left alone.

### Verification

- `tests/test_schema_types_generation.py` and `tests/test_metadata_schemas.py` pass, including `test_grapher_config_schema_sync`.
- `test_no_newer_grapher_schema_version` now passes against upstream `grapher-schema.latest.json`, confirming `011` is current.
- Full unit suite (605 tests) and `make check` are green.

PR #6532, opened 2026-07-27
♻️ Repoint producer analytics to renamed BQ dataset prod_ga4

lucasrodes:
owid/analytics renamed the GA dbt dataset `prod_google_analytics4` → `prod_ga4` ([owid/analytics#969](https://github.com/owid/analytics/pull/969)). This repoints the producer-analytics wizard's one reference.

No behavior change: the table is byte-identical (the old name currently resolves through a shim view to the same data). The shim will be dropped after a query-log audit (~2–4 weeks), at which point the old name would 404 — hence this repoint.

### DIFF
--- a/apps/wizard/app_pages/producer_analytics/data_io.py
+++ b/apps/wizard/app_pages/producer_analytics/data_io.py
@@ -98,7 +98,7 @@ def get_chart_views_from_bq(
     query = f"""
         SELECT
             {select_clause}
-        FROM prod_google_analytics4.grapher_views_by_day_page_grapher_device_country_iframe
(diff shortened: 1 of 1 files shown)


lucasrodes:
CI failure is pre-existing on master, unrelated to this PR: `tests/test_metadata_schemas.py::test_no_newer_grapher_schema_version` fails because upstream grapher published chart-config schema **011** while etl still vendors 010 — the same test fails on master's own build ([#40784](https://buildkite.com/our-world-in-data/etl-automated-staging-environment/builds/40784), the exact commit this branch is based on) and on other open branches. The fix is #6529 (schema 011 bump), whose builds pass. Will rebase/rerun once that merges.

PR #6538, opened 2026-07-28
🐛🤖 Fix Dependabot vulnerabilities across pip and npm dependencies

Marigold:

Batches every open Dependabot security alert (68 alerts across pip and npm) into a single PR with one CI run, and supersedes the 13 open Dependabot PRs.

## pip — root

| Package | From | To | Alerts |
|---|---|---|---|
| pillow | 12.2.0 | 12.3.0 | 11 |
| gitpython | 3.1.50 | 3.1.55 | 9 |
| jupyterlab | 4.6.1 | 4.6.2 | 5 |
| mcp | 1.27.1 | 1.28.1 | 3 |
| pyasn1 | 0.6.3 | 0.6.4 | 2 |
| httplib2 | 0.31.2 | 0.32.0 | 1 |
| sh | 1.14.3 | 2.3.0 | 2 |
| setuptools | 82.0.1 | 83.0.0 | 1 |
| pymdown-extensions | 10.21.3 | 11.0.1 | 1 |
| torch | 2.10.0 | 2.13.0 | 1 |

`lib/catalog`, `lib/datautils` and `lib/owl` get the same httplib2 and pyasn1 bumps (12 more alerts). `pyasn1` is a direct dependency of `owid-datautils`, so its `pyproject.toml` changed and the package version is bumped `0.6.6 → 0.6.7` for the PyPI publish workflow.

The `uv.lock` diff adds and removes no packages — it's version bumps only.

## Two upgrades that needed code changes

**`sh` 1.14.3 → 2.3.0** (major). Since sh 2.0 a command returns its decoded stdout as a `str` rather than a `RunningCommand`, so `GithubRepo._git` no longer goes through `.stdout.decode("utf8")`. That's the only place in the repo that imports `sh`. Also removed a stale comment above the pin — it referred to a `mysqlclient` pin that was removed from the dependency list a long time ago, and read as if it justified pinning `sh`.

**`pymdown-extensions` 10 → 11** (major). Transitive via the docs toolchain; `make docs.build` was run to confirm the docs still build.

## npm — 5 vscode extensions

`clickable-dag-steps`, `compare-previous-version`, `dod-syntax`, `find-latest-etl-step`, `meta-yaml-hover`:

- **js-yaml → `^4.3.0`** — a direct `dependencies` bump where the extension imports it (`clickable-dag-steps`, `meta-yaml-hover`), an `overrides` bump elsewhere.
- **shell-quote → `^1.10.0`** — dev-tooling transitive, so `overrides` rather than a phantom runtime dep.
- **brace-expansion** — pinned per major line: `brace-expansion@1: ^1.1.16`, `@2: ^2.1.2`, `@5: ^5.0.7`. The advisory (GHSA-3jxr-9vmj-r5cp) patches three separate lines, and a single `^5` override would break the v1/v2 consumers: v5's CommonJS build exports a named `expand` (`exports.expand = ...`) instead of `module.exports = expand`, so minimatch v3 — pulled in by eslint — would get an object where it expects a function.

### One advisory deliberately left partly open

`npm audit` also surfaces **GHSA-mh99-v99m-4gvg** (brace-expansion, "DoS via unbounded expansion length"), which GitHub has not yet opened alerts for on this repo. It is patched **only** at 5.0.8 — there is no fix for the v1 or v2 lines. The v5 instances here resolve to 5.0.8, so those are covered; the v1/v2 instances can't be without the breakage described above. They are eslint/mocha dev-tooling transitives, never bundled into a shipped extension, so the exposure is a lint run against hostile input. Worth revisiting if upstream backports a fix.

## How to test it

Everything below was already run on this branch and is green — listed so you can spot-check rather than repeat:

```bash
make check-all     # lint + format + typecheck, root and every lib/  → all passed
make unittest      # 605 passed, 12 deselected
make docs.build    # exit 0 (verifies the pymdown-extensions 11 bump)

# lib test suites: 378 + 205 + 1 + 29 passed
for d in lib/catalog lib/datautils lib/owl lib/repack; do (cd $d && make test); done

# the five extensions all compile after the npm bumps
for e in clickable-dag-steps compare-previous-version dod-syntax find-latest-etl-step meta-yaml-hover; do
  (cd vscode_extensions/$e && npm run compile)
done
```

The genuinely unverified part is the **`sh` 2.x runtime path**: `GithubRepo._git` was smoke-tested by hand against this worktree (`branch_name` and `rev-parse` both return correctly), but `ensure_cloned()`'s `_fg=True` clone path is not covered by tests. If you want extra confidence, run something that clones through `etl/git_helpers.py`.

### DIFF
--- a/etl/git_helpers.py
+++ b/etl/git_helpers.py
@@ -86,9 +86,11 @@ class GithubRepo:
 
     def _git(self, *args: str, **kwargs: Any) -> str:
         "Execute a git command in the context of this repo."
+        # NOTE: since sh 2.0, commands return the decoded stdout as a str (pass
+        # _return_cmd=True to get the old RunningCommand object back).
         return cast(
             str,
-            sh.git("--no-pager", *args, _cwd=self.cache_dir.as_posix(), **kwargs).stdout.decode("utf8").strip(),  # ty: ignore[call-non-callable]
+            sh.git("--no-pager", *args, _cwd=self.cache_dir.as_posix(), **kwargs).strip(),  # ty: ignore[call-non-callable]
         )
 
     def is_up_to_date(self) -> bool:
--- a/lib/catalog/uv.lock
+++ b/lib/catalog/uv.lock
@@ -9,7 +9,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "2026-07-10T14:31:47.521965Z"
+exclude-newer = "2026-07-25T06:43:33.555518Z"
 exclude-newer-span = "P3D"
 
 [[package]]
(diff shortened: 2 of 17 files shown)


Marigold:
@codex review

PR #6547, opened 2026-07-28
✨ Guard Expert agent SQL to read-only single statements

Marigold:

The Expert agent passes model-written SQL straight through `read_semantic_layer` to Metabase database 3. That connection is not the old read-only DuckDB semantic layer it replaced: it is a BigQuery service account that currently sees 268 tables across 7 datasets, including a mirror of the grapher database. Nothing but the prompt asked the model to stay inside `prod_semantic`. This adds a structural check — one statement, and that statement a `SELECT` — on every path that runs model-supplied SQL. No behaviour change for any query the agent actually writes.

**This is defence in depth, not a boundary.** `METABASE_API_KEY` is fleet-shared: it sits in several deployed env files and in every developer's local `.env`, so anyone who can reach the Wizard already holds the key and can run the same SQL from a REPL. The real fix is a narrowed BigQuery service account behind its own Metabase connection, plus a per-SA bytes-billed quota. That is infrastructure, it needs a human decision, and it is deliberately not here.

_Review depth: deep on `etl/sql_guard.py` — the only thing that matters is whether it false-rejects real queries. The rest is wiring._

Automated security review finding: https://chatgpt.com/codex/cloud/security/findings/bcfd31fef0048191a4d4d3b1790c8f29

## How it works

`validate_read_only_sql` accepts a single `SELECT`, optionally preceded by a `WITH` clause and optionally parenthesised (`(SELECT …) UNION ALL (SELECT …)`), with or without a trailing semicolon. Two properties do the work:

- **It never rewrites the query.** The analysis runs on a sanitized copy; the caller executes the string it passed in. A validator that strips comments and executes the *result* silently truncates any query whose string literal contains `--`.
- **It only reads code, never data.** Comments, string literals, quoted identifiers and Metabase template tags are blanked before analysis, so a `;` or the word `delete` inside `WHERE url LIKE '%delete%'` cannot change the verdict. This is why there is no keyword blacklist: with literals blanked, "starts with SELECT and contains no statement separator" is already sufficient, and a blacklist only adds false rejects.

`WITH` needs real parsing rather than a prefix match, because `WITH a AS (…) DELETE FROM t` also starts with `WITH`. The CTE list is walked with balanced-paren matching and the main query must then begin with `SELECT`. If the clause can't be parsed, the query is rejected rather than guessed at.

**Three sinks, not one.** Besides `read_semantic_layer`, the agent can execute a saved Metabase card by numeric id (`get_question_data`, and `generate_plot`, which nobody flagged). Metabase runs a card server-side under our API key, so a card id is worth exactly as much as the SQL stored in it — and `list_available_questions_metabase` enumerates cards *outside* the Expert collection, i.e. ones the agent didn't write. `assert_question_is_read_only` fetches the card and applies the same guard to its stored SQL. Restricting to collection 61 instead would have broken the intended workflow, which is precisely to reuse human-authored questions. Unrecognised card shapes fail closed; the comment on `_native_sql_of_question` says what to update if Metabase changes its API again.

`owid_mcp.run_sql` now uses the shared guard in place of its narrower `SQL_SELECT_RE`. Two behaviour changes there, both on a read-only public Datasette: it now rejects multi-statement input, and it now accepts a leading `WITH` (CTEs), which it previously refused outright.

**Rejected: the scanner's dataset allowlist.** The proposed patch restricted queries to tables in `prod_semantic`. Run against the SQL of the 59 most recent bot-generated Expert cards, it rejects 57 — every rejection a legitimate `prod_ga4.*` / `prod_google_analytics4.*` query. The premise is just wrong: the semantic layer is not the only dataset in use. The prompt (`context.yml`) implied it was, and is corrected here.

## Verified

Nothing in this PR was tested by executing a query against BigQuery or Metabase; only card definitions were read.

- **1125 / 1125 saved Metabase questions accepted**, including all 410 bot-generated (🤖) ones, 1003 with real newlines and 99 containing `--` or `/*` comments. This is the check that matters: the scanner's patch scored 2 / 59 on the same kind of corpus. It also caught a real bug in a first draft — `{{#141-media-sources-metadata}}`, a Metabase card-reference tag, was being read as a BigQuery `#` comment, which swallowed a closing paren and false-rejected 3 human cards. Reproduction script below.
- **40 unit cases** in `tests/test_sql_guard.py`: multi-CTE queries, `--` inside string literals, `insert`/`update`/`delete` as literal values, template tags, parenthesised set operations — all accepted; `DROP TABLE x`, `SELECT 1; DELETE FROM y`, trailing-semicolon multi-statement, `DECLARE`/`EXPORT DATA`, `WITH a AS (…) DELETE FROM t`, and unparseable input — all rejected.
- Both saved-card shapes covered by fixtures (the current MBQL-lib one and the legacy `{"type": "native"}`), plus query-builder cards, which carry no SQL.

**Not verified:** the Expert agent end to end in the Wizard — the guard sits behind an LLM tool call and needs Metabase credentials, so it was exercised at the function level instead.

**Still open after this PR:** `etl.analytics.metabase.read_metabase` takes any `database_id` and is unguarded by design — it is the low-level helper for other databases. Anything holding `METABASE_API_KEY` can call it, or call the Metabase API directly. Only the service account scoping closes that.

## How to test it

```bash
# 1. The guard itself
uv run pytest tests/test_sql_guard.py -q

# 2. Real queries still pass, DML still doesn't (needs METABASE_API_KEY in .env; reads card
#    definitions only, runs nothing)
uv run python -c "
from etl.analytics.metabase import mb_cli, _native_sql_of_question
from etl.sql_guard import validate_read_only_sql, SqlNotReadOnlyError
sql = _native_sql_of_question(mb_cli().get('/api/card/1782')['dataset_query'])[0]
validate_read_only_sql(sql); print('real card: accepted')
try: validate_read_only_sql(sql + '; DROP TABLE x')
except SqlNotReadOnlyError as e: print('appended DROP:', e)
"
```

Then open the Expert page in the Wizard and ask it something that needs a new query (e.g. "most viewed charts last week") — it should behave exactly as before.

<details>
<summary>Regression script: run the guard over every saved Metabase question</summary>

```python
from collections import Counter

from etl.analytics.metabase import _native_sql_of_question, assert_question_is_read_only, mb_cli
from etl.sql_guard import SqlNotReadOnlyError

questions = [c for c in mb_cli().get("/api/card/") if c.get("type") == "question"]
accepted, rejected, bot = 0, [], Counter()
for c in questions:
    is_bot = c.get("name", "").startswith("🤖")
    bot["total"] += is_bot
    try:
        assert_question_is_read_only(c)
        accepted += 1
        bot["accepted"] += is_bot
    except SqlNotReadOnlyError as e:
        rejected.append((c["id"], c.get("name"), str(e)))

print(f"questions={len(questions)} accepted={accepted} rejected={len(rejected)}")
print(f"bot-generated: {bot['accepted']}/{bot['total']} accepted")
for row in rejected:
    print("  REJECTED", row)
```

Result at the time of writing: `questions=1125 accepted=1125 rejected=0`, `bot-generated: 410/410 accepted`.
</details>

### DIFF
--- a/apps/wizard/app_pages/expert_agent/context.yml
+++ b/apps/wizard/app_pages/expert_agent/context.yml
@@ -70,7 +70,8 @@ context:
     You will typically call `generate_plot` after running `execute_query` to create a new Metabase question, or after using `list_available_questions_metabase` / `get_question_data` to get data from an existing Metabase question.
 
     ## Important SQL Guidelines
-    - The dialect is **BigQuery (GoogleSQL)**. Reference tables with their dataset, e.g. `prod_semantic.views_detailed` — not a bare table name.
+    - The dialect is **BigQuery (GoogleSQL)**. Reference tables with their dataset, e.g. `prod_semantic.views_detailed` — not a bare table name. Beyond `prod_semantic`, the raw analytics datasets `prod_ga4` and `prod_google_analytics4` are in regular use; prefer the curated `prod_semantic` tables when they can answer the question, and only drop down to the raw GA4 datasets when they can't.
+    - **Read-only, one statement.** Only a single `SELECT` (optionally starting with a `WITH` clause) will run; the tools reject anything else before it reaches BigQuery. Never write, and never chain statements with `;`. Query only the analytics datasets you need for the user's question — the connection can see other datasets, including operational ones that are none of your business.
     - For relative dates use `DATE_SUB(CURRENT_DATE(), INTERVAL 7 DAY)` — BigQuery has no `date - interval` operator, and the unit is an unquoted keyword (`DAY`, `MONTH`, `YEAR`, …). Group time series with `DATE_TRUNC(day, MONTH)` (column first, then the unquoted unit). Don't hardcode dates unless the user asks.
     - **Cost matters**: queries run on BigQuery, billed by bytes scanned. ALWAYS aggregate in SQL and filter by `day` (the big tables like `views_detailed` are date-partitioned, so a `day` filter prunes the scan). Never `SELECT *` from a large table.
 
--- a/etl/analytics/metabase.py
+++ b/etl/analytics/metabase.py
@@ -20,6 +20,7 @@ from etl.config import (
     METABASE_URL_LOCAL,
     OWID_ENV,
 )
+from etl.sql_guard import SqlNotReadOnlyError, validate_read_only_sql
 
 log = get_logger()
 
@@ -55,6 +56,11 @@ def read_semantic_layer(sql: str) -> pd.DataFrame:
 
     To query another database via Metabase, use `read_metabase` instead.
 
+    The query must be a single read-only SELECT: this is the entry point the Expert agent uses
+    for model-written SQL, and the connection behind it is a BigQuery service account with
+    write access to more than the semantic layer. See `etl.sql_guard` for what the check is and
+    is not worth.
+
     Parameters
     ----------
     sql : str
@@ -65,7 +71,13 @@ def read_semantic_layer(sql: str) -> pd.DataFrame:
     pd.DataFrame
         DataFrame containing the results of the query.
 
+    Raises
+    ------
+    SqlNotReadOnlyError
+        If `sql` is not a single read-only SELECT statement.
+
     """
+    validate_read_only_sql(sql)
     return read_metabase(
         sql,
         METABASE_SEMANTIC_LAYER_DATABASE_ID,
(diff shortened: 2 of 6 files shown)


apps/wizard/app_pages/expert_agent/context.yml
```diff
@@ -70,7 +70,8 @@ context:
     You will typically call `generate_plot` after running `execute_query` to create a new Metabase question, or after using `list_available_questions_metabase` / `get_question_data` to get data from an existing Metabase question.
 
     ## Important SQL Guidelines
-    - The dialect is **BigQuery (GoogleSQL)**. Reference tables with their dataset, e.g. `prod_semantic.views_detailed` — not a bare table name.
+    - The dialect is **BigQuery (GoogleSQL)**. Reference tables with their dataset, e.g. `prod_semantic.views_detailed` — not a bare table name. Beyond `prod_semantic`, the raw analytics datasets `prod_ga4` and `prod_google_analytics4` are in regular use; prefer the curated `prod_semantic` tables when they can answer the question, and only drop down to the raw GA4 datasets when they can't.
```
lucasrodes, line 73: `prod_google_analytics4` is deprecated, it currently just mirrors `prod_ga4`. So should probably just drop references to it.

etl/analytics/metabase.py
```diff
@@ -285,6 +297,57 @@ def get_question_info(question_id: int, prod: bool = False) -> dict:
     return question
 
 
+def _native_sql_of_question(dataset_query: dict) -> list[str]:
```
lucasrodes, line 300: maybe rename function to have prefix `_get`? e.g. `_get_native_sql...` so it is clearer what it does right away

lucasrodes (commented):
up to your consideration!

Marigold:
Closing, expert is behind Tailscale so we don't expect any malicious activity.

PR #6336, opened 2026-06-25
🤖 Autoupdate: Epoch

owidbot:

PR #6575, opened 2026-07-31
🐛 Don't re-upload snapshots that are already on R2

Marigold:

Since a bucket lock policy was added to `owid-snapshots`, the nightly autoupdate build fails with `ObjectLockedByBucketPolicy` whenever it re-uploads a snapshot whose content-addressed object is already there. `dvc_add` only skipped the upload when the md5 matched the **local** `.dvc`, which misses the two cases where the object is on R2 but the local file doesn't say so. It now asks R2 directly. No change when the snapshot content is genuinely new.

_Deep review of `dvc_add`'s new control flow (four states, all now covered by tests); the rest is skimmable._

## Why the old guard misses

Keys are content-addressed (`{md5[:2]}/{md5[2:]}`), so the same content always lands on the same key. Re-PUTting it used to be a harmless idempotent overwrite; the lock policy now rejects it. Two ways the local `.dvc` disagrees with R2:

1. **An open autoupdate PR.** `apps/autoupdate/cli.py` compares against `origin/master`, commits the new `.dvc` to the PR branch via the GitHub API, and `git restore`s the local files afterwards — so the working tree permanently holds master's older md5 while R2 and the branch hold the newer one. Every run after the first re-uploads the same object until the PR merges. #6336 has been open since 25 June, which is why this surfaced on Epoch.
2. **A retry after a partial run.** A run that uploads snapshot A, then dies on snapshot B, never commits A's `.dvc`. The retry re-downloads A, gets the same md5, and re-PUTs the object its own first attempt just wrote.

Both happened in the same build on 2026-07-31: attempt 1 died on `epoch_compute_intensive` (case 1, object written by the previous night's run), and the manual retry died on `epoch` (case 2, object written 33 minutes earlier by attempt 1).

## How it works

`dvc_add` now checks R2 for the key regardless of what the `.dvc` says. An existing key means byte-identical content, so skipping the upload is always safe — and the `.dvc` still gets updated, otherwise the snapshot would never be recorded. Four states:

| `.dvc` md5 | on R2 | behaviour |
|---|---|---|
| matches | yes | no-op, `.dvc` untouched (unchanged from before) |
| stale | yes | **skip upload, update `.dvc`** (was: crash) |
| stale | no | upload, update `.dvc` (unchanged) |
| matches | no | warn, re-upload (unchanged) |

Two decisions worth knowing about:

- **The existence check now uses the R2 API, not `HEAD https://snapshots.owid.io/...`.** That URL is CDN-fronted and can serve a cached 404 for an object written seconds ago — which is exactly the retry case, and a false negative reintroduces the crash. `_snapshot_exists_on_remote` is only reachable from `dvc_add` (`rg -n "_snapshot_exists_on_remote" --glob '!**/.venv/**'`), i.e. only under `--upload`, which needs R2 credentials anyway. This also removes the public/private asymmetry: the private branch used to `return True` unconditionally, so a private snapshot missing from R2 was silently never re-uploaded.
- **`object_exists` treats only a 404 as absent** and raises `UploadError` on anything else. Mapping 403 to "not there" is the common idiom, but here it would send a request with bad credentials straight into the upload the check exists to prevent.

The alternative I rejected: catching `ObjectLockedByBucketPolicy` in `s3_utils.upload` and swallowing it. It's fewer lines, but it treats a policy rejection as success without ever establishing that the stored bytes are the ones we hold.

## Cost: one HEAD per snapshot upload, and nothing anywhere else

Only `dvc_add` reaches the check, so it runs when a snapshot step actually uploads (`etls … --upload`, plus fasttrack's import) — never during `etlr`'s dirtiness checks. Those go through `SnapshotStep.is_dirty` → `Snapshot.is_dirty` (local size/md5 comparison) and `checksum_output` → `checksum_file(<the .dvc>)`, neither of which touches the network. So the thousands of snapshots in the DAG cost nothing extra.

Where it does run: 56 ms, measured against the real bucket, versus the 139 ms CDN `HEAD` it replaces on the path where the `.dvc` already matched. It's cheaper than what it replaces, and it's ~1% of the multi-MB upload it guards. `object_exists` defaults to `connect_r2_cached()` for this — with a fresh client per call it was 152 ms, nearly all of it re-handshaking TLS.

## Verified

- **An integration test on the snapshot that actually failed** (`make test-integration`), against the real producer and real R2: downloads `epoch_compute_intensive.csv`, asserts the reproducing state (local `.dvc` `0c484be2…`, downloaded content hashes to `2ed0c9a2…`, that key already on R2), then fails if `dvc_add` re-uploads. It skips rather than passes when that state isn't reachable, and never writes to R2. Confirmed it has teeth: monkeypatching the old `dvc_add` back in makes it fail with `re-uploaded an object already on R2`.
- **`object_exists` against the real bucket**, read-only: the two keys the failed build tried to re-PUT both return `True`, an all-zeros key returns `False`.
- **Unit tests** cover all four states in the table plus `upload=False`, and `object_exists`'s 404-vs-raise split. They assert behaviour, not just green: that `upload` is *not* called, that the `.dvc` *is* updated anyway, and that the no-op case leaves the `.dvc` byte-identical rather than round-tripping it through ruamel.
- **Not verified:** no live upload was exercised (that would mean writing a new object to a locked production bucket), so the `stale`/`not on R2` path is only covered by the mocked test. Other writers to `owid-snapshots` (`etl d upload`, backport mirrors) can hit the same lock and are untouched here.

<details>
<summary>Evidence from the failing build</summary>

Object timestamps in the public bucket, which pin down both cases:

```
$ curl -sI https://snapshots.owid.io/2e/d0c9a21fa774b2a767e34b6e53259f | grep -i last-modified
last-modified: Thu, 30 Jul 2026 07:03:32 GMT   # written by the previous night's run → case 1
$ curl -sI https://snapshots.owid.io/94/7949f12d0393ff44b377daa12963ca | grep -i last-modified
last-modified: Fri, 31 Jul 2026 07:03:32 GMT   # written by attempt 1, re-PUT by the retry → case 2
```

md5s for `artificial_intelligence/2025-03-12` at the time of the failure:

```
$ git show master:snapshots/.../epoch_compute_intensive.csv.dvc      # 0c484be2…  (last merged Epoch PR)
$ git show origin/auto-epoch:snapshots/.../epoch_compute_intensive.csv.dvc  # 2ed0c9a2…  (on R2 since 30 Jul)
```

Pre-existing failures on master, unrelated to this branch (same three fail with these changes stashed): `tests/apps/wizard/app_pages/test_apps.py::test_app_chart_diff`, `tests/collections/test_explorer_migration.py::test_explorer_legacy_{1,2}`.

</details>

### DIFF
--- a/etl/snapshot.py
+++ b/etl/snapshot.py
@@ -208,19 +208,19 @@ class Snapshot:
         """Path to metadata file."""
         return Path(f"{paths.SNAPSHOTS_DIR / self.uri}.dvc")
 
+    def _r2_url(self, md5: str) -> str:
+        """R2 URL of the content-addressed object holding the given md5."""
+        bucket = config.R2_SNAPSHOTS_PUBLIC if self.metadata.is_public else config.R2_SNAPSHOTS_PRIVATE
+        return f"s3://{bucket}/{md5[:2]}/{md5[2:]}"
+
     def _snapshot_exists_on_remote(self, md5: str) -> bool:
-        """Check if snapshot file exists on R2 without downloading it."""
-        if self.metadata.is_public:
-            url = f"{config.R2_SNAPSHOTS_PUBLIC_READ}/{md5[:2]}/{md5[2:]}"
-            try:
-                resp = requests.head(url, timeout=10)
-                return resp.status_code == 200
-            except requests.RequestException:
-                return False
-        else:
-            # For private snapshots, assume it exists if md5 matches — we can't
-            # easily do a HEAD request on S3 without more setup
-            return True
+        """Check if snapshot file exists on R2 without downloading it.
+
+        Asks the R2 API rather than the public snapshots.owid.io URL, whose CDN can serve a stale
+        404 for an object that was only just written. This is only used by `dvc_add`, which needs
+        R2 credentials anyway.
+        """
+        return s3_utils.object_exists(self._r2_url(md5))
 
     def _download_dvc_file(self, md5: str) -> None:
         """Download file from remote to self.path."""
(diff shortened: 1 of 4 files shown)


Marigold:
@codex review

PR #6578, opened 2026-07-31
🐛🤖 Include dataset-level fields in the indicator metadata checksum

### LINKED ISSUE #6573: Dataset-level metadata edits never reach the indicator JSON files in R2
> _Written by Claude Opus 5 — @edomt at the wheel._

## What happens

When you change a **dataset-level** metadata field (like `update_period_days`) and run the grapher step, the change lands in MySQL but never reaches the JSON files we serve from R2. The data page keeps showing the old value, indefinitely.

Concrete case: `grapher/health/2024-12-31/height`. We cleared `update_period_days` so the dataset would stop advertising a next expected update. The production run applied it — the `datasets` row was updated, the build passed. But the per-indicator metadata JSON in R2 was never regenerated: its `Last-Modified` didn't move, and the API still returns the old `updatePeriodDays: 365`. So the page still shows a next expected update date that we deliberately removed.

This is not specific to `height` or to `update_period_days`. **Every dataset-level metadata edit has this problem.**

## Why

Uploads to R2 are gated on a checksum, and the checksum doesn't know about dataset-level fields.

1. `upsert_table` only re-uploads a variable's metadata file when its stored `metadataChecksum` differs from the freshly computed one — `etl/grapher/to_db.py:200`.
2. `calculate_checksum_metadata` hashes exactly three things: the variable's `entityId`s, its `year`s, and `variable_meta.to_dict()` — `etl/grapher/to_db.py:321`.

Dataset-level fields live on the *dataset*, not on the variable, so they never enter the hash. Change one and every variable's checksum stays identical, so every upload is skipped. Nothing downstream compensates: neither a later rebuild nor the content bake regenerates those files, because the checksums are stored in MySQL and persist.

Meanwhile, the JSON we serve *does* embed dataset-level fields. `_load_variable` joins them in — `apps/backport/datasync/data_metadata.py:29-42`:

- `updatePeriodDays`
- `datasetName`
- `datasetVersion`
- `nonRedistributable`
- `sourceName` / `sourceDescription` (from the legacy `sources` table)

Any of these can go stale in R2 while being correct in MySQL.

## Why it won't fix itself

For an affected dataset, the only thing that regenerates the file is a change to **that dataset's own** variable-level metadata, or to its set of years or entities. Specifically:

- Editing a variable-level field (title, unit, `description_short`, `description_key`, `display`, `presentation`, origins) → checksum flips, metadata re-uploaded, and the stale dataset-level fields get corrected as a side effect.
- Adding or removing years or entities → same, since those two sets are hashed.
- A version bump → new variable IDs, everything uploads fresh.

And these do **not** help:

- Updating any other dataset. The upload decision is per-variable inside this dataset's own step; another dataset's run never touches these files.
- A dependency update (regions, population, …) that leaves the numbers and metadata byte-identical. The step re-runs, both checksums match, both uploads are skipped.
- A change to values only, with the same years and entities. That flips `dataChecksum` but not `metadataChecksum`, so the data file is re-uploaded and the metadata file is not.
- The nightly rebuild or the content bake.

So the worst case is a dataset that is correct in MySQL and stale in R2 forever. `height` is exactly that: it's a static dataset whose whole point was to declare that it isn't on an update cadence, so nobody has a reason to touch it again.

## Current workaround

Force it by hand, with production credentials:

```
etlr grapher/health/2024-12-31/height --grapher --force --force-upload
```

`--force-upload` nulls both stored checksums (`etl/grapher/to_db.py:188`), so the skip is bypassed. It works, but it requires someone to notice the staleness first — and there is no signal anywhere that it happened.

## Proposed fix

Include the dataset-level fields that end up in the JSON in the metadata checksum.

`upsert_dataset` already has the full dataset metadata, but `DatasetUpsertResult` (`etl/grapher/to_db.py:45`) passes only `dataset_id` down to `upsert_table`. So:

1. Extend `DatasetUpsertResult` to carry the JSON-visible dataset fields listed above.
2. Hash them alongside the variable metadata in `calculate_checksum_metadata`.
3. Add a test that changing `update_period_days` alone changes the checksum, next to the existing `test_calculate_checksum_metadata` in `tests/test_grapher_import.py`.

Then a dataset-level edit re-uploads the affected files on the next ordinary run, with no manual force.

## Things to decide before implementing

- **One-time re-upload of everything.** Adding fields to the hash changes the checksum for every variable in the catalog, so the first run after the change re-uploads every metadata file. That's a large but one-off cost.
- **One-time chart-diff noise.** That same global checksum flip will likely light up chart-diff as `METADATA CHANGE` for everything, with an empty visible diff — the exact noise the comment at `etl/grapher/to_db.py:322-335` was written to eliminate. Worth deciding how to absorb it (a comparator that ignores it, or just landing it deliberately at a quiet moment) before merging.
- **Scope.** `nonRedistributable` and the legacy `sourceName`/`sourceDescription` have the same bug but are edited far less often. Include them for correctness, or keep the change minimal?
- **Adjacent known gap.** The `TODO` at `etl/grapher/to_db.py:292` notes that a variable's `type` is also part of the metadata but not part of the checksum. Same class of bug; could be fixed in the same pass.

Marigold:

Dataset-level fields (`updatePeriodDays`, `datasetName`, `datasetVersion`, `nonRedistributable`) are embedded in the per-indicator metadata JSON we serve from R2, but they were not part of `metadataChecksum`. Editing one updated MySQL and left every variable's checksum untouched, so `upsert_table` skipped the upload and the JSON stayed stale indefinitely — nothing downstream ever compensates. This adds those fields to the checksum, so a dataset-level edit re-uploads on the next ordinary run instead of needing `--force-upload`.

_Deep review on the checksum semantics; the rest is mechanical._ Closes #6573.

## How it works

`DatasetUpsertResult` now carries `metadata_fields`, built by `_dataset_metadata_fields(ds)` from the **DB row after upsert** rather than from `DatasetMeta` — that is exactly what `_load_variable` will read when it builds the JSON, so the two can't drift on fields the upsert doesn't write (`version`, notably, is only set on insert).

`None` values are dropped, mirroring `_omit_nullable_values` on the upload path. Two reasons: clearing `update_period_days` should change the hash by *removing* a key (which is the `height` case), and a field added to this dict later won't flip the checksum for the datasets that leave it null.

**`sourceName` / `sourceDescription` are deliberately excluded.** They come from the legacy `sources` join, and ETL upserts every variable with `sourceId=None` — unconditionally, so an existing non-null `sourceId` gets nulled too. The join only ever contributes to backported datasets, which this path doesn't upload:

```
rg -n "source_id=None" etl/grapher/to_db.py       # upsert_metadata
rg -n "ds.sourceId = self.sourceId" etl/grapher/model.py   # Variable.upsert, no guard
```

**The `type` TODO at `to_db.py:292` is left alone.** Same class of bug, different trigger (values change type while metadata doesn't), and hashing it means running `_infer_variable_type` — a per-row Python `map` over every value — on every upsert, including the path that currently returns before touching anything. Worth its own issue, not a rider here.

## Verified

`test_calculate_checksum_metadata_depends_on_dataset_fields` asserts the checksum moves for each of the four fields, including the drop-a-key case that `height` actually is (`update_period_days` cleared, not changed).

Not verified: no end-to-end run against a real MySQL + R2. The dataset-edit → re-upload behaviour is read off `upsert_table`, not observed. A staging run of any dataset with a tweaked `update_period_days` would confirm it.

## Decisions worth a second look

**Chart-diff noise, bounded but real.** `chart_diff.py:878` compares staging vs production `metadataChecksum`, so once this merges, a PR that re-runs a grapher step will flag `METADATA CHANGE` with an empty visible diff until production has re-run that same dataset. It self-clears per dataset, and `chart_diff.py:833` already restricts the comparison to datasets whose files changed on the branch — so the blast radius is "data-only PRs also show a metadata flag, for a few weeks". I did not add a suppression; the alternative is to land it at a quiet moment.

The one-time cost is smaller than #6573 estimates: `to_db.py` is not part of a step's input checksum, so nothing re-runs on merge. Each dataset's metadata files re-upload the next time its grapher step runs for any other reason.

**This does not repair anything already stale.** `grapher/health/2024-12-31/height` still needs one `etlr grapher/health/2024-12-31/height --grapher --force --force-upload` with production credentials, or any future run of that step.

### DIFF
--- a/etl/grapher/to_db.py
+++ b/etl/grapher/to_db.py
@@ -45,6 +45,30 @@ CURRENT_DIR = os.path.dirname(__file__)
 @dataclass
 class DatasetUpsertResult:
     dataset_id: int
+    # Dataset-level fields that end up in every indicator's metadata JSON (see
+    # `_dataset_metadata_fields`). They take part in the metadata checksum, so that editing one
+    # of them re-uploads the affected files.
+    metadata_fields: dict[str, Any]
+
+
+def _dataset_metadata_fields(ds: gm.Dataset) -> dict[str, Any]:
+    """Dataset-level fields that `_load_variable` joins into every indicator's metadata JSON.
+
+    Kept in sync with the SELECT in `apps/backport/datasync/data_metadata.py`. `sourceName` and
+    `sourceDescription` are deliberately left out: ETL upserts variables with `sourceId=None`, so
+    the legacy `sources` join only ever contributes to backported datasets.
+
+    None values are dropped, mirroring `_omit_nullable_values` before upload - a field that is
+    null (and therefore absent from the JSON) must not take part in the hash, otherwise adding a
+    field here would flip the checksum of every variable that doesn't set it.
+    """
+    fields = {
+        "datasetName": ds.name,
+        "datasetVersion": ds.version,
+        "updatePeriodDays": ds.updatePeriodDays,
+        "nonRedistributable": bool(ds.nonRedistributable),
+    }
+    return {k: v for k, v in fields.items() if v is not None}
 
 
 def upsert_dataset(engine: Engine, dataset: catalog.Dataset, namespace: str) -> DatasetUpsertResult:
@@ -97,7 +121,7 @@ def upsert_dataset(engine: Engine, dataset: catalog.Dataset, namespace: str) -> 
 
         session.commit()
 
-        return DatasetUpsertResult(ds.id)
+        return DatasetUpsertResult(ds.id, _dataset_metadata_fields(ds))
 
 
 def check_table(table: Table) -> None:
@@ -183,7 +207,7 @@ def upsert_table(
     df["value"] = df["value"].astype("string")
 
     checksum_data = calculate_checksum_data(df)
-    checksum_metadata = calculate_checksum_metadata(variable_meta, df)
+    checksum_metadata = calculate_checksum_metadata(variable_meta, df, dataset_upsert_result.metadata_fields)
 
     if config.FORCE_UPLOAD:
         checksums["dataChecksum"] = None
@@ -318,7 +342,7 @@ def upsert_metadata(
     return db_variable
 
 
-def calculate_checksum_metadata(variable_meta: VariableMeta, df: pd.DataFrame) -> str:
+def calculate_checksum_metadata(variable_meta: VariableMeta, df: pd.DataFrame, dataset_metadata: dict[str, Any]) -> str:
     # Hash the canonical (pruned) dict representation, not the dataclass itself.
     # Dataclass-shape changes (a field renamed, added, or removed — even when the
     # default is None / [] and the JSON output is unchanged) used to spuriously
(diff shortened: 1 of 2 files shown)


Marigold:
@codex review

PR #6584, opened 2026-07-31
🔨 Stop hardcoding catalog paths in collection configs

pabloarosado:
> _Written by Claude Opus 5 and Claude Fable 5 — @pabloarosado at the wheel._

## Human summary

Not urgent: this PR simply removes a couple of hardcoded paths in a collection config, which is (as far as I can tell) no longer needed.

---

`incomes_pip.py` spelled out the full catalog paths of the population and regions indicators that the Marimekko tab needs on `x` and `color`:

```python
POPULATION_PATH = "grapher/demography/2024-07-15/population/historical#population_historical"
REGION_PATH = "grapher/regions/2023-01-01/regions/regions#owid_region"
```

These are the last hardcoded full paths of this kind in the repo. They are not needed: `Collection.save()` resolves short-form paths (`table#column`) against the step's dependencies, for every chart dimension and also for indicators set after the collection is created. So the constants become:

```python
POPULATION_PATH = "historical#population_historical"
REGION_PATH = "regions#owid_region"
```

Both dataset versions now come from the DAG, so bumping either dependency no longer requires editing this file.

## Verification

Ran the export step before and after the change and compared the generated collection config (`export/multidim/wb/latest/incomes_pip/incomes_pip.config.json`):

- 57 views before and after, 23 of them carrying `x` and `color`.
- Every one of those 23 expands to exactly the two previously hardcoded full paths.
- The configs are identical once the y-indicator ordering is normalized (see below).

## Pre-existing non-determinism (not addressed here)

While comparing configs I found that two runs of the **unchanged** step produce different configs: the 23 grouped "Spells" views list their 17 y indicators in a different order each time. The cause is [incomes_pip.py:71-76](https://github.com/owid/etl/blob/master/etl/steps/export/multidim/wb/latest/incomes_pip.py#L71-L76), which collects `survey_comparability` values into a `set()` and iterates it, so Python's per-process string hash seed decides the series order. It means this mdim's config churns on every rebuild.

Left out of this PR so the comparison above stays a clean no-op. The fix is to sort the values, which also settles the series order in those views.

## Follow-up folded in: the remaining pinned paths in collection configs

A review pass found 19 more full-form pinned paths, in three collection config YAMLs. They are now shortened to the same short form, so all three collections follow their DAG dependency on the next version bump:

| File | Paths |
|---|---|
| `etl/steps/export/multidim/ihme_gbd/latest/air_pollution.config.yml` | 9 |
| `etl/steps/export/multidim/un/latest/un_wpp.manual.config.yml` | 5 |
| `etl/steps/export/explorers/un/latest/un_wpp.manual.config.yml` | 5 |

All pinned versions matched the current DAG dependencies, so this is again a no-op today.

### Verification

Ran each export step before and after the change and compared the generated outputs:

- Air pollution mdim (`air_pollution.config.json`): byte-identical.
- UN WPP explorer (`population-and-demography.config.json` and the generated `.tsv`): byte-identical.
- UN WPP mdim (`population-and-demography.config.json`): identical except for the order of the two entries in its `dependencies` list, which is pre-existing run-to-run churn (the list is serialized from a Python set). All 941 distinct `catalogPath` values in the output are identical and fully expanded.

The only full-form paths left under `etl/steps/export/` are two in the dummy issue-reporting config (`multidim/dummy/latest/issue_reporting.config.yml`), pinned to `covid/latest` so there is no version to drift, and one commented-out line in the democracy explorer config.

### DIFF
--- a/etl/steps/export/explorers/un/latest/un_wpp.manual.config.yml
+++ b/etl/steps/export/explorers/un/latest/un_wpp.manual.config.yml
@@ -481,7 +481,7 @@ views:
       variant: estimates
     indicators:
       y:
-        - catalogPath: grapher/un/2024-07-12/un_wpp/deaths#deaths__sex_all__age_0__variant_estimates
+        - catalogPath: un_wpp/deaths#deaths__sex_all__age_0__variant_estimates
   - dimensions:
       indicator: infant_deaths
       age: "0"
@@ -489,7 +489,7 @@ views:
       variant: estimates
     indicators:
       y:
-        - catalogPath: grapher/un/2024-07-12/un_wpp/deaths#deaths__sex_female__age_0__variant_estimates
+        - catalogPath: un_wpp/deaths#deaths__sex_female__age_0__variant_estimates
   - dimensions:
       indicator: infant_deaths
       age: "0"
@@ -497,7 +497,7 @@ views:
       variant: estimates
     indicators:
       y:
-        - catalogPath: grapher/un/2024-07-12/un_wpp/deaths#deaths__sex_male__age_0__variant_estimates
+        - catalogPath: un_wpp/deaths#deaths__sex_male__age_0__variant_estimates
   # Projection views combine projection + estimates as two y-indicators so
   # grapher renders a solid→dashed transition on the line chart AND the map tab
   # auto-combines them into the full 1950–2100 series (see
@@ -606,7 +606,7 @@ views:
       variant: estimates
     indicators:
       y:
-        - catalogPath: grapher/un/2024-07-12/un_wpp/mortality_rate#mortality_rate__sex_all__age_0__variant_estimates
+        - catalogPath: un_wpp/mortality_rate#mortality_rate__sex_all__age_0__variant_estimates
           <<: *display_mortality_rate
   - dimensions:
       indicator: child_mortality_rate
@@ -615,7 +615,7 @@ views:
       variant: estimates
     indicators:
       y:
-        - catalogPath: grapher/un/2024-07-12/un_wpp/mortality_rate#mortality_rate__sex_all__age_0_4__variant_estimates
+        - catalogPath: un_wpp/mortality_rate#mortality_rate__sex_all__age_0_4__variant_estimates
           <<: *display_mortality_rate
 
   # Projection views combine projection + estimates as two y-indicators (projection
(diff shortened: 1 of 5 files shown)


pabloarosado:
@codex review

paarriagadap:

Thanks @pabloarosado — nice cleanup, and the before/after config comparison made this easy to check. I verified the resolution independently off this branch's staging server: 57 views, 23 carrying `x` and `color`, all 23 resolving to exactly the two paths that were hardcoded (variables 953903 and 900801, same catalog paths on production). A genuine no-op.

Two things from my side:

**1. I pushed a test.** `expand_paths` was only covered for `y` and `sortColumnSlug`, so nothing pinned the post-creation expansion this change now depends on — and `incomes_pip.py` is currently the repo's only `set_indicator(x=..., color=...)` caller. Added `test_view_expand_paths_covers_non_y_dimensions_set_after_creation`, which asserts `x`/`color`/`size` assigned in short form after the view is built come out as full paths. Confirmed it fails if the expansion is narrowed back to `y` only.

**2. A few hardcoded paths are still around, maybe for a follow-up PR.** Python step code is clean — I swept `etl/steps/**/*.py` and there are none left. But 19 full-form pinned paths remain in three collection config YAMLs:

| File | Count |
|---|---|
| `etl/steps/export/multidim/ihme_gbd/latest/air_pollution.config.yml` | 9 |
| `etl/steps/export/multidim/un/latest/un_wpp.manual.config.yml` | 5 |
| `etl/steps/export/explorers/un/latest/un_wpp.manual.config.yml` | 5 |

They look like the same shortening would work on them, if you fancy a follow-up.

pabloarosado:

Thanks @paarriagadap, both points taken:

1. Your test rebased cleanly onto the new work and passes locally along with the rest of `tests/collections/test_model_dimension_view.py`.
2. We folded the follow-up into this PR rather than opening a new one: all 19 pinned paths in the three config YAMLs you listed are now short-form. Verified the same way as the original change — ran each export step before and after and diffed the generated outputs. The air pollution mdim config and the UN WPP explorer config + TSV are byte-identical; the UN WPP mdim config is identical except for the order of the two entries in its `dependencies` list, which is pre-existing run-to-run churn from an unordered set (all 941 distinct `catalogPath` values in the output are unchanged and fully expanded). Details in the updated PR description.

The only full-form paths left under `etl/steps/export/` are two in the dummy issue-reporting config (pinned to `covid/latest`, so nothing to drift) and one commented-out line in the democracy explorer config.

paarriagadap (approved):
All good!

PR #6589, opened 2026-08-03
🐛 MDim chart matcher: twin-variable suspects + stale-decision fix

paarriagadap:

Follow-up to #6588, from the human review round of the Economic Inequality chart→MDim mapping. The reviewer flagged 3 unmatched charts with "the equivalent view is X" — all three turned out to be the same failure shape, plus the fix path exposed a preflight bug.

## 1. Twin-variable suspects (extract_and_match.py)

ID-based matching is structurally blind to a dataset that publishes the **same series in two tables**: WID and LIS both carry e.g. `inequality#share_top_10__…` (used by the standalone charts) and `incomes#share__…quantile_10…` (used by the MDim) — different variable ids, values verified identical on every entity-year (3765/3765 and 1060/1060 points). Such pairs land in `none`/`near_miss` with nothing pointing at the equivalence.

The extractor now reports **twin suspects**: an unmatched chart whose y variables come from the same dataset as a slot-compatible view with a similar column name (overlapping head tokens + ≥50% token overlap, quantile spellings normalized so `share_top_1` finds `quantile_richest_1pct`). Reported in the run summary and in `mdim_suggestions.md` with the ready-to-paste `overrides.csv` line — explicitly as *suspects to verify by comparing the two indicators' data via the public API*, never as automatic matches. On the Economic Inequality run this catches all 3 reviewer-found twins, ranked first.

## 2. Stale-decision fix (preflight.py)

`--decisions` treated an exported decision with an **empty target** as carrying no target info, so its status still applied. But a decision exported while the chart had *no proposed redirect* (flagged as "should point at X") then survived onto the same chart after an `overrides.csv` force — and, being `flagged`, silently dropped the freshly forced redirect from the CLI CSV. An empty reviewed target on a targeted row is now a mismatch like any other: the decision reads as stale and the row is kept, counted as undecided. Cost: exports predating the target columns read entirely as stale, which only raises the undecided warning — conservative, never excluding.

## Validation

- Economic Inequality run: 19 exact + 3 forced = 22 proposed redirects, 0 ambiguous/conflicts; the 19 pre-existing rows byte-identical.
- Twin detector control run (overrides removed): all 3 known twins surfaced with the correct view ranked first; 3 remaining suspects are genuine non-twins (multi-indicator or extrapolated variants) for a human to reject.
- Preflight with the reviewer's original export: 3 old flags → `STALE DECISION … treated as undecided`, rows kept; 0 excluded.
- `make check` + ruff clean.

## Open items

- **Unverified** — the twin heuristic's thresholds (head-token overlap, 50% Jaccard, top-2) are tuned on this one migration; the next migration run will tell whether they generalize.

### DIFF
--- a/.claude/skills/map-charts-to-mdim/SKILL.md
+++ b/.claude/skills/map-charts-to-mdim/SKILL.md
@@ -142,6 +142,17 @@ it `none` would assert an overlap check that came out the other way. Quality lab
 - `none` — no view shares the chart's indicators. **An accepted outcome** — only
   matched charts get redirects; don't force the rest.
 
+**Twin suspects.** ID matching is blind to one real equivalence: a dataset that
+publishes the same series in two tables (WID/LIS do — `inequality#share_top_10__…`
+for the standalone charts vs `incomes#share__…quantile_10…` for the MDIM; verified
+value-identical). The extractor flags these — unmatched charts whose y comes from
+the same dataset as a slot-compatible view with a similar column name — in the run
+report and in `mdim_suggestions.md` as **twin suspects**, with the exact
+`overrides.csv` line to use. A suspect is NOT a match: verify first that the two
+indicators' values are identical (fetch both
+`api.ourworldindata.org/v1/indicators/<id>.data.json` and compare every
+entity-year), then force it, citing the verification in the override note.
+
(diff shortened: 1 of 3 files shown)


paarriagadap:
@codex review

paarriagadap:
@codex review

PR #6604, opened 2026-08-04
🐛 Color region maps by name instead of per-chart palettes

paarriagadap:


paarriagadap:
Duplicate of #6603 — closing.

PR #6622, opened 2026-08-05
📜 skills: Note the CSV redirect CLI's limits on query params and explorers

paarriagadap:

Feedback on the chart→MDim redirect handover: the CSV script would need changes if source query params or explorer sources were involved. Neither case is reachable from `map-charts-to-mdim`'s own output (sources are always bare `/grapher/<slug>` paths), so this documents the boundary before someone hand-edits a CSV into it.

## What the note records

Traced through `createMultiDimRedirectsFromCsv.ts` and grapher's serving path:

- **Source query params.** They pass the CLI's source regex and land inside `multi_dim_redirects.source`. Chart sources bake to a flat *slug* → path map (`SiteBaker.bakeRedirects` calls `getGrapherToChartAndMultiDimRedirects` with an empty prefix), consulted only when the URL 404s — so `/grapher/foo?tab=map` bakes as the key `foo?tab=map` and matches no request. Param-conditional routing is the `sourceQueryParams` column, which the CLI never writes **and** which the admin route rejects outright on a `/grapher/` source. So it is a change on both sides, CLI and serving — not only the script. `preflight.py` already rejects a source carrying a query string.
- **Explorer sources.** Worse than inert. `/explorers/...` passes the regex, and with `sourceQueryParams` NULL the row bakes as an *unconditional* rule — and explorer redirects are checked on **every** request, not just 404s (`functions/explorers/[slug].ts`). The whole explorer would 302 to one MDim view. The CLI also never unpublishes the source (`getChartSlugsToUnpublish` skips non-`/grapher/` sources) and its chained old-slug migration early-returns. Explorers go through the admin bulk endpoint, i.e. `map-explorer-to-mdim`.

## Changes

- `map-charts-to-mdim/SKILL.md` — new "one bare chart path per row" block after the CLI-vs-admin-API paragraph; the old *"Never put query params on the source"* gotcha collapses into a pointer to it (it now duplicated the section, and its "serving matches the bare path" wording was vaguer than the actual mechanism).
- `map-explorer-to-mdim/SKILL.md` — one paragraph stating the admin bulk endpoint is the only way to apply an explorer redirect, and why the CSV CLI is not a substitute.
- `docs/guides/data-work/redirect-to-mdims.md` — a reader-facing note in the *Charts → They run the CLI* section: one bare chart path per row, and why neither case is fixable by editing the CSV.

Docs only; no script or behavior change. `make check` passes.

## Open items

- **Unverified** — every runtime claim is read from code on `origin/master` in both repos, not executed. Nobody has fed a query-param'd or `/explorers/` source to the CLI; the failure modes are derived. The explorer one (a live catch-all 302 on a still-published explorer) would only be proven by a staging run.
- **Not addressed** — the CLI itself is unchanged. If a migration ever genuinely needs either case, that is a Grapher-side ticket; none is filed.
- **Correction made while writing** — the first draft claimed an explorer-sourced redirect "would never fire", by analogy with chart redirects. The explorer edge handler contradicts that; the note now says it fires on every request.

### DIFF
--- a/.claude/skills/map-charts-to-mdim/SKILL.md
+++ b/.claude/skills/map-charts-to-mdim/SKILL.md
@@ -36,6 +36,32 @@ at it, and its bulk endpoint accepts explorer sources only. The CLI
 (`owid-grapher/devTools/createMultiDimRedirectsFromCsv.ts`) exists precisely for
 the chart case.
 
+**And only for that case: one bare chart path per row.** Two things the CSV
+pipeline cannot express — if a migration needs either, the CLI itself has to
+change first, so raise it with a Grapher developer instead of bending the CSV
+around it:
+
+- **Source query params.** The source column is a path. Anything after `?`
+  survives into `multi_dim_redirects.source` verbatim, and chart sources are baked
+  as a flat slug → path map that strips only the `/grapher/` prefix
+  (`getGrapherToMultiDimRedirects` → `getMultiDimRedirectTargets`), so the row
+  matches no request and the redirect never fires. Conditioning on the incoming
+  params is what the `sourceQueryParams` column does, and the CLI never writes it —
+  grapher honors it for `/explorers/` sources only (they bake to a query-param
+  decision tree resolved at the edge), and the admin route rejects it outright on a
+  `/grapher/` source. A per-params chart redirect therefore needs changes on both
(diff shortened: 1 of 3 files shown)


paarriagadap:
@codex review

PR #6632, opened 2026-08-07
📜 Clarify whether snapshot scripts should use click decorators

paarriagadap:

**This needs a decision from someone who knows the history — I'm raising the question, not asserting the answer.** The change implements what I believe the intended convention is; if it's wrong, the fix is to flip it and change the cookiecutter instead.

## The problem

Three places tell you three different things about whether a new snapshot script should use `click`:

| Source | Says |
|---|---|
| `CLAUDE.md` "Running Snapshot Steps" | forbids the `__main__` guard, and in the same sentence calls `run` "its **click command**" |
| `apps/wizard/etl_steps/cookiecutter/snapshot/` | emits a plain `def run(upload: bool = True) -> None:` — no click anywhere |
| `.claude/skills/create-snapshot/` | "no `click` decorators and no `if __name__ == "__main__":` block" |

The guidance loaded into every session is the one pointing the other way. It only ever forbids the `__main__` guard, and describing `run` as a click command reads as an endorsement of the decorators.

## Why it matters

Both forms work today, so nothing is broken. `etl/snapshot_command.py` imports the module, prefers a `main` over a `run`, and `_call_snapshot_function` branches on `isinstance(func, click.Command)` — click commands and plain functions are both invoked correctly. This is a convention question, not a bug report.

**The convention already changed, almost a year ago.** The snapshot cookiecutter stopped emitting click on **2025-08-12**, in the commit that introduced the `etls` command (#4517) — logically, since `etls` made the CLI wrapper redundant. No cookiecutter template contains click today: not the decorators, not even an `import click`. So click-decorated scripts written before that date were correct for their time.

What is left is a tail of scripts written *after* it. Of the 1,171 click-using scripts under `snapshots/`:

| | count |
|---|---|
| in version folders before 2025-08-12 (click was the convention) | 945 |
| **in version folders on/after 2025-08-12** | **125** |
| in `latest`/undated folders (continuously updated, so ambiguous) | 101 |

Cross-checking a different way: 65 scripts carry `@click` but no `__main__` guard — the half-modernized shape the current wording literally asks for — and the ones I sampled are dated 2026-06 and 2026-07. So new scripts are still being written with click, most plausibly because the guidance loaded into every session describes `run` as a click command.

## What this PR does

1. **Rewrites the CLAUDE.md bullet** to state the convention the cookiecutter already implements, and adds that existing click-decorated scripts keep working — so nobody reads it as a call to migrate 1,171 files.
2. **Adds the matching `@click` pattern to `detect-outdated-practices`**, scoped to `snapshots/**`. Without it, the guidance is unenforced: the detector has 11 patterns and none of them matched click, so `/check-outdated-practices` reported clean on a script full of decorators.

### Pattern design

Anchored on `@click.command`, not `@click.` generally. Measured over `snapshots/`:

| Anchor | Files flagged | Warnings emitted |
|---|---|---|
| `@click.` (every decorator) | 1,171 | 2,647 |
| **`@click.command` (this PR)** | **1,171** | **1,171** — one per file |

The narrower anchor misses nothing: every file carrying any `@click.` decorator also carries `@click.command`, and none use `@click.group`. Verified that the pattern emits exactly one warning per flagged file, and that the 32 legitimate click CLIs under `etl/` and `apps/` fall outside the `snapshots/**` scope. The snapshot cookiecutter template itself comes back clean.

The message asks for the **whole** decorator stack, because removing `@click.command` while leaving the `@click.option` lines above it breaks the script.

## The trade-off, stated plainly

The pattern fires on **1,171 of 1,229** snapshot scripts — 95% of the directory. That is a real cost, accepted deliberately rather than overlooked:

- **Most of those warnings are archaeological.** 945 sit in version folders predating 2025-08-12, on scripts that were correct when written and that nobody is going to reopen. Only ~125 (plus some of the 101 undated) are scripts written after the convention changed.
- **Most flagged files already carry a warning anyway.** 1,106 of the 1,171 also trip the existing `__main__` pattern, so the incremental noise is concentrated in the 65 files that currently look clean but are still click-decorated.
- **The risk is habituation.** A warning firing on almost every file in a directory trains people to ignore the squiggles, which devalues the other 11 patterns that point at genuine problems.

**Why the pattern is still worth keeping despite that:** no cookiecutter emits click — not the snapshot template, not any other, not even an `import click`. So the flagged set is *closed*. Every warning is on a hand-written or copied script, nothing generates fresh ones, and the count can only fall as files get touched. That is what separates this from a pattern fighting a live generator: the `dest_dir` warnings needed the fasttrack template fixed before the pattern made sense, whereas here the generator has been compliant since 2025-08-12.

A cleanup sweep of the ~125 post-2025-08-12 scripts would cut the noise a lot and can happen independently, whenever someone wants it — the pattern doesn't depend on it.

## What this PR deliberately does *not* do

- **No change to the 1,171 existing scripts.** They keep working; this is guidance plus a warning, not a migration.
- **No `.vsix` repackage.** #6630 already bumped the extension to 0.0.6 and packaged it, so building a second 0.0.6 artifact here would collide. The extension should be packaged and installed **once, after both PRs land** — until then this pattern exists in source but is not in anyone's editor.

## Merge order with #6630

Both PRs edit `vscode_extensions/detect-outdated-practices/src/extension.ts`, so whichever merges second needs a rebase. Two follow-ups belong to whoever does that:

- #6630 introduces `SNAPSHOT_SCOPE` (adding `apps/wizard/**/cookiecutter/**` to the snapshot-scoped patterns). After both land, this new pattern's `scope: 'snapshots/**'` should become `SNAPSHOT_SCOPE` so the cookiecutter template is covered too. The template is click-free today, so this is drift protection, not a current gap.
- Package and install the extension once, covering both changes.

## If the answer is the other way round

If click on `run()` is actually the preferred form, then both changes here are wrong and the real fix is in `apps/wizard/etl_steps/cookiecutter/snapshot/{{cookiecutter.namespace}}/{{cookiecutter.snapshot_version}}/{{cookiecutter.short_name}}.py` plus the `create-snapshot` skill. Say so on this PR and I'll turn it around.

## Still open

**Handed off** — someone with the context decides which of the two conventions is right. That is the only open question in this PR.

**Proposed** — a cleanup sweep of the ~125 click-decorated scripts in version folders dated on/after 2025-08-12, to cut the warning volume. Independent of this PR and not a prerequisite for it.

**Unverified**
- The counts come from `grep -rl` over `snapshots/**/*.py` and include archived snapshots, since the directory keeps them; I did not split active from archived.
- The before/after-2025-08-12 split uses each script's **version-folder date** as a proxy for when it was written, not git history. A script in an old version folder that was edited recently counts as "before", and the 101 in `latest`/undated folders can't be placed at all.
- I sampled 3 of the 65 `@click`-without-`__main__` files to date them, not all 65.
- The pattern was verified by reproducing its regex over the corpus and by compiling the extension (`npm run compile` succeeds). It has **not** been exercised inside a running VSCode instance, because the `code` CLI is unavailable in the environment this was written in — so the diagnostic has never actually been seen in an editor.
- `test_examples.py` (the manual fixture) gains no click example. It covers only 4 of the 11 patterns and already omits the other `snapshots/**`-scoped one (`__main__`), so extending it is a separate piece of work.


paarriagadap:

Closing — both commits here have been folded into #6630 rather than kept as a separate PR.

The reason for splitting them off was that #6630 was already large, and the click question needed a decision rather than a review. Two things removed that reason:

- **No cookiecutter emits click**, so the flagged set is closed: nothing generates fresh violations, and the warning count can only fall as scripts get touched. That settles the trade-off that made this feel like a separate decision.
- Both PRs edited `vscode_extensions/detect-outdated-practices/src/extension.ts`, so keeping them apart meant a guaranteed rebase and two competing `0.0.6` extension builds.

In #6630 the pattern also picks up the wider `SNAPSHOT_SCOPE` introduced there, so it covers the cookiecutter templates as well — drift protection, since they have been click-free since the commit that added `etls`.

The convention question itself is unchanged and still open for review, now on #6630: should new snapshot scripts be click-free? The evidence is carried over — the template dropped click on 2025-08-12 alongside `etls`, and 125 scripts in version folders dated after that still use it.

PR #6633, opened 2026-08-07
🔨🤖 Drop the SVG tester staleness check

sophiamersmann:

`is_stale` asked the wrong thing which commit is under test. It compared the results file's `grapherCommit` against the HEAD of `~/owid-grapher`, but that checkout is only moved to the new commit by the Grapher build step, which races the first owidbot run of the same build — and owidbot, a two-second step, normally wins. At that moment both sides still name the *previous* commit, so leftover results read as fresh. The case the check was written for, the same branch with a new commit, is precisely the one it was blind to; what it could catch instead was a suite whose good results it then dropped on the floor.

Rather than fix the comparison now, this removes it:

- `is_stale` and `get_head_commit` are gone, along with the `git` import that only they used.
- `load_suite_results` collapses into `read_verify_results` — no `{"status": "stale"}` sentinel — and `run` no longer needs the grapher checkout path.
- `has_results` is gone; `svg_tester_has_run` is now just "any suite produced a file".
- `make_suite_line` loses its `stale` branch (`_skipped_ (ignored a leftover results file)`).

A results file is now reported with its counts and report link whatever commit it was written for. The `unreadable` and `running` states are untouched.

The trade-off is that during a re-run the comment can show the previous commit's numbers without saying so — the same behaviour as before the check existed. Labelling those rows with the commit they came from, which is the fix that actually works, is described in `docs/svg-tester-comment-freshness.md` in `owid-grapher` and is deliberately not attempted here.

## Testing

`make check` passes. No tests referenced the removed helpers; the only caller of this module is `apps/owidbot/cli.py:95`, whose `grapher.run(branch)` call is unchanged.

## Open items

- The staleness problem itself is unfixed and stays unfixed until the labelling work happens.

### DIFF
--- a/apps/owidbot/grapher.py
+++ b/apps/owidbot/grapher.py
@@ -1,7 +1,6 @@
 import json
 from pathlib import Path
 
-from git import InvalidGitRepositoryError, NoSuchPathError, Repo
 from structlog import get_logger
 
 from etl.config import get_container_name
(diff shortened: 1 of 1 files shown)


sophiamersmann:
@codex review

PR #6637, opened 2026-08-07
📜 Loosen the open-items rule and move its detail out of CLAUDE.md

paarriagadap:

## Why

`CLAUDE.md` carried a 10-line section, *"Close every report with what's still open"*, mandating a three-bucket open-items block on every report. Having it there had it backwards in two ways:

- **The rigid part was the always-on part.** A required output format sitting in the agent guide gets applied literally, to every interaction — including work whose loose ends are "nothing's committed yet". Three labeled headings on top of two items is scaffolding, not information.
- **The useful part was already elsewhere.** What actually tends to dangle differs per workflow — a dataset update leaves content edits and follow-up-PR scope, an audit leaves failed fetches and capped sweeps — and five skills already spelled that out, each cross-referencing the CLAUDE.md heading by name. So the format loaded every session and the specifics loaded only when a skill ran.

This inverts it: a loose one-line reminder always on, the checklist read on demand, the per-workflow specifics untouched.

## CLAUDE.md: 10 lines → 1

Removed:

```markdown
## Close every report with what's still open

Multi-step work — updates, audits, reviews, migrations — rarely ends with everything closed.
End the report (and the PR body, when there is one) with an explicit open-items block; never
make the user ask "what's left?". Keep three buckets separate:

- **Handed off** — waiting on someone else. Name who acts and what the ask is, and include a
  locator (link, file, id) — an item the next person must re-derive is not handed off.
- **Proposed** — waiting on sign-off. State the exact change so "yes" is a complete answer.
- **Unverified** — nobody checked it: skipped or failed checks, surfaces left uncovered, capped
  or truncated sweeps. Silence here reads as "clean", which is the one wrong signal a report
  can send.

Re-state the whole block every time the work is revisited — carry open items forward rather
than reporting only the delta, and say explicitly when an item clears instead of dropping it
silently.
```

Added, as one bullet under Critical Rules:

```markdown
- **Say what's left open.** Multi-step work rarely ends with everything closed, so close the
  report (and the PR body) by saying what's still pending, who owns it, and what nobody
  checked. No fixed format — `.claude/docs/open-items.md` lists what tends to get dropped.
```

344 → 336 lines.

## New: `.claude/docs/open-items.md`

The same three categories, but described rather than labeled — *waiting on someone else* / *waiting on a decision* / *nobody checked it* — and framed as a checklist to think against, not a block to emit: *"Not an output format: there is no block to fill in and no headings to emit."* It also names the failure mode in the other direction, which the old wording invited:

> Over-applying this is its own failure. Three labeled headings on top of two loose ends is scaffolding, not information — a chat report with one thing pending needs one sentence.

Listed under Extended Documentation, so it loads when a pointer sends you there instead of every session.

## Skills

Each skill keeps its own dangler list verbatim — that was the part worth having — and now points at the doc instead of the deleted heading.

| Skill | Change |
|---|---|
| `update-dataset` | Points at the doc; keeps the formal PR-body requirement and the **fourth category** (deferred to a follow-up PR), since that list *is* the follow-up PR's scope |
| `review-data-pr` | 🟡 check softened from "does the body carry that block" to whether pending work is mentioned at all, judged "on whether a reader can tell what's outstanding, not on whether it uses any particular headings" |
| `check-empty-entities`, `check-hardcoded-years`, `edit-faust-metadata` | Heading `### Always close…` → `### Close…`; the dual-surface demand *"in the PR body as well as chat"* → *"a line or two in chat, written out in the PR body when there is one"* — the old phrasing is how a chat report ends up carrying a PR-body-shaped block |
| `pr-babysitter` | Three uses of "unverified" as a term of art now read in plain language; it never referenced the rule, so the vocabulary would have been orphaned |

## What deliberately stays structured

The audit skills' `## Report format` blocks (findings tables by severity), `review-data-pr`'s numbered §14 report template, and `update-dataset`'s PR-body requirement. Those are the cases where a written-out list earns its structure — a PR body outlives the chat, and a follow-up PR's scope can't be reconstructed later without redoing the analysis.

## Notes for review

- Docs and skill prose only — no Python, no pipeline effect. `make check` passes.
- No dangling references left: nothing in the repo still points at the removed heading.
- The change only takes effect for **new sessions** — a session that already loaded the old text keeps it in context until it ends.


paarriagadap:
@codex review

PR #6642, opened 2026-08-10
✨ Always link the SVG tester in owidbot comment

sophiamersmann:

The SVG tester link in the owidbot comment was only shown once at least one suite had written its results. But the first owidbot run of a build happens before the SVG tester step, so on a fresh build the link was missing exactly when you'd want to click through and watch it.

The link now sits next to the site-screenshots link unconditionally. The per-suite results block still appears only once there are results to show.

```
- **Site-screenshots:** https://github.com/owid/site-screenshots/compare/<branch>
- **SVG tester:** http://staging-site-<branch>/admin/svgtester
```

The `<details>` block at the bottom of the comment keeps its **SVG tester:** heading, so it repeats that label when results exist — left as is for now.

### DIFF
--- a/apps/owidbot/grapher.py
+++ b/apps/owidbot/grapher.py
@@ -21,8 +21,8 @@ def run(branch: str) -> str:
 
     results_by_suite = {suite: read_verify_results(svgs_repo / suite) for suite in SVG_TESTER_SUITES}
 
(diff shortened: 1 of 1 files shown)


sophiamersmann:
@codex review

PR #6438, opened 2026-07-14
✨ description_key becomes a free-form markdown string

Marigold:
> _Written by Claude Fable 5 — @Marigold at the wheel._

## Summary

`description_key` ("What you should know about this data") is no longer an array of bullet points — it becomes a **single free-form markdown string**, so data pages can use prose, paragraphs, headers, and lists where they help:

```yaml
description_key: |-
  Across the world, people are living longer.

  The measure shown is *period* life expectancy, which summarizes
  death rates across all age groups in one particular year.
```

A YAML list remains valid as **authoring sugar** (converted to a markdown list), so all ~400 existing `.meta.yml` files keep working unchanged and render exactly as before — one entry as prose, several entries as a bulleted list. No YAML migration needed; a follow-up PR may clean files up into prose where it reads better.

📖 **[Architecture overview](https://htmlpreview.github.io/?https://gist.githubusercontent.com/Marigold/39ae4beaaebf035daad7233eccb5d098/raw/description-key-string-architecture.html)** — the full pipeline, where the list→string conversion happens and why, legacy-data handling, and deployment order.

## How it works

- **Lists survive until after Jinja rendering.** Items may contain whitespace-control tags (`<%- … -%>`) that must render per-item; converting earlier lets Jinja eat the markdown list structure (a real bug caught during testing). The single conversion point is `update_variable_metadata`, which runs post-Jinja on the way to the grapher — MySQL and the indicator metadata JSON only ever see strings.
- Garden datasets and feather sidecars are unchanged on disk — no catalog rebuild.
- Legacy lists from old DB rows / API payloads convert at read time (datasync, `owid-catalog` charts API, schema.org export).
- Combining indicators (variable ops, MDim grouped views): an identical `description_key` passes through unchanged; distinct ones are concatenated as markdown blocks with whole-text dedup (previously per-bullet dedup — a deliberate simplification).
- Collection (MDim) configs convert list sugar at upsert (anchored `- *anchor` nesting is flattened first, as in dataset YAML).
- The `_validate_description_key` guard (string exploded into single-character bullets) is obsolete — a string is simply valid now — and removed.

## Live demo

The branch temporarily rewrote the long-run life expectancy `description_key` into prose with markdown headers as a staging demo of the format. That editorial change was **reverted before merge** (ec3fd09871) — this PR ships the format change only, no content changes.

## Companion PR — merge order matters

[...]

PR #6650, opened 2026-08-12
🐝 Republish stale R2 indicator metadata as descriptionKey strings

Marigold:

R2 indicator metadata files are only rewritten when a variable is upserted, so every dataset not re-run since #6438 still serves `descriptionKey` as a legacy array — **88,214 variables across 396 datasets**, while the grapher DB has been string-only since the migration. owid-grapher absorbs the difference by normalizing at four ingress points, and those shims can only be removed once no published file holds an array. This adds `etl d republish-description-key`, which closes that gap without re-running a single ETL step.

_Skim: one new command, no existing code paths touched. The judgement worth checking is the approach — patch one field in place rather than rebuild the file — argued below._

**How:** for each variable, fetch the published metadata file, and if `descriptionKey` is a list, rewrite **only that field** with `description_key_to_string`. Everything else — `dimensions`, `display`, `origins`, `metadataChecksum`, `updatedAt` — is left byte-identical, so the file remains exactly what the last ETL run produced apart from this field's shape.

**Why patch instead of rebuild.** The obvious alternative is to rebuild the file from the DB via `dm.variable_metadata()`, the way an upsert does. I tried that first and rejected it:

- it needs each variable's `.data.json` to recompute `dimensions` — 88k data downloads for a field that isn't changing;
- it also rewrites `updatedAt` and `metadataChecksum`, so the diff stops being "one field" and starts being something a reviewer has to reason about.

The patch is also the safer of the two on rendering: `description_key_to_string` is the same conversion owid-grapher's `normalizeDescriptionKey` already applies to these files at runtime, so the rewritten file says exactly what the browser renders today. Verified on 40 stale variables that the locally converted text is byte-identical to the string MySQL holds — **40/40 match** — so this also brings R2 into agreement with the DB rather than inventing a third version of the text.

## Verified

Ran end to end against a staging namespace (`s3://owid-api-staging/…`), 60 variables:

- **Shape**: 60/60 arrays became strings.
- **Nothing else moved**: diffing a rewritten file against its production counterpart field by field, the only difference is `descriptionKey`.
- **Text is unchanged**: the written string equals `description_key_to_string(production array)` exactly — i.e. what grapher computes at runtime.
- **Rendering**: the affected data page renders the same 5 bullets, the chart fetches the rewritten file (`200`), and the browser console is clean.
- **Idempotent**: a second run reports `already-string=60` and writes nothing.
- **Production dry run** (read-only) over 200 candidates reports `would-rewrite=200`, confirming the selection and the production target resolve correctly.

Not verified: no production file has been written — the full run is a separate, deliberate step (commands in the appendix).

## This has been run against production

All 146,645 published files now hold a string; **zero arrays remain**, verified by reading every object from R2.

| Run | Result |
|---|---|
| 50-file trial | `rewritten=50`, only `descriptionKey` changed in all 50, text byte-identical to the runtime conversion |
| Main run (`--cutoff 2026-07-21`) | `rewritten=90,259`, `already-string=51`, **0 failed, 0 changed-underneath**, ~36 min at 16 workers |
| Stragglers | `rewritten=211` — see below |
| Final sweep over all 146,645 | **146,645 strings, 0 arrays, 0 missing** |

**The cutoff missed 211 files, and this is the lesson worth keeping.** `--cutoff 2026-07-21` selects on `datasets.dataEditedAt`, which is only a proxy for when a *file* was written. Those 211 belonged to datasets edited on 21 July between 08:37 and 12:04 — hours before #6438 landed at 13:06 UTC — so their files were pre-migration while their timestamps were not. A second way it under-selects: `upsert_table` skips variables whose checksums are unchanged, so a re-run dataset can still hold stale files. The command's own content test is the sound one, so `--cutoff` is now documented as a speed optimisation that can miss files, not the recommended path.

## How to run it

Dry run by default; `--execute` is the only thing that writes.

```bash
# 1. read-only: how many files still hold an array (no cutoff — content decides)
ENV_FILE=.env.prod.write .venv/bin/etl d republish-description-key

# 2. small live batch first
ENV_FILE=.env.prod.write .venv/bin/etl d republish-description-key \
    --limit 100 --execute --state-file /tmp/republish.txt

# 3. the rest, resumable — rerunning skips ids already in the state file
ENV_FILE=.env.prod.write .venv/bin/etl d republish-description-key \
    --workers 16 --execute --state-file /tmp/republish.txt
```

Drop `ENV_FILE` to target your own namespace, or set `STAGING=<branch>` for a branch's staging server.

**Verify against R2, not the API.** The CDN serves these files long past their stated `s-maxage` without revalidating — one object was still serving the old shape 90 minutes after being rewritten — so a spot check through `api.ourworldindata.org` makes a correct run look broken. Use `?nocache` if you must go through HTTP. And give any verification script the same retries the command has: a 24-worker scan without them reported 200 phantom "missing" files that a serial re-check found present.

### Where this sits in the migration

| Egress from ETL | State before this PR |
|---|---|
| `variables.descriptionKey` (MySQL) | string-only — 0 arrays in 108,867 rows |
| mdim configs | string-only — 18 carry the field, 0 arrays |
| explorer configs | field never appears |
| **R2 indicator metadata** | **mixed — arrays wherever a dataset predates the migration** |

After a full run, the last row joins the others and owid-grapher can drop `normalizeDescriptionKey` along with its call sites in `db/model/Variable.ts`, `grapher/src/core/loadVariable.ts` (×2), `types/src/dbTypes/Variables.ts` and `utils/src/MultiDimDataPageConfig.ts`, and tighten `parseVariableDescriptionKey` to reject an array instead of converting it. That is a separate PR in owid-grapher, and it should land only after the run is confirmed complete.

This is independent of migrating the ~3,861 `description_key` YAML lists in `etl/steps/**` to strings. Those are an authoring convenience that never reaches an egress in list form; they can stay as long as we like without blocking any of the above.

### Failure handling

Fetch and upload errors are logged per variable, counted, and the command exits non-zero with the count so a rerun retries them; only successfully processed ids are written to the state file. Progress is logged every 500 variables with a running tally.

</details>

### DIFF
--- a/apps/cli/__init__.py
+++ b/apps/cli/__init__.py
@@ -76,6 +76,7 @@ SUBGROUPS = {
             "map-datasets": "apps.utils.map_datasets.cli",
             "scan-chart-diff": "apps.utils.scan_chart_diff.cli",
             "profile": "apps.utils.profile.cli",
+            "republish-description-key": "apps.utils.republish_description_key.cli",
         },
     },
 }
--- a/apps/utils/republish_description_key.py
+++ b/apps/utils/republish_description_key.py
@@ -94,8 +94,12 @@ def _write_metadata(variable_id: int, metadata: dict[str, Any], etag: str) -> No
 def _candidate_ids(cutoff: str | None, limit: int | None) -> list[int]:
     """Variable ids to inspect, newest dataset first so a partial run covers the visible ones.
 
-    The cutoff is only a way to skip variables whose file is already known to be fresh — the
-    file itself is always what decides whether a rewrite happens.
+    The cutoff only skips work; the file's own content always decides whether it is rewritten.
+    It is a poor filter and is off by default: `datasets.dataEditedAt` says when the dataset was
+    touched, not when a given variable's file was written, and `upsert_table` skips variables
+    whose checksums did not change. A date-only cutoff is worse still — the first production run
+    used `< '2026-07-21'` and missed 211 files whose datasets were edited that same morning,
+    hours before the migration landed at 13:06 UTC.
     """
     where = "v.descriptionKey IS NOT NULL AND v.descriptionKey != ''"
     if cutoff:
@@ -155,8 +159,10 @@ def _rewrite_one(variable_id: int, execute: bool) -> str:
 @click.option(
     "--cutoff",
     default=None,
-    help="Only inspect variables whose dataset was last edited before this date (e.g. 2026-07-21). "
-    "Skips files already known to be fresh; omit to inspect every variable.",
+    help="Only inspect variables whose dataset was last edited before this timestamp (e.g. "
+    "'2026-07-21 13:06:56'). A speed optimisation that can MISS files: the dataset's edit time "
+    "is only a proxy for the file's age, and a dataset re-run does not rewrite variables whose "
+    "checksums are unchanged. Prefer omitting it — the file's own content is the sound test.",
 )
 @click.option("--limit", type=int, default=None, help="Only inspect this many variables.")
 @click.option("--workers", type=int, default=16, show_default=True, help="Concurrent requests.")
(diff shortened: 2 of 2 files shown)


Marigold:
@codex review

PR #6653, opened 2026-08-12
🐝 Add seaborn for static visualization styling

paarriagadap:

Adds `seaborn` to the `dev` dependency group so that `export://static_viz` steps can share one visual style.

## Why

The static visualizations we hand off to Figma are currently styled ad hoc, one step at a time. Standardizing on seaborn's `ticks` style preset and `deep` palette gives them a common look and makes the emitted SVGs predictable to work with downstream.

## Placement

Alongside `matplotlib` in the `dev` group rather than in `[project] dependencies`, because only `export://static_viz` steps import it — no data step in `snapshot`/`meadow`/`garden`/`grapher` does. That matches how `matplotlib`, `plotly` and `kaleido` are already declared.

## Footprint

One package. `numpy`, `pandas`, `matplotlib` and `scipy` — everything seaborn depends on — were already in the environment:

```
+ seaborn==0.13.2
```

The `uv.lock` diff looks larger than that because `uv` reformats the dev group's dependency list, but exactly one package entry is added.

## Open items

- Nothing blocking. This is deliberately separate from the static viz that motivated it, so the dependency change can be reviewed on its own.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,6 +142,7 @@ dev = [
     "sqlacodegen>=3.1.0",
     "codespell>=2.4.1",
     "playwright>=1.60.0",
+    "seaborn>=0.13.2",
 ]
 
 [project.optional-dependencies]
--- a/uv.lock
+++ b/uv.lock
@@ -13,7 +13,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "2026-07-26T10:51:54.004645Z"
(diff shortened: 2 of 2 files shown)


Marigold (approved):
Up to you, but isn't seaborn pretty much dead? The last version is from [January 2024]((https://seaborn.pydata.org/whatsnew/v0.13.2.html))

paarriagadap:
@Marigold Ah! Didn't realize that. Anyway, it's mostly based on Matplotlib, and it makes it look nicer. I will think if it's worth keeping, but for now I merge. Thanks!

PR #6669, opened 2026-08-13
📜 Record the static chart templates' header and footer frame bands

paarriagadap:

## What this is

Docs-only. It records a structural change made in the design team's **Charts (2026)** Figma file, so the two skills that build on those templates don't read a stale band.

## The Figma change

`Static Chart Template_Horizontal` (850×638) kept its seven elements as loose children of the frame, while `Static Chart Template_Vertical` and the mobile templates group theirs. It now matches the vertical, name for name:

```
Static Chart Template_Horizontal   850×638
├─ Frame 5   header, 0→134, vertical auto-layout, hug height
│  ├─ Frame 2   title row  ├─ title (fill)  └─ Frame 1 → logo
│  └─ Frame 4   subtitle row → subtitle (fill)
└─ Frame 8   footer, 540→638, bottom-pinned (MIN/MAX constraints)
   ├─ Note (fill)  ├─ Data source (fill)
   └─ Frame 7   tagline ↔ CC-BY licence (space-between)
```

Nothing moved. The vertical's grid — 16.2162 side padding, 16.2162 title↔logo gap, fill widths — reproduces the horizontal's existing positions exactly, so the wrappers went around the elements rather than nudging them. Measured drift across all seven elements is ≤0.0002 px, and the re-rendered frame is pixel-identical to the baseline apart from one antialiased pixel differing by 8/255.

Two properties deliberately differ from the vertical, because matching them would have moved something:

| Property | Vertical | Horizontal | Why |
|---|---|---|---|
| `Frame 5` itemSpacing | 5.4054 | **6** | the horizontal's own title→subtitle gap is 6 px |
| `Frame 1` paddingTop | 6.0811 | **0.0811** | its logo sits 6 px higher against a 58 px title (vs 60 px) |

Everything else — layout mode, padding, alignment, sizing modes, constraints, empty fills, clipping, and every child's fill/hug mode — matches the vertical property-for-property across all six frames.

## Why the docs needed a line

Both skills document each template's chart band. The horizontal now exposes header/footer frames a script can read, and those frames include 16 px of inner padding on the chart side — so `header.y + header.height` returns **134** where the docs list the subtitle bottom at **118**. Same relationship on the vertical (frames 116/997, text 100/1013), where the table already used frame bounds.

- `create-figma-chart/SKILL.md` — a note under the band table spelling out frame band vs text band for both 850-wide templates, and which convention each row uses.
- `create-static-viz/TEMPLATES.md` — the horizontal's frame blocks recorded alongside the vertical's, with the slot table left as-is.

No recommended chart area changes from the restructure itself: every slot y stayed correct, because no element moved.

## Re-verified after the 2026-08-14 template edits, and two corrections

The Templates page was edited again after the restructure, so every documented figure was re-measured against the live file — all four static templates, DI, IG square, IG portrait, and both small-chart templates. **41 assertions, 41 passing**, after these two fixes:

**1. The Vertical template's subtitle grew to two lines.** It went from one line at 15px to two at 16px, so both desktop templates now carry the same subtitle style. That pushed its header block from `0→116` to `0→136` and took 20px off its chart band:

| | Before | Now |
|---|---|---|
| Vertical subtitle | 15px, one line, 18 tall | **16px, two lines, 38 tall** |
| Vertical header block `5332:94` | 0→116 | **0→136** |
| Vertical chart band | 116 → 997 | **136 → 997** |

Corrected in the band table, in TEMPLATES.md's Vertical section, and in the band note. A chart placed at the old 116 now lands inside the header instead of below it. The titles still differ — 29px line height on Horizontal against 30px on Vertical, so 58 vs 60 — which accounts for the whole 134 vs 136 gap between the two header blocks.

**2. The mobile logo x was the title's right edge, not the logo.** TEMPLATES.md recorded `x=444`; the logo actually starts at **460**, one 16px auto-layout gap after the 428-wide title. Pre-existing error, unrelated to the restructure, fixed here since the sweep surfaced it.

Everything else held: the Horizontal slot table, both mobile footers (`Frame 15`, 38 tall, rows 21px apart, ids `25343:276`/`25343:275`), DI's 118 → 508, IG square's 118 → 488, IG portrait's 135 → 640, and both small-chart bands (header bottom 44; `H − 10` guided, `H − 23` pull).

## Needs master merged in, and one correction that comes with it

#6664 is concurrent work touching both of these files. A trial merge is **clean** — no conflicts, and the merged band-table section reads in a sensible order — but it leaves one statement that is no longer true, which git cannot catch:

> | Static Horizontal (`5332:75`) | rows are **absolutely positioned** | they stay put; the note→source gap opened from **5.41px to 19.41px** |

The horizontal's footer is now the same bottom-pinned auto-layout `Frame 8` as the vertical's, so that row and the `DESIGNED_GAP` re-spacing workaround under it no longer apply to it. #6664 also carries a second claim that the 2026-08-14 subtitle edit has since retired — that the two desktop templates disagree on subtitle size (16px vs 15px). They now agree, and share one style.

Four spots on #6664's branch will need trimming after the merge, none of them reachable from this branch: its footer-behavior table and the `DESIGNED_GAP` block under it, its template-index row ("differ in ways that bite — footer re-spacing behavior and subtitle size"), and its subtitle-size bullet in TEMPLATES.md.

Measured on throwaway clones of both templates, shrinking the note from two lines to one:

| | Horizontal | Vertical |
|---|---|---|
| note→source gap | stays 5.405 (was: opened to 19.41) | stays 5.406 |
| licence bottom edge | unchanged, 621.62 | unchanged, 1078.81 |
| footer frame top | moves down 14 → band grows | moves down 14 |

So both desktop templates now reflow on their own and keep every footer row on its designed edge. The follow-on rule in that section still holds — the band's bottom moved, re-fit the chart — it just reads off `footer.y` now instead of needing a manual re-space.

**Sequencing:** #6664 merges first (it is out of draft, this one is not), then master gets merged in here and that table plus the workaround are corrected in this PR, since the restructure is what invalidated them.

## Left open

- **The templates will likely be adjusted further by design.** When that happens the node IDs and pixel figures here go stale — both docs already carry the "verify against the actual clone with `get_metadata`" caveat, and that remains the operative instruction. This PR is a snapshot, not a contract.
- Naming is still not unified across the template family: the 850-wide pair uses `Frame 5`/`Frame 8` (full-bleed, padding inside), mobile uses `Frame 14`/`Frame 15` (inset at x=16, y=16). The horizontal was matched to the vertical because that was the target; unifying all four is a separate decision.
- Verification was metadata plus a pixel diff. Nobody has yet opened the frame in Figma and eyeballed the layers panel.
- The Figma edit itself is live in the file and outside version control — this PR can only record it, not revert it.


paarriagadap:

Closing this in favor of #6664, so there is one PR to merge instead of two.

Both content commits were cherry-picked over (`6df83bb75`, `6ae528aae`), keeping every dimension recorded here: the frame-band vs text-band distinction on the two 850-wide templates, the Horizontal's `25398:753` / `25398:769` blocks, the Vertical's corrected `136 → 997` band, and the mobile logo at x=460. Only one conflict, in the band table, where this branch's Vertical row met #6664's added rows — resolved keeping both.

The reconciliation this PR's description predicted is done there too (`ac6310322`). Before deleting either claim I re-measured against the live file, since one of them came from an observed incident rather than an assumption:

- Both desktop footers are bottom-pinned auto-layout `Frame 8`s with every child flowed, so the manual re-space no longer applies to the Horizontal. It survives as a condition to *detect* rather than as history, because pages cloned before the restructure still behave the old way.
- The subtitle-size divergence is gone: both are 16px over two lines, 38 tall. The remaining difference is title line height, 29px against 30px — the entire 134 vs 136 gap between the header bands.

Two things went further than this PR proposed, at the owner's request: the two skills' overlapping template docs were split by ownership (measurements in `TEMPLATES.md`, operations in `create-figma-chart`), and the running history was removed from both in favor of a single dated `Last verified` stamp per file. So the "left open" item here about node IDs going stale is now handled by that stamp plus the re-verify instruction rather than by a changelog.

Nothing from this branch is lost — it is all in #6664, verified against the live file on 2026-08-14.

PR #6672, opened 2026-08-14
🐛🤖 Clean up orphaned chart_configs when deleting ghost variables

sophiamersmann:

`cleanup_ghost_variables` deletes variables with raw SQL and clears every table that
references them, but it never touches `chart_configs`. Every FK into `chart_configs`
points *outward* from the referencing row:

- `variables.grapherConfigIdETL` / `grapherConfigIdAdmin` → `chart_configs.id` (`ON DELETE RESTRICT`)
- `multi_dim_x_chart_configs.chartConfigId` → `chart_configs.id`

So `DELETE FROM variables` succeeds and the indicator's config row is simply stranded —
one leaked row per deleted variable that had an ETL grapher config. Same for each mdim
view config, since the function also deletes `multi_dim_x_chart_configs` rows directly.
Production currently holds ~1,600 of these, growing in bursts that line up with dataset
re-versions.

Grapher's own writers delete both sides — `deleteVariablesVariableIdGrapherConfigETL`
nulls the column and deletes the config row, `cleanUpOrphanedChartConfigs` does the same
for mdim views. This function now does too: it collects the config ids before the deletes
and removes them after the variables are gone (the FKs are `RESTRICT`, so the order
matters).

The `NOT EXISTS` guard on `multi_dim_redirects` is there because a redirect can point at
an mdim view config, and that FK is `RESTRICT` as well — without the guard, that rare case
would fail the whole grapher step instead of leaking a single row.

## Still open

- **The ~1,600 rows already in production are not cleaned up by this PR.** That needs a
  one-off delete of `chart_configs` rows unreferenced by `charts`, `variables`,
  `narrative_charts`, `multi_dim_x_chart_configs`, `explorer_views` and
  `multi_dim_redirects` — best placed as an `owid-grapher` migration. Not written yet.
- **No test.** `tests/` has no MySQL harness and this function's SQL is MySQL-only (the
  `IN :param` style relies on pymysql expanding lists), so covering it means new
  integration infra. The `DELETE … NOT EXISTS` semantics were verified by hand against
  sqlite; the change itself has not been run against a real database.
- **Unverified:** how many of the existing orphans are the mdim-view leg rather than the
  indicator-config leg. The public Datasette mirror carries a filtered subset of
  `chart_configs`, so that count needs a prod/staging MySQL run.

### DIFF
--- a/etl/grapher/to_db.py
+++ b/etl/grapher/to_db.py
@@ -468,6 +468,21 @@ def cleanup_ghost_variables(engine: Engine, dataset_id: int, upserted_variable_i
                 )
                 return False
 
+        # collect the chart_configs rows that only these variables (and their mdim views) point at
+        rows = con.execute(
+            text(
+                """
+            SELECT grapherConfigIdETL FROM variables WHERE id IN :variable_ids AND grapherConfigIdETL IS NOT NULL
+            UNION
+            SELECT grapherConfigIdAdmin FROM variables WHERE id IN :variable_ids AND grapherConfigIdAdmin IS NOT NULL
+            UNION
+            SELECT chartConfigId FROM multi_dim_x_chart_configs WHERE variableId IN :variable_ids
+        """
+            ),
+            {"variable_ids": variable_ids_to_delete},
+        ).fetchall()
+        config_ids_to_delete = [row[0] for row in rows]
+
         # then variables themselves with related data in other tables
         # delete relationships
         con.execute(
@@ -527,6 +542,22 @@ def cleanup_ghost_variables(engine: Engine, dataset_id: int, upserted_variable_i
             ),
             {"dataset_id": dataset_id, "variable_ids": variable_ids_to_delete},
         )
+
+        # delete the now-unreferenced configs (the multi_dim_redirects guard is there because
+        # a redirect can point at an mdim view config)
+        if config_ids_to_delete:
+            con.execute(
+                text(
+                    """
+                DELETE FROM chart_configs
+                WHERE id IN :config_ids
+                  AND NOT EXISTS (
+                    SELECT 1 FROM multi_dim_redirects r WHERE r.viewConfigId = chart_configs.id
+                  )
+            """
+                ),
+                {"config_ids": config_ids_to_delete},
+            )
 
         con.commit()
 


sophiamersmann:
* **#6672** <a href="https://app.graphite.com/github/pr/owid/etl/6672?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/etl/6672?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sophiamersmann:
@codex review

etl/grapher/to_db.py
```diff
@@ -528,6 +543,22 @@ def cleanup_ghost_variables(engine: Engine, dataset_id: int, upserted_variable_i
...
+        # delete the now-unreferenced configs (the multi_dim_redirects guard is there because
+        # a redirect can point at an mdim view config)
+        if config_ids_to_delete:
+            con.execute(
+                text(
+                    """
+                DELETE FROM chart_configs
+                WHERE id IN :config_ids
+                  AND NOT EXISTS (
+                    SELECT 1 FROM multi_dim_redirects r WHERE r.viewConfigId = chart_configs.id
```
sophiamersmann, line 555: I don't know much about redirects, so don't fully understand why this is necessary

Marigold (approved):
I don't know much about redirects either, but the rest looks good.

etl/grapher/to_db.py
sophiamersmann: hey @rakyi, can you have a quick look? the background is that we currently have 1000s of orphaned chart configs in the db because this ETL function doesn't clean up chart configs.

Claude added this extra multi dim redirects check and I'm not sure why. I guess the question is: can an mdim view config that's being deleted still be pointed to by a multi dim redirect? If so, don't we need to take that into account on the grapher side as well?

sophiamersmann:
@Marigold, I’m closing this because I just had a chat with @rakyi, and we thought it made more sense to rewrite this code to go through the admin API instead, so the ETL doesn’t directly write to the DB. Martin will bring this up for discussion on Monday, so we can discuss it then!

PR #6677, opened 2026-08-14
🐛 Fix string boolean in labor force participation MDim config

sophiamersmann:

Schema validation of MDim views flagged three views of the labor force participation MDim:

```
'false' is not of type 'boolean'    hideRelativeToggle
  age_group=_15_plus__sex=male_vs_female
  age_group=_15_64__sex=male_vs_female
  age_group=_15_24__sex=male_vs_female
```

The `group_views` config for the male-vs-female views set `hideRelativeToggle` to the string `"false"` instead of the boolean `False`.

## Still open

- **The schema check never saw this config.** `validate_schema()` runs only at collection *create* time (`etl/collection/core/utils.py:96`); `Collection.save()` runs many other checks but never re-validates against the schema. So anything added or edited after creation — `group_views` configs and the `for view in c.views:` edit loops that most MDim steps have — reaches MySQL unvalidated. Adding `self.validate_schema()` to `save()` (gated off for explorers, as it already is at create time) would catch this class at build time. Not done here because it's unknown how many existing MDim steps that would newly fail — it needs a run of every MDim export step first.
- **Other MDims not swept** for the same string-boolean pattern beyond the `hideRelativeToggle`/`hasMapTab` grep. The `"false"` strings in the LIS explorer steps are fine — `etl/collection/explorer/core.py:205` stringifies booleans anyway.

### DIFF
--- a/etl/steps/export/multidim/worldbank_wdi/latest/labor_force_participation_rate.py
+++ b/etl/steps/export/multidim/worldbank_wdi/latest/labor_force_participation_rate.py
@@ -33,7 +33,7 @@ def run() -> None:
                 "choices": ["male", "female"],
                 "choice_new_slug": "male_vs_female",
                 "view_config": {
-                    "hideRelativeToggle": "false",
+                    "hideRelativeToggle": False,
                     "selectedFacetStrategy": "entity",
                     "hasMapTab": False,
                     "chartTypes": ["LineChart", "DiscreteBar", "SlopeChart"],


sophiamersmann:
* **#6677** <a href="https://app.graphite.com/github/pr/owid/etl/6677?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/owid/etl/6677?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sophiamersmann:
@codex review

paarriagadap:
Hi @sophiamersmann! I actually don't quite follow what Claude says, but the fact is that the fix, even if correct, doesn't make the toggle appear.

cc @lucasrodes 

@sophiamersmann The fix is right, but heads up: the change doesn't actually show up on the MDim on staging. I went looking for the relative toggle on the three `male vs. female` views and it isn't there. Worth knowing why before you take the absence as a sign the fix didn't work.

## The MDim page doesn't read view configs from MySQL

`chart_configs.full` on the branch's staging DB holds the corrected boolean for all three views — that part of the fix landed:

| view | staging DB | production DB |
|---|---|---|
| `age_group=_15_plus__sex=male_vs_female` | `false` (boolean) | `"false"` (string) |
| `age_group=_15_64__sex=male_vs_female` | `false` (boolean) | `"false"` (string) |
| `age_group=_15_24__sex=male_vs_female` | `false` (boolean) | `"false"` (string) |

But the browser never sees that row. An MDim page hydrates each view from `/grapher/by-uuid/<fullConfigId>.config.json` (`site/multiDim/api.ts`, `cachedGetGrapherConfigByUuid`), which a Cloudflare Pages function serves out of the R2 grapher-configs bucket. On a staging box nginx routes `location /` to `wrangler pages dev`, so it holds there too — the Node app's own DB-backed by-uuid route in `adminSiteServer/mockSiteRouter.ts` is only reachable under `/admin`.

And that endpoint is still serving the pre-fix config:

```
staging DB fullMd5 → d73c035a7144…   (boolean false)
served ETag        → e9631b5dacfd…   (string "false")
```

That served ETag is byte-identical to **production's** `fullMd5` for the same UUID. So the branch's object is missing from `by-branch/<branch>/config/by-uuid/`, and the function fell back to the production bucket (`functions/_common/grapherTools.ts`). No error surfaces anywhere — the page just renders production's config on a staging URL.

## It's these three configs, not staging in general

I compared the staging DB's `fullMd5` against the served ETag for the 40 most recently updated MDim view configs on that server. **37 match; the 3 mismatches are exactly the ones this PR changes**, all updated in the DB at `2026-08-14 15:19:19` and all still serving the old object. So the branch bucket is populated and working — this one update just didn't propagate.

Most likely cause: the staging box has `GRAPHER_CONFIG_R2_BUCKET` and `GRAPHER_CONFIG_R2_BUCKET_PATH` set but no R2 credentials, so the R2 write that `updateChartConfigInDbAndR2` fires on MDim upsert can't land, while the DB write succeeds. The DB→R2 reconciliation job (`yarn syncGraphersToR2`) only runs on the production Buildkite queue, so nothing catches it up afterwards. A staging rebuild ran at 15:33, after the 15:19 DB update, and the object is still stale — so re-triggering one probably won't help.

## Where that leaves the PR

Production's R2 *is* in sync with production's DB for these UUIDs (`fullMd5` == served ETag), so the prod write path works and merging should surface the toggle. Grapher's `canToggleRelativeMode` shows "Display relative change" on the line and slope tabs once the browser gets a real boolean, and the resolved config has no `log` y-scale and no pinned `minTime`/`maxTime`, so nothing else blocks it. Not on the bar tab — `DiscreteBar` isn't in `showAbsRelToggle`'s allow-list.

Options, roughly in order of appeal:

1. Ask someone on the grapher side to run `syncGraphersToR2` against that staging box, or to confirm whether missing staging R2 credentials is expected. That's the only route to seeing it on staging.
2. Verify locally — local dev serves `/grapher/by-uuid/…` from the DB, so running the export step against a local DB would show the toggle.
3. Merge on the DB-level verification and confirm on production.

Also worth considering as a follow-up: this is a second way for an MDim config change to pass every check and still not reach anyone, alongside the `validate_schema()` gap you flagged in the description. A build-time check that the config the page will actually fetch matches the config we just wrote would catch both.

## What I haven't established

- Why the R2 write fails *silently* rather than erroring the admin API call. Missing credentials is an inference from the empty settings plus the 37/40 split — I didn't read the staging app logs.
- Whether the admin preview (`/admin/grapher/<catalogPath>`) shows the correct config. It renders server-side from the DB, but its client still fetches views by UUID, so I'd expect it to hit the same stale object. Unconfirmed.
- Whether other open PRs touching MDim view configs are affected the same way. I only audited this branch.

sophiamersmann:
@paarriagadap, I don't quite understand why this needs to be checked so thoroughly downstream, since we should clearly just use the right data type in the config? I'm probably missing something here, so feel free to close without merging if this is more complicated than I think!

paarriagadap:
Hi @sophiamersmann! I'm just saying that while the fix is right, it is not showing the relative view toggle, as we should expect. So I was trying to understand the reasons for that.

sophiamersmann:
@paarriagadap, for context, this is what I'm trying to fix: https://owid.slack.com/archives/C5JJW19PS/p1786965774923909

paarriagadap:
@sophiamersmann Yes, no problem with that.

I was flagging that the toggle didn't appear automatically even if set correctly, but now I realized that it did after running `etlr labor_force_participation_rate --export` manually in this branch.

image

So we can merge, and I will check if similar errors happen in the future.

Thanks!

paarriagadap (approved):

PR #6678, opened 2026-08-14
📊 Update OECD Time Use Database and refresh static viz

paarriagadap:

Refreshes the OECD Time Use Database to its 2026 edition and rebuilds the "How do people spend their time?" static visualization on top of it.

The previous version was a legacy backport (`oecd/2020-11-05/time_use_across_activities__oecd_gender_data__2020` — private, no catalog path). It is replaced by a normal public ETL chain, so the data is in the catalog and the chart is reproducible from a step rather than from a one-off file. The old dataset is archived.

## The chain

```
snapshot://oecd/2026-08-14/time_use.xlsx        the OECD's own workbook, published 2026-04-30
  → data://meadow/oecd/2026-08-14/time_use
    → data://garden/oecd/2026-08-14/time_use
      → data://grapher/oecd/2026-08-14/time_use
      → export://static_viz/oecd/2026-08-14/time_use_by_country
```

## What garden publishes

**One table, the source's own hierarchy** — 30 activity columns across five top-level categories, plus `survey_year` and `age_of_reference`. 26 countries × 3 breakdowns (total, men, women), one survey per country.

**Only surveys from 2010 onwards**, which is 26 of the source's 35 countries. The OECD runs a single survey per country and does not repeat it, so a year cutoff drops countries rather than years — and what it buys is that nothing downstream can put a 1999 survey beside a 2024 one. Every chart and the static viz on this dataset is a cross-country comparison, so that mixing is the failure mode rather than an edge case.

It costs India (1999) and China (2008), and with them most of the coverage outside high-income Europe. Weighed against that: six countries sit on 2010 itself so the cut lands on a cluster; it buys eleven years of recency for nine others; and survey year does not tilt what the data says, the correlation with any top-level category being +0.09 at most. The cutoff is applied **after** the output checks, because one of them reproduces the previous edition's published values for China 2008 — a survey it removes. It also removes every age-of-reference exception the source flags (Australia 15+, China 15–74, Lithuania 20–64), so what is published is 15-to-64 throughout; `sanity_check_cutoff` asserts that, since it is what lets the chart say "aged 15 to 64" unqualified.

**No display groups.** A `time_use_chart_groups` table carrying ten chart-oriented groups was built and then removed: each of the four categories the chart draws IS one of the source's own top-level categories, so the ten were only ever summed back up to columns `time_use` already has — agreeing to 3e-05 minutes — and none of their 33 indicators was used by any chart. The one thing the chart adds is folding the source's small religious/civic/uncategorized category into unpaid work, which is a presentation choice and lives in the export step.

## Metadata

The snapshot description is the producer's own "About" paragraph. Everything reader-facing that goes beyond it is `description_key`, which none of the indicators had before: the OECD's own caveat about differences in survey design, diary days sampled and activity grouping; the age-bracket weighting, by both of the two methods the workbook documents; Mexico's instrument, which asks about the previous week rather than keeping a diary; and that each country's estimates are its most recent survey *in the OECD's database*, which can be older than its own latest survey. Four more go where they apply — Japan's TV group also covering print and its "other" group also covering non-commuting travel, and the minutes some countries leave unassigned inside a category.

## The static visualization

**Two frames, from one step and one render.** Desktop 850×1095 on the vertical static-chart template, mobile 540×824 on `Static Chart Template_Mobile (example 2)` — each as SVG and PNG. One bar per country either way, the day split into the OECD's four top-level categories and ranked by paid work or study.

They differ only in where the key goes. The desktop names each category over the run of bars it belongs to, with its activities listed underneath — the better reading, and why that frame keeps it. It needs the segment to be wide enough to hold the words, and 508px of mobile content does not give it: the narrowest category spans 56px against a 105px name, so all four names would overhang the bars they name by more than they cover. So the mobile frame puts the key in **four rows above the plot** — a colour chip at the bar's own fill and opacity, the category's name, then what it holds. All 26 countries and all 104 values survive; it was the per-segment key that did not fit, not the data.

The mobile template has no Note slot, so the Note is drawn inside the chart band's foot, on the footer's own rhythm — 8px of ink to the `Data source:` row, the same gap that row leaves the licence.

The step fills each template's own slots at its sizes and line heights, so the render previews the frame rather than only its proportions; both chart bands are derived from the templates' auto-layout rhythm rather than recorded. Type sizes are declared in whole template pixels, the way the templates state their own — declaring them in points was producing sizes like 12.153px. The frame sizes come from `etl.static_viz`'s template registry rather than being retyped, and the margin is the templates' own 16: both frames carried 16.216 from an earlier generation of the templates, worth 0.43px of content width, which the crop-snap had been absorbing silently.

Figma: page **20260817 How do people spend their time? (Pablo A)** (`25524:5`), frames **`how-do-people-spend-their-time`** (`26879:11`) and **`how-do-people-spend-their-time-mobile`** (`27518:8`), each with its unstyled import parked to its left so the page reads original-then-edited twice. The frame name is the durable join — it is the kebab-case slug the website exports the PNG by.

## Chart 2541

`time-women-and-men-spend-on-leisure` was remapped to the new variable IDs (`1207880 → 1296121`, `1207882 → 1296123`) and its FAUST corrected on staging:

- the subtitle said readers aged **15-65** — the reference population is 15 to 64;
- it said surveys ran **2009 to 2016 "but some countries have older surveys"** — now 2010 to 2024, and with the cutoff there are no older surveys;
- the x axis carried its unit and the y axis did not, because the y dimension had a `unit: ""` override blanking it.

Being a scatter with no entity selection, it also drops from 35 points to 26.

## Checks added along the way

- **`check_contrast`** measures every seam, in-bar label and legend colour on each build, against floors plus pinned allowances for the two shortfalls grapher's 0.8 bar opacity costs. It fails three ways: a new shortfall, a regression below an allowance, and an allowance left stale after a fix.
- **`check_members_are_the_largest`** fails if a category's key names a smaller member than one it hides — it was naming volunteering (a median 3.0 min/day) and omitting household travel (17.3, reported by 25 of 26 countries).
- **Meadow asserts the source's missing-value markers.** One `errors="coerce"` was swallowing three different ones, including `(see notes)` on Japan's household travel, which means the minutes were counted under another activity.
- **One column, one value form, and no shorter fallback.** Two mobile rows printed a bare `162` where the compact form did not fit — a different unit, with nothing on the frame to say which. A segment too narrow for its column's form is now left unlabelled, and the step fails loudly if a column can find no form at all.
- **The row legend is asserted to fit rather than wrapped.** A key row that wraps stops reading as a row, so the fix is a shorter member name, not a second line.
- Garden asserts the day closes, that no published column is entirely empty, and reproduces four of the previous edition's published values for China 2008.

## Verification

The live `url_download` file is byte-identical to the snapshot (md5 `fe7d069c…`), all 35 survey-year labels parse to the right end year, and the values reconcile against independent publications: Mexico's unpaid work to 0.6% of INEGI's own ENUT 2024 figure, US sleeping to 9.0 h against BLS ATUS, Japan's 7h42m below the Statistics Bureau's 7h54m for ages 10+ as the narrower age range predicts. No magnitude, sign, wrong-year or entity error was found.

`verify_page.js` ran on both frames — all three documented slices, full coverage of its five row groups. Both pass `box-alignment` exactly (+0/+0), `margins`, `within-frame`, `dead-fills`, `text-hierarchy` and `source-line-weight`; every plot text node on both sits on the 12/13/14/15/16 ladder at zero drift; the page census holds exactly the four intended objects. Two rows do not pass and neither is a defect:

- The desktop's **`text-floor`** flags the template's own tagline and licence at 11px, under the format's 12px floor. That is inherited furniture and CHECKS.md says so; the failing ranges are exactly those two rows. The mobile frame passes, its footer being 14px.
- The mobile's **`gap`** reads 14.12 top and 5.27 bottom, because the in-band Note occupies the band's foot by design and the row measures footer-top to chart-bottom. The desktop's `gap`, which did read 18.44/17.01, now reads **14 and 14** — the band is inset from the ink at both ends rather than from the layout boxes.

`colour-vision` and `grayscale-seams` come back SKIPPED on both, because the script will not audit a translucent fill and the bars sit at grapher's 0.8. That gap is closed by the step instead of left open: `check_contrast` composites at that opacity on every build and asserts the grayscale seams plus a Viénot–Brettel–Mollon dichromacy simulation of every touching pair.

## Open before merge

- [ ] **Chart-diff approval.** Chart 2541's remap and its FAUST edit both ride to production this way, and it is the only red check.
- [ ] **Design sign-off** on both frames. The palette is settled: Camel, Midnight Blue, Light Teal and Maroon, chosen from a search of all 10,626 four-colour subsets of the library against the grayscale-seam, in-bar-label and header-text floors. One allowance is pinned and stated — the Light Teal/Maroon seam falls to 1.13:1 under red-green colour blindness, which is why the seam check now runs the simulation too.
- [ ] A 1px rule under each category name, spanning its own segment, was built and rendered but not shipped — the header lost its bracket rules deliberately, so putting one back is a design call.
- [ ] Topic-page image swap, and the surrounding prose: the cited survey years all moved (United States 2018→2024, Japan 2016→2021, Mexico 2014→2024, and nine others), and Japan now tops paid work.
- [ ] The 302-wide PNG export for the website.
- [ ] One note to the OECD: `volunteering` is a hard `0` for Latvia and Lithuania where every other non-report is `..`, and the workbook's activity sheet still documents Japan's 2016 survey while the data are 2021.
- [ ] Figma comments are not visible over MCP, so the design review is not covered either way — it needs a human to open the page.


paarriagadap:
@codex review

paarriagadap:
@codex review

paarriagadap:
@codex review

paarriagadap:
@codex review

paarriagadap:
@codex review

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,216 @@
...
+    - TV & radio: watching TV or listening to the radio at home.
+    - Seeing friends: visiting or entertaining friends.
+    - Other leisure: sports, participating in or attending events, and other leisure activities.
+
+    Groups that cover "the rest" of one of the OECD's top-level categories are computed as remainders from that category's total, so minutes that a country does not allocate to any detailed sub-activity remain included.
+
+dataset:
+  update_period_days: 365
+  owners:
+    - Pablo Arriagada
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid, and checked against the file this PR deletes: `etl/steps/data/garden/oecd/2020-11-05/time_use_across_activities__oecd_gender_data__2020.meta.yml` on master lists exactly one owner, `Diana Beltekian  # backport`. She is also in the `schemas/dataset-schema.json` owner enum, so restoring her validates.

Two things worth flagging alongside it, because they affect who should be listed first rather than whether to restore:

- That entry carries the inline `# backport` marker, i.e. it was assigned when the legacy dataset was mechanically backported, not chosen as an accountable owner.
- This PR does not version-bump that step. It retires the private backport entirely and builds a new public chain at `oecd/2026-08-14/time_use`, which is why the owner list was written fresh rather than extended.

Neither is a reason to drop the name, so the finding stands: the list should carry both. Accepted — it is not in this commit and will land in a follow-up on this branch, with the ordering (which name is the accountable owner) confirmed by a human first.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,216 @@
+definitions:
+  common:
+    unit: minutes per day
+    short_unit: min
+    display:
+      numDecimalPlaces: 0
+    presentation:
+      topic_tags:
+        - Time Use
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid. Confirmed by grepping the file at `b2b953f21c`: `attribution_short` appears nowhere in `time_use.meta.yml`. It is set on the snapshot origin (`snapshots/oecd/2026-08-14/time_use.xlsx.dvc`), and that does not inherit into `presentation` — Grapher fills `variables.attributionShort` from the indicator presentation metadata, so as written every indicator in both tables reaches Grapher without it.

The fix is the one you name: `attribution_short: OECD` under the `definitions.common.presentation` block, so it covers `time_use` and `time_use_chart_groups` in one place.

Accepted — not in this commit; it will land in a follow-up on this branch before the PR leaves draft.

etl/steps/data/garden/oecd/2026-08-14/time_use.py
```diff
@@ -0,0 +1,206 @@
...
+    ds_meadow = paths.load_dataset("time_use")
+    tb = ds_meadow.read("time_use")
+
+    sanity_check_inputs(tb)
+
+    #
+    # Process data.
+    #
+    tb = paths.regions.harmonize_names(tb, country_col="country", countries_file=paths.country_mapping_path)
+
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid, and the gap is narrower than it looks but real. Checked at `b2b953f21c`: `description_processing` is set once in `time_use.meta.yml`, on the `time_use_chart_groups` table, and it documents the activity grouping only — nothing in either table documents the year.

So the chain you describe is unguarded end to end: `parse_survey_end_year` maps a label like `2009/10` to `2010`, garden keeps the original `survey_year` alongside it as context, and the grapher step then drops both `survey_year` and `age_of_reference` because they are strings rather than indicators. A reader downstream sees a single year with no way to tell it is the end of a two-year fieldwork window, and no way to recover the original label.

Accepted. The note belongs in `definitions.common` so it reaches both tables, and should say that the year is the survey period's end year and that survey years differ by country. Not in this commit; it will land in a follow-up on this branch.

etl/steps/export/static_viz/oecd/2026-08-14/time_use_by_country.py
```diff
@@ -0,0 +1,1727 @@
...
+    This is a chart-side sum of this step's own display groups, which is why it lives here rather than
+    in garden: garden owns the mapping from the OECD's 30 activities into the ten groups, and
+    `CATEGORIES` is how *this* step groups those ten. If the four-category version is the one that
+    ships, promote the sums to garden so the aggregate is in the catalog too.
+    """
+    assert [group["label"] for group in MAIN_CATEGORY_GROUPS] == [category["name"] for category in CATEGORIES], (
+        "The four-category groups must stay in the same order as CATEGORIES, since they are its totals."
+    )
+    for group, category in zip(MAIN_CATEGORY_GROUPS, CATEGORIES):
+        tb[group["column"]] = tb[list(category["columns"])].sum(axis=1)
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Agreed on the substance, and it is already tracked rather than overlooked: the PR description carries it as an explicit open item ("Promote the category sums into garden"), and `add_main_category_totals` says the same in its own docstring — "If the four-category version is the one that ships, promote the sums to garden so the aggregate is in the catalog too."

The deferral is deliberate. The four-category layout is still waiting on design sign-off, and two of the open questions on this PR could still change what the segments are — the in-bar label contrast at 0.8 opacity, and whether the fifth top-level category stays folded into unpaid work. Promoting the sums now would mint catalog indicators and Grapher variable IDs for a shape that may not be final, and unwinding that costs a retirement plus a chart remap rather than an edit.

So: it will be done, on this branch, once the layout is signed off and before the PR leaves draft. Leaving this thread open as the tracking record rather than resolving it, since the change has not been made yet.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Kept, in c094668543 — with a deliberate choice about order that is worth stating rather than burying.

The name is back with its `# backport` marker, but **second**. `etl/owners.py` lists it under `OWID_LEGACY_CREATORS`, described there as former contributors whose names appear on legacy backported datasets *per grapher's `createdByUserId`* — so it was assigned mechanically by the backport heuristic, not chosen. This PR also retires that step rather than version-bumping it: the new chain is built from a 2026 snapshot. Dropping the name would erase the lineage the archived dataset records; making it the accountable owner would move live responsibility onto someone who did not build this. Second place keeps the credit without the accountability.

If the convention here is that the previous primary always stays first, say so and I will swap them.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Fixed in c094668543. Confirmed the gap first: the string appeared nowhere in the garden metadata, and `presentation.attribution_short` does not inherit from the origin — so every indicator in both tables was reaching Grapher without one.

Set as `attribution_short: OECD` under `definitions.common.presentation`, which covers both tables in one place. Verified in the rebuilt catalog rather than in the YAML: `time_use.sleeping` and `time_use_chart_groups.paid_work` both now carry it.

etl/steps/data/garden/oecd/2026-08-14/time_use.py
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Fixed in c094668543, and the gap was as described: `description_processing` was set once, on the chart-groups table, and documented only the activity grouping. Nothing said that `parse_survey_end_year` dates a survey by the year fieldwork **ended** — so with `survey_year` and `age_of_reference` dropped downstream as strings, the year read as a point observation with no way back to the fieldwork window.

One thing worth recording for anyone doing this next: a table's own `description_processing` **replaces** the shared one rather than appending to it, so putting the note in `definitions.common` alone would have left the chart-groups table without it. The text is therefore in both places, and I verified in the built catalog that each table carries the dating note and that the grouping note survives underneath it.

paarriagadap:
@codex review

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,227 @@
+definitions:
+  survey_dating: &survey_dating |-
+    Countries run their time-use surveys in different years, and some run them across two: we date each
+    country by the year its fieldwork ended, so a survey labelled 2009/10 is shown as 2010. The survey's
+    own label is kept in the "Survey year" indicator.
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid, and confirmed independently before this review landed. Three checks against `c094668543`:

- `etl/steps/data/grapher/oecd/2026-08-14/time_use.py:22` drops `survey_year` and `age_of_reference` outright, so neither reaches Grapher.
- The "Survey year" indicator the note points at is real, but only in garden — `time_use.meta.yml` defines it at `tables.time_use.variables.survey_year`.
- Worth adding: for the second table the claim is wrong twice over. `build_chart_groups` builds `time_use_chart_groups` from `country`, `year` and `sex` alone, so that table never carried `survey_year` in garden either — yet the duplicated copy of the note inside `groups_processing` makes the same promise.

Since `description_processing` is what Grapher renders to readers, the note sends every published indicator after something that is not there.

The fix is to stop claiming it — drop the third sentence and keep the first two, which are accurate on their own and carry the part that matters:

    Countries run their time-use surveys in different years, and some run them across two: we date each
    country by the year its fieldwork ended, so a survey labelled 2009/10 is shown as 2010.

It has to change in **both** places: the `survey_dating` anchor in `definitions.common`, and the copy pasted into `groups_processing`, which exists because a table-level `description_processing` replaces the shared one rather than appending to it. Changing only the anchor would leave the chart-groups table still promising the indicator.

Accepted — not applied in this commit; it will land in a follow-up on this branch before the PR leaves draft.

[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Fixed in 46568e99b8, and the claim was wrong in a second way worth recording: the chart-groups table never carried `survey_year` in garden either — `build_chart_groups` builds it from country, year and sex alone — so on that table the sentence pointed at something that has never existed, published or not.

Dropped from **both** copies. The duplication is the trap named in the previous commit biting again: a table-level `description_processing` replaces the shared one rather than appending, so the text lives twice, and editing only the `survey_dating` anchor would have left the chart-groups table still promising the indicator. Verified in the rebuilt catalog rather than the YAML — neither table mentions it, both keep the dating note, and the grouping note still sits underneath on the table that owns it.

The two remaining sentences are true of what ships: how a country is dated, and what a 2009/10 label becomes.

paarriagadap:
@codex review

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,225 @@
...
+          Average time spent on religious and spiritual activities and civic obligations. {definitions.averaged}
+      other_uncategorized_activities:
+        title: Time spent on uncategorized activities
+        description_short: |-
+          Average time spent on activities without a category. {definitions.averaged}
+      survey_year:
+        title: Survey year
+        unit: ""
+        short_unit: ""
+        description_short: The year(s) in which the country's most recent national time-use survey was conducted.
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid. Both open with an article that the sentence does not need:

- `survey_year` — "The year(s) in which the country's most recent national time-use survey was conducted."
- `age_of_reference` — "The age range of the population covered by the country's time-use survey."

Neither is a sentence whose subject is that noun phrase, and neither "The" is part of a proper name, so the rule applies straightforwardly. Rewritten to "Year(s) in which…" and "Age range of…".

One correction to the finding, for the record: these two are not new in this commit — they have been in the file since the garden step was first written. The commit under review only moved the text around them.

Worth noting where these land, since it does not change the verdict but does change the stakes: both indicators are garden-only. `grapher/oecd/2026-08-14/time_use.py:22` drops `survey_year` and `age_of_reference`, so no reader ever sees either description — this is catalog hygiene rather than a reader-facing fix.

Applied and verified locally; not yet committed to the branch, so leaving this thread open until it is pushed.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,225 @@
+definitions:
+  survey_dating: &survey_dating |-
+    Countries run their time-use surveys in different years, and some run them across two: we date each
+    country by the year its fieldwork ended, so a survey labelled 2009/10 is shown as 2010.
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid. "labelled" appears twice — once in the `survey_dating` anchor and once in the copy pasted into `groups_processing` — and both are now "labeled".

Both had to change for the same reason the previous two fixes to this note did: a table-level `description_processing` replaces the shared one rather than appending, so the text is deliberately duplicated and a single-site edit would have left the chart-groups table on the British spelling.

I also swept the rest of the user-facing metadata for other British forms rather than fixing only the one reported — `colour`, `behaviour`, `centre`, `organis*`, `favour`, `analyse`, `programme`, `travelling`, `modelling`, `catalogue`, `licence` — across both `time_use.meta.yml` and `snapshots/oecd/2026-08-14/time_use.xlsx.dvc`. These two were the only hits; "traveling" was already American.

Verified in the rebuilt grapher dataset that both tables' processing notes now read "labeled" and neither contains "labelled".

Applied and verified locally; not yet committed to the branch, so leaving this thread open until it is pushed.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,225 @@
...
+      housework_and_shopping:
+        title: Time spent on housework and shopping
+        description_short: |-
+          Average time spent on routine housework and shopping. {definitions.averaged}
+      other_unpaid_work:
+        title: Time spent on other unpaid work
+        description_short: |-
+          Average time spent on care work, volunteering, travel related to household activities, and other unpaid or uncategorized activities. {definitions.averaged}
+      personal_care:
+        title: Time spent on personal care other than sleeping and eating
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid, and this is the substantive one of the three. `build_chart_groups` computes the group as

    tb["personal_care"] = tb_wide["personal_care"] - tb_wide["sleeping"] - tb_wide["eating_and_drinking"]

so it subtracts the OECD's whole 3.2 column, which is eating **and** drinking — while the title said "other than sleeping and eating". The gap is not hypothetical either: the sibling group in the same table is titled "Time spent eating and drinking", so the two titles together implied drinking sat in both, or in neither.

Retitled to "Time spent on personal care other than sleeping, eating, and drinking" — serial comma to match the surrounding descriptions in this file.

Two things I checked so the fix does not overreach:

- The `time_use` table's own `personal_care` (the OECD's top-level category 3, not a residual) is a different indicator and keeps its title unchanged.
- The group's `description_short` — "personal, household, and medical services, and travel related to personal care" — already describes what the group *contains* accurately, so it needed no change.

Verified in the rebuilt grapher dataset: all three sex variants now carry the corrected title.

Applied and verified locally; not yet committed to the branch, so leaving this thread open until it is pushed.

paarriagadap:

The three round-5 fixes are now on the branch as **1d1a608cd7** — the earlier replies were written while they sat uncommitted in the worktree, so this is the commit they refer to.

Verified in the built catalog rather than the YAML: the new `personal_care` title on all three `personal_care__sex_*`, "labeled" in both tables' processing notes with no "labelled" left, and the two subtitles without their leading "The". Render untouched — SVG still `8e1fd3fcbb23555f6ca3cf7258937f0c`, verifier 0 clip paths / 267 named layers, 31 tests, `make check` clean.

Also worth recording against the staging question: `buildkite/etl-automated-staging-environment` #43172 failed once on `c094668543` and has passed on both commits since — including `46568e99b8`, which changed garden metadata and therefore re-ran the same garden→grapher→MySQL invalidation the failure was suspected of. That path has now been exercised twice, green. The log itself was never read (no Buildkite token here), so it is recorded in the PR body as unexplained rather than closed.

paarriagadap:
@codex review

etl/steps/data/garden/oecd/2026-08-14/time_use.py
```diff
@@ -0,0 +1,265 @@
...
+    tb_wide = tb.pivot(
+        index=["country", "year", "sex", "survey_year", "age_of_reference"],
+        columns="activity_code",
+        values="minutes",
+    )
+    tb_wide = tb_wide.rename(columns=CODE_TO_COLUMN)
+    return tb_wide.reset_index()
+
+
+def build_chart_groups(tb_wide: Table) -> Table:
```
paarriagadap: Maybe we no longer need this?

etl/steps/data/garden/oecd/2026-08-14/time_use.py
```diff
@@ -0,0 +1,265 @@
...
+        ("leisure", "4."),
+        ("other_activities", "5."),
+    ]:
+        children = [
+            column for code, column in CODE_TO_COLUMN.items() if code.startswith(prefix) and code.count(".") == 1
+        ]
+        gap = (tb_wide[parent] - tb_wide[children].fillna(0).sum(axis=1)).abs()
+        unallocated = tb_wide.loc[gap > 1.0, ["country", "sex"]].to_records(index=False).tolist()
+        if unallocated:
+            log.warning(f"{parent}: minutes unallocated to sub-activities for {unallocated}")
```
paarriagadap: Check if this is not a problem

snapshots/oecd/2026-08-14/time_use.xlsx.dvc
```diff
@@ -0,0 +1,29 @@
+meta:
+  origin:
+    producer: OECD
+    title: OECD Time Use Database
+    description: |-
+      The OECD Time Use Database provides information on how women and men spend their time. Data are currently available for 30 OECD countries. Differences in survey features, the number of diary days sampled, and the categorization of activities may all affect the cross-country comparability of results. To improve the comparison of time use across countries, the samples are restricted to populations aged 15-64 and activities are aggregated into five main categories: (1) Unpaid work; (2) Paid work or study; (3) Personal care; (4) Leisure; and (5) Other time use. Bulgaria, Croatia, China, India, and South Africa are also included.
```
paarriagadap, line 6: Keep only the first paragraph

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,283 @@
...
+  # The caveats every indicator carries. The first is the OECD's own, from the database's landing page;
+  # the rest name the countries it flags, and what each flag means for a reader comparing bars.
+  comparability: &comparability |-
+    Time-use surveys are run differently in different countries. The OECD notes that differences in how the surveys are designed, in how many diary days each one samples, and in how each one groups activities all limit how comparable the results are between countries.
+  mexico_instrument: &mexico_instrument |-
+    The OECD flags Mexico's estimates as not fully comparable. Mexico measures time use with a questionnaire that asks how long people spent on a set list of activities over the previous week, rather than with a diary of a single day, and recall over a week tends to raise the time reported for frequent activities such as housework and care.
+  survey_vintage: &survey_vintage |-
+    Each country's estimates come from a single survey, and the survey years differ widely. They are the most recent surveys in the OECD's database, which is not always updated when a country runs a new one.
```
paarriagadap: Where can you find this? which is not always updated when a country runs a new one.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,283 @@
...
+      - *mexico_instrument
+      - *survey_vintage
+    presentation:
+      topic_tags:
+        - Time Use
+      attribution_short: OECD
+  averaged: &averaged |-
+    Minutes are a daily average over the whole week, including both weekdays and weekends.
+  groups_processing: &groups_processing |-
+    Countries run their time-use surveys in different years, and some run them across two: we date each
```
paarriagadap: Why do you repeat the text from `survey_dating` here?

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,283 @@
...
+    country by the year its fieldwork ended, so a survey labeled 2009/10 is shown as 2010.
+    We include each country whose most recent survey is from 2010 onwards, which is 26 of the 35 countries the OECD publishes. The source runs one survey per country and does not repeat it, so this leaves out countries rather than years, and it avoids showing a survey from the 1990s beside one from 2024.
+
+    We group the OECD's detailed activities into the categories shown in our chart "How do people spend their time?":
```
paarriagadap: Is this still needed with the chart only showing the main categories?

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
```diff
@@ -0,0 +1,283 @@
...
+
+    Groups that cover "the rest" of one of the OECD's top-level categories are computed as remainders from that category's total, so minutes that a country does not allocate to any detailed sub-activity remain included.
+
+    Japan and Sweden do not report time spent on research and homework separately. Because paid work is computed as the rest of the OECD's "paid work or study" category, their homework minutes are counted in paid work rather than in education.
+
+dataset:
+  update_period_days: 365
+  owners:
+    - Pablo Arriagada
+    - Diana Beltekian  # backport
```
paarriagadap: Drop Diana, she's no longer here

etl/steps/data/garden/oecd/2026-08-14/time_use.py
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Correct, and it's gone — along with the whole `time_use_chart_groups` table. Two measurements settled it:

- **Each of the four categories the chart draws IS one of the source's own top-level categories.** The ten groups were only ever summed back up to numbers `time_use` already carries: `paid_work + education` rebuilds code 1, `sleep + eating_and_drinking + personal_care` rebuilds code 3, and so on. Measured, the reconstruction agreed to **3e-05 minutes**.
- **0 of the 33 `time_use_chart_groups` indicators is used by any chart** (checked against this branch's staging DB — the only two used indicators of this dataset are `leisure__sex_men/women`, from `time_use`).

So the table was a detour. The static viz now reads codes 1, 3 and 4 straight and adds 2 + 5 — one addition, and the one thing the chart does on top of the source: folding the small religious/civic/uncategorized category into unpaid work. That's a presentation choice, so it belongs in the chart rather than the catalog.

**The chart is unchanged, and not approximately:** all 163 text nodes identical, largest path coordinate moved **7e-06 px**, PNG byte-identical. So the Figma frame needs nothing.

etl/steps/data/garden/oecd/2026-08-14/time_use.py
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Not a bug — it's the source's own reporting — but the warning was too quiet to let you conclude that, which is the real finding here. Full extent, on the published data:

| Country | Category | Unassigned | Share of category |
|---|---|---|---|
| Poland | unpaid work | 37 min/day (women 47, men 24) | 15% (women 16%, men 13%) |
| Japan | unpaid work | 22 min/day (women 36, men 10) | 18% (women 17%, men 21%) |
| Sweden | paid work / leisure | ∓1 min (women only) | 0.3% — rounding |

Those minutes stay inside their own top-level category, so **no category total is affected** and nothing the chart draws changes. What was missing was visibility, so the warning now reports the amounts and shares rather than just country/sex pairs, and readers get it as a WYSK bullet on `unpaid_work`.

snapshots/oecd/2026-08-14/time_use.xlsx.dvc
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Done — the description is now the producer's "About" paragraph alone.

One thing moved rather than going with it: the **age-bracket weighting**, which was the adversarial review's one confirmed factual error. The old text credited all eleven weighted-average countries to Eurostat, where the workbook names two methods — Bulgaria, Estonia, Finland, Germany and Norway weight five-year brackets using their own national statistical institutes' figures. It's now a WYSK bullet, which is the better home for it anyway.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

**South Korea.** Its estimates here are from 2014, while Statistics Korea has since published 2019 and 2024 rounds (the 2024 survey is its 6th, fielded March 2024).

Fair challenge — as written it asserted something about the OECD's process that we infer rather than know. The bullet now names the evidence instead:

> Each country's estimates come from a single survey, and the survey years differ widely — from 2010 to 2024. They are the most recent surveys in the OECD's database, which can be older than a country's own latest survey: South Korea's estimates here are from 2014, while Statistics Korea has since published 2019 and 2024 rounds.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Because a table's own `description_processing` **replaces** the shared one rather than adding to it — so the second table had to carry `survey_dating`'s text again or lose it.

Moot now: with one table there's one block and nothing to repeat. The duplication went with `groups_processing`.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

No — it went with the groups themselves (see the `build_chart_groups` thread).

The homework note went too. It described a `paid_work`/`education` split that no longer exists; with the display groups gone, a country not reporting homework separately is just the source's own missing value rather than something this step does.

etl/steps/data/garden/oecd/2026-08-14/time_use.meta.yml
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Done — the owners list is a single entry now.

PR #6683, opened 2026-08-17
🐛🤖 Fix Dependabot vulnerabilities

Marigold:
> _Written by Claude Opus 5 — @Marigold at the wheel._

Clears all 20 open Dependabot security alerts in one batch.

## What changed

| Package | From → To | Alerts | Where |
|---|---|---|---|
| `gitpython` | 3.1.55 → 3.1.59 | 9 (6 high, 3 medium) | `pyproject.toml` |
| `cryptography` | 48.0.1 / 49.0.0 → 50.0.0 | 4 high | root + `lib/{catalog,datautils,owl}` |
| `pypdf` | 6.14.2 → 6.16.0 | 2 medium | `pyproject.toml` |
| `js-yaml` | 4.3.0 → 4.3.1 | 5 high | 5 `vscode_extensions/` |

`pyopenssl` 26.2.0/26.3.0 → 26.4.0 comes along as a transitive dependency of `cryptography`. `lib/datautils` version is bumped to 0.6.9 because its `pyproject.toml` changed (PyPI publish rejects a repeated version).

For `js-yaml`: `meta-yaml-hover` and `clickable-dag-steps` import it in `src/`, so it's bumped in `dependencies`. The other three (`find-latest-etl-step`, `dod-syntax`, `compare-previous-version`) only pull it in transitively via eslint/mocha, so their existing `overrides` pin was bumped instead — no phantom runtime dependency added.

## Verification

- `make check-all` (lint, format, typecheck across root and every `lib/`) — passes
- `make unittest` — 665 passed
- All 5 touched extensions compile/lint cleanly
- GitPython 3.1.56+ added guards on unsafe option forwarding; grepped our usage — we only touch `Repo()`, branches, commits and diffs, none of the guarded APIs (`Repo.archive`, `IndexFile.checkout/from_tree`, `Repo.init` templates, submodules)

Supersedes Dependabot PRs #6654, #6640, #6635, #6634, #6609, #6608 — closed in favour of this batch.

## Still open

- Nobody exercised the vscode extensions interactively; they compile, but no runtime smoke test was run.
- `npm audit` still reports unrelated findings in the extensions that GitHub has not flagged as alerts — left alone, out of scope here.

PR #6685, opened 2026-08-17
🔨🤖 Fix PyPI publish, and keep GitHub Action pins from going stale

Marigold:

Fixes the failed `owid-datautils` 0.6.9 release ([run](https://github.com/owid/etl/actions/runs/32005753555/job/95314808261)), then removes the reason it could happen.

```
Checking lib/datautils/dist/owid_datautils-0.6.9-py3-none-any.whl:
ERROR InvalidDistribution: Invalid distribution metadata: '2.5' is not a valid metadata version
```

## What broke

hatchling 1.32.0 (released 2026-08-11) began emitting `Metadata-Version: 2.5`. Our `lib/*/pyproject.toml` declare an unpinned `requires = ["hatchling"]`, so CI builds with the latest. The pinned `pypa/gh-action-pypi-publish` v1.14.0 bundles twine 6.1.0 + packaging 25.0, which don't know metadata 2.5 and fail at `twine check` before the upload is attempted.

Nothing about the dependency bumps in #6683 caused this — that PR was just the first `lib/*/pyproject.toml` version bump since hatchling 1.32.0 landed. The break needed a frozen publisher *and* a floating build backend; either one alone would have been fine.

## Changes

**1. Bump the action to v1.14.2** (twine 7.0.0 + packaging 26.2) — the minimal fix, kept as its own commit so it stays available if we ever want to fall back to the action.

**2. Add `.github/dependabot.yml` for the `github-actions` ecosystem.** This is the actual root cause: the repo enforces SHA pinning (`.github/zizmor.yml` sets `"*": hash-pin`) but has no version-update config, and actions almost never receive security advisories — so Dependabot's *security* updates, the only kind running here, never touch them. Every action pin was frozen indefinitely:

| Action | Pinned | Latest |
|---|---|---|
| `astral-sh/setup-uv` | v8.2.0 | v10.0.1 |
| `actions/checkout` | v6.0.3 | v7.0.1 |
| `actions/setup-python` | v6.2.0 | v7.0.0 |
| `actions/stale` | v10.3.0 | v11.0.0 |
| `zizmorcore/zizmor-action` | v0.5.6 | v0.6.2 |

Five of ten were behind, most by a major version; `gh-action-pypi-publish` was the sixth and the only one whose staleness became load-bearing. Grouped into a single PR, checked weekly. A 7-day cooldown (required by zizmor's `dependabot-cooldown` audit) keeps us from adopting a release the instant it appears, leaving time for a compromised one to be yanked. Other ecosystems are deliberately left out — security updates already cover pip and npm without config, and blanket version updates were removed in 2022 (`c7d5ad0c0d`) for being too noisy. That removal was a *pip, weekly, ungrouped* config, so this doesn't reverse it.

**3. Publish with `uv publish` instead of the action**, dropping a third-party action — and a twine frozen inside it — from the release path entirely.

## Verification

- Built `lib/datautils` locally and confirmed the wheel carries `Metadata-Version: 2.5`; `twine check` fails on 6.1.0 and passes on 7.0.0
- Diffed the action's `requirements/runtime.txt` across tags to confirm the pin change (`twine==6.1.0`/`packaging==25.0` → `twine==7.0.0`/`packaging==26.2`)
- PyPI accepts metadata 2.5 — hatchling 1.32.0's own published wheel uses it
- Authentication is unchanged by change 3: the last successful run (`30496926603`) shows OIDC trusted publishing with no token, and this PR keeps the same workflow filename, `pypi` environment and `id-token: write` permission, so the existing PyPI publisher config applies as-is
- `uv publish` uploads PEP 740 attestations by default (`--no-attestations` is the opt-out), so the provenance the action generated is preserved
- Confirmed `uv publish` defaults to `dist/*` relative to cwd, matching the `cd` the build step already does
- zizmor passes locally on both changed files, and all CI checks are green — including GitHub's own validation of `.github/dependabot.yml`

## Still open

- **`owid-datautils` 0.6.9 is not on PyPI.** The build succeeded but the upload never ran, and this PR doesn't touch a `lib/*/pyproject.toml`, so the push trigger won't fire. After merge someone needs to run *Publish OWID packages to PyPI* via `workflow_dispatch` with `package: datautils`. Nothing in the repo is broken meanwhile — `uv.lock` resolves datautils from the local path, not PyPI.
- **Change 3 is unproven end-to-end.** No `uv publish` has run against PyPI from this workflow yet; the OIDC exchange is verified only by reasoning from the previous run's config. The `workflow_dispatch` above is the first real test, and a failure there is recoverable by reverting to commit `95aca95` (the bumped action).
- `pinact` isn't installed on this machine, so the v1.14.2 SHA was resolved by hand via the GitHub API (annotated tag → commit `dc37677`) rather than by `pinact run`. Moot if change 3 stands, since the action reference is gone.
- `lib/*` still declare `requires = ["hatchling"]` unpinned. Left alone on purpose: Dependabot doesn't manage `build-system.requires`, so pinning it would create a hand-tended pin with no automation behind it — the same failure mode this PR is removing.
- `actions/setup-python` is probably now redundant (uv provisions its own Python), but removing it changes how the wheel is built, so it's out of scope here.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,11 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    # Don't adopt a release the moment it appears: if an action's publishing
+    # account is compromised, the malicious version is usually yanked within
+    # days. Long enough to matter, short enough that pins still move.
+    cooldown:
+      default-days: 7
     groups:
       github-actions:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,8 +15,16 @@
 # Other ecosystems are intentionally left out. Dependabot security updates
 # already cover pip and npm without any config, and blanket version updates
 # across this repo's dependency tree were removed in 2022 (c7d5ad0c0d) for
-# being too noisy. Grouping every action into a single monthly PR keeps this
-# to roughly one review a month.
+# being too noisy. Grouping every action into a single PR keeps the volume
+# low regardless of interval — there are only ~10 actions here, and they
+# release a couple of times a month between them.
+#
+# Weekly rather than monthly, because for actions these version updates are
+# also the de facto *security* mechanism. Dependabot's real security updates
+# ignore this schedule and fire on advisory publication, but actions almost
+# never get advisories, so nothing else will ever move these pins. Combined
+# with the cooldown below, weekly caps worst-case adoption of a fix at about
+# two weeks; monthly would have stretched it past five.
 version: 2
 updates:
   - package-ecosystem: github-actions
@@ -24,7 +32,7 @@ updates:
     # would each need their own entry — there are none today.
     directory: /
     schedule:
-      interval: monthly
+      interval: weekly
     # Don't adopt a release the moment it appears: if an action's publishing
     # account is compromised, the malicious version is usually yanked within
     # days. Long enough to matter, short enough that pins still move.
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,33 @@
+# Dependabot version updates.
+#
+# Deliberately scoped to GitHub Actions only.
+#
+# Actions must be pinned to immutable commit SHAs (see
+# .github/instructions/github-actions.instructions.md). Pinning is a
+# supply-chain control we want to keep, but it freezes each action until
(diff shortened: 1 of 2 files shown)


Marigold:
@codex review

PR #6705, opened 2026-08-18
🐝 collections: Pin grapher schema version in MDIM configs

lucasrodes:

MDIM view configs carry no schema version — neither their own `$schema` nor an inherited `grapherConfigSchema`. Grapher only migrates a view config when it can see a version ([`multiDim.ts#L274`](https://github.com/owid/owid-grapher/blob/master/adminSiteServer/multiDim.ts#L274)), so today every one of them silently skips migration and is assumed to already match the latest schema. Nothing is broken at `011`; the next breaking schema change is where it would bite.

`grapherConfigSchema` was already declared in `multidim-schema.json`, but `Collection` had no such field and `dataclass_from_dict` drops unknown keys — so writing it in yaml was a no-op. Hence the code change alongside the pins.

| File | Change |
|---|---|
| `etl/collection/utils.py` | `resolve_grapher_schema()` — accepts `"011"` or the full URL, falls back to `DEFAULT_GRAPHER_SCHEMA` |
| `etl/collection/model/core.py` | `Collection.grapher_schema`, validated at init, sent as `grapherConfigSchema` on upsert |
| `schemas/multidim-schema.json` | `grapherConfigSchema` → `grapher_schema`, `pattern`-enforced |
| 75 × `etl/steps/export/multidim/**/*.yml` | `grapher_schema: "011"` |

```yaml
grapher_schema: "011"                                                            # short
grapher_schema: https://files.ourworldindata.org/schemas/grapher-schema.011.json  # full
```

**Quote the short form.** Our loader is YAML 1.1, so a bare `011` is octal and reaches Python as `9`. Rather than zero-pad that back into the wrong version, it raises and says to quote it.

`Explorer` raises if the field is set — explorers reach Grapher via the legacy TSV path, which has no `grapherConfigSchema`, so it would do nothing. The equally inert entry in `explorer-schema.json` is removed.

### The one pin this PR corrects, and why

`migration/latest/migration_flows.py` pinned `$schema: 005` in its view config. That pin was **provably false**: the config uses `chartTypes`, which does not exist in schema 005 — the 005→006 migration is what *creates* it (from the old `type`/`hasChartTab`). ETL has been validating this config against 011 all along (`multidim-schema.json` `$ref`s + `additionalProperties: false`), so a genuinely 005-shaped body would fail CI. The pin was stale copy-paste, not a record.

Keeping it means Grapher runs the 005→011 migrations over an 011-shaped config. Most are harmless deletes of absent fields, but `migrateFrom007To008` fires on `colorScale.customNumericValues` and prepends `customNumericMinValue ?? 0`:

```
authored:      [0, 1000, 3000, …, 1000000, 0]
after 007→008: [0, 0, 1000, 3000, …, 1000000, 0]   ← spurious leading bin
```

So the pin is dropped and those views inherit the collection-level `011`. **This changes the map legend** — worth eyeballing on staging. This MDIM has no row in production's `multi_dim_data_pages`, so nothing live is affected today; it is latent on the upsert path.

(The trailing `0` is authored in the step and looks wrong independently of this. Left alone.)

**This is the exception, not the rule** — a pin that contradicts its own config body is stale rather than historical. Documented as such in `/sync-grapher-schema`, `docs/guides/grapher-schema-sync.md`, `mdims.md` and the schema guidelines.

All four `005` pins in the repo were false in the same way — every one uses `chartTypes`, and `energy_prices` also uses `map.region` (created by the 006→007 rename). They are dropped, and now inherit the collection-level `011`.

Every pinned config in the repo was run through grapher's own `migrateGrapherConfigToLatestVersion` to check what the migrations actually do to it:

| Pin | Files | Contradicted by body? | Altered by migration? | This PR |
|---|---|---|---|---|
| 005 | `migration_flows` | yes (`chartTypes`) | **yes** — bins corrupted | pin dropped |
| 005 | `natural_disasters/shared` (×2), `energy_prices`, `vaccination_coverage` | yes (`chartTypes`, `map.region`) | no | pin dropped |
| 008 | 10 × `education/*` (18 dicts) | no | no | **left as-is** |

Only `migration_flows` changes on the wire; the other three are provably no-ops (22 configs checked, 1 altered), removed so the collection-level pin is the single source of truth.

The `education` `008` pins are the one case where the pin is *not* contradicted by its config body, so per the rule above they are left alone. Worth noting they are still shadowing the collection pin, and our CI validates those configs against 011 — so if they have in fact been edited since 008, they are a latent version of the `migration_flows` bug. Happy to drop them too, but that is a judgement call about 10 actively-maintained MDIMs rather than a provable fix.

### Notes for review

- **`test_multidim_configs_pin_grapher_schema` enforces the pin** on new MDIMs. The default alone would work without it; it is there so we don't drift back to the current state. Happy to drop it if that is too strict.
- **Do not bump these pins on a schema version bump** — that would re-create the bug. Documented in `/sync-grapher-schema` and `docs/guides/grapher-schema-sync.md`.
- **Unrelated drive-by, needed for green CI:** `test_dataset_schemas` full-scans whenever a PR touches `schemas/`, and that scan already failed on master (`garden/covid/2024-11-05/github_stats.meta.yml` puts a Jinja template in `display.timeInterval`). Wrapped that enum in the standard `oneOf` + `pattern: "<%"` escape hatch in `definitions.json`.

### Verification

- 673 unit tests + `make check` green. New tests cover the resolver, model round-trip, schema validation, and the combine path used by multi-table MDIMs.
- All 75 configs still parse and build a `Collection`, identical to master (3 partial configs whose choices come from the table fail on both).
- Live `grapher-schema.latest.json` `$id` is `011`, so the pin is current and the migration is a no-op. Had prod been behind `011`, `migrateGrapherConfigToLatestVersionAndFailOnError` would have thrown on every upsert.
- **Not verified:** nobody has run the steps. `migration_flows` and one `education` MDIM (to confirm no visible change there) should be upserted to staging.


lucasrodes:
@codex review

PR #6715, opened 2026-08-19
📜 Update create-figma-chart for the restructured Charts templates

paarriagadap:

The design team restructured every template in the yearly Charts file, which moved a number of things `create-figma-chart` had recorded as stable. This re-measures them and folds in the conventions a designer's rework of a shipped page demonstrated.

Found while building a static chart from the food-trade bespoke visualization; the frames themselves are not part of this PR.

## Template geometry

- **The logo is a sibling of the header, not a child.** It no longer contributes to header height, so the band tracks the copy alone. What still constrains a title is the title node's own width, which is narrower than the content box to clear the logo (737.84 against 818 on Static Vertical, 428 against 508 on the 540-wide set) — that is the number a candidate title has to be measured against.
- **The 850-wide pair lost its wrappers' inner padding.** The content box reads 16/818 directly rather than 0/850, and the frame band *is* the text band — the 16px correction the file used to require on that pair is gone. Bands are now 118 → 559 (Horizontal) and 118 → 1015.81 (Vertical).
- **New footer ids and heights** on that pair, and **every footer is auto-layout** now.
- **The small-chart templates carry a real frame fill**, so the hidden-fill-plus-backing-vector repair is no longer needed.
- Header and footer frames were renamed wholesale, which is the concrete case for resolving them structurally (topmost/bottommost auto-layout child) rather than by name. The structural resolver survived untouched; every name-based lookup would have returned `null`.

## Two properties a filled clone has to check for itself

Both are invisible until the texts are in, and neither shows up in a screenshot.

- **Does the header actually hug?** A header left at `primaryAxisSizingMode: "FIXED"` with a `FILL` / `layoutGrow: 1` child does not reflow — `headerBottom` stays at the placeholder's value however short the copy is, and the slack inflates the flexible child's *box* instead. A one-line title left a subtitle box 67px tall for 19px of ink, which buries ~48px of dead air under the subtitle once you fit to the frame band.
- **Which way does a footer grow?** `constraints.vertical` is `MIN` on every template except static mobile 1, so a footer that gains a line grows *downward out of the frame* rather than upward into the band. Re-pin the bottom by hand; it is a no-op where the constraint is `MAX`, so it can be unconditional.

## Design conventions

From diffing a designer's rework of a page this skill produced:

- **Abbreviate units in data labels, spell them out in prose.** `1.3 million tonnes (48%)` becomes `1.3M t (48%)` on the mark, while the note still says "26,000 tonnes". Keep digit groups as digits. On a sankey this recovered 59px of label-column width on both frame sizes.
- **Label every mark's identity, but only the values worth reading.** Keep the name on every mark and drop the value below a cut-off chosen on the marks rather than the numbers.
- **Bold only the names in a multi-author credit** — the template ships one bold run, so several names bolds the connectives too.

Plus the two mechanical consequences: a shortened label shrinks the chart group (fix with a wider render, never a rescale, which would take every label off the ladder), and a label that loses its pair partner is left off-centre on its mark.

## Tooling

`verify_templates.js` now also reports header sizing (`headerSizing.reflows`), the footer's `constraintV` / `growsOutOfFrame`, and the logo's placement — so this class of change surfaces in the first diff rather than mid-build.

`BESPOKE-SVG.md` additionally records that an imported group's ink is inset from the SVG canvas (so the render must be solved against the group, not the declared size), and how to render a bespoke component against a filtered dataset when its config cannot express the change — with the warning that doing so is a data decision that has to be quantified and signed off, not a styling tweak.

## Checks

`make check` passed on commit. `codespell` clean over the three markdown files. Frontmatter re-verified as parsing, with the full description intact and `internal: true`.

## Still open

- Nobody has reviewed the rendered frames these lessons came from; Figma comments are not readable over MCP, so design review is pending rather than clean.
- Two parallel sessions were folding lessons into other skills at the same time; if any of them touched `create-figma-chart`, the regions to check are `SKILL.md`'s node-map and Step 7 tables, `GUIDELINES.md`'s Text and Direct labeling sections, and `BESPOKE-SVG.md`'s render-to-band section.


paarriagadap:

Superseded by #6717, which merges this branch in full alongside the DI Chart Library sweep and the designer-rework lessons. Every line added here is present there — the only edits were in one `SKILL.md` block where this branch and #6716 documented the same designer-rework behavior from different runs; those overlapping claims were consolidated rather than duplicated, and the merge notes in #6717 spell out exactly what changed.

Closing so the work lands once. Reopen if anything looks missing.

paarriagadap:

Superseded by #6717, which carries all of this plus the Chart Library lessons from a parallel session.

Verified before closing rather than assumed: every substantive line added here is present in #6717 — `GUIDELINES.md` and `BESPOKE-SVG.md` verbatim, `SKILL.md` and `verify_templates.js` with some passages reworded but nothing dropped. Two of the rewordings are corrections worth noting: Static Vertical's header/footer widths now both read 818 (this PR recorded 817.57, which the design team has since fixed), and the script gained a per-template `EXPECT` block.

PR #6716, opened 2026-08-19
📜🤖 Fold a designer's DI rework into create-figma-chart

paarriagadap:

A designer replicated a data-insight page that `create-figma-chart` had produced, applied seven changes and left notes explaining each one. This folds those in as general rules, with the measurements they were derived from.

Docs only — two markdown files under `.claude/skills/create-figma-chart/`. No code, no pipeline steps.

### Annotations — the substantive change

The knockout mechanic was one recipe applied to every annotation: wrap the text in a hugging auto-layout frame filled with the template's canvas color. It becomes **three tiers, chosen per annotation**:

1. no treatment when the block crosses no chart ink,
2. a **3px canvas-colored `OUTSIDE` stroke on the text** when it crosses furniture — the new default,
3. the opaque hugging frame only where a halo cannot carry the words.

Checked on the rendered frames before rewriting, rather than taking the instruction on trust: the frame is sized to the *longest* line, so it blanks a block-shaped hole in the gridline behind it — including alongside every shorter line, where there is no text at all. The stroke interrupts the gridline only at the letterforms. The frame recipe and its two traps (`clipsContent = false`, `paddingBottom`) survive, demoted to tier 3, which is also where the descender-clipping gotcha now belongs.

### The rest

- **Line charts.** On a single-series chart the end label goes *below* the last point, right-aligned on the content edge, not beside it. Beside-the-line reserves the label's full width as right margin; moving it took the plot from 485 to 518.6 on a 540 frame — 7% more plot, with the end dot carrying the attachment. In a two-line `name / value` label, step the **weight**, not the size, so both lines stay on the ladder rung the rest of the plot sits on.
- **Dots.** Every value written out gets a dot, including one labelling a plateau. The precision argument for withholding it — a flat run of years does not belong to any single year — loses to the ambiguity it creates, since the reader's first question about a number is which mark it belongs to.
- **Arrows.** An annotation's arrow takes the annotation gray, bound as a style on both the filled head and the 1px curve, never the series color: colored words tie a sentence to a series, an arrow ties it to a point. And default to placing an annotation in clear space with an arrow rather than tucking it against its mark — which also dissolves the axis-label collision that proximity creates, where a line of annotation ends up level with a y-axis tick and `0.6` reads as part of the sentence.
- **Reading a rework, as method.** A rework may arrive as a replica *beside* your frame rather than an edit of it. Consequences now recorded: where a written note and the executed frame disagree the note is the rule (a stated 16px margin against an executed 14px, which put ink outside the content band); a replica inherits your frame name, so the page carries the export slug twice and would export two files with one name; and the `chart` group does not survive, so `chart.height` — which the fit step measures the band gap from — is gone, and any re-entry script must resolve the plot from its named subgroups.
- **One API trap the rework exposed:** a text node carries a *scaled* `strokeWeight` after `rescale()`, so assigning the 3px halo without setting the weight gives a sub-pixel halo indistinguishable from none.

### Still open

- **Not verified with the pixel probe.** The skill's own acceptance test for arrowheads is a hide-and-diff pixel measurement, because rotated groups make vector distances unreliable. The tip gaps quoted here (3–5px) are derived from `absoluteBoundingBox` and a 6× crop instead — inside the documented 3–7px band, but not the prescribed method. Probing would have meant toggling visibility on a colleague's frame.
- **The Figma page still carries two frames with the same export slug** — the version this skill built and the rework. Whichever ships keeps the bare slug; the other needs renaming. Not done here, since which one ships is a design call.
- **The shipped frame does not yet contain these fixes.** The rework embodies all seven; the frame this skill produced has none of them. Either that frame gets updated or the rework becomes the deliverable.
- **Nobody has checked this against the design team's own guidelines file.** The rules here come from one rework of one line chart; the tier model in particular is a generalisation from two annotations on a single page, and the bar-chart and map cases are untested against it.


paarriagadap:

Superseded by #6717, which merges this branch in full alongside the DI Chart Library sweep and the restructured-templates work. Every line added here is present there — the only edits were in the `SKILL.md` designer-rework block, where this branch and #6715 described reworks seen on different runs. Their duplicate-frame-name bullets said the same thing twice and are now one bullet carrying both sets of specifics; the bullet here on following the written note and #6715's on reading the rule off the reworked frame looked contradictory, so both are scoped ("where the note states a precise intent" vs "where the note is vague") under a bridging line. The merge notes in #6717 record this.

Closing so the work lands once. Reopen if anything looks missing.

PR #6756, opened 2026-08-25
🤖 Autoupdate: Epoch

owidbot:


lucasrodes:

The staging build was failing on `meadow/artificial_intelligence/2026-01-30/frontiermath`, not on the Epoch CSVs themselves. Pushed a fix in bfd5a97.

**Cause:** on 2026-09-01 Epoch's `epoch_capabilities_index.csv` (inside `benchmark_data.zip`) started shipping `deepseek-r1-0528-qwen3-8b` twice — two byte-identical rows — which trips `tb_registry.format(["model_version"])`:

```
ValueError: Index has duplicate keys: Index(['deepseek-r1-0528-qwen3-8b'], name='model_version')
```

That matches the CI history exactly: green Aug 25–31, red from Sep 1. The registry snapshots confirm it (867 rows and no duplicates on Aug 31 → 900 rows with the duplicate on Sep 1). `Model version` is the only column populated in every row, so there is no other field that distinguishes the pair, and no composite key to fall back on.

**Fix:** deduplicate after narrowing to the two columns the step reads. Only rows indistinguishable in the data we use get collapsed; one version carrying two *different* names still reaches `format()` twice and fails loudly there. The row is unused either way — that model has no FrontierMath score.

The rest of the update is unaffected: both Epoch CSVs still have identical headers (47 and 34 columns, same names and order), and `epoch` / `epoch_compute_intensive` / `epoch_regressions` / `epoch_aggregates_affiliation` all built fine — 3 new models (MiniMax-M3, LongCat-2.0, Qwen3.8-2.4T-A95B) and DeepSeek-V4-Pro-0813 withdrawn upstream.

One thing worth a human eye before merging: chart-diff reports **no charts for review** despite the data changes, so the regression-line charts (`epoch_regressions`) haven't actually been eyeballed — worth checking with **Show all charts** enabled, since the regression endpoints shift with every autoupdate.

lucasrodes:

Correction to my comment above: the registry duplicate was only *one* of two red checks. Full picture:

| check | cause | fix |
|---|---|---|
| `Build ETL catalog` | duplicated row in Epoch's `epoch_capabilities_index.csv` | bfd5a97 — confirmed `failed → passed` on builds 43007 and 43013 |
| `Run ETL integration tests` | `test_vendored_grapher_schema_is_current` — this branch's vendored `schemas/grapher-schema.011.json` was stale | a2a2e0e — merged master (the branch was 18 commits behind) |

The second one is not epoch-related at all. Upstream mutated the grapher schema in place (a `sortBy` description edit dropping "marimekko"); master synced it on 2026-09-01 in #6785, and every branch whose last master merge predated that sync fails this test deterministically — `feature-chart-critic-llm`, `metadata-diff` and `data-add-longrun-us` each went red and then green on their next push today for the same reason. Worth knowing for future autoupdate PRs: the daily "Merge master into auto-epoch" step is what normally keeps this green, and the merge did not happen on 2026-09-02 (build 42958 for that merge commit died at "Create staging container", exit 128).

Also added in 17968ed: `Saudi Arabia` in `epoch_compute_intensive_countries.countries.json`. It has been in the source data without a mapping entry, so every run logged it as a missing country; its OWID name is identical, so the data does not change.

Not verified by me: `tests/test_snapshot.py::test_dvc_add_skips_upload_of_snapshot_already_on_r2` fails in my local environment with `401 Unauthorized` on `HeadObject`, but it fails identically on `master` there — local R2 credentials, not this branch. It passed on CI for every other branch today.

Still open before merging: the charts. chart-diff reports no charts for review despite the data changes, so the `epoch_regressions` lines have not actually been eyeballed — worth opening chart-diff with **Show all charts** enabled, since the regression endpoints move with every autoupdate.

PR #6765, opened 2026-08-26
🐛 Follow grapher's chart column renames for the ETL layer

pabloarosado:

Stacked on [#6511 "Charts as first-class citizens in ETL, addressed by chart config UUID"](https://github.com/owid/etl/pull/6511). One commit, no behaviour change.

[owid/owid-grapher#6826 "ETL-authored chart configs, addressed by chart config UUID"](https://github.com/owid/owid-grapher/pull/6826) renamed two columns of the `charts` table after the ETL branch was last refreshed:

- `charts.configIdETL` → `charts.patchConfigIdETL` (an authored layer, like `charts.patchConfigId` and `variables.patchConfigIdETL`)
- `charts.catalogPath` → `charts.etlConfigCatalogPath` (the step that owns the ETL layer; not an identifier)

Without this, the ETL branch cannot load a chart from a grapher running that PR: chart-diff fails with `Unknown column 'charts.configIdETL'` as soon as a chart is listed (observed on `staging-site-chart-diff-layers`).

This renames the ORM columns and every read of them in chart-diff, chart-sync and `etl chart-config-id`, and makes the probe that checks whether production already has the columns look for the new names.

## Test plan

- Unit tests for chart-diff, the grapher model and the chart-config-id CLI pass; `make check` clean.
- Against a staging server running the grapher branch, chart-diff and chart-sync (dry run) were exercised end to end on the chick-culling chart: unrelated production edits are not listed, a rebuild or a staging admin edit is listed, a production edit after the staging server was created shows as a conflict, and an approval expires when production is edited afterwards.

### DIFF
--- a/apps/chart_config_id/cli.py
+++ b/apps/chart_config_id/cli.py
@@ -134,8 +134,8 @@ def lookup(config_file: Path, slug: str | None, chart_id: int | None, env_name: 
     console.print(f"Looking up {target} in [bold]{owid_env.conf.DB_NAME}[/bold] on {owid_env.conf.DB_HOST}")
 
     # Deliberately a plain SELECT of the columns we need rather than loading the ORM `Chart`: some
-    # environments (production, until the grapher release lands) lack the `catalogPath` /
-    # `configIdETL` columns the model declares, and selecting those would fail with "Unknown column".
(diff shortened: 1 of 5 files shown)


pabloarosado:
@codex review

PR #6778, opened 2026-08-31
🎉 chart-critic: LLM review of published charts

Marigold:

Adds `etl chart-critic`: send a published chart's render, its metadata and a plain numeric summary to an LLM, and get back what would mislead a reader.

```bash
etl chart-critic --sample 20                          # 20 charts, sampled by readership
etl chart-critic --top 50 --output /tmp/critic.html   # most-viewed, with an HTML report
etl chart-critic --slugs life-expectancy --repeat 3   # named charts, three passes
etl chart-critic --mdim-views 2                       # + declared multi-dim views
etl chart-critic --eval                               # nine known-answer cases, nonzero on regression
etl chart-critic --changed-since 1 --digest f.txt     # a Slack-ready digest of what changed
```

## Why

It targets the class of error our checks are structurally blind to: one that is **stable over time and statistically unremarkable**, where the only tell is knowing something about the world.

owid-issues#2507 is the motivating case. The UN swapped female and male across SDG 16.2.2, so our data said 95% of adults trafficked for sexual exploitation were male. The series was smooth, internally consistent and identical across our 2023, 2024 and 2025 imports — no jump, no outlier, nothing to diff. The issue's own post-mortem put it plainly: *"Catching this class of problem needs a plausibility check on the direction of a breakdown, not a diff."*

**Anomalist keeps its job.** Time-series, upgrade-diff and distribution detection are not duplicated here; this only asks for judgement about published output.

## Does it work?

The swapped 16.2.2 values are still in our catalog (`un_sdg/2025-10-29`, female share 5.4%) alongside the corrected import (`2026-07-31`, 94.5%), which makes a real blind A/B possible. Given the render and nothing else — no mention of sex, swapping, or that an error exists — `gemini-3.7-flash` answered:

> "The series labels for Male and Female appear to be accidentally swapped… UNODC and SDG 16.2.2 data consistently show that the vast majority (around 90% or more) of detected victims of trafficking for sexual exploitation are female, not male."

On the corrected arm it said nothing about a swap. `gemini-3.1-flash-lite` got it too, for $0.00018.

Then sweeps over **200 charts** sampled by readership: **11 claims, 10 verified true, 1 false positive** (a chart-config sentinel misread, caught and guarded the same day). Nine are filed on #6779 — including CAR's life expectancy oscillating between 19 and 57 years behind the site's fifth most-viewed chart, three charts whose UK coal share exceeded 100%, COVID death rates carrying the unit "doses", and two charts whose subtitle names a different currency year than the indicator.

**One has since been fixed, and the fixture is how we found out.** The UK electricity data was corrected (#6782); the `share-elec-by-source` case flipped to FAIL, which is what a known-answer set is for. It is now kept as a clean case.

## Precision is fine. Recall is the honest weak spot.

Counting every cached pass: **eight of the nine fixtures answer identically on 15/15 passes**, and one lands **1 in 15**. So the critic is far more deterministic than "LLMs are flaky" suggests — the variance is concentrated in a single hard case, the World COVID vaccination series that opens at 10.14% before anyone was vaccinated. It is neither the min nor the max, so it never reaches the numeric summary, and in the image it is the left endpoint of one line among thirteen.

That case is marked `flaky` and excluded from `--eval`'s gate, reported as a recall probe instead. A gate that fails seven days in ten is a gate people learn to click past.

## What was measured, not guessed

Each is a comment in the code, because each cost a wrong turn:

1. **The prompt must state today's date**, or recent data gets flagged as future-dated. It did so twice.
2. **The instruction belongs in the user turn, not `instructions=`.** Moving it into the system prompt silently trebled the miss rate: 0/3 and 0/3 versus 2/3 and 3/3, with the clean control unaffected either way.
3. **Temperature 0 is worse here** (1/3 → 0/3 on a known finding) and the model varies regardless, so it stays at the default.
4. **`--repeat` is the recall dial**, and per the numbers above it buys attempts at the hard case rather than lifting a general fog.
5. **Axis bounds must not be shown.** `yAxis: {"max": 0}` is grapher's auto-scale default, and passing it through produced three confident false positives across two sweeps and zero true findings.
6. **Aggregate-vs-members context was tried and reverted.** Africa at 465% of the four countries that report is legitimate, and separating that from a real defect needs a coverage threshold — a deterministic check, which belongs in Anomalist.

## Two layers of cache, because the failure modes differ

Bundles are keyed on the fetching code's fingerprint plus the chart URL (24h TTL); reviews on the prompt fingerprint, the model, the exact bytes the model was shown, and the pass index. Two bugs found here are worth knowing:

- **All cached passes are replayed, not the first `--repeat` of them.** Replaying a fixed slot that happened to miss froze a chart as "clean" forever. Using everything already paid for means the cache can only add findings.
- **Every image counts in the key.** `--views 1` and `--views 2` shared an entry, so whichever ran first answered for both.

## The daily digest

`--changed-since 1 --digest f.txt` reviews the charts whose configuration changed and writes a Slack-ready message; `--post` sends it to `#we-need-to-correct-it` as owidbot, for the scheduled path where no human sits between the file and the channel. The pipeline is owid/ops#647, and it needs this merged first.

It is quiet by design: nothing to say produces no message. Findings are fingerprinted **per indicator** for data-level claims, so one defective column reported on five charts is one finding — the UK coal share was found three times on three charts from one column of one ETL step. `--post` refuses to run without `SLACK_API_TOKEN` rather than letting `send_slack_message` print to stdout, because a green job over a silent channel is the failure nobody notices.

Format follows #analytics-bites: linked title, the claim, then severity, `kind`-level, views/day and an admin edit link.

## Cost

~$0.003–0.01 per chart with the render attached on `gemini-3.7-flash`, so a pass over all ~4,280 published charts is roughly $12–40; `--cheap` (flash-lite) is about a fifteenth of that. The digest is ~$0.20/day for ~17 changed charts, plus ~$0.20 for the gating eval. Cost is printed after every run, priced through `apps/utils/llms/costs.py` with published rates as a fallback.

**The bill is not the constraint.** How many findings a human will read is — which is why an empty result is a first-class outcome, the digest caps at five, and the report shows a chart image only next to a finding.

## The recurring theme, and how the code answers it

A broken prompt, a stale cache, a dropped chart and a genuinely clean catalog all produce **identical empty output**. Most of the review effort on this PR went there:

- the fixtures exist because a prompt regression looks exactly like a quiet day;
- a chart that could not be fetched no longer counts as a fixture pass, and no longer reads as "clean" in the HTML report;
- a CSV that 5xxs fails the chart, while the non-redistributable 200-with-JSON-body still degrades gracefully — those were the same code path;
- a failed mdim or chart-config lookup is distinguished from "there is nothing there", and the run says which;
- the command exits **2** when any chart went unreviewed, after every output is written.

## How to test it

```bash
etl chart-critic --eval                                  # ~$0.20, a few minutes; expect 8/9 and exit 0
etl chart-critic --slugs life-expectancy-vs-gdp-per-capita --repeat 3
etl chart-critic --sample 5 --output /tmp/critic.html     # open it; images appear only beside findings
```

The eval's one expected miss is the case marked `flaky`. For the digest path, `--changed-since 1 --digest /tmp/d.txt` **without** `--post` writes what it would send and sends nothing; it needs a production DB, since `chart_configs.updatedAt` on a staging copy records when the copy was made.


Marigold:
All seven findings were real and all seven are fixed in 5425d58. Notes where the fix differed from the suggestion, or where verifying it changed something else:

**`bundle.py:82` — use `etl.http.session`.** Done, and the premise behind the spoofed UA turned out to be wrong. The comment claimed the endpoints need a *browser-like* UA; in fact only the default `python-urllib` one is rejected. `owid-etl (python 3.11.15)` gets 200 on `.metadata.json`, `.csv` and `.png`, so the shared session costs nothing and makes the traffic attributable in our own logs. Retries kept, with 404 mapped to `ChartGone` inside `_fetch` now that `requests` raises a different exception shape.

**`cli.py:268` — extra views in the review cache key.** `content_hash` is variadic now and takes every image the model is shown. Verified: `--views 1` and `--views 2` on the same chart produce two distinct cache entries where they previously produced one.

**`mdim.py:45` — enumeration failure vs. not-an-mdim.** New `MdimLookupError`; `config()` returns `None` only when the slug genuinely isn't a published mdim. `--mdim-views N` now aborts with "could not read the multi-dim registry" instead of quietly reviewing the bare slug once, and `--sample` aborts rather than treating every mdim as an ordinary chart.

**`cache.py:132` — stale PNG.** The PNG is deleted whenever a refreshed bundle has none, so a `--no-image` or render-failed refresh can't leave the old image beside new numbers.

**`mdim.py:112` — urlencode.** Fixed, but latent: I audited all 40 published mdims and **zero** dimension slugs contain a character needing escaping. The values I expected to bite (`Richest 1%`, `No spells`) are display labels, not URL slugs. Encoding is still correct — an unescaped value would surface as a bundle failure rather than a wrong answer — the comment just no longer claims an observed bug.

**`report.py:95` — merged params.** Both the render and the link fall back to the merged view now. The merge itself moved into `critic.issue_params()` so the terminal, the digest and the report can't drift apart.

**`cli.py:504` — fixture status gate.** Fixed, and it exposed a second problem: with the gate in place, one fixture failed. Measuring every cached pass showed eight of the nine cases are deterministic (15/15 correct each) and one lands **1 in 15** — while its own note claimed "one pass in three". So a hard gate would have failed a daily scheduled run most days. That case is now marked `flaky` and reported as a recall probe outside the gate; the other eight gate normally.

Thanks — the cache-key and fixture-gate ones were both cases where the failure mode is silent, which is this tool's whole problem.

Marigold:
Round two, all four fixed in 7300a60 — including one that was a regression from my own previous round.

**`cli.py:534` (P1) — `dimension_keys()` on every finding.** Correct, and worse than described: this made the database-free `--slugs` mode succeed *only when it found nothing*, after the reviews had been paid for. The `MdimLookupError` I added last round turned a best-effort lookup into a fatal one at link time. Fixed by taking link keys from the reviewed target's own params — those were either enumerated from the mdim config or typed by the user, so they need no lookup — and consulting the registry only as a tolerated bonus for a dimension the model names that the target did not. The asymmetry is deliberate and commented: a failed lookup during *selection* still aborts, because there the user asked for something we cannot deliver.

**`cli.py:186` (P2) — `%` and `+` in link values.** Fixed. Both fixes were mine and they contradicted each other, which is the more interesting failure: encoding at fetch time and filtering at link time were written a round apart and never met.

**`report.py:115` (P2) — "No issues found" on a fully failed run.** Fixed: failed charts get their own "Not reviewed" section and suppress the success headline, which now reads "No issues found in the N charts that were reviewed — but M could not be reviewed at all." Verified on a synthetic one-chart run where the only chart failed.

**`digest.py:83` (P2) — dedup across slugs.** Implemented rather than documented away, since the docstring already claimed indicator-level dedup. `chart_facts()` resolves each *flagged* chart's indicators (`chart_dimensions` → `variables.catalogPath`) in one query, and data-level findings key on that set; chart-level findings stay per-chart, since a wrong subtitle really is specific to one chart even when the data is shared. Two details worth flagging:

- the catalog path has its version stripped (`/2026-06-30/` → `/`), or tomorrow's re-run of the same step re-posts yesterday's finding under a new key;
- `stamp()` had to take the same mapping. A key written per-chart and looked up per-indicator matches nothing, so the digest would have re-posted every finding every day — a silent bug that only shows up on day two.

Verified: two charts sharing an indicator collapse to one finding, chart-level findings on the same shared indicator stay separate, a version bump does not re-post, and the whole thing falls back to per-chart without a database rather than dropping findings.

Unrelated to the review, in the same commit: `severity` and `confidence` had no field descriptions, so the model was choosing them with no rubric while the digest ranked on severity first. They now carry the project's finding bar. Measured against the previous prompt on the fixture set it changes almost nothing, so it is a definition rather than a demonstrated improvement.

`--eval` 8/9, exit 0 — the one miss is the case marked `flaky` and excluded from the gate.

Marigold:

Round three: three implemented in 8acd0b7, one I think does not apply. The first one is worth reading even if you skip the rest, because my initial fix for it was wrong and `--eval` is what caught that.

**`bundle.py:200` (P1) — CSV failures swallowed.** Correct, and it is the worst shape of bug in this tool: `data_available=False` with `status="ok"` meant the report and the digest could describe a chart as clean when the model never saw a single value.

My first attempt routed every `HTTPError` to a hard failure, and the fixture set immediately went 7/9 — the `suicide-rate-in-1980-vs-2023` "aumber" typo, a case that had passed 15/15 across three prompt revisions. The reason is the thing your finding half-anticipated: **non-redistributable data answers `403` with a JSON body**, not `200`.

```
suicide-rate-in-1980-vs-2023 -> 403 | application/json | {"status":403,"error":"This chart contains non-redistributable data…"}
democracy-index-eiu          -> 403 | application/json | {"status":403,"error":"This chart contains non-redistributable data…"}
literacy                     -> 200 | text/csv         | entity,code,year,…
```

The IHME/GBD family lives there, and those charts are still reviewable on metadata and image — which is literally how that finding was ever made. So the discriminator is the **body, not the status**: a 403 carrying an `error` document is policy and degrades; a bare 403, a rate limit, or a 5xx past its retries is operational and fails the chart. Verified all three: the real IHME chart degrades with the typo still visible, a simulated 503 raises, a 403 with no JSON body raises.

**`chart_config.py:49` (P2) — lookup failures indistinguishable from "no config".** Agreed, and the same class as the mdim one from round one. It now raises `ConfigUnavailable`, but the *bundle* records it as a note rather than aborting — `--slugs` is documented to work without a database, so a hard failure would break the mode. The result is that a review which could not see the chart's configuration says so, in the notes and in the report, instead of looking like a review that saw everything.

**`digest.py:135` (P2) — dedupe on shared indicators, not whole sets.** You are right, and this is a real gap in my own previous fix: keying on the joined set gives `A` and `A|B`, so the identical finding still takes two slots. Now one key per indicator, matched on **any** overlap, with the claim's significant words in every key so two different problems on charts that share an indicator do not collapse into each other. Verified: a column read alone by chart A and alongside another by chart B collapses to one finding, a chart with an unrelated indicator does not, the state file holds across days in both directions, and chart-level findings on shared data stay separate.

**`digest.py:241` (P1) — attribution line.** I do not think this applies. AGENTS.md's rule is scoped to content going out **under a human's identity**, and states the exception explicitly: *"An agent posting under its own bot account, e.g. `chatgpt-codex-connector[bot]`, skips the line: the platform already attributes the content."* `--post` posts as **owidbot**, its own bot account, so the platform attributes it — and the footer already reads "Posted by `etl chart-critic` — an LLM reading each chart, its metadata and its values."

The rule does bite the other path, where a person pastes the `--digest` file into Slack under their own name. That is a human action rather than something the CLI can enforce, so it is written into the operating notes for whoever posts it that way.

`--eval` 8/9, exit 0, $0.23 — the one miss is the case marked `flaky` and excluded from the gate.

Issue #6668, opened 2026-08-13
Update and adapt the `owid_energy` step and the `energy_data` repos


pabloarosado:
Decide what to do with the `energy_data` repos, now that the data has changed significantly (after we no longer use the substitution method).

One option could be to archive the `energy_data` repos, and change the README to link to the Energy mix, Electricity mix, and Fossil fuel mdims, once there's a way to download their full data.

Once `owid_energy` is updated, consider retiring `dag/energy_legacy.yml`.

pabloarosado:
@Marigold maybe we already have the three links to the mdim export zip folder (or whatever format). If that's confirmed I'd just need to rewrite the README :) (and archiving the repos).

Marigold:
@pabloarosado I polished the [mdim download package](https://github.com/owid/etl/pull/6491) so that when merged, all mdim steps (run with `--export`) will publish their own. Here are [examples](https://claude.ai/code/artifact/aea4039e-09f2-449e-be72-d0338185efe4) of the three mdims, could you take a look, please?

pabloarosado:
> [@pabloarosado](https://github.com/pabloarosado) I polished the [mdim download package](https://github.com/owid/etl/pull/6491) so that when merged, all mdim steps (run with `--export`) will publish their own. Here are [examples](https://claude.ai/code/artifact/aea4039e-09f2-449e-be72-d0338185efe4) of the three mdims, could you take a look, please?

Hey Mojmir, thanks for doing this! I have checked the mdim download package. I think it's getting there, but I also think it's not yet optimal. Here are some comments and ideas that came to mind:
- [ ] I'd rephrase a few things a bit in the README. It's a bit too verbose and sometimes redundant.
- [ ] A zip folder with a csv, a markdown, and a json seems to not be optimal either for humans or for machines. If we want to stick with it, I wonder if we need the metadata.json inside the zip folder at all. We could instead add the link to the full metadata inside the markdown, when giving the detailed data about an indicator. My intuition is that humans would ignore it, and bots will fetch the metadata directly from the API.
- [ ] An alternative (or addition) would be to offer an excel file with 4 tabs:
  - Data: same as the csv.
  - Metadata: one row per indicator, where one of the columns is the origins joined in one line.
  - Sources: one row per origin.
  - Readme: a few rows with the generic info (but much less verbose than in the status quo).
  Note that currently we can easily produce excel files with data and metadata, with table.to_excel(); we could extend the logic to include sources and readme. 
[...]

PR #6783, opened 2026-09-01
✨ Rework the download-package readme

Marigold:
## 👨 Summary

Rework data download readme following pablo's suggestions. The best way to see what's going on is to check side-by-side comparison [here](https://claude.ai/code/artifact/772f26df-08f3-40cf-8285-277ffd7f434c).

I haven't changed `.metadata.json` or added Excel format (created a separate issue from it).

**The reworked readme, rendered before/after:** https://claude.ai/code/artifact/772f26df-08f3-40cf-8285-277ffd7f434c

Applies the review feedback in #6668 to the complete-dataset readme. On electricity-mix it goes from 85 kB to 54 kB over the same 46 indicators.

- **Per-indicator headings shift a level.** An indicator was `##`, a sibling of the `## Detailed information about each time series` that introduces it, so the whole second half of the document was flat. They are now `###`, their sub-sections `####`. `#### Notes on our processing step` is the one heading left alone — it was already a level too deep, reading as a child of the last source block, and now lines up with its siblings.
- **The full citation is gone.** It restated producer names the Sources section already gives in far more detail, and across 46 indicators it was a fifth of the file. `metadata_column_entry` still emits `citationLong`, so it stays in `metadata.json`. What is left under `#### How to cite this data` is the in-line citation.
- **`Source:` moves up** with `Last updated`, `Date range` and `Unit`. Upstream prints it immediately after the full citation, where it reads as a trailing fragment of it; with the full citation gone it sat next to the short one looking like a duplicate.
- **Sources come last, after the indicators.** It is reference material a reader arrives at from a citation, not something to read past on the way to the column they opened the file for.
- **"Next update" → "Next expected update".** The date is the producer's stated update period added to our last update — our expectation, not a promise.
- **"How is this data described by its producer - Ember (2026), Energy Institute … (2026), Pinto et al. (2023)?"** → **"How this data is described by its producers"**.
- **"About the data" → "How we process data at Our World in Data"**, and the sub-heading of that name goes away. The old heading announced a description of the dataset and delivered generic text about OWID's processing; that text is what the section is now called.
- **The CSV section is a bulleted list** and says which time columns this package has, rather than hedging "either Year or Day". A chart download's readme is one static string and has to hedge; this one is built per package, and a mixed-resolution collection has a `Frequency`/`Time` pair the hedge does not describe at all.

The `metadata.json` paragraphs and the processing text keep their existing wording — only the headings around them changed.

Two of Pablo's items are deliberately not here: `metadata.json` stays in the zip, and the Excel variant is on hold until we know whether anyone would use it.

## Keeping in sync with grapher

`download_package_format.py` is a port of `readmeTools.ts`, and the readme half has now deliberately moved ahead of it. None of these changes are MDIM-specific, so the same rework is in owid/owid-grapher#7136 for the chart download; the module docstring lists the divergences so they can be deleted once that lands.

## Testing

Six unit tests cover the heading levels, the dropped full citation and moved `Source:` line, the Sources section's position, the per-package time-column description, and the DoD stripping.

Checked end to end against the real thing: the published `electricity-mix.complete-dataset.zip` readme was reproduced byte-for-byte from the same indicator metadata through this code path, then regenerated with the changes applied and diffed side by side.

## What's left open

The published packages keep their old readmes until a forced production run — the deploy's `build_mdims_and_explorers` runs without `--force`, and ETL's change detection cannot see library-code changes.

### DIFF
--- a/etl/collection/download_package.py
+++ b/etl/collection/download_package.py
@@ -827,13 +827,53 @@ def _zip_timestamp(build_date: date) -> tuple[int, int, int, int, int, int]:
     return (build_date.year, build_date.month, build_date.day, 0, 0, 0)
 
 
-def _readme(title: str, page_url: str, column_sections: list[str], sources_section: str = "") -> str:
+def _time_column_lines(time_header: str, frequency_column: bool) -> str:
+    """The bullet describing the CSV's time column(s), for the shape this package
+    actually has.
+
+    Upstream hedges -- "the third column is either Year or Day" -- because a chart
+    download's readme is one static string. This one is built per package, so it can
+    just say which it is, and can describe the Frequency/Time pair that a
+    mixed-resolution collection gets and no chart download has.
+    """
+    if frequency_column:
+        return (
+            '- "Frequency" and "Time" — the resolution a row is reported at ("annual", "monthly", ...) and '
+            'the timepoint at that resolution ("2025", "2026-06"). Collections that report the same metric at '
+            "more than one resolution keep it in a single column and tell the rows apart here."
+        )
+    if time_header == "Day":
+        return '- "Day" — the timepoint, as a date string in the form "YYYY-MM-DD".'
+    return '- "Year" — the timepoint, as an integer year.'
+
+
+def _readme(
+    title: str,
+    page_url: str,
+    column_sections: list[str],
+    sources_section: str = "",
+    time_header: str = "Year",
+    frequency_column: bool = False,
+) -> str:
     """Ported from `constructReadme`'s multi-column branch (readmeTools.ts),
     with the tolerance-column paragraph dropped (a complete-dataset package has
     no tolerance columns) and three MDIM-only additions that have no counterpart
     there: that the package covers every dimension combination, that column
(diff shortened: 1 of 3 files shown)


pabloarosado (approved):
Hi Mojmir, thanks for this work. Everything looks good. Just some minor typos and suggestions (feel free to disregard the suggestions):
* Typo: "etc.." (appears twice).
* Typo: "stich".
* Typo: missing period after "Read about our data pipeline"
* In the indicator's info, the capitalization of OWID is wrong in the Source, it should be "Our World in Data" (not "In"); multiple sources should be separated with ";" (I think Sophia fixed these things recently).
* I see that the section title is "Source" or "Sources" depending on the number of sources. But the text below always says "These are the sources...". I'd just say "Sources" always; the overcomplication is not necessary (otherwise fix both section title and text).
* "All data and visualizations on Our World in Data rely on data sourced from one or several original data providers." This is redundant with the previous paragraph. I'd just start that paragraph with "Preparing this data involves several processing steps…"

PR #6792, opened 2026-09-02
📊 Fix Dependabot vulnerabilities and clear version-update PRs

Marigold:

Clears all 12 open Dependabot security alerts and the one outstanding version-update PR, so the Dependabot PR list ends up empty.

## Security alerts

| Package | From → to | Severity | Advisory |
|---|---|---|---|
| `transformers` | 5.8.1 → 5.10.1 | high | `save_pretrained` path traversal via chat template names |
| `hydra-core` | 1.3.2 → 1.3.6 | high | code execution via `hydra.utils.instantiate` on untrusted config |
| `tornado` | 6.5.7 → 6.5.8 | medium + low | multipart `split()` memory-amplification DoS; cookie attribute injection re-opened via the legacy `**kwargs` path in `set_cookie` |
| `pypdf` | 6.16.0 → 6.16.2 | medium ×2 | long runtimes / large memory when retrieving outlines and extracting XForm objects |

`tornado` was flagged in four lockfiles — root, `lib/catalog`, `lib/datautils`, `lib/owl` — all relocked. It's a direct dependency only in `lib/datautils`, whose constraint moved to `>=6.5.8`; elsewhere it's transitive (Jupyter) and nothing capped it.

`transformers` is transitive via `sentence-transformers`. It's an ML library and the wizard's embeddings depend on it, so it's pinned to the minimum patched release (5.10.1) rather than latest (5.16.1) — that avoids an eight-minor-version behaviour jump for a security fix. `tokenizers` and `safetensors` stayed put as a result. Verified `sentence_transformers` and `apps/wizard/utils/embeddings` still import cleanly.

`owid-datautils` goes to 0.6.10 because its `pyproject.toml` changed and `publish-owid-packages.yml` rejects a republish of an existing version.

## Version updates

Supersedes #6687 (github-actions group), applied here so it shares one CI run:

- `actions/checkout` v6.0.3 → v7.0.1
- `actions/setup-python` v6.2.0 → v7.0.0
- `astral-sh/setup-uv` v8.2.0 → v10.0.1
- `actions/stale` v10.3.0 → v11.0.0

All four are major bumps, so I checked the breaking changes against how we use them:

- checkout v7 blocks checking out fork PRs on `pull_request_target` / `workflow_run` — no workflow here uses either trigger.
- setup-uv v10 disables caching by default on `pull_request_target` / `workflow_run` / `release` — same, none of those triggers exist; the two workflows that disable caching deliberately are unaffected, and `sync-grapher-schema.yml` sets `enable-cache: true` explicitly.
- setup-python v7 removed the `pip-install` input — not used anywhere in `.github/`.
- stale v11 is an ESM/Node-24 migration with no input changes.

Each SHA was verified against the corresponding tag in the action's repo. `pinact` isn't installed locally, so the pins were applied by hand from Dependabot's own (already-pinned) diff and then checked with `gh api repos/<action>/git/ref/tags/<tag>`.

## Verification

`make check-all` passes (root plus every `lib/`).

## Still open

- Nobody exercised the four bumped workflows end to end. `buildkite.yml` and `zizmor.yml` run on this PR, but `publish-owid-packages.yml`, `deploy-docs-cf.yml` (master/dispatch paths) and `sync-grapher-schema.yml` only fire after merge — the publish workflow in particular will run on merge because `lib/datautils/pyproject.toml` changed, which doubles as its first live test of checkout v7 / setup-python v7 / setup-uv v10.
- `transformers` deliberately stops at 5.10.1; a later bump to 5.16.x is a separate, opt-in decision.


Marigold:
@codex review

PR #6793, opened 2026-09-02
📜🤖 create-figma-chart: move Steps 1 and 3 into reference/, holding the spine's 64 KB cap

paarriagadap:

## Why

`SKILL.md` had **19 bytes** of headroom against the 64,000-byte cap `verify_docs.py --structure` enforces. That is not a comfortable margin, it is an exhausted budget: the file has grown about **600 bytes a day** since the big split on 2026-08-20, so the next lesson write-back breaks the build.

It is already costing quality rather than just blocking commits. #6787 had 62 bytes to spend on its Step 2 tripwire, spent 46 of them, and the compression produced a pointer to a `GOTCHAS.md` entry that did not exist. That was caught in review, but the pressure that caused it is still there.

**The cap is not the thing to change, and the file says so itself.** It was set on 2026-08-28 at that day's size plus 450 bytes, in the commit that raised it from 62 KB. The paragraph that did the raising is explicit:

> A cap that moves whenever it is hit is not a cap: the discipline below is unchanged, and the next addition pays for itself by deduplicating rather than by another raise.

So this PR frees space instead. There is room under the combined 145,000-byte cap to raise the per-file one, and taking it would re-arm the same ratchet inside a fortnight.

## What

Steps 6, 7, 8 and 8c are already pure routing — **426 bytes each on average**, a pointer into the reference file that carries the detail. Steps 1 and 3 never got that treatment and held **35% of the spine** between them. This finishes the refactor rather than inventing a structure.

| Step | Before | After | New home |
|---|---|---|---|
| Step 1, resolve the chart | 10,837 | ~1,900 | `reference/RESOLVING-THE-CHART.md` |
| Step 3, export the SVGs | 11,639 | ~2,200 | `reference/EXPORTING.md` |

The spine keeps the **decision** and hands off the per-case detail:

- **Step 1** keeps a seven-row input-to-URL table, one line per route, plus the two rules that must not be skipped whatever the input: never shorten the producer's name, and check every selected entity actually renders. A run reads the one row for the input it was handed instead of carrying twelve prose cells about narrative-chart uuid recovery and Datasette mirror lag.
- **Step 3** keeps the two-exports framing, the ordering rule that avoids a re-export, the `curl` block and a short `solve_export.py` pointer. The aspect solve and its per-axis insets, `imFontSize`, the `imType=square` alternative and the silent `tab=` fallback move out.

Both new files carry the original prose **verbatim**. Nothing was rewritten in transit, and nothing is duplicated between spine and reference.

## Result

| | Before | After |
|---|---|---|
| `SKILL.md` | 63,981 / 64,000 | **47,013 / 64,000** |
| Headroom | 19 bytes | **16,987 bytes** |
| Eagerly-read pair | 115,927 / 145,000 | **98,959 / 145,000** |

No cap changed. At the current write-back rate that is about a month of runway, and the pattern for the next one is now uniform across all nine steps.

## Verification

- `verify_docs.py --against origin/master` — **zero `LOST` lines**, which is the whole safety argument for a move this size. Two `REWORDED` findings, both deliberate: the budget paragraph now records that the cap held and how, and a stale `~180 KB` figure for `reference/` is corrected to `~510 KB`.
- `verify_docs.py --structure` — clean. Both new files are reachable from the spine's file map, and the two markdown links that moved a directory down were re-depthed.
- `test_inline_script.py` — 76/76.
- codespell on the three files — clean.
- `make check` — clean.

Script sizes are untouched, since no script changed. The largest documented `use_figma` call is still `verify_page.js --rows type,geometry` at 40,664 chars, 81% of the 50,000 cap.

## Open items

- The other two fat sections are the **preamble at 11,188 bytes** and **Round-trip budget at 8,317**. Neither is a step, so neither has an obvious reference file to move to, and both are arguably spine material. Left alone deliberately; worth a look if the file tightens again.
- This does nothing about the growth *rate*, only the runway. If write-backs keep landing at 600 bytes a day, the same conversation is due in about a month, and the answer then is likely to be that some of what accumulated is no longer worth its bytes.

### DIFF
--- a/.claude/skills/create-figma-chart/SKILL.md
+++ b/.claude/skills/create-figma-chart/SKILL.md
@@ -81,10 +81,12 @@ Two more sibling files own a route each, and both replace rather than supplement
 
 This page is the **spine**: the step order, the checkpoints, and the routing. The detail for each
 step lives in [`reference/`](reference/) and is read *at* that step, not up front — the whole set is
-~180 KB and no run needs all of it.
+~510 KB and no run needs all of it.
 
 | Read | When | Covers |
 |---|---|---|
+| [reference/RESOLVING-THE-CHART.md](reference/RESOLVING-THE-CHART.md) | Step 1 | Every input form's route to an SVG URL and the trap in each, where each template text comes from, and the entity-completeness check with the right baseline. |
+| [reference/EXPORTING.md](reference/EXPORTING.md) | Step 3, before the embed export | The canvas-versus-chart aspect solve and its per-axis insets, `imFontSize`, the `imType=square` alternative and when it wins, and the silent `tab=` fallback. |
 | [reference/NODE-MAP.md](reference/NODE-MAP.md) | Step 5, before cloning anything | The yearly Charts file's node ids, the ten templates, per-family slot sizes, header sizing and the band table. Run `scripts/verify_templates.js` from here **every run** — a `DRIFT` verdict stops the run. |
 | [reference/TEXTS.md](reference/TEXTS.md) | Step 6 | Filling the template's text slots, and the header reflow that makes the band measurable. |
 | [reference/FITTING.md](reference/FITTING.md) | Step 7 | Measuring the band, importing the embed, unwrapping and scaling. The local-SVG restyle route. |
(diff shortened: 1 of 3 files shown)


paarriagadap:
@codex review

PR #6801, opened 2026-09-02
📊 Maintain GCO cancers attributable to infections dataset

lucasrodes:
> _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Ref owid/owid-issues#2478 · No new data at the source: the three JSON files behind the GCO "Cancers Attributable to Infections" tool are byte-identical to our 2024-09-06 snapshot (server `Last-Modified: 31 Aug 2023`; the tool still applies the de Martel et al. 2020 fractions to GLOBOCAN 2020 incidence). So this PR keeps the existing version (no new variable IDs, no chart remap) and does the maintenance the update checklist asks for.

| Chart | PROD | STAGING |
|---|---|---|
| Share of new cancers caused by infections | [prod](https://ourworldindata.org/grapher/share-of-new-cancers-caused-by-all-known-infections-by-type) | [staging](http://staging-site-data-gco-cancer-infections/grapher/share-of-new-cancers-caused-by-all-known-infections-by-type) |
| Share of new cancers caused by all known infectious agents | [prod](https://ourworldindata.org/grapher/share-of-new-cancers-caused-by-all-known-infectious-agents) | [staging](http://staging-site-data-gco-cancer-infections/grapher/share-of-new-cancers-caused-by-all-known-infectious-agents) |
| Share of new cancers caused by Helicobacter pylori bacteria | [prod](https://ourworldindata.org/grapher/share-of-new-cancers-caused-by-helicobacter-pylori-bacteria) | [staging](http://staging-site-data-gco-cancer-infections/grapher/share-of-new-cancers-caused-by-helicobacter-pylori-bacteria) |

`etl diff` against the published catalog: data identical except the deliberate removal of the source's attributable fraction for all agents (a duplicate of our share); otherwise metadata only.

**What changed**
- Garden sanity checks on the aggregates we compute: World equals the sum of countries, the both-sexes rows we add for cancer sites equal men + women, attributable cases never exceed new cases, shares within 0–100%, expected agents/cancers/years/country count. A soft warning flags that the source's own both-sexes totals differ from men + women by more than 1% in The Gambia and Zambia (source values kept as published).
- Step code brought to the current shape (`run()`, `paths.create_dataset`, `paths.regions.harmonize_names` / `add_aggregates`); snapshot script without `click` boilerplate (produces the same file).
- Indicator text rewritten in plain language. Titles follow one pattern per measure: "Number of new cancer cases (excluding non-melanoma skin cancer) caused by hepatitis B virus among women", "Share of new cases of Kaposi sarcoma caused by infections", "New cancer cases per 100,000 men (age-standardized)". Agent and site labels are no longer lowercased blindly ("hpv", "helicobacter pylori"; the source's "Non-Hodkin" typo is corrected in display text only); the all-cancers total reads "cancer cases (excluding non-melanoma skin cancer)"; "both sexes" is left implicit and men/women spelled out (style guide: women/men for all ages; confirmed by the owner). Titles say "caused by" for brevity; descriptions, where there is room, say "attributed to". `paf` (the source's fraction, all cancers by agent) keeps "(population attributable fraction)" in its internal title only; its public title matches the `attr_cases_share` pattern, and the two agree to 1e-5 where they overlap. `description_processing` on the three computed indicators; `attribution_short`. Draft `description_key` bullets are left commented out pending the owner's review. Chart titles/subtitles are pinned on the charts and are not affected.
- Snapshot `.dvc` carries a NOTE recording the September 2026 source check (nothing new; look for a GLOBOCAN 2022-based release next time).

**Checklist**
- [x] Source checked for a newer release: none (JSON files unchanged since 2023-08-31); NOTE recorded in the snapshot `.dvc`.
- [x] Chain re-run from the branch; `etl diff` vs the published catalog: data identical except `paf`, which is now blank for "all infectious agents" (3 unused duplicate indicators removed, per review); everything else metadata only.
- [x] Sanity checks on our aggregates (World = sum of countries; both sexes = men + women; attributable ≤ total; ranges).
- [x] Staging built green; the three charts render; new indicator text live on the staging API.
- [x] Codex review: no findings.
- [x] Owner decisions: men/women wording kept; "caused by" in titles, "attributed to" in descriptions; "(excluding non-melanoma skin cancer)" kept in the all-cancers labels; `description_key` bullets reviewed by the owner and enabled, plus a fifth on the four sites IARC treats as 100% attributable; owner set to @lucasrodes.
- [ ] Optional: a second topic tag (Global Health) next to Cancer in `definitions.common`.
- [ ] Review by another data scientist before merging (per the update checklist).
- [x] Reminder text updated so next year's assignee checks the source first (owid/owid-issues#2527, cron unchanged).

### DIFF
--- a/etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
+++ b/etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
@@ -4,69 +4,78 @@ definitions:
     presentation:
       topic_tags:
         - Cancer
-
+      attribution_short: IARC
     processing_level: minor
     display:
       numDecimalPlaces: 1
-  sex: |-
-    <% if sex == "both" %>all individuals<% elif sex == "males" %>males<% elif sex == "females" %>females<% endif %>
-  footnote: |-
-    To allow for comparisons between countries and over time, this metric is [age-standardized](#dod:age_standardized).
+    description_key:
+      - Estimates come from the International Agency for Research on Cancer (IARC). They apply the population attributable fractions from the study "Global burden of cancer attributable to infections in 2018" (de Martel et al., 2020) to the GLOBOCAN 2020 cancer incidence estimates.
+      - |-
+        Ten infectious agents classified as carcinogenic by IARC are covered: Helicobacter pylori, human papillomavirus (HPV), hepatitis B and C viruses, Epstein-Barr virus (EBV), human herpesvirus type 8, human T-cell lymphotropic virus, the liver flukes Opisthorchis viverrini and Clonorchis sinensis, and Schistosoma haematobium.
+      - Non-melanoma skin cancers are excluded because their registration is often incomplete and inconsistent across countries.
+      - Estimates carry uncertainty, especially for countries without population-based cancer registries or with limited data on how common these infections are.
+  sex: <% if sex == "both" %>both sexes<% elif sex == "males" %>men<% elif sex == "females" %>women<% endif %>
+  people: <% if sex == "both" %>people<% elif sex == "males" %>men<% elif sex == "females" %>women<% endif %>
+  agent: |-
+    <% if agent == "All infectious agents" %>all infectious agents<% elif agent == "HPV" %>human papillomavirus (HPV)<% elif agent == "EBV" %>Epstein-Barr virus (EBV)<% elif agent == "Hepatitis B virus" %>hepatitis B virus<% elif agent == "Hepatitis C virus" %>hepatitis C virus<% elif agent == "Human T-cell lymphotropic virus" %>human T-cell lymphotropic virus<% elif agent == "Human herpesvirus type 8" %>human herpesvirus type 8<% else %><< agent >><% endif %>
+  cancer: |-
+    <% if cancer == "All cancers but non-melanoma skin cancer (C00-97, but C44)" %>all cancers (excluding non-melanoma skin cancer)<% else %><< cancer.lower().replace("hodkin", "hodgkin").replace("hodgkin", "Hodgkin").replace("burkitt", "Burkitt").replace("kaposi", "Kaposi").replace("t-cell", "T-cell") >><% endif %>
+  footnote: To allow for comparisons between countries and over time, this metric is [age-standardized](#dod:age_standardized).
+  processing_regions: |-
+    For world regions, Our World in Data sums the values of the countries in the region with data. For cancer sites, the source only reports men and women separately; the both-sexes values are the sum of the two.
 
+# Learn more about the available fields:
+# http://docs.owid.io/projects/etl/architecture/metadata/reference/
 dataset:
   update_period_days: 365
-
-
   owners:
     - Tuna Acisu
+    - Lucas Rodés-Guirao
+
 tables:
   gco_infections:
     variables:
-      attr_cases:
-        title: |-
-          Number of new << cancer.lower() >> cases attributable to << agent.lower() >> among {definitions.sex}
-        description_short: |-
-          Estimated number of new << cancer.lower() >> cases attributable to << agent.lower() >> among {definitions.sex}.
+      cases:
+        title: Number of new cases of {definitions.cancer} among {definitions.sex}
+        description_short: Estimated number of new cases of {definitions.cancer} among {definitions.sex}.
         unit: cases
+        description_processing: "{definitions.processing_regions}"
+        processing_level: major
 
-      asir_att:
-        title: |-
-          Age-standardized incidence rate of << cancer.lower() >> attributable to << agent.lower() >> among {definitions.sex}
-        description_short: |-
-          Estimated age-standardized incidence rate of << cancer.lower() >> attributable to << agent.lower() >> among {definitions.sex}.
-        unit: per 100,000 people
-        presentation:
-          grapher_config:
-            note: |-
-              {definitions.footnote}
+      attr_cases:
+        title: Number of new cases of {definitions.cancer} attributable to {definitions.agent} among {definitions.sex}
+        description_short: Estimated number of new cases of {definitions.cancer} among {definitions.sex} that are attributable to {definitions.agent}.
+        unit: cases
+        description_processing: "{definitions.processing_regions}"
+        processing_level: major
+
+      attr_cases_share:
+        title: Share of new cases of {definitions.cancer} among {definitions.sex} attributable to {definitions.agent}
+        description_short: Estimated share of new cases of {definitions.cancer} among {definitions.sex} that are attributable to {definitions.agent}.
+        unit: "%"
+        short_unit: "%"
+        description_processing: |-
+          Computed by Our World in Data as the number of attributable cases divided by the total number of new cases. {definitions.processing_regions}
+        processing_level: major
+
       paf:
-        title: |-
-          Proportion of << cancer.lower() >> cases among {definitions.sex} attributable to << agent.lower() >>
-        description_short: |-
-          Proportion of << cancer.lower() >> cases among {definitions.sex} attributable to << agent.lower() >>.
-        unit: '%'
-        short_unit: '%'
+        title: Population attributable fraction of {definitions.cancer} among {definitions.sex} due to {definitions.agent}
+        description_short: Estimated share of new cases of {definitions.cancer} among {definitions.sex} that are attributable to {definitions.agent}, as published by the source.
+        unit: "%"
+        short_unit: "%"
 
       asir:
-        title: |-
-          Age-standardized incidence rate of << cancer.lower() >> among {definitions.sex}
-        description_short: |-
-          Estimated age-standardized incidence rate of << cancer.lower() >> among {definitions.sex}.
-        unit: per 100,000 people
+        title: Age-standardized incidence rate of {definitions.cancer} among {definitions.sex}
+        description_short: Estimated number of new cases of {definitions.cancer} per 100,000 {definitions.people} in a year, age-standardized.
+        unit: cases per 100,000 people
         presentation:
           grapher_config:
-            note: |-
-              {definitions.footnote}
-      attr_cases_share:
-        title: |-
-          Share of new << cancer.lower() >> cases among {definitions.sex} attributable to << agent.lower() >>
-        description_short: |-
-          Share of new << cancer.lower() >> cases among {definitions.sex} attributable to << agent.lower() >>.
-        unit: '%'
-        short_unit: '%'
-      cases:
-        title: |-
-          Number of new << cancer.lower() >> cases among {definitions.sex}
-        description_short: |-
-          Estimated number of new << cancer.lower() >> cases among {definitions.sex}.
-        unit: cases
+            note: "{definitions.footnote}"
+
+      asir_att:
+        title: Age-standardized incidence rate of {definitions.cancer} attributable to {definitions.agent} among {definitions.sex}
+        description_short: Estimated number of new cases of {definitions.cancer} attributable to {definitions.agent} per 100,000 {definitions.people} in a year, age-standardized.
+        unit: cases per 100,000 people
+        presentation:
+          grapher_config:
+            note: "{definitions.footnote}"
--- a/etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
+++ b/etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
@@ -20,13 +20,12 @@ definitions:
   # Readable labels for the three dimensions. The source labels are kept as dimension values (so indicator short names
   # do not change); only the displayed text is rewritten. "Other Non-Hodkin lymphoma" is a typo in the source.
   cases_of: |-
-    <% if cancer == "All cancers but non-melanoma skin cancer (C00-97, but C44)" %>cancer cases<% else %>cases of << cancer.lower().replace("hodkin", "hodgkin").replace("hodgkin", "Hodgkin").replace("burkitt", "Burkitt").replace("kaposi", "Kaposi").replace("t-cell", "T-cell") >><% endif %>
+    <% if cancer == "All cancers but non-melanoma skin cancer (C00-97, but C44)" %>cancer cases (excluding non-melanoma skin cancer)<% else %>cases of << cancer.lower().replace("hodkin", "hodgkin").replace("hodgkin", "Hodgkin").replace("burkitt", "Burkitt").replace("kaposi", "Kaposi").replace("t-cell", "T-cell") >><% endif %>
   agent_phrase: |-
     <% if agent == "All infectious agents" %>infections<% elif agent == "HPV" %>human papillomavirus (HPV)<% elif agent == "EBV" %>Epstein-Barr virus (EBV)<% elif agent == "Hepatitis B virus" %>hepatitis B virus<% elif agent == "Hepatitis C virus" %>hepatitis C virus<% elif agent == "Human T-cell lymphotropic virus" %>human T-cell lymphotropic virus<% elif agent == "Human herpesvirus type 8" %>human herpesvirus type 8<% else %><< agent >><% endif %>
   # Sex is only spelled out for men and women; "both sexes" is the default and is left implicit.
   among_sex: <% if sex == "males" %> among men<% elif sex == "females" %> among women<% endif %>
   people: <% if sex == "males" %>men<% elif sex == "females" %>women<% else %>people<% endif %>
-  nmsc: <% if cancer == "All cancers but non-melanoma skin cancer (C00-97, but C44)" %> Non-melanoma skin cancers are excluded.<% endif %>
   footnote: To allow for comparisons between countries and over time, this metric is [age-standardized](#dod:age_standardized).
   processing_regions: |-
     For world regions, Our World in Data sums the values of the countries in the region with data. For cancer sites, the source only reports men and women separately; the both-sexes values are the sum of the two.
@@ -43,21 +42,21 @@ tables:
     variables:
       cases:
         title: Number of new {definitions.cases_of}{definitions.among_sex}
-        description_short: Estimated number of new {definitions.cases_of}{definitions.among_sex}.{definitions.nmsc}
+        description_short: Estimated number of new {definitions.cases_of}{definitions.among_sex}.
         unit: cases
         description_processing: "{definitions.processing_regions}"
         processing_level: major
 
       attr_cases:
         title: Number of new {definitions.cases_of} caused by {definitions.agent_phrase}{definitions.among_sex}
-        description_short: Estimated number of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}.{definitions.nmsc}
+        description_short: Estimated number of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}.
         unit: cases
         description_processing: "{definitions.processing_regions}"
         processing_level: major
 
       attr_cases_share:
         title: Share of new {definitions.cases_of} caused by {definitions.agent_phrase}{definitions.among_sex}
-        description_short: Estimated share of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}.{definitions.nmsc}
+        description_short: Estimated share of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}.
         unit: "%"
         short_unit: "%"
         description_processing: |-
@@ -68,7 +67,7 @@ tables:
         # The source's own attributable fraction. It only covers the all-cancers total by agent, for countries; the
         # "attr_cases_share" indicator above covers cancer sites, regions and both sexes.
         title: Share of new {definitions.cases_of} caused by {definitions.agent_phrase}{definitions.among_sex} (population attributable fraction)
-        description_short: Estimated share of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}, as published by IARC.{definitions.nmsc}
+        description_short: Estimated share of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}, as published by IARC.
         unit: "%"
         short_unit: "%"
         presentation:
@@ -76,7 +75,7 @@ tables:
 
       asir:
         title: New {definitions.cases_of} per 100,000 {definitions.people} (age-standardized)
-        description_short: Estimated number of new {definitions.cases_of} per 100,000 {definitions.people}, age-standardized so that populations with different age structures can be compared.{definitions.nmsc}
+        description_short: Estimated number of new {definitions.cases_of} per 100,000 {definitions.people}, age-standardized so that populations with different age structures can be compared.
         unit: cases per 100,000 {definitions.people}
         presentation:
           grapher_config:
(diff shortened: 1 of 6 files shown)


lucasrodes:
@codex review

lucasrodes:
@paarriagadap I checked and there is no update to be done on this dataset, so no thorough review is expected. I took this PR to bring the etl step up to speed with current good practices, added some sanity checks, and tried to improve the metadata around it.

Regarding metadata, it's mostly the yaml organization and abstracting some bits. The final render remains almost the same.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
...
+  agent_phrase: |-
+    <% if agent == "All infectious agents" %>infections<% elif agent == "HPV" %>human papillomavirus (HPV)<% elif agent == "EBV" %>Epstein-Barr virus (EBV)<% elif agent == "Hepatitis B virus" %>hepatitis B virus<% elif agent == "Hepatitis C virus" %>hepatitis C virus<% elif agent == "Human T-cell lymphotropic virus" %>human T-cell lymphotropic virus<% elif agent == "Human herpesvirus type 8" %>human herpesvirus type 8<% else %><< agent >><% endif %>
+  # Sex is only spelled out for men and women; "both sexes" is the default and is left implicit.
+  among_sex: <% if sex == "males" %> among men<% elif sex == "females" %> among women<% endif %>
+  people: <% if sex == "males" %>men<% elif sex == "females" %>women<% else %>people<% endif %>
+  footnote: To allow for comparisons between countries and over time, this metric is [age-standardized](#dod:age_standardized).
+  processing_regions: |-
+    For world regions, Our World in Data sums the values of the countries in the region with data. For cancer sites, the source only reports men and women separately; the both-sexes values are the sum of the two.
```
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🔴 This sentence is attached unconditionally, so it ships on all 204 count/share variables — but it's only true on 60 of them.

I rendered every dimension combination through `_metadata_for_dimensions` to count it:

| where the sentence lands | variables | is it true? |
|---|---|---|
| a cancer site, `sex=both` | 60 | ✅ yes — this is the case it was written for |
| **all-cancers, `sex=both`** | **12** | ❌ **no** — the published value is IARC's own both-sexes total |
| men-only / women-only variants | 132 | ❌ irrelevant — no both-sexes summing happened |
| sex-specific sites (cervix, vagina, vulva, penis), `sex=both` | 12 | ❌ misleading — `both` is one sex's value, not a sum of two |

The 12 all-cancers rows are the ones that matter most, for two reasons.

First, the claim is contradicted by this step's own soft check. For `cancer == ALL_CANCERS` the source ships `both` directly and `tb_sites` excludes it, so we pass IARC's figure straight through — and `sanity_check_inputs` warns that exactly this figure disagrees with men + women by >1%:

```
warning  Source both-sexes attributable cases differ from men + women by more than 1% in: ['The Gambia', 'Zambia']
```

So on the Gambia and Zambia data pages the note tells the reader the value is men + women while the step has just logged that it isn't.

Second, blast radius: variable `987291` (`attr_cases_share__sex_both__agent_all_infectious_agents__cancer_all_cancers…`) is in that group, and it's the indicator behind two of the three published charts — `share-of-new-cancers-caused-by-all-known-infections-by-type` and `share-of-new-cancers-caused-by-all-known-infectious-agents`.

The region half of the string is accurate everywhere — I checked the member gaps, and "the countries in the region **with data**" is exactly right (the source covers 185 countries, so continental totals legitimately omit Hong Kong, Macao, Greenland, Mayotte and the small territories). It's only the sex clause that needs gating. Splitting the definition in two would do it:

```yaml
  processing_regions: |-
    For world regions, Our World in Data sums the values of the countries in the region with data.
  processing_both_sexes: |-
    <% if sex == "both" and cancer != "All cancers but non-melanoma skin cancer (C00-97, but C44)" %> For this cancer site the source only reports men and women separately, so the both-sexes value is the sum of the two.<% endif %>
```

and then `description_processing: "{definitions.processing_regions}{definitions.processing_both_sexes}"` on `cases` / `attr_cases`, with `attr_cases_share` keeping its computed-by-OWID lead sentence in front.

That still leaves the sex-specific sites saying "the sum of the two" where one addend is always absent. Cheap to sharpen — "the sum of the values reported for men and women" — or leave it; your call, since it's arithmetically the sum of what exists.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
...
-        title: |-
-          Number of new << cancer.lower() >> cases among {definitions.sex}
-        description_short: |-
-          Estimated number of new << cancer.lower() >> cases among {definitions.sex}.
-        unit: cases
+            note: "{definitions.footnote}"
+
+      asir_att:
+        title: New {definitions.cases_of} caused by {definitions.agent_phrase} per 100,000 {definitions.people} (age-standardized)
+        description_short: Estimated number of new {definitions.cases_of} attributed to {definitions.agent_phrase} per 100,000 {definitions.people}, age-standardized so that populations with different age structures can be compared.
```
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 Word order: the rate phrase lands after the agent, so this reads as a rate *of infections*.

Rendered for `sex=both`, `agent=All infectious agents`, `cancer=All cancers…`:

> New cancer cases (excluding non-melanoma skin cancer) **caused by infections per 100,000 people** (age-standardized)

"caused by infections per 100,000 people" parses as though infections were what's being counted per 100,000. It reads worse on the sites, where two prepositional phrases stack up:

> New cases of hepatocellular carcinoma **caused by hepatitis B virus per 100,000 people** (age-standardized)

All 66 `asir_att` variables have this shape. The old title avoided it by putting the rate first (`Age-standardized incidence rate of X attributable to Y among Z`), so this is a regression from the rewrite rather than a pre-existing wart.

Suggestion — keep the new plain-language vocabulary but move the rate back in front and drop the trailing parenthetical:

```yaml
title: Age-standardized rate of new {definitions.cases_of} caused by {definitions.agent_phrase}{definitions.among_sex}, per 100,000 {definitions.people}
```

> Age-standardized rate of new cancer cases (excluding non-melanoma skin cancer) caused by infections, per 100,000 people

`asir` on line 77 doesn't have the problem — nothing sits between the noun and the rate — so this is `asir_att` only.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
...
   owners:
-    - Tuna Acisu
+    - Lucas Rodés-Guirao
+
 tables:
   gco_infections:
     variables:
+      cases:
+        title: Number of new {definitions.cases_of}{definitions.among_sex}
+        description_short: Estimated number of new {definitions.cases_of}{definitions.among_sex}.
```
paarriagadap, line 50: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 `description_short` restates the title instead of adding to it. On `cases` it's exact for **59 of 59** variables:

> **title:** Number of new cancer cases (excluding non-melanoma skin cancer)
> **description_short:** Estimated number of new cancer cases (excluding non-melanoma skin cancer).

Across the dataset, 110 of 336 rendered variables are exact restatements, and the rest of `attr_cases` / `attr_cases_share` differ from their title only in where the sex phrase sits ("… among men that are attributed to X" vs "… caused by X among men"). So effectively all 204 count/share variables echo their own title. `asir` and `asir_att` are the two that don't — they add the age-standardization explanation, which is exactly the right use of the field.

This matters because `description_short` is the chart-subtitle fallback and the line under the data-page heading: it's the one sentence we get to explain the indicator, and here it's spent repeating the heading the reader just read.

Suggestion — say what the title can't, i.e. that these are modelled estimates and which vintage they come from:

> Estimated number of new cancer cases in 2020, excluding non-melanoma skin cancers. Estimates are modelled from GLOBOCAN 2020 incidence data.

Same treatment on `attr_cases` (line 52) and `attr_cases_share` (line 59). Not a blocker — but it's the difference between a subtitle and an echo, and this PR is the natural place to fix it.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
...
-        unit: '%'
-        short_unit: '%'
+        # The source's own attributable fraction. It only covers the all-cancers total by agent, for countries; the
+        # "attr_cases_share" indicator above covers cancer sites, regions and both sexes.
+        title: Share of new {definitions.cases_of} caused by {definitions.agent_phrase}{definitions.among_sex} (population attributable fraction)
+        description_short: Estimated share of new {definitions.cases_of}{definitions.among_sex} that are attributed to {definitions.agent_phrase}, as published by IARC.
+        unit: "%"
+        short_unit: "%"
+        presentation:
+          title_public: Share of new {definitions.cases_of} caused by {definitions.agent_phrase}{definitions.among_sex}
```
paarriagadap, line 82: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 `title_public` strips "(population attributable fraction)", which is the only thing in the title distinguishing this indicator from `attr_cases_share`. For the three combos where both exist — `agent = All infectious agents` × `cancer = All cancers…` × 3 sexes — both now render exactly:

> Share of new cancer cases (excluding non-melanoma skin cancer) caused by infections

and the values agree to 7.6e-06, so under the public name the two are indistinguishable in indicator search and the data catalog.

**Scope, so you can weigh it:** it's three variable pairs, and no published chart is affected. `paf` covers 30 variables (10 agents × 3 sexes, countries only) against `attr_cases_share`'s 59, and they overlap only on the all-agents row — because the source reports the denominator `cases` *only* there, so `attr_cases / cases` is NaN for the other nine agents. Chart 8061 sits on `paf` at `agent = Helicobacter pylori`, where no share exists to collide with. And `title` keeps the parenthetical, so the admin picker can still tell them apart.

**Suggested fix: don't publish `paf` where `agent == "All infectious agents"`.** That removes a genuinely duplicate indicator rather than renaming it. The obvious alternative — restore the parenthetical in `title_public` — would put "population attributable fraction" back in front of readers, which is the kind of jargon this PR is removing everywhere else, so it trades one small problem for another.

Worth logging as a follow-up rather than doing here: the reason `paf` has to exist for the other nine agents at all is that we don't broadcast that agent-independent denominator across agents in garden. If we did, `attr_cases_share` would cover all ten agents *and* gain region-level values (which `paf` can never have, since rates aren't aggregated), and `paf` could go away entirely — chart 8061 included.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.py
```diff
@@ -1,65 +1,155 @@
...
+    "HPV",
+    "Helicobacter pylori",
+    "Hepatitis B virus",
+    "Hepatitis C virus",
+    "Human T-cell lymphotropic virus",
+    "Human herpesvirus type 8",
+    "Opisthorchis viverrini and Clonorchis sinensis",
+    "Schistosoma haematobium",
+}
+EXPECTED_NUM_CANCERS = 21  # 20 infection-related cancer sites plus the all-cancers total.
```
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 This asserts the *count* while `EXPECTED_AGENTS` just above asserts the *set* — and for cancers the count is the weaker guard, because the reader-facing labels are derived from these exact strings in `gco_infections.meta.yml`:

```yaml
<< cancer.lower().replace("hodkin", "hodgkin").replace("hodgkin", "Hodgkin").replace("burkitt", "Burkitt").replace("kaposi", "Kaposi").replace("t-cell", "T-cell") >>
```

I checked all 20 site labels render correctly today, including the `hodkin → hodgkin → Hodgkin` two-step and the fact that `hodkin` is not a substring of `hodgkin`, so `Non-Hodgkin lymphoma of gastric location` survives untouched. So the chain is right — it's just undefended.

A *rename* upstream keeps the count at 21 and slips straight through: `.lower()` on an acronym-initial label would mangle it, and any new site whose capitalization isn't in the replace chain ships lowercased into titles. Since the count assertion passes, nothing would surface it.

Suggestion: mirror `EXPECTED_AGENTS`, so the failure names the label that moved — which is precisely what you need in order to update `cases_of`:

```python
EXPECTED_CANCERS = {
    ALL_CANCERS,
    "Adult T-cell leukemia and lymphoma",
    "Anus squamous cell carcinoma",
    "Bladder carcinoma",
    "Burkitt lymphoma",
    "Cardia gastric cancer",
    "Cervix uteri carcinoma",
    "Cholangiocarcinoma",
    "Hepatocellular carcinoma",
    "Hodgkin lymphoma",
    "Kaposi sarcoma",
    "Larynx cancer",
    "Nasopharynx carcinoma",
    "Non-Hodgkin lymphoma of gastric location",
    "Non-cardia gastric cancer",
    "Oral cavity cancer",
    "Oropharyngeal carcinoma",
    "Other Non-Hodkin lymphoma",  # NOTE: the source's typo; corrected in display text only.
    "Penis carcinoma",
    "Vagina carcinoma",
    "Vulva carcinoma",
}
```

then `assert set(tb["cancer"]) == EXPECTED_CANCERS`. That list is verbatim from the current meadow table, so it should drop in as-is.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
     presentation:
       topic_tags:
         - Cancer
```
paarriagadap, line 6: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 `topic_tags` still lists only `Cancer`, but the PR checklist marks this as done:

> - [x] Optional: a second topic tag (Global Health) next to Cancer in `definitions.common`

Either the edit was lost somewhere or the box was ticked ahead of it. Worth reconciling before merge — otherwise the description becomes a record of something that didn't ship, and the next person reading it will assume the tag is there.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
...
+    # (the tool's methodology statement, the agent list in the data, the footnote and subtitle on charts 8059-8061),
+    # but they have not been reviewed by the dataset owner. Uncomment after review to enable them.
+    # description_key:
+    #   - Estimates come from the International Agency for Research on Cancer (IARC). They apply the population attributable fractions from the study "Global burden of cancer attributable to infections in 2018" (de Martel et al., 2020) to the GLOBOCAN 2020 cancer incidence estimates.
+    #   - |-
+    #     Ten infectious agents classified as carcinogenic by IARC are covered: Helicobacter pylori, human papillomavirus (HPV), hepatitis B and C viruses, Epstein-Barr virus (EBV), human herpesvirus type 8, human T-cell lymphotropic virus, the liver flukes Opisthorchis viverrini and Clonorchis sinensis, and Schistosoma haematobium.
+    #   - Non-melanoma skin cancers are excluded because their registration is often incomplete and inconsistent across countries.
+    #   - Estimates carry uncertainty, especially for countries without population-based cancer registries or with limited data on how common these infections are.
```
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 `description_key` ends up empty on every indicator because it's commented out — confirmed on staging: `descriptionKey IS NULL` for **336 of 336** variables. Two reasons to close this out in the PR rather than defer it:

**The repo's no-commented-out-code rule covers metadata files too.** Nine lines of disabled drafts read as unfinished work to whoever opens this file next; the PR thread or owid-issues#2478 is the place to park them for the owner's review.

**More substantively, this dataset needs a WYSK more than most.** IARC assigns a **100% attributable fraction** to four of the sites — `Cervix uteri carcinoma`, `Kaposi sarcoma`, `Anus squamous cell carcinoma` and `Adult T-cell leukemia and lymphoma` all come out at exactly 100.0% in the built data. A reader who lands on "100% of new cervical cancer cases are caused by infections" needs to be told that's an IARC modelling assumption rather than an observation, and right now nothing on the page says so.

For what it's worth I checked the drafted bullets and they hold up — including the "ten infectious agents" count, which is right (nine categories, because *O. viverrini* and *C. sinensis* share one, and hepatitis B and C are listed separately). They'd cover the provenance, the agent list and the skin-cancer exclusion; a fifth bullet on the 100% fractions would close the remaining gap.

If the owner review can't land before merge, enabling the four as drafted still beats shipping none — they only paraphrase text we already publish on charts 8059–8061.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
```diff
@@ -4,69 +4,87 @@ definitions:
     presentation:
       topic_tags:
         - Cancer
-
+      attribution_short: IARC
     processing_level: minor
     display:
       numDecimalPlaces: 1
```
paarriagadap, line 11: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

🟡 This sits on `definitions.common`, so it applies to all 336 variables — including the 145 whose unit is `cases`. From staging:

| numDecimalPlaces | unit | variables |
|---|---|---|
| 1 | cases | 145 |
| 1 | % | 89 |
| 1 | cases per 100,000 women | 48 |
| 1 | cases per 100,000 men | 44 |
| 1 | cases per 100,000 people | 10 |

So the World total renders as `17,970,624.0 cases`, and every country-level count carries a spurious decimal.

Pre-existing — it was already on `common` before this PR, so not a regression — but the block is being edited anyway and the fix is small: move `display.numDecimalPlaces` off `common` and onto the share and rate indicators (`attr_cases_share`, `paf`, `asir`, `asir_att`), leaving `cases` and `attr_cases` at the default of 0.

paarriagadap (approved):
Hi Lucas!

It looks good at first glance, but I also decided to leave it to `/review-data-pr` so it can find more stuff, especially in the metadata, which is harder to grasp given all the specific names of diseases. I selected the most relevant comments.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Fixed in 1a699a1596. The definition is split as you suggested: the regions sentence stays unconditional, and the both-sexes sentence is emitted only when `sex == "both"` and the cancer is a site (not the all-cancers total), worded "the sum of the values reported for men and women" so the single-sex sites read correctly. Re-rendered all 204 count/share indicators: the sentence is present on exactly the 60 site-level both-sexes variables and absent everywhere else; data unchanged.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Fixed in 4849f6fdc1, following your shape: `Age-standardized rate of new {cases_of} caused by {agent}, per 100,000 {people}`, and `asir` gets the same shape for consistency. The `among men/women` phrase is dropped from these two rate titles because `per 100,000 men/women` already carries the sex. Rendered all 102 rate indicators: all start with the rate phrase, no leftovers. Examples: "Age-standardized rate of new cancer cases (excluding non-melanoma skin cancer) caused by infections, per 100,000 people"; "Age-standardized rate of new cases of hepatocellular carcinoma caused by hepatitis B virus, per 100,000 people".

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Done as you suggested, in 0cc13ecac1 + 5ea1df3e8c: the garden step blanks `paf` for `agent == "All infectious agents"`, so the three duplicate indicators (prod ids 987289, 987326, 987453; used by no chart, explorer or article) are no longer published, and the grapher upsert will remove them as ghosts. `paf` stays for the nine individual agents (chart 8061 on H. pylori unaffected). Two output assertions guard it: blank for all agents, present for every individual agent in every country (regions excluded, since the source has no fraction there; the first version of that assertion forgot this and broke the build for one commit). `etl diff` now reports the intended data change on `paf` only. Your follow-up (broadcast the denominator so `attr_cases_share` covers all agents and regions, then retire `paf` and move chart 8061) is logged for a separate PR.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.py
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Done in cbe7124113: `EXPECTED_CANCERS` is now the 21-label set (your list, verbatim, with the NOTE on the source's "Non-Hodkin" typo) and the check asserts set equality, so the failure message names the label that moved. Tested by renaming one label in the meadow table: the check fails with `{'Other non-Hodgkin lymphoma', 'Other Non-Hodkin lymphoma'}`.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Done in 04257e39ac, with the inverse arrangement: `numDecimalPlaces: 1` stays on `definitions.common` because shares and rates are the majority (188 of 333 indicators), and the two count indicators (`cases`, `attr_cases`, 145 indicators) override it to 0. Verified on the rebuilt grapher dataset: every `cases` indicator renders with 0 decimals, every `%` and per-100,000 indicator with 1.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

The checklist item is unticked on the PR ("- [ ] Optional: a second topic tag (Global Health) …"), so the description and the file agree: the tag has not been added, and whether to add it is still an open, optional decision for the owner.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Agreed that the short descriptions mostly restate the titles, but this is the pattern the dataset already had in production ("Estimated number of new X cases among Y." under the same title), so it is not a regression of this PR. The owner's call is to leave them as they are for now and keep this PR to its maintenance scope; a provenance sentence (IARC fractions applied to GLOBOCAN 2020 incidence) can be added in a later pass together with the WYSK bullets.

etl/steps/data/garden/cancer/2024-09-06/gco_infections.meta.yml
lucasrodes: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Done in 6d92584efe after the owner reviewed the drafts: the four bullets are enabled (no commented-out block left) and a fifth one is added: "For four cancer types (cervical cancer, Kaposi sarcoma, anal squamous cell carcinoma, and adult T-cell leukemia and lymphoma), IARC attributes every case to an infection, so their attributable share is 100% by construction rather than an observed value." Verified on the rebuilt dataset that all 333 indicators carry the five bullets, and that those four sites are indeed at exactly 100% in every country and both sexes.

PR #6804, opened 2026-09-02
🐛 Fix provider analytics: prod DB guard, collection post recovery, top-chart image

pabloarosado:

Stacked on [#6395 "🐝 Update data provider analytics script for 2026"](https://github.com/owid/etl/pull/6395). Fixes found while generating the 2026-H1 data provider reports:

- **Require the production grapher DB** (assert in `Report.__init__`). The analytics semantic layer is built from production, so grapher-DB reads (`multi_dim_redirects`, `origins_variables`, chart configs) against a staging snapshot silently drop any redirect or re-versioned dataset newer than the snapshot. This was the cause of the missing mdim views (Ember ~1M views, Global Carbon Project ~36.5k, Luxembourg Income Study ~10.3k in H1); reports must be generated with `ENV_FILE=.env.prod`.
- **Recover posts that embed a producer's mdims or explorers by live slug** (`get_post_views_of_producer_collections`, sharing a new `_get_post_views_citing_slugs` helper with the redirected-slug recovery). The chart-id lookup cannot see collections (no chart id) and the redirected-slug recovery only covers dead slugs, so articles re-pointed to mdim slugs after chart→mdim migrations were dropped (e.g. all of LIS's publications).
- **Guard against NaN daily post views** when a producer has no qualifying posts (previously crashed report generation).
- **Top-chart image now matches the #1 entry of the chart list**: mdim view URLs support `.png` export with query params, so the image is picked from charts and mdim views combined instead of standalone charts only.
- **Skip redirected sources that fan out to multiple target views** (with a warning). A sunset explorer mapped view-by-view into mdims shares one source path across many redirect rows; the previous deduplication credited its whole multi-provider traffic to one arbitrary view (Ember's electricity generation view had absorbed all ~52k H1 views of the retired Energy Data Explorer). Such views are now disclosed separately in the affected report, matching how mixed explorers were reported while live.
- **The "features your data in N interactive charts" count is now the current catalog**: all live standalone charts + all mdim views using the producer's data (each has its own URL and analytics) + live explorers featuring it (one each, mixed ones included; their views remain disclosed separately). Previously N only counted content with traffic in the period, a flow-derived number behind a catalog claim. View totals are unchanged (zero-traffic rows add nothing) and the top-charts list keeps only content with period traffic.

Validated by regenerating all ten H1-2026 reports against production, reconciling them against the 2025 reports (chart-level and post-level), and diffing before/after each code change to confirm only the intended numbers moved.


pabloarosado:
@codex review

antea04 (approved):
LGTM! I can't test it (no prod credentials), but the code changes make sense and the resulting reports looked good as well. Merging into the main feature branch now and will request your review to merge into master :)

PR #6820, opened 2026-09-04
🐛🤖 chart-critic: stop re-reporting what has been fixed, explained or already said

Marigold:

The digest re-posted a finding to #we-need-to-correct-it the morning after the chart had been fixed, and a separate one that had been investigated and found to be correct. Three independent causes, one per commit-worthy change; no behaviour change to what the critic looks for.

## 1. A `--changed-since` sweep reviewed charts at their pre-edit state

`--changed-since 1` selects charts *because* their configuration changed in the last 24 hours. The bundle cache keeps a chart for 24 hours. With a daily cron the previous run's bundle is always a few minutes inside the TTL, so the selection criterion and the cache guaranteed each other wrong: every finding was a report on yesterday's chart.

That is what reached the channel — the subtitle of `oil-prices-inflation-adjusted` said "constant 2023 US$" against an indicator on 2025; it was corrected during the day, and the correction is what nominated the chart for the next morning's sweep, which then read the pre-correction bundle and raised it again.

A sweep of just-changed charts now re-fetches (`--cache-ttl` still overrides). It is ~17 charts a day, so nothing worth caching is being given up.

## 2. Dedup was keyed on the model's wording

`digest_state.json` is what stops day two repeating day one, but a finding's key carried the claim's own significant words, and the model rewords every pass:

```
Sep 3  The subtitle states prices are in constant 2023 US$, while the indicator
       metadata specifies constant 2025 US$.
Sep 4  The chart subtitle states prices are measured in constant 2023 US$,
       contradicting the indicator metadata base year of 2025.
```

Two keys, so the state file recognised nothing. Meanwhile `_merge` — which folds the repeat passes of a *single* run together — already had a wording-robust test for "the same finding". Those two notions now share one definition (`critic.same_claim`): the key identifies the *subject* (indicator for a data finding, chart for a chart-level one) and the claim is compared separately.

The state file's format changes accordingly, and old entries are read into the new shape rather than dropped, so nothing already posted comes back on deploy day.

## 3. The model was never shown where we document anomalies

The Central African Republic's life expectancy swinging between 14.7 and 57.4 years was filed as an error (#6779). It isn't one — it is the UN's crisis-mortality adjustment, and #6813 documents it in the indicator's `description_key`, for readers, who hit the same spike.

So the bundle now carries `descriptionKey` and `descriptionProcessing`, and the prompt says that a discontinuity our metadata accounts for is documented rather than wrong. This makes the fix for a false positive be *documenting the data* — which we want anyway — rather than a suppression list maintained for the critic alone, which would leave the reader with the unexplained spike.

Measured on `life-expectancy-vs-gdp-per-capita`, 5 passes each on the same bundle:

| | CAR flagged |
|---|---|
| as published today | 5/5 |
| with #6813's bullet spliced in | 0/5 |

An earlier draft of the prompt named CAR as an example. It worked — and it was a suppression list in prose, silencing that one chart while teaching nothing about the next one, so it is stated as a rule about the metadata instead. Cost: ~520 extra input tokens a chart, about $0.0004.

## Checks

- `etl chart-critic --eval` → **9/9** ($0.24), against 8/8 of nine before (the ninth is the flaky COVID case, which landed this time — one run is not evidence it stopped being flaky).
- New unit tests for the dedup: a reworded repeat, two different findings on one chart, the same finding on a chart sharing an indicator, and the old state format.
- Editing `bundle.py`/`critic.py` invalidates the bundle and review caches by design, so the next run re-fetches and re-reviews once.

## Also here, unrelated to the critic

`CLAUDE.md` no longer tells an agent to post `@codex review` on every PR. Asking for a review pass is the author's call, and the `pr-babysitter` skill still covers it when the loop is wanted.

## Still open

- **The fixture for #6779 will go FAIL** once #6813 is deployed, since the critic will correctly stop reporting CAR. The case says so and says what to do (flip it to a guard case). Doing it here would put the gate red until #6813 lands, so it belongs with that PR.
- **Why the review cache missed** while the bundle cache hit — which is what let a reworded claim exist at all on Sep 4 — is inferred, not confirmed: most likely a manual sweep during Sep 3 refreshed the bundle and cached its passes without stamping the digest state. The runner's `.cache/chart_critic` was not inspected. Either cause is fixed by the two changes above.
- Nobody has checked whether other daily owidbot jobs share the same cron-versus-TTL shape.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -188,8 +188,6 @@ gh pr edit <number> --body "..."
 
 **Cleaning up after merge**: `etl pr-clean` lists local branches whose PR was merged or closed (it checks the GitHub PR state, so squash-merges are detected), then deletes the selected branch(es). For branches created in a worktree (`etl pr "..." --worktree`), it also removes the worktree and copies that worktree's Claude sessions back into the main repo's `~/.claude/projects/` dir so they stay resumable.
 
-**Post `@codex review` as a separate PR comment** (not in the PR description) when the PR is ready for a review pass. Do not repost it after every push/update unless the user asks or the changes are substantial enough to warrant a fresh review.
-
 To run the full **review → wait → fix → re-review** loop hands-off (and watch CI) in the background while you keep working, use the `pr-babysitter` skill — it spawns a background agent that triggers Codex, judges and fixes the valid findings, and loops to a cap (never merges). Fire it proactively after pushing a substantial chunk to a PR branch.
 
 ### Commit Message Emojis
--- a/apps/chart_critic/bundle.py
+++ b/apps/chart_critic/bundle.py
@@ -33,7 +33,30 @@ GRAPHER_URL = "https://ourworldindata.org/grapher"
 
 # Fields worth showing the model. The long forms (titleLong, citationLong, fullMetadata) add
 # tokens without adding judgement.
-COLUMN_FIELDS = ("titleShort", "descriptionShort", "unit", "shortUnit", "timespan", "type", "conversionFactor")
+#
+# ``descriptionKey`` and ``descriptionProcessing`` are here for one specific reason: they are where
+# OWID documents the data that *looks* wrong and is not. The critic filed the Central African
+# Republic's life expectancy oscillating between 14.7 and 57.4 years as an error (owid/etl#6779);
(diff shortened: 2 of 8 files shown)


Marigold:
@codex review

PR #6825, opened 2026-09-04
🔨🤖 Validate indicator grapher_config against the patch schema

sophiamersmann:

An indicator's `presentation.grapher_config` is a patch applied on top of a chart's own config, never a complete chart, but `_validate_grapher_config` validated it against the full grapher schema and worked around the `dimensions` requirement by emptying `required` in place on a `@cache`d dict. Upstream publishes a `.patch.json` variant of every schema that requires only `$schema`, so validate against that instead. The URL is derived from each config's own `$schema`, so validation follows whatever version a config pins; `DEFAULT_GRAPHER_SCHEMA` stays the non-patch URL, since that is the value written into configs and what grapher's config migrations key on.

This is behaviour-preserving. The patch schema requires `$schema`, which the `setdefault` a line above already guarantees, and every other constraint is byte-identical. I validated all 1,424 `grapher_config` blocks in the repo both ways and none of them disagree.

- **Fix.** Emptying `required` in place poisoned the shared cached dict for every other caller of that URL, including `etl/indicator_upgrade/schema.py`, which reads the real `required` in three places. Not reachable today, since those run in separate processes.
- **Fix.** `get_schema_from_url` had no `raise_for_status`, so a missing schema surfaced as `JSONDecodeError` at line 1 column 1 rather than a 404 naming the URL. That matters more now there is no fallback path.
- **Fix.** The sync workflow's `ls schemas/grapher-schema.*.json` matches the new patch file too, and a two-line `$VENDORED` breaks `basename` and `curl -o`. Schema-change detection would have stopped working silently the moment the vendored patch file landed, so the glob now excludes it. Detection stays keyed off the full file, which is fine because the two move together upstream and `--refresh` pulls both.
- The multidim and explorer `$ref`s stay on the full schema. They reference individual properties rather than the root, so top-level `required` never applies to them, and one of them is `#/properties/$schema`, whose `const` is correctly the non-patch URL.
- The patch schema is vendored so a unit test can assert it differs from the full one in `required` and `$id` only. Nothing reads the vendored copy at runtime.

Left open:

- `compute_inheritance_patch` borrows the full schema's `required` into the indicator baseline so a bare `{}` patch validates. It can now validate against the patch schema instead and shrink that borrow to `$schema`.

### DIFF
--- a/.claude/skills/sync-grapher-schema/SKILL.md
+++ b/.claude/skills/sync-grapher-schema/SKILL.md
@@ -15,16 +15,17 @@ metadata:
 
 # Sync Grapher Schema
 
-The grapher chart-config schema is owned by the web team in [`owid-grapher`](https://github.com/owid/owid-grapher/tree/master/packages/%40ourworldindata/grapher/src/schema) and published at `https://files.ourworldindata.org/schemas/grapher-schema.NNN.json`. **It is mutated in place without version bumps** (e.g. dumbbell plots landed in `.010` directly), so when it changes upstream, four things in this repo need to follow:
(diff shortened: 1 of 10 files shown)


sophiamersmann:
@codex review

Marigold (approved):
Looks good I guess. Keeping schema in sync with grapher is getting more and more complex.

sophiamersmann:
As discussed

PR #6827, opened 2026-09-04
🔨🤖 Stop vendoring the grapher schema

Marigold:

ETL kept a vendored copy of the grapher chart-config schema and three derivatives of it: a hand-mirrored `grapher_config` block inside `dataset-schema.json`, `$ref`s into it from the multidim and explorer schemas, and generated TypedDicts. Keeping all four level with upstream needed a cron, four drift tests, a docs guide and a skill, and every in-place upstream change needed a human to propagate it by hand. This deletes all of it and leaves chart-config validation to the grapher admin API.

_Deep review on the premise, skim on the diff._ Nearly all 3,957 deleted lines are the vendored schema and generated code. The thing to weigh is the premise, and the merge ordering below.

## The premise this rests on

Grapher does not validate config content today — its write paths call only `migrateGrapherConfigToLatestVersionAndFailOnError`, which checks the `$schema` version string. `migrate.ts:21` says so in a comment: *"the given config is not actually validated against the schema!"* There is no ajv in the repo.

Its own schema README already specifies the missing half:

> A write must migrate the config to the latest version and reject it if it fails validation. A read only migrates, never validates. Reads can skip validation because every stored config passed it on the way in.

So the read path's justification for skipping validation rests on a premise that isn't true yet. **This PR should not merge until that write-side validation exists in grapher** (`validateViewConfigSchemas` in `multiDim.ts` already walks every view config; `charts.ts` and `variables.ts` are the other two write paths). Left as a draft for that reason.

Worth noting the ETL-side validation being deleted here was partly circular anyway: ETL stamps `$schema` from its own `DEFAULT_GRAPHER_SCHEMA` before sending (`apps/chart_sync/admin_api.py:273`), so grapher's one check is always satisfied by ETL's own constant.

## How it works

Three schema blocks become open objects — `presentation.grapher_config` in `dataset-schema.json`, and `views[].config` in the multidim and explorer schemas — each with a description pointing at the grapher schema and saying where validation now happens. `$defs/indicatorConfig.display` in the multidim schema went the same way; it was the one grapher `$ref` outside a view config.

Everything ETL owns is untouched: dimensions, choices, `common_views`, dependencies, `download_package`, view metadata. Those never reach grapher as a config object, so nothing upstream could validate them. Confirmed still enforced:

```
data.default_selection must be array
data.title must be object
```

Deliberate calls:

- **The `grapher_schema` pins stay**, along with `DEFAULT_GRAPHER_SCHEMA` and `test_no_newer_grapher_schema_version`. Grapher keys its config migrations on the version, so ETL still has to stamp a truthful one — that's a string, not a schema.
- **The generated TypedDicts go, replaced by `dict[str, Any]`.** They carried no runtime meaning and near-zero static meaning: every annotation was `ViewConfig | dict[str, Any]`, so any dict already satisfied it, and every assignment went through a `cast`. `ViewMetadata` derives from `dataset-schema.json` rather than the grapher schema, so it could have been kept by keeping the 560-line generator for one type — not worth it, but that's the call to push back on if you disagree.
- **`_validate_grapher_config` becomes `_stamp_grapher_config_schema`.** It did two jobs; only the validation goes. Dropping the `setdefault` too would have shipped indicator configs with no `$schema`, which grapher rejects outright.
- **Explorers get nothing in exchange.** ETL sends TSV and grapher's `PUT /explorers/{slug}` checks only the slug, so `explorer-schema.json`'s refs were the only structural check those view configs ever got. Accepted on the grounds that explorers are being deprecated; if that timeline slips this is the part to revisit.

## Verified

- `make unittest` green (868 passed). The two failures in `tests/collections/test_explorer_migration.py` under a bare `pytest` are integration-marked and need built datasets on disk — they fail the same way on master.
- Behavioural check that the change does what it claims: a view config carrying `{"titel": "typo", "madeUpField": 42}` now passes `Collection.validate_schema()`, while malformed ETL-owned structure is still rejected (the two errors quoted above).
- No consumer of the deleted file remains — every surviving mention is a version string, not a file read:
  ```
  rg -n "grapher-schema\.[0-9]{3}\.json" --glob '!etl/steps/**' --glob '!snapshots/**' . | grep -v files.ourworldindata.org
  ```
- **Not verified:** no staging run yet, since the grapher-side validator doesn't exist to run against. Once it does, the check that matters is that a deliberately broken config fails the upsert.

<details>
<summary>Consumers removed, and what replaced each</summary>

| Consumer of the vendored schema | Replacement |
|---|---|
| `dataset-schema.json` embedded `grapher_config` (65 properties, mirrored by hand) | open object; admin API validates |
| `multidim-schema.json` — 58 `$ref`s + `indicatorConfig.display` | open object |
| `explorer-schema.json` — 55 `$ref`s | open object |
| `scripts/generate_schema_types.py` → `schema_types.py` | `dict[str, Any]` |
| `chart_upsert._validate_chart_config` | admin API |
| `grapher/helpers._validate_grapher_config` | admin API; the `$schema` stamp is kept |
| `Collection.validate_schema`'s `https_handler` | removed — no https `$ref`s remain in any ETL schema |
| `test_conflict_resolver` fixture reading the vendored file | a small inline schema with defaults; the test asserts defaults are not written into the config, which a handful of `default`s exercises as well as the real ~38 |

Also removed on the grapher side, once this lands: `sync-grapher-schema-to-r2.yml`'s `repository_dispatch` into owid/etl, the `ETL_SCHEMA_SYNC_TOKEN` secret, and the sibling-PR merge-ordering rules in the schema README — all of which existed only to feed the vendoring.

</details>


Marigold:

Parked until the grapher side lands. This PR deletes ETL's own config validation, so merging it before grapher validates would leave nothing checking chart configs anywhere.

Waiting on owid/owid-grapher#7237 — *validate grapher configs at every write route* — which adds `ingestGrapherConfig` (ajv) to every admin route that accepts a config. It's stacked on owid/owid-grapher#7236, so that one goes first.

Two things from that stack that change this PR's surroundings:

- **#7236 removes the patch schema.** `dimensions` is no longer required, so a single document validates a config layer and a rendered chart alike. That supersedes #6825, which exists only to validate indicator configs against `.patch.json`.
- **#7246 adds immutable per-revision schema names** (`grapher-schema.<version>.<revision>.json`). Doesn't affect the `DEFAULT_GRAPHER_SCHEMA` pin this PR keeps, but it's worth reading before the next version bump.

Rebased onto master in the meantime. Two things the rebase surfaced, beyond the `etl/collection` → `etl/viz/chart` move:

- Master made `DEFAULT_GRAPHER_SCHEMA` *derived* from the vendored file's `$id` (`vendored_grapher_schema_id()`). That auto-merged without a conflict and would have made `import etl.config` raise once the file was deleted — it's a plain constant again here.
- The two tests pinning that derivation are dropped, along with the new `--bump-version` references in the schema guide.

`make unittest` green after the rebase, and the behavioural check still holds: a view config carrying `{"titel": "typo", "madeUpField": 42}` now validates, while malformed ETL-owned structure is still rejected.

PR #6830, opened 2026-09-04
📊 frontiermath: move to Epoch's corrected v2 problem set

lucasrodes:
Tracks: owid/owid-issues#2519

| | |
|---|---|
| Chart | [ai-frontiermath-over-time](https://ourworldindata.org/grapher/ai-frontiermath-over-time) · [staging](http://staging-site-data-frontiermath-epoch-v2/grapher/ai-frontiermath-over-time) |
| Source | `frontiermath.csv` → `frontiermath_tiers_1_3_v2.csv` (same `benchmark_data.zip`) |
| Producer note | ["On 2026-06-12, we released a major update, addressing errors in 42% of problems"](https://epoch.ai/benchmarks/frontiermath-tiers-1-3-v2) |

## Why

The file we read is marked superseded in the archive's own `benchmark_metadata.csv` (2 of its 68 benchmarks are, and this is one), Epoch's v1 page is titled "FrontierMath Tiers 1-3 (v1, old)", `epoch.ai/benchmarks/frontiermath` no longer resolves to it (it now redirects to the Tier 4 v2 page), and the file has gained no model since 2026-05-28. The daily autoupdate kept refreshing the snapshot because other files in the zip move, so nothing surfaced it.

## This is a new series, not a refresh

Of 101 model versions in v1, only **32** appear in v2 (69 lost, 72 gained), and **all 32 score higher** on the corrected set:

| | median | mean | max |
|---|---|---|---|
| change | +11.9pp | +13.8pp | +32.8pp |

`benchmark_metadata.csv` gives the old set a `score_ceiling` of **0.57** against v2's **1.00**, which matches the 42% error rate — the flawed problems capped what any model could score. Splicing the two would be wrong, so the metadata now says explicitly that scores are not comparable across versions.

## New step version, not an in-place swap

Since this body was first written the change became a dated version: `2026-09-07` for meadow, garden and grapher, with the `2026-01-30` steps archived via `etl archive-dag`. Reusing the old variable id would have served rebased numbers under the same identity to the chart, the data API and the CSV download. The snapshot deliberately stays at `2026-01-30`, which is what keeps the dataset inside the scheduled Epoch autoupdate PR — `etl autoupdate` walks the live DAG for `snapshot://` nodes and reads each `.dvc` for its `autoupdate` block.

## Chart config (done on staging)

`selectedEntityNames` pins 12 entities; **6 no longer exist** in v2 and will silently drop out:

- `Gemini 3 Pro` · `Claude 3.5 Sonnet (Jun 2024)` · `o1-mini (Sep 2024), medium` · `Claude 3.5 Sonnet (October 2024)` · `Gemini 2.5 Deep Think (Aug 2025)` · `GPT-5.5 Pro, high`

The 6 that survive: `o1 (Dec 2024), high` · `o3-mini (Jan 2025), high` · `o4-mini (Apr 2025), high` · `GPT-5 (Aug 2025), high` · `GPT-5.2 (Dec 2025), xhigh` · `Claude Opus 4.8, max`.

The chart's `note` also still reads "300 problems" and needs to become 295.

v2 has good vendor coverage for replacements — best score per family: Claude Fable 5.1 (90.2%), Qwen 3.8 Max (74.7%), Gemini 3.7 Flash (71.6%), Grok 4.6 (66.0%), DeepSeek V4 Pro (64.6%) — and the frontier now tops out at **GPT-6 Astra, max: 93.7%** (2026-09-03), which the current chart cannot show at all.

## Guard added

The meadow step now reads `benchmark_metadata.csv` and asserts our source file is not marked `superseded_by`, so the next reissue fails the build rather than freezing a chart for three months.

## Verified before pushing

All eight of the garden step's sanity checks pass against the v2 file: keys non-null, registry de-duplicates cleanly, scores within 0–1 (0.0–93.7%), `(release_date, model_version)` unique, no entity-name collisions, no fully-empty columns, and every model version covered by Epoch's registry (0 fallbacks to raw version strings). Schema is identical — 13 columns, same names.

## Not done

- Nobody has looked at the rendered chart yet; the y-axis will need rechecking since the attainable ceiling moved from ~57% to 100%.
- The Tier 4 v2 and Erdős cuts are still unused, as are Epoch's other 15 own-run benchmarks and the ECI score that sits in a file we already download.


etl/steps/data/meadow/artificial_intelligence/2026-01-30/frontiermath.py
```diff
@@ -16,12 +20,21 @@ def run() -> None:
...
-        tb = archive.read("frontiermath.csv", safe_types=False)
+        tb = archive.read(BENCHMARK_FILE, safe_types=False)
         # Epoch's model registry ships in the same archive and maps every model version string used across
         # its benchmark files to a curated human-readable name.
         tb_registry = archive.read("epoch_capabilities_index.csv", safe_types=False)
+        # The archive also indexes its own benchmarks, marking retired ones in "superseded_by". Reading a
+        # superseded file is how this chart silently stopped gaining models for three months, so refuse to.
+        tb_benchmarks = archive.read("benchmark_metadata.csv", safe_types=False)
+
+    superseded = tb_benchmarks.loc[tb_benchmarks["source_file"] == BENCHMARK_FILE, "superseded_by"].dropna()
```
lucasrodes, line 33: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

This passes vacuously if `benchmark_metadata.csv` stops listing `BENCHMARK_FILE` at all (file renamed, row dropped, or `source_file` values change shape): `.loc[...]` returns an empty frame, `.dropna()` keeps it empty, and the assert is satisfied. Pin the row count too:

```python
matches = tb_benchmarks.loc[tb_benchmarks["source_file"] == BENCHMARK_FILE]
assert len(matches) == 1, f"Expected exactly one benchmark_metadata row for {BENCHMARK_FILE}, got {len(matches)}."
superseded = matches["superseded_by"].dropna()
```

Today it matches exactly one row (`FrontierMath-Tiers-1-3-v2-Private`, `score_ceiling` 1.0).

etl/steps/data/garden/artificial_intelligence/2026-01-30/frontiermath.meta.yml
```diff
@@ -22,18 +22,19 @@ tables:
...
-          - The benchmark has four difficulty tiers. This indicator shows accuracy on Tiers 1–3 (300 problems). Tier 4 contains 50 exceptionally difficult problems and is not included here.
+          - FrontierMath is a set of 338 original math problems written by experts, covering many areas of advanced mathematics. Many problems are difficult enough that human specialists might need hours or days to solve them.
+          - The benchmark has four difficulty tiers. This indicator shows accuracy on Tiers 1–3 (295 problems). Tier 4 contains 43 exceptionally difficult problems and is not included here.
+          - In June 2026, Epoch AI reissued FrontierMath after correcting errors in 42% of its problems. This indicator uses the corrected problem set. Scores on it are not comparable with scores on the original set, where the flawed problems held every model well below 100%; models evaluated on both scored around 12 percentage points higher on the corrected set.
```
lucasrodes, line 30: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Epoch's changelog on the v2 page: "We released v2, which corrected 123 problems in Tiers 1-3 and 12 problems in Tier 4. We also removed 5 problems from Tiers 1-3 and 7 problems from Tier 4." That is 147/350 = 42%, which is why Epoch writes "addressing errors", not "correcting". Twelve problems were dropped rather than fixed, so "correcting errors in 42% of its problems" overstates a little. Suggestion:

`In June 2026, Epoch AI reissued FrontierMath after finding errors in 42% of its problems — most were corrected and a few removed. This indicator uses the corrected problem set. …`

(Em dash with spaces for the aside, per the style guide.)

etl/steps/data/garden/artificial_intelligence/2026-01-30/frontiermath.meta.yml
```diff
@@ -22,18 +22,19 @@ tables:
         title: Share of FrontierMath problems solved correctly by AI models
         unit: "%"
         short_unit: "%"
-        description_short: FrontierMath benchmark evaluates models on 300 difficult, research-level problems in advanced mathematics (Tiers 1–3), which can take expert mathematicians hours or days to work through.
+        description_short: FrontierMath benchmark evaluates models on 295 difficult, research-level problems in advanced mathematics (Tiers 1–3), which can take expert mathematicians hours or days to work through.
```
lucasrodes, line 25: > _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

Pre-existing, but the line is already changing: "FrontierMath benchmark evaluates models…" drops the article where the grammar needs it — the subject is the noun phrase "The FrontierMath benchmark", which is the case where the "no leading The" rule says to keep it. Either `The FrontierMath benchmark evaluates models on 295 …` or `FrontierMath evaluates models on 295 …`.

lucasrodes (commented):
> _Written by Claude Fable 5.1 — @lucasrodes at the wheel._

**`/review-data-pr` — a same-version repoint, not a version bump.** Two files change; the snapshot, DAG, variable id and chart are untouched, so the snapshot-field, DAG and archive checks don't apply. Pipeline correctness below rests on the staging build (Buildkite green; dataset 7342 rebuilt 17:04 UTC, variable 1144186 upserted 17:01; data-diff report produced).

**Verified**
- PR-body numbers reproduce from the committed zip: 32 shared model versions (69 lost, 72 gained), all 32 higher on v2, median +11.9 pp / mean +13.8 / max +32.8; `score_ceiling` 0.57 → 1.00; top score GPT-6 Astra, max at 93.7% on 2026-09-03.
- `benchmark_metadata.csv` marks exactly `frontiermath.csv` and `frontiermath_tier_4.csv` as superseded; the new guard passes on v2.
- Epoch's v2 page matches `description_from_producer` verbatim. Its changelog gives 123 corrected + 5 removed in Tiers 1–3 (→ 295) and 12 corrected + 7 removed in Tier 4 (→ 43), i.e. 147/350 = 42%. Every URL in the `.dvc` and meta returns 200; `epoch.ai/about` states CC BY.
- Every v2 model version resolves in Epoch's registry (0 fallbacks), `(release_date, model_version)` unique, and `mean_score` equals Epoch's declared score column `Best score (across scorers)` in all 104 rows.
- The staging chart SVG draws all 12 currently pinned entities and its x-axis ends in 2026 — grapher takes `zeroDay` from the variable (now 2024-01-25), not from the stale copy in the chart's dimension display.
- No published article or data insight mentions FrontierMath (gdocs sweep empty); `find-chart-references` finds only chart 9020 and its key-chart slot on the AI topic page. No outdated patterns, codespell clean, no explorer/MDim/narrative-chart references, no 📊 commits on master since the branch.

**🔴 Before merge**
- Chart 9020's `note` still reads "(300 problems)" — on staging too (the entity selection was edited there at 17:31, the note wasn't). Once merged, production shows the 295-problem corrected set under a note naming the old set, contradicting the data page. Fix it on staging so Chart Diff carries it with the merge, e.g. *"This chart shows performance on Tiers 1–3 (295 problems) of the FrontierMath problem set that Epoch AI corrected in June 2026. Scores are not comparable with the original set."* Then eyeball the render once more; `yAxis.max: 100` is right now that the ceiling is 1.00.

**🟡 Suggestions**
- `presentation.attribution_short` was never set on this dataset (NULL on prod and staging). Add `attribution_short: Epoch AI` under `definitions.common.presentation` while here.
- Inline below: the guard passes vacuously when the metadata file has no row for `BENCHMARK_FILE`; "correcting errors in 42%" vs Epoch's "addressing"; the article in `description_short`.
- Same version and same variable id for what the PR itself calls a new series: every consumer of id 1144186 (chart, data API, CSV download) gets different numbers with no visible break. The alternative is a new step version plus an explicit indicator upgrade. With one chart and a daily autoupdate the in-place move is defensible — record the choice in the PR body so it's on the record.
- PR body: open with `Tracks: owid/owid-issues#2519` (the current Epoch update ticket). The title carries a doubled `📊 📊`. `epoch.ai/benchmarks/frontiermath` actually redirects to the **Tier 4** v2 page, not Tiers 1–3.
- `@codex review` hasn't been requested.

**🟢 Informational**
- `minTime: 1612` (= 2024-06-20, the first v1 point) now hides two v2 evaluations: GPT-3.5 Turbo (0%, Jan 2024) and GPT-4 Turbo (0.7%, Apr 2024). Probably fine as framing; drop the pin if the axis should start at the data.
- `.dvc` still says `date_published: 2026-04-17` and "Accessed 17 Apr 2026" in `citation_full` against `date_accessed: 2026-09-04` — autoupdate only bumps `date_accessed`. Not this PR's file; systemic.
- owid-issues `update-epoch-data.yml` (bi-monthly cron) covers this dataset. Its body could mention the new superseded guard, so the next person reads a red build as "repoint, don't refresh".
- "around 12 percentage points" in `description_key` is a hard-coded prose figure. It should stay stable because v1 is frozen, but it drifts if Epoch re-runs a shared model on v2.

**Optional add-ons, not run** — `check-empty-entities` full sweep: moot, the only surfaces are the chart and its key-chart slot and all 12 pinned entities render. `adversarial-data-review` (~25–45 web calls): low value here, Epoch is the sole producer of FrontierMath scores so there is no independent source; say so if you want it anyway.

**Still open**
- Chart note fix and render recheck on staging (🔴) — before merge.
- Versioning decision written into the PR body.
- Tier 4 v2, Erdős, Epoch's other 15 own-run benchmarks and the ECI score: unused, deliberately deferred (from the PR body).
- Nobody has compared the rendered chart against Epoch's own v2 leaderboard by eye.

PR #6840, opened 2026-09-08
🎉 vdem: ever had a woman head of state or head of government

bastianherre:
> _Written by Claude Opus 5 — @bastianherre at the wheel._

Draft, for exploring the data on staging before review.

## What this adds

We had one "ever had a woman leader" indicator, for the **chief executive** — whichever of the two offices actually holds power. This splits the question into the two offices, so presidents and prime ministers can be asked about separately, and adds a democratically elected variant for each:

- `wom_hos_ever` / `wom_hos_ever_dem` — ever had a woman **head of state** (president, monarch)?
- `wom_hog_ever` / `wom_hog_ever_dem` — ever had a woman **head of government** (prime minister)?

Each also gets a country count and a share of countries, giving 12 new indicators in total. Out of the 179 countries V-Dem lists in 2025:

| | yes | no | share |
|---|---|---|---|
| woman head of state | 70 | 109 | 39.1% |
| … democratically elected | 43 | 136 | 24.0% |
| woman head of government | 76 | 103 | 42.5% |
| … democratically elected | 58 | 121 | 32.4% |
| woman chief executive | 75 | 104 | 41.9% |
| … democratically elected | 55 | 124 | 30.7% |

## Three decisions worth a look

**"Democratically elected" means both a democracy and an elected office.** A woman counts only if she held the office in a year when the office was filled through multi-party elections *and* the country was an electoral or liberal democracy. Both conditions are needed. Asking only whether the country was a democracy would count hereditary monarchs, since many democracies have an unelected head of state — Elizabeth II and Margrethe II would appear as democratically elected women heads of state. Asking only whether the office was filled through multi-party elections would count Isabella II of Spain, who reigned by inheritance alongside an elected legislature.

This definition is now applied to the chief executive too, where the code had only ever tested whether the country was a democracy, although the metadata described the election condition as well. No country's answer changes there; five move from unknown to "no", because the added condition also settles the negative case.

**For the head of government, the gender is carried over from the head of state where one person holds both offices.** V-Dem only fills its head-of-government columns when the two are different people, so without this the gender would be missing for 53% of observations and 115 of 179 countries would read as unknown. Where V-Dem fills both columns the two never disagree, so the carry-over only ever adds information. Thailand 1897 is the clearest case: the queen regent held both offices while the king travelled, and appears in the head-of-state column alone.

**Unclear cases are read conservatively.** An unrecorded gender counts as a man, an unrecorded political regime as non-democratic, an unrecorded way of filling the office as not elected. Every country-year therefore comes out as yes or no, where 20, 17 and 17 countries were previously undecided — and since this can only add to the "no" side, no country's "yes" changed. Most unclear genders are years where V-Dem names the officeholder but leaves the gender field empty: 70 people in all, every one of them a man, among them the sultans of Oman, the sheikhs of Kuwait, the grand dukes of Luxembourg and Pope Pius IX. The rest are collective bodies or successions of short-lived officeholders, almost entirely nineteenth-century. Before this, two years under President Sadi Carnot were enough to leave France undecided on ever having had a woman head of state, for every year from 1887 onwards.

## Changes to indicators already published

**Women-leader indicators are no longer extended back in time from the historical entity a country was part of.** Who led a country is a fact about that country rather than about the power that ruled its territory earlier, and for the cumulative indicators the imputation actively misled: Ukraine inherited Catherine the Great from Russia, and so read as having had a woman head of state from its first year in the data. The years before a country appears in V-Dem's own data are now left empty. Women in parliament keeps its imputation.

This shortens the three "Gender of the …" charts for about 40 countries, which is the main thing to check on staging. The country counts and shares are unaffected, as they already skipped imputed countries — I compared both published share indicators against production and got 237 rows each with no differences.

**The population variants of these indicators are removed** (`population_wom_hoe_ever`, `population_wom_hoe_ever_demelect`), and none were added for the new offices. These indicators are reported per country and as a share of countries; the number of people living in such countries is not something we need. No chart references them.

## Notes on the code

Every published indicator keeps its name, and therefore its variable ID — verified table by table against the published catalog: 12 additions, 2 deliberate removals, no renames. The democratically elected variant stays suffixed `_dem` at country level and `_demelect` in the counts, matching how the chief-executive indicators are already published, rather than unifying the two and orphaning charts.

The three offices are declared once and the six indicators are generated from that list, so they cannot drift apart. Sanity checks assert that no country-year is left undecided, that a country never stops being flagged once a woman has held the office, and that the democratically elected variant is never 1 where the plain one is 0.

https://claude.ai/code/session_01HSDtmgvHf1SnKawudPncPW

---
_Generated by [Claude Code](https://claude.ai/code/session_01HSDtmgvHf1SnKawudPncPW)_


paarriagadap:
@codex review

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem_clean.py
```diff
@@ -8,6 +8,20 @@
 
 from etl.data_helpers import geo
 
+# The three leadership offices, and for each the V-Dem flag giving the gender of whoever held it at
+# any point during the year, plus the flag saying whether the office was filled through multi-party
+# elections. Used to build the "ever had a woman leader" indicators.
+OFFICES = {
+    "hos": {"gender": "v2exfemhos", "elected": "v2elmulpar_osp_hos"},
+    "hog": {"gender": "v2exfemhog", "elected": "v2elmulpar_osp_hog"},
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

The code observation is accurate, but the consequence described here does not apply to this change.

**It is pre-existing.** `estimate_hog_indicators` is byte-identical to `master` on this branch — the `vdem_clean.py` hunks in this PR start at lines 8, 69, 841 and 1131, and none of them touch that function. The `v2expathhs` / `v2expathhg` mix is genuinely there, and it is even internally inconsistent: the three branches that set `v2elmulpar_osp_hog` all read `v2expathhs`, while the high/low helper reads `v2expathhg` only in its `== 8` branch and falls back to `v2expathhs` for `== 7` and `== 6`. But it predates this PR.

**It is inert for the new indicator.** Against the built dataset, the multi-party-election condition excludes zero country-years from `wom_hog_ever_dem`: there is no row where the head of government is a woman, the country is an electoral or liberal democracy, and `v2elmulpar_osp_hog` is 0. Correcting the flag could therefore not add a single "yes" to this indicator, so the undercount described here does not occur. The headline figures reproduce exactly (2025, 179 countries: 76/103 for `hog`, 58/121 for `hog_dem`), and the pre-existing `share_countries_wom_hoe_ever_demelect` series is byte-identical to production across all 237 years.

So: worth a separate issue to make the head-of-government selection path consistent, since other consumers of `v2elmulpar_osp_hog` could be affected. Not a blocker for this PR, and not something it introduced. Left for the author.

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
```diff
@@ -3045,6 +3155,78 @@ tables:
...
+          <%- elif category == "yes" %>
+          Number of countries that had a woman head of state elected democratically
+          <%- elif category == "no" %>
+          Number of countries that never had a woman head of state elected democratically
+          <%- endif -%>
+        description_short: |-
+          <%- if category == '-1' -%>
+          Countries for which we do not know whether they had ever had a woman head of state while being a democracy, by the end of the given year.
+          <%- elif category == "yes" %>
+          Countries that had ever had a woman head of state while being an electoral or a liberal democracy, by the end of the given year. A woman counts for a given year if she held the office at any point during that year.
```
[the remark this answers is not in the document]
paarriagadap: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Valid — the aggregate `description_short` understates what is being counted.

The country-level indicator requires all three conditions: woman, office filled through multi-party elections, and electoral or liberal democracy — `((tb[column_woman] == 1) & was_elected & is_democracy)`. `wom_hos_ever_dem`'s own `description_processing` spells out why both of the last two are needed, noting that asking only about democracy "would count hereditary monarchs, since many democracies have an unelected head of state". The aggregate `description_short` then says only "while being an electoral or a liberal democracy", which read literally describes exactly the broader population the data excludes.

The data confirms the gap matters: the UK, Denmark, the Netherlands and New Zealand all come out "no" on the democratically elected head-of-state variant, despite having had a woman head of state while being a democracy. A reader going by the subtitle alone would expect them to be counted.

This is worth fixing because `description_short` is what renders as the chart subtitle, so it is the text a reader actually uses to interpret the number; the titles do say "elected democratically", but that qualification does not carry into the subtitle.

Fields affected, all six `demelect` aggregates:

- `num_countries_wom_hos_ever_demelect`, `num_countries_wom_hog_ever_demelect` — new in this PR
- `share_countries_wom_hos_ever_demelect`, `share_countries_wom_hog_ever_demelect` — new in this PR
- `num_countries_wom_hoe_ever_demelect`, `share_countries_wom_hoe_ever_demelect` — pre-existing, same phrasing, so worth fixing in the same pass for consistency

The `-1` branch is looser still ("while being a democracy") and needs the same treatment.

No code change is needed — only the `description_short` wording. Left for the author.

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
```diff
@@ -541,7 +548,51 @@ tables:
...
+        description_processing: |-
+          We derive this indicator as follows:
+
+          1) Take the gender of the head of state in a given year from V-Dem's `v2exfemhos` indicator, which refers to any point during the year rather than to December 31.
+          2) If 1) evaluates to "female", the office was filled through multi-party elections, and the country was an electoral or liberal democracy, we assign the value 1 to the indicator for the given year and all following years.
+
+          Both conditions in 2) are required. Asking only whether the country was a democracy would count hereditary monarchs, since many democracies have an unelected head of state. Asking only whether the office was filled through multi-party elections would count monarchs of countries that had an elected legislature alongside a hereditary throne.
+
+          To assess whether the office was filled through multi-party elections, we use V-Dem's `v2elmulpar_osp_hos` indicator. To assess whether the country was a democracy, we use our political regime classification, which is based on V-Dem's Regimes of the World indicator.
```
paarriagadap, line 589: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Three of the new sentences attribute variables to V-Dem that we construct ourselves.

`v2elmulpar_osp_hos` is not a V-Dem indicator. It is built in `estimate_hos_indicators` from `v2elmulpar_osp_ex` / `v2elmulpar_osp_leg` (also ours) together with `v2expathhs`, `v2xex_elecreg`, `v2xlg_elecreg` and `v2ex_legconhos`. I checked both the v16 codebook and the raw file that the snapshot ships:

- the codebook has **no** occurrence of `v2elmulpar_osp_hos`, `v2elmulpar_osp_hog` or `v2elmulpar_osp_hoe`, and **no** `v2…hoe…` token anywhere in its 507 pages;
- none of those names is a column in V-Dem's own data file either — nothing in it contains "hoe" at all.

The same applies in two more places:

- line 687 (`wom_hog_ever_dem`): "we use V-Dem's `v2elmulpar_osp_hog` indicator" — built in `estimate_hog_indicators`.
- line 792 (`wom_hoe_ever_dem`): "the V-Dem indicator `electmulpar_hoe_row_owid`, which is based on V-Dem indicator `v2elmulpar_osp_hoe`" — both are ours, and the `_owid` suffix on the first is there precisely to say so.

What V-Dem does publish, and what the codebook documents, is `v2elmulpar_osp` plus the inputs these are derived from: `v2expathhs`, `v2expathhg`, `v2exhoshog`, `v2ex_hosw`, `v2ex_hogw`, `v2exfemhos`, `v2exfemhog`. The four "We derive this indicator from V-Dem indicators: `v2exfemhos`, `v2exfemhog`, `v2ex_hosw`, `v2ex_hogw`" sentences elsewhere in the file are correct — this is only about the election-flag sentences.

A reader who follows these names into the codebook will not find them, and the text credits the producer with a methodological choice that is ours. Suggested shape for all three:

> To assess whether the office was filled through multi-party elections, we use our own indicator for this office, derived from V-Dem's `v2elmulpar_osp` together with `v2expathhs` and `v2expathhg`.

Worth noting the underlying cause, since it will keep producing this: the codebase has two conventions that disagree. `v2elmulpar_osp_hoe` is renamed to `electmulpar_hoe_row_owid` to mark it as ours, but `v2exfemhoe` keeps a `v2` prefix while being equally derived — built in `estimate_gender_hoe_indicator` by taking the gender of whichever office holds more power — and it is published under the title "Woman chief executive". Renaming it on the same `_owid` pattern would make the distinction visible at the point where these sentences get written.

For reference, the concept is V-Dem's even though the variable is not. The codebook defines it once, in the polyarchy-index section, as "the head of the executive — either the head of state or head of government — whoever has more relative power over the appointment and dismissal of cabinet ministers as measured by `v2ex_hosw`", which is exactly what the step implements.

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem_impute.py
```diff
@@ -29,6 +29,22 @@
     "corr_jud_vdem",
     "v2mecorrpt",
     "v2xnp_client",
+    # Who led a country is a fact about that country, not about the power that ruled it earlier, so
```
paarriagadap, line 32: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Dropping the imputation here also moves eight published population indicators, which the description does not mention. A local `etl diff REMOTE` against the catalog shows `population_hos__category_{man,woman,-1}`, `population_hog__category_{man,-1}` and `population_hoe__category_{man,woman,-1}` all modified.

The reason is that the two aggregation paths treat imputed rows differently, deliberately: `make_table_countries` drops them before counting, while `make_table_population_counts` keeps them, because population figures only exist for today's borders. So the country counts are untouched, as the description says — but the population buckets shift, with those countries' pre-independence population moving out of man/woman and into "no data". It is material in the older years: the world "no data" bucket is 204M in 1900, 147M in 1950 and 210M in 1970.

None of the eight is used by a chart, so there is no reader-facing impact. It is just the one published series family this change actually moves, and worth a line so it is not a surprise later.

Two smaller things while in this file:

- The count series move by more than the description says. It states that five countries go from unknown to "no" for the chief executive; on the built output `num_countries_wom_hoe_ever` goes from 75/87/17 to 75/104/0 in the latest year and `_demelect` from 55/114/10 to 55/124/0, with the unknown category at zero from 1991 onward. No country's "yes" changed, which is the important part and holds exactly.
- There is an unused `a = {"lala", "lele"}` a few lines below this list, predating the change.

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
```diff
@@ -586,7 +639,58 @@ tables:
...
+        title: Country had a woman head of government democratically elected at least once
+        description_short: |-
+          Denotes whether a country ever had a democratically elected woman head of government.
+        description_processing: |-
+          We derive this indicator as follows:
+
+          1) Take the gender of the head of government in a given year from V-Dem's `v2exfemhog` indicator, which refers to any point during the year rather than to December 31. Where the head of state was also the head of government, given by V-Dem's `v2exhoshog` indicator, we use the gender of the head of state instead.
+          2) If 1) evaluates to "female", the office was filled through multi-party elections, and the country was an electoral or liberal democracy, we assign the value 1 to the indicator for the given year and all following years.
+
+          Both conditions in 2) are required. Asking only whether the country was a democracy would count leaders who never stood for election. Asking only whether the office was filled through multi-party elections would count leaders of countries whose elections were neither free nor fair.
```
paarriagadap, line 685: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

This paragraph is accurate for the head of state but describes a safeguard that never fires for the head of government.

Counting country-years on the built output: for the head of state the multi-party-election condition excludes 314 of them across 9 countries — Australia, Barbados, Canada, Denmark, Luxembourg, Netherlands, New Zealand, Philippines and the United Kingdom — which is exactly the monarchs and governors-general the paragraph is there to keep out. For the head of government it excludes nothing at all: no country-year has a woman, an electoral or liberal democracy, and a zero election flag. `wom_hog_ever_dem` is determined by the regime classification alone.

That makes sense on reflection — a head of government in a democracy has essentially always come through an elected legislature or a direct election — but a reader comparing the two variants cannot tell that one condition is doing the work and the other is inert. Either say so, or drop the justification here and keep it on the head-of-state variant where it bites.

The equivalent sentence on the chief executive is nearly inert too: the condition binds 3 country-years, all Philippines, and changes no country's answer.

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
```diff
@@ -541,7 +548,51 @@ tables:
...
-          {definitions.aux.description_imputes}
+          {definitions.aux.description_no_imputes}
+        unit: ""
+
+      # Had female HOS ever
+      wom_hos_ever:
+        title: Country had a woman head of state at least once
+        description_short: |-
+          Denotes whether a country ever had a woman head of state.
+        description_from_producer: |-
```
paarriagadap, line 559: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Small consistency point across the four new indicators, worth settling in one pass.

`description_from_producer` appears on `wom_hos_ever` and `wom_hog_ever` but not on either `_dem` variant, and not on the existing `wom_hoe_ever`. Where it does appear the two differ: this one carries only the clarification, while `wom_hog_ever` carries the full question and response list.

Worth considering whether it belongs on these at all. The text describes the producer's yearly gender question, but the indicator it now sits on is the cumulative "ever had" derivation, so a reader opening "Source description" gets a definition of something else. Dropping it from the derived indicators and leaving it on the raw `v2exfemhos` / `v2exfemhog` entries would be more consistent than adding it to the other two.

Separately, the interpolated blocks come in different orders between siblings: `wom_hos_ever` puts the conservative-reading note before the "only estimated for countries listed in the most recent year" sentence, `wom_hos_ever_dem` after. And the new indicators say "countries listed in V-Dem in the most recent year of the data" where the existing `wom_hoe_ever` says "countries that V-Dem lists in the most recent year of the data".

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem_clean.py
```diff
@@ -1131,35 +1148,133 @@ def _estimate_indicator(tb, column_name, hos_col, hog_col):
...
+        tb[_column_anytime(office)] = tb[columns["gender"]].astype("float64")
+
+    # Carry the head-of-state gender over to the head of government where one person holds both.
+    mask = (tb[_column_anytime("hos")].notna()) & (tb["v2exhoshog"] == 1)
+    tb.loc[mask, _column_anytime("hog")] = tb.loc[mask, _column_anytime("hos")]
+
+    # V-Dem's own head-of-state and head-of-government flags never contradict each other when the
+    # two offices are held by the same person, so the carry-over cannot overwrite anything.
+    both = (tb["v2exhoshog"] == 1) & tb["v2exfemhos"].notna() & tb["v2exfemhog"].notna()
+    assert (tb.loc[both, "v2exfemhos"] == tb.loc[both, "v2exfemhog"]).all(), (
```
paarriagadap, line 1180: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

This assert holds, but it covers much less than it looks like it does.

Only 23 rows satisfy its condition — same person in both offices *and* both gender columns filled — and there are indeed zero disagreements among them. The carry-over it is guarding fills roughly 14,520 rows. So the check confirms the convention on about 0.2% of the cases it licenses.

That is not an argument against the carry-over, which is clearly right and which the data supports everywhere it can be tested. It is just worth knowing that the assert would not catch the producer changing this convention, since almost no row exercises it. If you wanted real cover, asserting on the count itself — that the same-person rows with both columns filled stay small and stay consistent — would at least fail loudly if the shape of the input changed.

etl/steps/data/garden/democracy/2026-03-17/vdem/vdem.meta.yml
paarriagadap, line 1: > _Written by Claude Opus 5 — @paarriagadap at the wheel._

Two file-level leftovers, both predating this change but cheap to fix while the file is open.

`dataset.owners` is missing — the `dataset:` block carries only `title` and `update_period_days`. The convention is for it to list the accountable owner first.

And `description_imputes` links the imputation file from the previous version's folder rather than this one. The link resolves, so it is cosmetic, but it points readers at a file that is a year out of date relative to the data they are looking at.

PR #6848, opened 2026-09-08
🔨 Use create_snapshot(data=...) instead of df_to_file in snapshot scripts

lucasrodes:

`Snapshot.create_snapshot(data=df, upload=upload)` already writes the dataframe to `snap.path` and calls `dvc_add`, so the `df_to_file(df, file_path=snap.path)` + `snap.dvc_add(upload=upload)` pair was a hand-rolled copy of it. 125 snapshot scripts and the wizard cookiecutter already use `create_snapshot`; these 22 call sites were the stragglers.

Two things in `etl/snapshot.py` that this exposed:

- `create_snapshot` wrote the dataframe via `owid.datautils.dataframes.to_file` — the deprecated shim that warns on every call — instead of `owid.datautils.io.df_to_file`.
- **`SnapshotMeta.save()` rewrote a `.dvc` even when nothing had changed.** `yaml.safe_load` parses an unquoted `date_accessed: 2026-03-20` into a `datetime.date` while `Origin` keeps it as a string, so the "no change" check always missed and the file was reserialized: dates re-quoted, long lines reflowed, `outs` re-indented. 851 of the repo's 8253 `.dvc` files carry an unquoted date, and `create_snapshot` calls `save()` on every run — so those diffs surfaced in whichever dataset update happened to come next. Without this fix, 13 of the 19 `.dvc` files behind the migrated scripts would have been reformatted on their next run. Regression test added, verified to fail without the fix.

## Verification

`make check` and 875 unit tests pass. `who/2023-07-14/standard_age_distribution.py` (the one migrated script that runs fully offline) was run end to end with `--skip-upload`: output md5 `5c6bc0b649c343483f4c8d07fbb540c5`, 333 bytes — an exact match for what its `.dvc` records, with the `.dvc` left untouched.

## Open items

- **The other 20 migrated scripts were not executed** — they need network, API keys (BLS, FRED, Media Cloud), or PDF parsing. The edit is mechanical and identical across all of them, and none of their `.dvc` files change under `save()`, but nobody ran them.
- **482 legacy backport scripts left alone.** They hand-roll the same thing as `df.reset_index().to_feather(snap.path)` + `dvc_add`, all in 2014–2023 version folders. Same collapse would apply; deliberately out of scope to keep this diff reviewable.
- **`snapshots/aviation_safety_network/2022-10-12/aviation_statistics.py` is an orphan** — pre-existing, not caused by this PR. Its two `.dvc` files were removed when the step was archived (`dag/archive/main.yml:145`), so the script can only raise `FileNotFoundError`. Migrated for consistency; deleting the leftover `.py` is a separate call.

### DIFF
--- a/apps/wizard/app_pages/fasttrack/fast_import.py
+++ b/apps/wizard/app_pages/fasttrack/fast_import.py
@@ -8,7 +8,6 @@ from pathlib import Path
 import pandas as pd
 from git.repo import Repo
 from owid.catalog import Dataset, DatasetMeta, Origin, Table
-from owid.datautils import io
 from rich.console import Console
 from sqlalchemy.orm import Session
 from structlog import get_logger
@@ -147,8 +146,7 @@ class FasttrackImport:
 
         # upload snapshot
         snap = self.snapshot
-        io.df_to_file(self.data, file_path=snap.path)
-        snap.dvc_add(upload=True)
+        snap.create_snapshot(data=self.data, upload=True)
 
         # save metadata
         self.__snapshot_path = snap.metadata_path
--- a/etl/snapshot.py
+++ b/etl/snapshot.py
@@ -1,3 +1,4 @@
+import datetime as dt
 import re
 import tempfile
 import time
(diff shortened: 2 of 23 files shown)


lucasrodes:
@codex review

PR #6852, opened 2026-09-09
🎉 Add skill to log data producer report emails

pabloarosado:
> _Written by Claude Fable 5.1 — @pabloarosado at the wheel._

Adds `/log-producer-interactions`, a skill that records the analytics reports we send to data producers, and the producers' replies, in the shared Notion log: one row per email, incremental (it starts from the latest logged date per producer), deduplicated on the Gmail message ID, with quoted replies, gateway banners and signatures stripped from the stored body.

Context: the data producer relations project sends periodic analytics reports to producers (`etl/scripts/data-producer-reports/`). The log is the shared answer to "which producer got which report, when, and what did they say back", which is otherwise spread across individual inboxes and impossible to see when the person who sent it is away.

**Scope is deliberately narrow, and the skill enforces it:**

- It searches **only the addresses recorded in the contacts table**. No domain searches, no producer-name searches. An earlier version had both, and they returned everything an institution had ever sent, most of it unrelated to us.
- It logs **only threads that share a report**. Data issues, release announcements, introductions and anything personal are named in one line in the run's report and never written to Notion. Data issues belong in Front, which the whole team can act in.
- It **shows the proposed rows and waits for approval** before creating any.
- Each run sees the caller's own mailbox only, which is also why the log cannot pretend to be a complete history of a relationship: calls, Slack, and colleagues' threads are invisible to it.

Changes:
- `.claude/skills/log-producer-interactions/SKILL.md`: the procedure (internal skill), including why the scope is what it is, so it does not drift back.
- `.claude/skills/log-producer-interactions/scripts/strip_quotes.py`: prints the messages of a saved Gmail thread with quoted text, signatures and prepended `CAUTION:`-style gateway banners removed.
- `.env.example`, `etl/config.py`: new `NOTION_DATA_PRODUCER_INTERACTIONS_TABLE_URL`.
- `CLAUDE.md`: entry in the skill index.

No producer names, contact names, email addresses or Notion URLs are in the repo; they live in Notion and in `.env`. Examples in the skill use an invented institution.

Open items:

- **Everyone who runs this needs the new `.env` entry**; without it the skill stops and says so.
- The skill has **not been run in this narrowed form**. The one real run happened under the earlier, broader design, whose rows have since been removed from the log; the first run after merge is the real test.
- The env var and the Notion database are still named "interactions" although the log is now reports-only. Renaming the variable would mean every user editing their `.env`, so it is left alone on purpose.
- Not covered here: having the report script append its own "report sent" row, which would remove the mailbox step for outgoing mail entirely, and scheduling the skill. Both need the ETL Notion integration to be granted access to the log database.
- Nobody has reviewed the skill against a second mailbox, so the "identify the caller" and watermark logic is exercised on one account only.

### DIFF
--- a/.claude/skills/log-producer-interactions/SKILL.md
+++ b/.claude/skills/log-producer-interactions/SKILL.md
@@ -1,29 +1,35 @@
 ---
 name: log-producer-interactions
 description: >-
-  Sync the current user's Gmail exchanges with OWID's data producers into the shared Notion
-  "Data Producer Interactions" log, one row per email, incrementally and without duplicates.
-  Use when the user says "log my producer emails", "update the interactions log", "sync producer
-  interactions", after sending analytics reports, or when someone asks what is pending with a
-  data producer. Also flags emails from producers that are not yet in the contacts table, and
-  emails that look like open questions or promises, so the team can track them with Status and
-  Owner. Runs against the caller's own mailbox only.
+  Log the analytics reports sent to OWID's data producers, and the producers' replies, into the
+  shared Notion reports log: one row per email, incrementally and without duplicates. Use after
+  sending a round of producer analytics reports, or when the user says "log the report emails" or
+  asks whether a producer ever replied to a report. Searches only the addresses recorded in the
+  contacts table, only for report threads, and asks before writing anything. Runs against the
+  caller's own mailbox only.
 metadata:
   internal: true
 ---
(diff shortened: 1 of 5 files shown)


pabloarosado:
@codex review

pabloarosado:
@codex review

PR #6853, opened 2026-09-09
🔨 Rename data provider to data producer

pabloarosado:
## Human summary
Consistently use "data producer" instead of "data provider" (at least internally). It's not that one term is significantly better than the other, but currently we use both indistinctively, and it's confusing, so I just picked producer, which I think is slightly preferrable.
In other public-facing surfaces, like data pages, we still use "provider" sometimes, but I think that's fine. This PR is mostly to have internal consistency to avoid confusion.

---
> _Written by Claude Fable 5.1 — @pabloarosado at the wheel._

Rename "data provider" to "data producer" wherever the word means the organisation whose data we use. The metadata model has always said producer (`origin.producer`, `description_from_producer`), while prose, comments, our own corrections schema and the Notion glue said provider about two to one. This aligns everything internal with the model. No reader-facing text changes.

- Corrections: the `provider` key of `.corrections.yml` becomes `producer` (139 rows, plus `schemas/corrections-schema.json`, `etl/data_corrections.py`, `apps/corrections/cli.py`, tests, and the two skills that mention the key).
- Notion glue: `NOTION_DATA_PROVIDERS_CONTACTS_TABLE_URL` becomes `NOTION_DATA_PRODUCERS_CONTACTS_TABLE_URL` (update your local `.env`), and `etl/notion.py` follows the already renamed Impact Highlights column "Data producer(s) related".
- Reports script: `etl/scripts/data-provider-reports/` becomes `etl/scripts/data-producer-reports/`, `providers.yml` becomes `producers.yml` (key `PRODUCERS`), and the period subfolder in Drive is named "<period> Data producer reports". The one existing subfolder (H1-2026) was renamed in place in Drive, so its ID and every link to it or to the files inside are unchanged; the status sheet links by file ID and is unaffected.
- Developer-facing docs and guidance: the public ETL docs and the dataset schema guidance.
- Comments and docstrings in the report script, HTTP helper, catalog API, and the latest versions of a few steps and snapshots.

Deliberately left as is: all reader-facing text on ourworldindata.org and in indicator metadata (readers understand "provider" and "producer" as synonyms there); text quoted from producers (OECD "providers of development cooperation", UNAIDS, WDI); unrelated senses (LLM, VSCode, email and OTT providers, energy providers); the region-grouping "provider" terminology (`add-provider-regions`, `(Provider)` suffixes), which is coupled to grapher identifiers; superseded step versions; the style guide, which is synced from Notion; and the `#proj-data-providers` channel name until the channel is renamed.

Open item: everyone needs to rename the env var in their `.env`. The new skill in [#6852 "Add skill to log data producer interactions"](https://github.com/owid/etl/pull/6852) references the renamed variable, so this PR should merge first.


pabloarosado:
@codex review

PR #6855, opened 2026-09-09
📜🤖 Document that merging reruns the pipeline

edomt:
> _Written by Claude Opus 5 — @edomt at the wheel._

Adds one paragraph to `CLAUDE.md`'s Git Workflow section: merging to master triggers the production ETL run, which rebuilds every step whose checksum changed and performs the `grapher://` upserts.

## Why

This is documented nowhere in the repo — not `CLAUDE.md`, `.claude/docs/`, `.claude/rules/`, any skill, or `docs/`. `.github/workflows/buildkite.yml` only *destroys* staging on PR close; the production rebuild is a Buildkite pipeline outside this repo, so there was no trace of it here to find. The only merge-related line in `CLAUDE.md` was `etl pr-clean` deleting branches.

The cost of the gap: after a metadata-only or code change, agents report "the affected datasets are now dirty and still need their `grapher://` upserts" as leftover work, and propose coordinating re-runs across dataset owners. That's noise — it's just what merging does. It came up on #6854, where the caveat spanned eight namespaces and was entirely self-resolving.

The paragraph also keeps the two cases that *do* still need a human: a `--force` re-run where no checksum changed (upstream data patched out-of-band), and ghost-variable remapping after a version bump.

Docs only — no code or behavior change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -189,7 +189,7 @@ git push
 gh pr edit <number> --body "..."
 ```
 
-**Merging reruns the pipeline.** Merging to master triggers the production ETL run, which rebuilds every step whose checksum changed and performs the `grapher://` upserts. You do **not** need to re-run steps for a change to take effect, and it isn't an open item worth reporting — editing a step's code or metadata is enough, because `etlr`'s change detection picks it up on that run. Run steps locally to *verify* output before merge, not to publish it. The exceptions still worth flagging: a `--force` case where no checksum changed (upstream data patched out-of-band), and ghost-variable remapping after a version bump.
+**Merging reruns the pipeline.** Merging to master triggers the production ETL run (a Buildkite pipeline), which rebuilds every step whose checksum changed and performs the `grapher://` upserts. You do **not** need to re-run steps for a change to take effect, and it isn't an open item worth reporting — editing a step's code or metadata is enough, because `etlr`'s change detection picks it up on that run. Run steps locally to *verify* output before merge, not to publish it. The exceptions still worth flagging: a `--force` case where no checksum changed (upstream data patched out-of-band), and ghost-variable remapping after a version bump.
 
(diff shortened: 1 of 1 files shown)


edomt:
@lucasrodes Just to confirm this is correct?

CLAUDE.md
```diff
@@ -189,6 +189,8 @@ git push
 gh pr edit <number> --body "..."
 ```
 
+**Merging reruns the pipeline.** Merging to master triggers the production ETL run, which rebuilds every step whose checksum changed and performs the `grapher://` upserts. You do **not** need to re-run steps for a change to take effect, and it isn't an open item worth reporting — editing a step's code or metadata is enough, because `etlr`'s change detection picks it up on that run. Run steps locally to *verify* output before merge, not to publish it. The exceptions still worth flagging: a `--force` case where no checksum changed (upstream data patched out-of-band), and ghost-variable remapping after a version bump.
```
lucasrodes: this is great! only thing i can think of adding is a reference to "Buildkite" as the place where the pipeline is triggered. I considered maybe even adding a pointer to the specific job that runs the etl pipeline (https://github.com/owid/ops/blob/main/.buildkite/etl/deploy.yml), but that maybe that's too much and it could  easily change in the future (and be an invalid link).

So, in short, maybe just worth adding a note that this is executed in Buildkite could be enough

lucasrodes (approved):
thanks for adding this, Ed! This is a good example of something we data folks would assume to be known, but it is not so obvious for someone contribute less often to the repo!

CLAUDE.md
edomt: > _Written by Claude Fable 5.1 — @edomt at the wheel._

Added — the sentence now says the production ETL run is a Buildkite pipeline. Left out the link to the ops deploy job for the reason you give: it would go stale.

PR #6877, opened 2026-09-14
🐛🤖 Fail fast when DB_HOST is set but empty

Marigold:

A blank `DB_HOST` reaches SQLAlchemy and surfaces roughly 240 lines later as:

```
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError)
(2003, "Can't connect to MySQL server on '' ([Errno -2] Name or service not known)")
```

raised from `select_dirty_steps`, before any step runs. The traceback names no step and reads like a database outage rather than a configuration problem. This adds one assert so it fails on a readable line instead.

## Why the value can only come from an .env file

Neither code path in `config.py` can produce an empty host:

- `load_STAGING()` maps `""` and `"0"` to `None`, so the `STAGING` override does not run for a blank value.
- `get_container_name()` always returns a `staging-site-…` string.

`env.get("DB_HOST", "localhost")` returns `""` only when the key is **present with no value** — so the value was read from an .env file that had `DB_HOST=` in it. On a staging server that means the container's .env was read while it was being rewritten by a concurrently starting build.

## Scope

Six staging builds failed this way over the last ~11 days. This does not prevent any of them — the fix for that is on the ops side. It makes the next one diagnosable immediately.

Behaviour is otherwise unchanged: an absent `DB_HOST` still defaults to `localhost`.

## Testing

Exercised all three paths through `ENV_FILE`:

| `.env` contents | Result |
|---|---|
| `DB_HOST=` | `AssertionError: DB_HOST is set but empty. Expected a hostname; check the .env this process loaded.` |
| `DB_HOST=somehost` | imports cleanly, `DB_HOST == 'somehost'` |
| key absent | imports cleanly, `DB_HOST == 'localhost'` |

Also confirmed the `STAGING` path is unaffected (`STAGING=some-branch` → `staging-site-some-branch`).

## Still open

Nothing in this PR addresses the cause. That is build-level concurrency on the staging servers, handled separately in owid/ops.


Marigold:
@codex review

